PR #13, opened 2017-05-03
Updated flight mode documentation for some auto modes

hamishwillee:
This creates separate and updated flight mode docs for takeoff, land, hold, rtl. It also moves them into their own folder and fixes up links appropriately.


en/flight_modes/land.md
```diff
@@ -0,0 +1,43 @@
...
+--- | ---
+[MPC_LAND_SPEED](../advanced_config/parameter_reference.md#MPC_LAND_SPEED) | The rate of descent during landing. This should be kept fairly low as the ground conditions are not known.
+
+## Fixed Wing (FW)
+
+The vehicle turns and lands at the location at which the mode was engaged. 
+
+> **Note** The vehicle doesn't necessarily know where the ground is, and will therefore follow a fixed trajectory for the whole landing approach (no flare).
+
+Landing is affected by the following parameters:
```
hamishwillee: @dagar FW LAND mode simply follows a trajectory to ground. Would it be correct to say therefore that most of the parameters here (e.g. FW_LND_HVIRT, FW_LND_FLALT, etc are not relevant). Can you confirm which ones are/are not relevant and I'll delete them from this list.

en/flight_modes/takeoff.md
```diff
@@ -0,0 +1,56 @@
...
+--- | ---
+[RWTO_TKOFF](../advanced_config/parameter_reference.md#RWTO_TKOFF) | Runway takeoff with landing gear.
+-->
+
+> **Warning** FW takeoff is still being clarified. The available information is listed below.
+>
+>  FW takeoff splits into two types
+>
+>  1. Runway takeoff has the following phases
+>      - "throttle ramp"
```
hamishwillee: Is this just "ramp up throttle until you go to next phase"? When does that happen 
- on movement? 
- On some parameter? 
- at some speed?

What params affect the final throttle setting and its rate of ramp up?

en/flight_modes/takeoff.md
```diff
@@ -0,0 +1,56 @@
...
+>      - "clamped to runway" until min airspeed ([FW_AIRSPD_MIN](../advanced_config/parameter_reference.md#FW_AIRSPD_MIN) * [RWTO_AIRSPD_SCL](../advanced_config/parameter_reference.md#RWTO_AIRSPD_SCL))
+>      - "takeoff" until alt > navigation altitude ([RWTO_NAV_ALT](../advanced_config/parameter_reference.md#RWTO_NAV_ALT)) 
+>      - "climbout" until alt agl > [FW_CLMBOUT_DIFF](../advanced_config/parameter_reference.md#FW_CLMBOUT_DIFF)
+>    There's a bit more detail for roll limitation until climbout, and some heading specifics. 
+>
+>  1. Normal/non runway takeoff
+>     - full throttle climbout until altitude error < [FW_CLMBOUT_DIFF](../advanced_config/parameter_reference.md#FW_CLMBOUT_DIFF) then regular navigation
```
hamishwillee: - What is a full throttle climbout? Specifically, I am envisaging that this just means ramping throttle to max and heading forward until you start getting lift off. 
  - Is there a stage where you start playing with pitch etc?
- What params affect the throttle value and rate of ramp up?

en/flight_modes/takeoff.md
```diff
@@ -0,0 +1,56 @@
...
+[RWTO_TKOFF](../advanced_config/parameter_reference.md#RWTO_TKOFF) | Runway takeoff with landing gear.
+-->
+
+> **Warning** FW takeoff is still being clarified. The available information is listed below.
+>
+>  FW takeoff splits into two types
+>
+>  1. Runway takeoff has the following phases
+>      - "throttle ramp"
+>      - "clamped to runway" until min airspeed ([FW_AIRSPD_MIN](../advanced_config/parameter_reference.md#FW_AIRSPD_MIN) * [RWTO_AIRSPD_SCL](../advanced_config/parameter_reference.md#RWTO_AIRSPD_SCL))
```
hamishwillee: Does "clamped to runway" just mean ... "you're moving on the runway, haven't got enough airspeed to take off"?

en/flight_modes/takeoff.md
```diff
@@ -0,0 +1,56 @@
...
+>      - "takeoff" until alt > navigation altitude ([RWTO_NAV_ALT](../advanced_config/parameter_reference.md#RWTO_NAV_ALT)) 
+>      - "climbout" until alt agl > [FW_CLMBOUT_DIFF](../advanced_config/parameter_reference.md#FW_CLMBOUT_DIFF)
```
hamishwillee: - What is AGL?
- I think this basically says that you keep climbing until you reach this height (or within the tolerance for this height) - correct?
- What happens next? Do we then loiter?
- Is there a stage where we start changing the FW attitude to head up steeper?

en/flight_modes/takeoff.md
```diff
@@ -0,0 +1,56 @@
...
+<!-- What about MT_TKF_THR_MIN and friends? Also RWTO_TKOFF and friends? https://docs.px4.io/en/advanced_config/parameter_reference.html#runway-takeoff 
+
+Takeoff is affected by the following parameters:
+
+Parameter | Description
+--- | ---
+[RWTO_TKOFF](../advanced_config/parameter_reference.md#RWTO_TKOFF) | Runway takeoff with landing gear.
+-->
+
+> **Warning** FW takeoff is still being clarified. The available information is listed below.
```
hamishwillee: @dagar . Some questions required to complete this section on TAKEOFF mode

1. How does the system know you want to do a Runway takeoff or a Normal takeoff? Is this parameter driven, and if so, which one.
2. Are there any hard rules about which type to use? Ie why not do a runway takeoff in a normal location

en/flight_modes/rtl.md
```diff
@@ -0,0 +1,45 @@
...
+
+Parameter | Description
+--- | ---
+[RTL_RETURN_ALT](../advanced_config/parameter_reference.md#RTL_RETURN_ALT) | Minimum RTL return altitude in meters (default: 60m). The vehicle will ascend to this altitude before returning. If already above this value the vehicle will return at its current altitude.
+[RTL_DESCEND_ALT](../advanced_config/parameter_reference.md#RTL_DESCEND_ALT) | Altitude at which the vehicle will slow or stop its initial descent (default: 30m)
+[RTL_LAND_DELAY](../advanced_config/parameter_reference.md#RTL_LAND_DELAY) | Time to hover at `RTL_DESCEND_ALT` before landing (default: 0.5s). If set to -1 the system will loiter at `RTL_DESCEND_ALT` rather than landing.
+[RTL_MIN_DIST](../advanced_config/parameter_reference.md#RTL_MIN_DIST) | Minimum distance to trigger rising to a safe altitude. If the system is horizontally closer than this distance to home it will land straight on home instead of raising to the return altitude first.
+
+
+## Fixed Wing (FW)
```
hamishwillee: @dagar My understanding is that RTL can land or circle. If it does land (because you have set RTL_LAND_DELAY>0) then it will do so using flare etc (the same land as you get in mission)
- is that correct?

I think we should separately document this landing because it is used in mission as well. Then link to it. OK?

en/flight_modes/land.md
dagar: This is where it gets convoluted. The autopilot likely doesn't know the actual height of the ground. When you jump into LAND as a mode it's the same landing behavior, but it internally created the land waypoint at altitude 0. As a result you probably aren't going to hit the flare logic (FW_LAND_FLALT) unless you actually get to FW_LAND_FLALT over the land altitude. If you have a distance sensor (and it's enabled), then it will be used instead.

If we used HOME altitude it would sometimes be right, sometimes be wrong.

It seems like we should be explaining FW LAND logic somewhere, and then reference it for LAND, MISSION, and RTL.

en/flight_modes/takeoff.md
dagar: It's a 2 second ramp up to RWTO_MAX_THR.

All throttle settings respect  [FW_THR_MIN, FW_THR_MAX]

en/flight_modes/takeoff.md
dagar: It would be 100% by default, but constrained to FW_THR_MAX (always), and RWTO_THR_MAX for runway takeoff.

For runway takeoff there's a stage called "clamped to runway" where pitch is kept at RWTO_PSP. Clamped to runway lasts until airspeed is > RWTO_AIRSPD_SCL * FW_AIRSPD_MIN.

For takeoff and climbout it's forced to >= 10 degrees pitch, but can increase beyond that.

en/flight_modes/takeoff.md
dagar: Yes, with pitch fixed, no roll, and heading hold.

en/flight_modes/takeoff.md
dagar: AGL = above ground level

Yes, forced climb out until you reach that altitude.

Once runway takeoff completes it's effectively regular waypoint navigation until the takeoff waypoint is accepted.

I'm not sure what you mean by "head up steeper".

en/flight_modes/takeoff.md
dagar: RWTO_TKOFF enables/disables runway takeoff and it defaults to off.

Not really, but the implication is either runway takeoff or hand/catapult launch (with or without launch detect).

en/flight_modes/rtl.md
dagar: Yes, it just sets a land waypoint at your home position.

hamishwillee:
Note, there may still be outstanding work to be done on top of this from closed issue https://github.com/PX4/Firmware/issues/7028

en/flight_modes/land.md
hamishwillee: @dagar 

> It seems like we should be explaining FW LAND logic somewhere, and then reference it for LAND, MISSION, and RTL.

Probably. But I'm still working out the logic. I will ping you on slack if this isn't right. Promise :-)

> This is where it gets convoluted. The autopilot likely doesn't know the actual height of the ground. When you jump into LAND as a mode it's the same landing behavior, but it internally created the land waypoint at altitude 0. As a result you probably aren't going to hit the flare logic (FW_LAND_FLALT) unless you actually get to FW_LAND_FLALT over the land altitude. If you have a distance sensor (and it's enabled), then it will be used instead.

So what I think you are saying is that (please confirm):
1. Landing logic is always the same/similar - e.g. for Mission, LAND and RTL, but there are different assumptions about where ground is in each case, and because of that flaring may or may not happen.
2. FW landing has multiple phases. The first phase is basically a fixed trajectory(FW_LND_ANG)  towards the ground. At a certain height (FW_LAND_FLALT) above the specified land altitude you'll try to flare [as per this diagram](https://github.com/PX4/Firmware/blob/master/Documentation/fw_landing.png). 
- for LAND mode you set the land altitude at 0 (sea level). Because the land is likely to be significantly above sea level you never descend to FW_LAND_FLALT so you never engage the flare logic - just proceed on your initial fixed trajectory. 
  - If you have a distance sensor this will be used. At that point the landing will do the proper flaring, because you know where land is.
- FOR RTL, you may not land because the default behaviour is to circle. However if you do land, you will use the HOME position as your landing altitude. That should be correct because you know when you took off what the height was. As a result the flaring will be above this point, and will be engaged.
- FOR Mission, you specify the height of your end landing point (????) so that will be the point that is used for ground. A proper landing with flare should happen. Correct?

So, what parameter is used to specify the LAND ALTITUDE - from which FW_LAND_FLALT is calculated.

IF THE ABOVE IS CORRECT, then I'm nearly good to document this :-) What would be great from you  [a version of  this diagram](https://github.com/PX4/Firmware/blob/master/Documentation/fw_landing.png) labeled with our parameters. 

On that note, can you confirm this definition of FW_LND_HVIRT:
```
FW_LND_HVIRT - Virtual horizontal line/altitude used to calculate the flare trajectory.

The FW_LND_HVIRT represents the notional sub-ground altitude where the vehicle would reach the desired post-flare touchdown position, if it were to maintain its initial descent trajectory. This negative altitude (expressed as a positive number) is used as part of the algorithm for calculating the flare path. 
```
If correct, then the diagram should also show the intersection of the FW_LND_HVIRT and the touchdown position.

en/flight_modes/takeoff.md
hamishwillee: > Once runway takeoff completes it's effectively regular waypoint navigation until the takeoff waypoint is accepted.

@dagar What if I have no mission? If I just select takeoff mode with no waypoints will the vehicle just loiter (circle) at the height?

PR #27, opened 2017-06-20
Document the additional radio settings

hamishwillee:
This documents additional radio settings like spektrum binding, copying trims, setting AUX channels.


hamishwillee:
@dagar @PX4/docs-team @DonLakeFlyer

This is an update to the "Additional Radio Settings" doc. I have cobbled this together using information from Don, information on various forums, and experimentation. I believe it to be correct but there are bits I can't test end-to-end (e.g. Spektrum bind). Can you please sanity check?

dagar:
This looks good, although I've never actually used the RC to param thing.
For a Spektrum transmitter all you need to do is bind in QGC, then power on your transmitter while holding the bind button down.

dagar:
For params I think the general idea is you'd use it to adjust tuning parameters.

hamishwillee:
@dagar Thanks - what do you mean by "tuning parameters"? ie a single example I can link to instead of the geofence parameter.

Thanks for the bind information - I will add that!

mhkabir:
@hamishwilee, the tuning parameters are the PID gains, for example those under Multicopter Attitude Control. "MC_ROLL_P", etc.

Did you test the radio-to-parameter functionalit though? I think it was broken at some point, and may still be so on master.

hamishwillee:
@dagar Thanks for that - I added the bind instructions and merged. You can see live [here](https://docs.px4.io/en/config/radio.html#spectrum-bind)

@mhkabir Thanks very much. 
> the tuning parameters are the PID gains, for example those under Multicopter Attitude Control. "MC_ROLL_P", etc.

I linked to the tuning guides we already have in a note [here](https://docs.px4.io/en/config/radio.html#param-tuning-channels):  "This feature is provided to enable manual in-flight tuning: Multicopter PID Tuning Guide, Fixedwing PID Tuning Guide."

> Did you test the radio-to-parameter functionality though? I think it was broken at some point, and may still be so on master.

No. Even if it is broken in master, I wouldn't "not document it" unless there was a clear statement that we weren't going to fix it. What I would do (If I could find an open issue) is provide a note/warning in the docs at this point that the functionality is broken.

FYI, this merge also included some fixes to the linked tuning docs (e..g [here](https://docs.px4.io/en/advanced_config/pid_tuning_guide_multicopter.html)). This fixes heading levels so the menu works properly and cross links to the parameter docs for all/most parameters - makes it a bit more usable.

PR #31, opened 2017-06-25
Update quick_start_pixhawk_mini.md to improve connection docs

hamishwillee:


en/assembly/quick_start_pixhawk_mini.md
```diff
@@ -130,7 +133,11 @@ external safety button, as shown below.
 
 ## Motors
 
-![Pixhawk Mini - port for motors/servos](../../images/pixhawk_mini_port_main_out.png)
+The [Airframe Reference](airframe_reference.html) shows the default output port to motor/servos mappings for common airframe types.
```
hamishwillee: @dagar This is the fix to this [discuss comment](http://discuss.px4.io/t/pixhawk-mini-pwm-on-pin-3-is-a-constant-1500-us-resolved/3656/3). I decided to put it up in the top section rather than repeating in each sub-section. There is a chance that people will jump straight to a vehicle section and miss this, but I think very repetitive to have in each section.

hamishwillee:
This is not quite ready to merge. It seems to me that this motor connection info is needed for every single flight controller and is always the same. Therefore we should have a stand alone document for this and embed it (or link to it). Seem reasonable @dagar @bkueng?

hamishwillee:
@dagar @bkueng Ignore that last note - I have given this a significant update and will merge shortly.

The change is to remove all our "standard connection diagram" and "motor" docs and replace them with links to the airframe reference (e.g. see [here](https://github.com/PX4/px4_user_guide/blob/f6836d3eab051143af04ea42bea30eebb21d59cf/en/assembly/quick_start_pixhawk.md#motors)). This is the only way to ensure that people will get an accurate map of vehicle connections and understand that it **can** differ across frames. It also means that we have only got to maintain information in one place.

PR #39, opened 2017-07-20
MISSION flight modes doc

hamishwillee:


en/flight_modes/mission.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Additional Detail
+
+Missions are usually created in [QGroundControl](https://docs.qgroundcontrol.com/en/PlanView/PlanView.html) and uploaded prior to launch (they may also be created in another GCS or by a developer API, and/or uploaded in flight). 
+
+If MISSION mode is activated when no mission is stored, the vehicle will enter [HOLD mode](../flight_modes/hold.md). 
+
+Missions can be paused by activating [HOLD mode](../flight_modes/hold.md). The mission will then continue from the current mission command when you reactivate the MISSION flight mode.
+
+Missions will complete and reset once the vehicle is disarmed. If the mission completes before disarming it will enter [HOLD mode](../flight_modes/hold.md) (in this case the mission is not reset).
```
hamishwillee: @AndreasAntener I know that if the vehicle disarms the mission will reset to the first step. Do you happen to know if there is any way to reset a mission "manually" back to start point - so for example say I finish all the steps but I'm still flying (or I've landed but not disarmed) and I want to start the mission again..

hamishwillee:
@dagar @MaEtUgR @tubeme @sanderux @
I've started working through other flight mode docs. Currently just done [Mission](https://github.com/hamishwillee/px4_user_guide/blob/b94cbb7ad89c778caeeb0a2b1cf463e48485e021/en/flight_modes/mission.md). Would any of you be able to review?

I have *assumed*:
* if there is no mission you'll go into HOLD mode - is this correct?
* if the mission ends in the air you go into HOLD mode - is that correct?
* All the vehicle types behave the same in this mode "at high level". So for example
  * if there is no mission they will all go to HOLD (and behave as they would in HOLD).
  * They will all follow the mission, though precise handling of each waypoint depends on vehicle characteristics. Is that correct - ie there is no specific mission handling at the top level?
  * VTOL has the same high level behaviour. Within the mission it obeys the characteristics of its "mode" -FW or MC.
* Normally I separate out the parameters that are relevant to each vehicle, but in this case there are quite a lot and I have just linked the the section in parameters. I think that is OK in this case. If it would benefit to separate them out, might need some help to work out which are relevant to each vehicle type. 

Questions
* Is there any other high level characteristics of MISSION mode that need to be captured (ie Pause and no mission present are covered)?
* I am still clarifying whether we can reset a mission.

sanderux:
> if there is no mission you'll go into HOLD mode - is this correct?

No, it will enter mission mode but:
- When on the ground it will refuse takeoff
- When in the air it will loiter (so no manual control)

> if the mission ends in the air you go into HOLD mode - is that correct?

It will loiter

> All the vehicle types behave the same in this mode "at high level". 

yes

> They will all follow the mission, though precise handling of each waypoint depends on vehicle characteristics. Is that correct - ie there is no specific mission handling at the top level?

yes

> VTOL has the same high level behaviour. Within the mission it obeys the characteristics of its "mode" -FW or MC.

yes

> Normally I separate out the parameters that are relevant to each vehicle, but in this case there are quite a lot and I have just linked the the section in parameters. I think that is OK in this case. If it would benefit to separate them out, might need some help to work out which are relevant to each vehicle type.

If we get this https://github.com/PX4/Firmware/issues/5920 going it would make everything much easier.

sanderux:
> Is there any other high level characteristics of MISSION mode that need to be captured (ie Pause and no mission present are covered)?

If a mission doesnt start with a takeoff command and the vehicle is landed at start, it will add a takeoff to minimum takeoff altitude

> I am still clarifying whether we can reset a mission.

You could click the first waypoint in qgc and tell it to go there, that will reset the sequence. but i'm not sure if that will also reset the loop counters

hamishwillee:
>> I am still clarifying whether we can reset a mission.

> You could click the first waypoint in qgc and tell it to go there, that will reset the sequence. but i'm not sure if that will also reset the loop counters

What's the difference?

sanderux:
If there is a jump command in the mission with a repeat x times amount set, i do not know if clicking the first waypoint will reset the counter on the jump command.

You could test that in SITL

hamishwillee:
>> If there is a jump command in the mission with a repeat x times amount set, i do not know if clicking the first waypoint will reset the counter on the jump command.

> You could test that in SITL

Thanks. Will do. 

> If a mission doesnt start with a takeoff command and the vehicle is landed at start, it will add a takeoff to minimum takeoff altitude

Is "it" PX4 or QGC?

sanderux:
>> If a mission doesnt start with a takeoff command and the vehicle is landed at start, it will add a takeoff to minimum takeoff altitude

> Is "it" PX4 or QGC? 

PX4

hamishwillee:
OK, I think the information has been integrated
@sanderux FYI, loop counters are indeed NOT reset by moving the current waypoint. This implies only way to reset mission fully is to wait for vehicle to disarm OR to re-upload a mission (which should do the same thing).

@dagar For MC in mission mode, if the vehicle mission is landed but does not have TAKEOFF defined one is added. 
1. Is this true for FW, 
1. If so, what are characteristics of the TAKEOFF? - ie is it a "Runway takeoff"? https://docs.px4.io/en/flight_modes/takeoff.html#fixed-wing-fw

en/flight_modes/mission.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Description
+
+Missions are usually created in a ground control station (e.g. [QGroundControl](https://docs.qgroundcontrol.com/en/PlanView/PlanView.html)) and uploaded prior to launch. They may also be created by a developer API, and/or uploaded in flight. Individual mission commands are handled in a way that is appropriate for each vehicle's flight characteristics (for example loiter is implemented as *hover* for copter and *circle* for fixed-wing). VTOL vehicles follow the behavior and parameters of fixed-wing when in FW mode, and of copter when in MC mode.
+
+At high level all vehicle types behave in the same way when MISSION mode is engaged:
+
+1. If a mission is stored and PX4 is flying it will execute the [mission/flight plan](../flying/missions.md) from the current step.
+1. If a mission is stored and PX4 is landed:
+   * On copters PX4 will execute the [mission/flight plan](../flying/missions.md). If the mission does not have a `TAKEOFF` command then PX4 will add one, flying the vehicle to the minimum altitude before executing the rest of the flight plan.
```
hamishwillee: @sanderux This is accurate if landed and at beginning of mission - ie PX4 takes off even if there is no TAKEOFF command and then executes rest of mission. 

What if we're landed half way through a mission, but not disarmed. The user switches from a mode back into mission. What does PX4 do - take off then execute rest of mission? Do nothing?

A minor point of clarification here too. You referred to a takeoff command being added if needed. If you download the mission after it has been added would this TAKEOFF be included in the mission listing? If not then I'd assume that you could simply say that the vehicle takes off (ie not that a command is added).

hamishwillee:
I think this is now ready to go (modulo one minor point for @sanderux . Anyone have further feedback/objection to merging? If no response, I will do so tomorrow.

en/flight_modes/mission.md
sanderux: @hamishwillee the takeoff is not added to the mission. in stead the code checks if the vehicle is in a landed state before navigating to a waypoint and will perform a takeoff if so. that means it will also happen half way through

hamishwillee: Thanks very much. Updated and I'll merge now.

PR #40, opened 2017-07-26
Remove unused FW airframes

JonReacher:
The latest version of PX4 only has 2 standard plane airframes available, Standard AETR and Bormatec Maja. This will remove confusion, especially confusing AERT with AETR, which can lead to unsafe accidental arming condition.


LorenzMeier:
Thanks!

hamishwillee:
@JonReacher  Cheers! The airframes reference is build from source and propagates here and to the DevGuide. 
1. Did you regenerate this from sources - or more specifically have the entries that generate these entries been removed? (otherwise they'd get added again next time we build this)
2. Same change needs to happen in DevGuide right?

PR #47, opened 2017-08-09
Fix copypaste error in parameter names

bgat:
The parameter names in the summary for Step 3 weren't the ones referenced in the body text. The body text references were correct.


hamishwillee:
Thanks @bgat !

PR #49, opened 2017-08-16
[WIP] Camera hardware

hamishwillee:
We need documents in user guide on "how to set up cameras". Once complete, this will be it. I've "seeded" it by copying [Camera Trigger](https://dev.px4.io/en/tutorials/camera_trigger.html) tutorial from the devguide.

This will have to cover:
1. Overview of what is covered - integration of still camera for use during missions/survey (and also available for manual triggering). Not video/FPV (link to that if we have docs)
2. Explanation of mechanisms for physically connecting cameras
   - No info in any docs, but "hints" in parameters
     1: GPIO
     2: Seagull MAP2 (over PWM)
     3: MAVLink (forward via MAV_CMD_IMAGE_START_CAPTURE)
     4: Generic PWM (IR trigger, servo)
3. Lists of known good/tested /recommended cameras
4. Explanation of how to configure cameras (ie the parameters)
5. Link to other information on *using* cameras.
6. Links to MAVLink camera API.

Note:
- How to use a camera in a mission would be a separate "mission" based document. We would link to it from here
- The camera mission interface would probably be here. 
- POSSIBLY this should also include gimbal integration.


hamishwillee:
This doesn't talk about specific hardware - users really need an example or two. 
- Do we know of any existing material on connecting cameras we can use/borrow?
- Do we know lists of cameras that will work with our mechanism(s)
- Explain mechanisms

en/peripherals/camera.md
```diff
@@ -0,0 +1,177 @@
...
+--- | ---
+[TRIG_POLARITY](../advanced_config/parameter_reference.md#TRIG_POLARITY) | Relevant only while using the GPIO interface. Sets the polarity of the trigger pin. Active high means that the pin is pulled low normally and pulled high on a trigger event. Active low is vice-versa.
+[TRIG_INTERVAL](../advanced_config/parameter_reference.md#TRIG_INTERVAL) | Defines the time between two consecutive trigger events in milliseconds.
+[TRIG_ACTIVATION_TIME](../advanced_config/parameter_reference.md#TRIG_ACTIVATION_TIME) | Defines the time in milliseconds the trigger pin is held in the "active" state before returning to neutral. In PWM modes, the minimum is limited to 40 ms to make sure we always fit an activate pulse into the 50Hz PWM signal.
+
+The full list of parameters pertaining to the camera trigger module can be found on the [parameter reference](../advanced_config/parameter_reference.md#camera-trigger) page.
+
+## Command interface 
+
+**TODO : NEEDS UPDATING updating**
```
hamishwillee: Says "needs updating" - does this mean "from the standard" (ie the specific values have changed), or does it mean that there are new commands we need to add?

en/peripherals/camera.md
```diff
@@ -0,0 +1,177 @@
...
+1 | Works like a basic intervalometer that can be enabled and disabled by using the MAVLink command `MAV_CMD_DO_TRIGGER_CONTROL`. See [command interface](camera_trigger.md#command-interface) for more details.
+2 | Switches the intervalometer constantly on.
+3 | Triggers based on distance. A shot is taken every time the set horizontal distance is exceeded. The minimum time interval between two shots is however limited by the set triggering interval.
+4 | triggers automatically when flying a survey in Mission mode.
+ 
+> **Info** If it is your first time enabling the camera trigger app, remember to reboot after changing the `TRIG_MODE` parameter.
+
+## Trigger hardware configuration
+
+You can choose which AUX pins to use for triggering using the `TRIG_PINS` parameter. The default is 56, which means that trigger is enabled on AUX 5 and AUX 6.
```
hamishwillee: So presumably we have to have a note to users that depending on their firmware some AUX pins may already be used?

Can we use a MAIN pin for Pixhawk Mini? or anything else that doesn't have an IO board/AUX?

LorenzMeier:
@hamishwillee Can we revive this or close it?

hamishwillee:
We should revive this. It needs to be made "approach agnositic" - ie to also cover camera manager, and be less driver-centric. There are questions inline too. I just need to find the time to chase people up.

en/peripherals/camera.md
hamishwillee: @bkueng Do you happen to know the answer to this? Also, who are good people to talk about using Camera? I'm pretty sure I understand this, but was hoping to talk to someone who has recent experience.

en/peripherals/camera.md
```diff
@@ -0,0 +1,186 @@
...
+
+[MAV_CMD_DO_TRIGGER_CONTROL](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_TRIGGER_CONTROL) - Accepted in "command controlled" mode (`TRIG_MODE` 1).
+
+Command Parameter | Description
+--- | ---
+Param #1 | Trigger enable/disable (set to 0 for disable, 1 for start)
+Param #2 | Trigger cycle time in milliseconds (sets the `TRIG_INTERVAL` parameter.)
+Param #3 | Sequence reset (set to 1 to reset image sequence number, 0 to keep current sequence number)
+
+[MAV_CMD_DO_DIGICAM_CONTROL](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_DIGICAM_CONTROL) - Accepted in all modes. This is used by the GCS to test-shoot the camera from the user interface. The trigger driver does not yet support all camera control parameters defined by the MAVLink spec.
```
hamishwillee: FYI This is deprecated. Should remove. Also need to add info about the MAVLink Camera API.

en/peripherals/camera.md
bkueng: > Can we use a MAIN pin for Pixhawk Mini? or anything else that doesn't have an IO board/AUX?

That should work too. It should automatically select the MAIN pins 5 and 6 on boards w/o IO.
However, there is a limitation: the other (aux) pins can only be used for actuators/mixer outputs if TRIG_PINS is 56 (see https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/init.d/rcS#L454).

>  Also, who are good people to talk about using Camera?

I'm not sure if @dogmaphobic has experience?

hamishwillee:
I have merged this, even though it is not as complete as I would like. Will add issue for further updates.

PR #64, opened 2017-09-20
mention the use of a telem port and capacitor for llv3 on aero

ChristophTobler:


en/flight_controller/intel_aero.md
```diff
@@ -67,19 +67,28 @@ After setting up the PX4 development environment, follow these steps update the
...
-| 2   | SCL        | - (Power enable) |
-| 3   | SDA        | - (Mode control) |
-| 4   | GND        | SCL              |
-| 5   | -          | SDA              |
-| 6   | -          | GND              |
+| pin | Aerofc TELEMETRY | Lidar Lite V3    |
+| --- | ---------------- | ---------------- |
+| 1   | VCC              | VCC              |
+| 2   | TX               | - (Power enable) |
+| 3   | RX               | - (Mode control) |
```
zehortigoza: connecting 'mode control' to RX port could cause it change to PWM mode maybe is better advise user to leave this unconnected.

ChristophTobler: Ah no, - should mean that it is not being used.

en/flight_controller/intel_aero.md
```diff
@@ -67,19 +67,28 @@ After setting up the PX4 development environment, follow these steps update the
 
 ## Connecting a Lidar Lite range finder
 
-The following instructions are for a Lidar Lite V3 connected via I2C. The I2C port on the Aero (labled compass) is used for the external magnetometer (part of the GPS). Therefore a I2C splitter has to be used to connect the Lidar Lite (see picture).
-
-![](../../assets/hardware/Aero_I2C_splitter.JPG)
+The following instructions are for a Lidar Lite V3 connected via I2C. The I2C port on the Aero (labled compass) is used for the external magnetometer (part of the GPS). Therefore, either a I2C splitter has to be used to connect the Lidar Lite (see picture) or the Telemetry port can be used (recommended as it is not being used). In addition it is recommended to use a electrolytic capacitor to reduce spikes in the distance readings (see [here](https://static.garmin.com/pumac/LIDAR_Lite_v3_Operation_Manual_and_Technical_Specifications.pdf) on page 3).
```
hamishwillee: This is all confusing. 
1. It says "The following instructions are for a Lidar Lite V3 connected via I2C." but then your pinouts cover both using telemetry and i2c ports
2. Is the capacitor used in both cases or only with I2C?
3. I would remove "(Power enable)" and "(Mode Enable) from the table - makes it obvious "not connected". ONly leave if it is useful for someone to know this when making this particular set of connections.
4. In section title use first letter capitalisation "Connecting a Lidar Lite Range Finder"

I can help you reword if you can confirm the above (that this is for both cases).

ChristophTobler: @hamishwillee 
1. Both ports support I2C
2. Yes
3. done
4. done
Could you please check if the new commit is more clear?

hamishwillee:
This is good. I have merged. Thank you.

One side point, this wording is legitimate English, but "clunky": " it is recommended to use ...". It is usually better to write something like "Intel recommend a capacitor is used ..."

PR #65, opened 2017-09-21
Create standalone docs with info on offboard mode

hamishwillee:


LorenzMeier:
@hamishwillee Could you identify a sparring partner for review? I stop scaling and I need to hand off non-core repos like docs to a submodule maintainer.

hamishwillee:
Sure @LorenzMeier , I was going to ask Julian. The only thing I need your help on is https://github.com/PX4/px4_user_guide/pull/62

en/flight_modes/offboard.md
```diff
@@ -0,0 +1,42 @@
...
+>  * This mode is automatic (RC control is disabled [by default](../advanced_config/parameter_reference.md#COM_RC_OVERRIDE) except to change modes).
+>  * The vehicle must be armed before this mode can be engaged.
+
+## Description
+
+Offboard mode is primarily used for controlling vehicle movement and attitude, and supports only a very limited set of MAVLink commands (more may be supported in future). It can be used to:
+* Control vehicle position or velocity (`SET_POSITION_TARGET_LOCAL_NED`). 
+* Control vehicle attitude/orientation (`SET_ATTITUDE_TARGET`)
+
+> **Note** Velocity and attitude commands must be resent regularly at a rate of around (TBD Hz) or PX4 will stop the vehicle.
```
hamishwillee: @julianoes Can you confirm the update rate, and which messages it applies to. I would guess velocity and attitude, but not position, on the basis that if a vehicle has been told to go to a location that is OK, but if it is being told to move "forward" you probably want your code to actively command that on a regular basis or stop. (i.e. more defensive approach).

hamishwillee:
@julianoes I've selected you as reviewer since you actually use this mode. @TSC21 , I suspect you do too. Would appreciate your comments. 

I am aware that this mode, unlike the others is primarily for developers. I'm grouping it here anyway.

en/flight_modes/offboard.md
julianoes: The timeout is 0.5 seconds according to:
https://github.com/PX4/Firmware/blob/master/src/modules/commander/commander.cpp#L151

Therefore the update rate needs to be > 2 Hz.

en/flight_modes/README.md
```diff
@@ -73,8 +73,6 @@ Throttle Command is mapped direct to Motor Speed.
...
 - [Mission](../flight_modes/mission.md): The vehicle follows a programmed mission, which is usually planned and uploaded using a ground control station (GCS).
 - [Follow Me](../flight_modes/follow_me.md) (Multicopter-only): The vehicle autonomously follows a user with an Android phone/tablet running *QGroundControl*.
-- **Offboard** In this mode the position, velocity or attitude
-  reference / target / setpoint is provided by a companion computer
-  connected via serial cable and MAVLink. The offboard setpoint can be
-  provided by APIs
-  like [MAVROS](https://github.com/mavlink/mavros) or [Dronekit](http://dronekit.io/).
+- [Offboard](../flight_modes/offboard.md): The vehicle obeys a position, velocity or attitude
+  setpoint provided over MAVLink (often from a companion computer connected via serial cable). The setpoint can be
+  provided by APIs like [DroneCore](http://dronecore.io/) or [MAVROS](https://github.com/mavlink/mavros).
```
julianoes: Nice. One addition: I think acceleration is in theory also possible but in reality not really supported.

en/flight_modes/offboard.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Offboard Parameters
+
+*Offboard mode* is affected by the following parameters:
+
+Parameter | Description
+--- | ---
+<span id="COM_OF_LOSS_T"></span>[COM_OF_LOSS_T](../advanced_config/parameter_reference.md#COM_OF_LOSS_T) | Time-out (in seconds) to wait when offboard connection is lost before triggering offboard lost failsafe (`COM_OBL_ACT` and `COM_OBL_RC_ACT`)
+<span id="COM_OBL_ACT"></span>[COM_OBL_ACT](../advanced_config/parameter_reference.md#COM_OBL_ACT) | Mode to switch to if offboard control is lost when *not* connected to RC (Values are - 0: [Land](../flight_modes/land.md), 1: [Hold](../flight_modes/hold.md), 2: [Return ](../flight_modes/rtl.md)).
+<span id="COM_OBL_RC_ACT"></span>[COM_OBL_RC_ACT](../advanced_config/parameter_reference.md#COM_OBL_RC_ACT) | Mode to switch to if offboard control is lost while still connected to RC control (Values are - 0: *Position*, 1: [Altitude](../flight_modes/altitude.md), 2: *Manual*, 3: [Return ](../flight_modes/rtl.md), 4: [Land](../flight_modes/land.md)).
```
julianoes: I would need to look at the code and play with this to confirm. I'll let you know once I get the chance.

en/flight_modes/offboard.md
```diff
@@ -0,0 +1,42 @@
+# Offboard Flight Mode
+
+The vehicle obeys a position, velocity or attitude setpoint provided over MAVLink (often from a companion computer connected via serial cable or wifi). The setpoint can be provided by APIs like [DroneCore](http://dronecore.io/) or [MAVROS](https://github.com/mavlink/mavros).
+
+> **Note** 
+>  * This mode requires position information (GPS or optical flow).
```
TSC21: I will add also **visual** information - Visual-Inertial Odometry or Motion Capture.

en/flight_modes/offboard.md
```diff
@@ -0,0 +1,42 @@
+# Offboard Flight Mode
+
+The vehicle obeys a position, velocity or attitude setpoint provided over MAVLink (often from a companion computer connected via serial cable or wifi). The setpoint can be provided by APIs like [DroneCore](http://dronecore.io/) or [MAVROS](https://github.com/mavlink/mavros).
+
+> **Note** 
+>  * This mode requires position information (GPS or optical flow).
+>  * This mode is automatic (RC control is disabled [by default](../advanced_config/parameter_reference.md#COM_RC_OVERRIDE) except to change modes).
+>  * The vehicle must be armed before this mode can be engaged.
```
TSC21: Not only that - to change to this mode, a stream of setpoint msgs should be sent prior to changing the mode.

en/flight_modes/offboard.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Description
+
+Offboard mode is primarily used for controlling vehicle movement and attitude, and supports only a very limited set of MAVLink commands (more may be supported in future). It can be used to:
+* Control vehicle position or velocity (`SET_POSITION_TARGET_LOCAL_NED`). 
+* Control vehicle attitude/orientation (`SET_ATTITUDE_TARGET`)
+
+> **Note** Velocity and attitude commands must be resent regularly at a rate of around (TBD Hz) or PX4 will stop the vehicle. 
+
+Other operations, like taking off, land, return to launch, arming, are best handled using the appropriate modes. Operations like uploading, downloading missions can be performed in any mode.
```
TSC21: Takeoff and land is also possible with the appropriate bitmasks.

en/flight_modes/offboard.md
```diff
@@ -0,0 +1,42 @@
...
+
+Offboard mode requires an active connection to a remote system. If the connection is lost the vehicle will attempt to land or perform some other failsafe action (after a timeout ([COM_OF_LOSS_T](#COM_OF_LOSS_T)). The action is defined in the parameters [COM_OBL_ACT](#COM_OBL_ACT) and [COM_OBL_RC_ACT](#COM_OBL_RC_ACT).
```
TSC21: As I said above, that same connection is established when sending that same setpoint stream.

en/flight_modes/README.md
hamishwillee: What do you mean by "not really"? It looks like it is [here](https://github.com/PX4/Firmware/blob/master/src/modules/mavlink/mavlink_receiver.cpp#L939)

en/flight_modes/offboard.md
hamishwillee: Isn't this ultimately just position information - albeit obtained from an inertial source?
If so, would the following text be better?:

> * This mode requires position information - e.g. GPS or intertial (optical flow/visual-intertial odometry, etc.)

en/flight_modes/offboard.md
hamishwillee: Can you please expand on this/why? 
Are you saying that PX4 needs to have already received setpoint messages before changing to the mode? Are those messages ignored - ie if received before the vehicle enters the mode?
I *guess* what you are saying is that offboard mode needs an active connection so you need to send messages before you switch to offboard mode or you'll immediately timeout - and the messages are ignored. Correct?

en/flight_modes/offboard.md
hamishwillee: Thanks you. I'll fix that!

en/flight_modes/offboard.md
hamishwillee: OK, see request for clarification above. What if you're in offboard most but you don't want to go anywhere - how do you keep the offboard mode alive? - Just send setpoints to move to the current position?

en/flight_modes/offboard.md
hamishwillee: @julianoes recommended just using the supported modes for that. I tend to agree - lots of thought has gone into landing/takeoff on various vehicles, so while you can manually reinvent the wheel by sending appropriate setpoints, easier to just invoke "takeoff" and "land" mode.
Do both of you think perhaps this would be better as:
> other operations, like taking off, land, return to launch, arming, are best handled using the appropriate modes (you can takeoff and land using setpoint commands, but the respective modes are safer and easier to use).

en/flight_modes/offboard.md
hamishwillee: No urgency. These are the important ones I can see that most users (ie not developers) would actually want to play with.

en/flight_modes/README.md
hamishwillee: Ping @julianoes

en/flight_modes/offboard.md
hamishwillee: Pint @TSC21

en/flight_modes/offboard.md
hamishwillee: Ping @TSC21

en/flight_modes/offboard.md
TSC21: Not just position. It can include orientation. Visual or Mocap are not inertial sources (inertial sources integrate IMU data). So it will be **pose**  instead of position:
> **Note** 
>  * This mode requires position or pose information -  e.g. GPS, optical flow, visual-inertial odometry, mocap, etc.

en/flight_modes/offboard.md
TSC21: > I guess what you are saying is that offboard mode needs an active connection so you need to send messages before you switch to offboard mode or you'll immediately timeout - and the messages are ignored. Correct?

That is correct. To change to this mode there needs to be an active setpoint msg stream.

en/flight_modes/offboard.md
TSC21: Yes agreed

en/flight_modes/offboard.md
TSC21: if you don't want to go anywhere you need to send a setpoint stream with the current position of the vehicle.

en/flight_modes/README.md
TSC21: @hamishwillee it is handled on the mavlink receiver but then it is not used in any controllers.

hamishwillee:
Note to me. FW not supported.

hamishwillee:
@dagar @TSC21 Thanks for your feedback! I think this is just about done now - I added the information about stream required for moving into mode and remaining in mode, and the fact that the mode isn't supported for Fixed Wing.

Last question/confirmation around the time out/what is an active connection. My understanding is that the 2Hz stream rate is what defines an "active connection". If the stream rate drops below 2Hz then the vehicle will stop obeying commands and start the timeout for loss of active connection. However it won't leave the mode/failsafe until the timeout expires (if the rate goes up again the timeout resets and the vehicle will start obeying setpoints again). Is that correct?

hamishwillee:
@TSC21 @julianoes , there are some good questions [here](https://github.com/PX4/Firmware/issues/8021#issuecomment-335457882) about PX4 offboard support. Do you happen to know the answers to:

- Does PX4 support all the coordinate frames in [SET_POSITION_TARGET_LOCAL_NED](http://mavlink.org/messages/common#SET_POSITION_TARGET_LOCAL_NED), and if not, which are supported/not supported?
- Regarding the `type_mask` in SET_POSITION_TARGET_LOCAL_NED I know that we don't support acceleration (afx, afy, afz) in any controller. 
  - Do we support all the other options?
  - What combinations are supported for a particular frame - ie can we specify a position and velocity at the same time?
  - Are there any type masks that aren't supported for particular frames?
- Are there any other limitations? For example, can we just specify vx, and not vy, vz?

julianoes:
> Does PX4 support all the coordinate frames in SET_POSITION_TARGET_LOCAL_NED, and if not, which are supported/not supported?

For velocity setpoints, the coordinate frame is copied here:
https://github.com/PX4/Firmware/blob/master/src/modules/mavlink/mavlink_receiver.cpp#L996-L997

The two frames supported are `MAV_FRAME_LOCAL_NED` and `MAV_FRAME_BODY_NED` according to https://github.com/PX4/Firmware/blob/8a0a8e20e1ce1a764300c99d1a28fbe6db15cfae/src/modules/mc_pos_control/mc_pos_control_main.cpp#L1592-L1604, https://github.com/PX4/Firmware/blob/22733b150172dba6a09c91c4b44c97b65ef39031/msg/position_setpoint.msg#L12-L13.

julianoes:
> - Regarding the type_mask in SET_POSITION_TARGET_LOCAL_NED I know that we don't support acceleration (afx, afy, afz) in any controller.
>     - Do we support all the other options?
>     - What combinations are supported for a particular frame - ie can we specify a position and velocity at the same time?
>     - Are there any type masks that aren't supported for particular frames?

We don't support acceleration as such but we use these acceleration setpoints like a thrust setpoint, see https://github.com/PX4/Firmware/blob/8a0a8e20e1ce1a764300c99d1a28fbe6db15cfae/src/modules/mc_pos_control/mc_pos_control_main.cpp#L2540.

I think we support the rest but there might be edge cases with various combinations. This needs proper testing to be sure. We could do that step by step using DroneCore.

hamishwillee:
@julianoes Thanks for those clarifications - I have updated and merged. 

FYI, with respect to "We don't support acceleration as such but we use these acceleration setpoints" I more or less just paraphrased this, though I don't quite understand the difference - presumably constant thrust maps to constant acceleration initially but then gets capped as the motors reach max thrust?

> I think we support the rest but there might be edge cases with various combinations. This needs proper testing to be sure. We could do that step by step using DroneCore.

Agree!

PR #68, opened 2017-09-27
aero: add instructions to use optical flow

ChristophTobler:
@zehortigoza @lucasdemarchi  is v1.6 or higher correct?
We also need to check the PX4 side and the default parameters like `EKF2_AID_MASK`,  `EKF2_HGT_MODE` and `EKF2_RNG_AID` for the next release.

### DIFF
--- a/en/flight_controller/intel_aero.md
+++ b/en/flight_controller/intel_aero.md
@@ -93,3 +93,16 @@ The pinout for the Lidar Lite V3 is as follows
 ![](../../assets/hardware/Aero_I2C_splitter.JPG)
 
 ![](../../assets/hardware/Aero_LidarLite.JPG)
+
+## Using Optical Flow on the Aero
+
+The Intel Aero comes with a preinstalled optical flow binary on the compute board (Linux OS version 1.6 or higher) which enables the Aero to stably fly based on optical flow velocity estimation. In order to use optical flow, a range sensor has to be installed first (see above).
+
+To use the optical flow, run the following command in a console on the Aero compute board:
+```
+systemctl start aero-optical-flow
+```
+If you want to start the optical flow binary at boot, use
+```
+systemctl enable aero-optical-flow #use disable to undo
+```
--- a/en/flight_controller/intel_aero.md
+++ b/en/flight_controller/intel_aero.md
@@ -96,7 +96,7 @@ The pinout for the Lidar Lite V3 is as follows
 
 ## Using Optical Flow on the Aero
 
-The Intel Aero comes with a preinstalled optical flow binary on the compute board (Linux OS version 1.6 or higher) which enables the Aero to stably fly based on optical flow velocity estimation. In order to use optical flow, a range sensor has to be installed first (see above).
+The Intel Aero comes with a preinstalled optical flow binary on the compute board (Linux OS version 1.6 or higher), which enables the Aero to stably fly based on optical flow velocity estimation. In order to use optical flow, a range sensor has to be installed first (see above).
 
 To use the optical flow, run the following command in a console on the Aero compute board:
 ```
@@ -106,3 +106,10 @@ If you want to start the optical flow binary at boot, use
 ```
 systemctl enable aero-optical-flow #use disable to undo
 ```
+
+In addition, the following parameter values should be set in the flight controller.
+
+| Parameter     | Value |
+| ------------- | ----- |
+| EKF2_AID_MASK | 2     |
+| EKF2_HGT_MODE | 2     |


lucasdemarchi:
If we change `EKF2_AID_MASK` to 3 and the user didn't add the lidar or didn't start the optical flow binary, what happens?  Same for the other parameters... There isn't an option "use rangefinder if you have, otherwise fallback to baro".

So I think we should add in the instructions what parameters need to change.

zehortigoza:
there is not date to release 1.6 yet maybe @lfelipe has more information about that.

As Lucas said, I think is better add instructions to user set those parameters than enable by default.

ChristophTobler:
@lucasdemarchi 
>I would not fly like this as it will stop when you loose connection :)

You are right! Thanks :)

@lucasdemarchi @zehortigoza 
I my opinion, the goal has to be chaning no parameters. In the case of `EKF2_AID_MASK`, there is no issue if you have enabled flow as well and just use GPS with no lidar attached/not running the optical flow binary. You can also use optical flow with `EKF2_HGT_MODE` set to baro (as long as you have a lidar estimating the terrain). But performance will not be as good. I have to check if we can get range aid (`EKF2_RNG_AID`) in a shape where we can have it as a default.

en/flight_controller/intel_aero.md
```diff
@@ -93,3 +93,16 @@ The pinout for the Lidar Lite V3 is as follows
 ![](../../assets/hardware/Aero_I2C_splitter.JPG)
 
 ![](../../assets/hardware/Aero_LidarLite.JPG)
+
+## Using Optical Flow on the Aero
+
+The Intel Aero comes with a preinstalled optical flow binary (OS version 1.6 or higher) which enables the Aero to stably fly based on vision. In order to use optical flow, a range sensor has to be installed first (see above).
```
hamishwillee: This is the user guide - if you say "OS" users are likely to assume PX4 version (we're currently on 1.6.x) Is this what you mean? In that case I'd change to "PX4 v1.6". If not, specify the OS.

en/flight_controller/intel_aero.md
```diff
@@ -93,3 +93,16 @@ The pinout for the Lidar Lite V3 is as follows
...
+
+To use the optical flow, just run
+```
+systemctl start aero-optical-flow
+```
+on the Aero. If you want to start the optical flow binary at boot, use
```
hamishwillee: The Aero has a NuttX and compute board right? So shouldn't we say NuttX?

en/flight_controller/intel_aero.md
hamishwillee: - which should be "that"
- the use of "vision" seems imprecise. What does this actually provide - complete replacement to GPS?, velocity from optical flow.?

en/flight_controller/intel_aero.md
```diff
@@ -93,3 +93,16 @@ The pinout for the Lidar Lite V3 is as follows
...
+```
+systemctl enable aero-optical-flow #use disable to undo
+```
```
hamishwillee: Really depends on your responses above, but perhaps could re-write this similar to below:
----

## Using Optical Flow on the Aero

> **Note** In order to use optical flow, a range sensor must be installed (see above).

The Intel Aero comes with a preinstalled optical flow binary (PX4 v1.6 or higher) that enables the Aero to fly using optical flow for navigation rather than GPS.

To use the optical flow run the following command on PX4 console:
 ```
 systemctl start aero-optical-flow
 ```
If you want to start the optical flow binary at boot, use
 ```
 systemctl enable aero-optical-flow #use disable to undo
 ```

en/flight_controller/intel_aero.md
ChristophTobler: No, it's the Linux verison

hamishwillee:
@ChristophTobler Yes, much better thank you. 

As you mentioned up the top, if parameters have to be set to get this to work properly they should also be explained/listed.

ChristophTobler:
@hamishwillee Yes. I'm hoping that the user doesn't need to change anything with the new release. But I can add it for now.

en/flight_controller/intel_aero.md
```diff
@@ -93,3 +93,23 @@ The pinout for the Lidar Lite V3 is as follows
 ![](../../assets/hardware/Aero_I2C_splitter.JPG)
 
 ![](../../assets/hardware/Aero_LidarLite.JPG)
+
+## Using Optical Flow on the Aero
+
+The Intel Aero comes with a preinstalled optical flow binary on the compute board (Linux OS version 1.6 or higher) which enables the Aero to stably fly based on optical flow velocity estimation. In order to use optical flow, a range sensor has to be installed first (see above).
```
hamishwillee: Make this
" higher) that enables the Aero to stably fly based on optical flow velocity estimation."  (which change to that, OR have a comma immediately before "which".

en/flight_controller/intel_aero.md
```diff
@@ -93,3 +93,23 @@ The pinout for the Lidar Lite V3 is as follows
...
+To use the optical flow, run the following command in a console on the Aero compute board:
+```
+systemctl start aero-optical-flow
+```
+If you want to start the optical flow binary at boot, use
+```
+systemctl enable aero-optical-flow #use disable to undo
+```
+
+In addition, check if the following two parameters are set correctly on the flight controller.
```
hamishwillee: So Aero refers to the NuttX part as the "Flight controller"? If so, cool. 
I would probably word this as:

"In addition, the following parameter values should be set in the flight controller.

hamishwillee:
That works for me. I have proposed 2 minor English changes. Up to you if you want to include. Happy for this to be merged otherwise.

ChristophTobler:
@hamishwillee l'll merge it and adapt the parameter situation when needed.

hamishwillee:
THanks @ChristophTobler !

PR #74, opened 2017-10-11
Add VTOL Transition Testing doc from DevGuide

hamishwillee:
This pulls in the VTOL testing doc from the DevGuide. I have done very basic update just for style. 

I see this as complementary to the [Back Transition Tuning](https://docs.px4.io/en/config_vtol/vtol_back_transition_tuning.html) doc. What I like about it is that it has specific instructions for checking that "the basics" are working on the bench and for running in flight tests. What I don't like about it is that it appears to be out of date - e.g. `VT_ARSP_BLEND` is not in the back transition guide, and that guide also has other behaviours that aren't covered in the testing.

@sanderux Would you be able to help with review? How out of date is this doc? Is it useful? How can we usefully include it? Are there any "clearly wrong" bits - e.g. is transition switch still set on AUX1 ...

If it is kind of useful but will require more work than we have time for we could add a note up the top stating the main flaws and requesting help in update. We should also link to the back transition doc.


sanderux:
This is pretty out of date.
AUX1 is no longer in use and the preferred method (and default by QGC) is VTOL_TAKEOFF and VTOL_LAND.

hamishwillee:
@sanderux  Before we talk about the specific content ...
1. Is it useful to have "how to test" information for VTOL transitions?
2. If it is useful, should it be in a standalone doc as done here or as part of https://docs.px4.io/en/config_vtol/vtol_quad_configuration.html (say).

My feeling is that it is useful to have this "kind" of document, and to have it in a separate doc. If you agree then lets fix the doc, if not, let me know and we can just delete it and move on. Thoughts?

> AUX1 is no longer in use and

According to the docs here https://docs.px4.io/en/config_vtol/vtol_quad_configuration.html#flight--transition-mode-switch you can assign the transition channel to any switch using RC_MAP_TRANS_SW - so I can certainly replace that bit about AUX. 

>  "the preferred method (and default by QGC) is VTOL_TAKEOFF and VTOL_LAND."

Are you saying the preferred method for testing is to create a mission that uses  http://mavlink.org/messages/common#MAV_CMD_NAV_VTOL_TAKEOFF and MAV_CMD_NAV_VTOL_LAND , presumably with some waypoint in the middle - ie the current mission is wrong?

Is the "manual inflight" section OK?

hamishwillee:
I am closing this because it is very out of date and the relevant material is covered in  https://docs.px4.io/en/config_vtol/vtol_quad_configuration.html.

At some point it may be worth changing the https://docs.px4.io/en/config_vtol/vtol_quad_configuration.html to use a similar style to this document, as that is easier to follow.

Issue #75, opened 2017-10-11
Multicopter PID Tuning Guide review


nicolaerosia:
Hello,

I'm trying to understand the [Multicopter PID Tuning Guide](https://docs.px4.io/en/advanced_config/pid_tuning_guide_multicopter.html) and I have some questions:

**"Set all MC_XXX_P to zero (ROLL, PITCH, YAW)"** if these parameters do what they advertise, I don't understand how this will work since they will output 0.
_"The gains actually have an intuitive meaning, e.g.: if the MC_ROLL_P gain is 6.0, the copter will try to compensate 0.5 radian offset in attitude (~30 degrees) with 6 times the angular speed, i.e. 3 radians/s or ~170 degrees/s."_ working this example and replacing 6 with 0, it will output 0 degrees/s.

**"Keep the multi rotor in your hand and increase the thrust to about 50%, so that the weight is virtually zero. "**
In which mode? Manual Stabilized?

**I Gain Tuning
If the roll and pitch rates never reach the setpoint but have an offset, add MC_ROLLRATE_I and MC_PITCHRATE_I gains, starting at 5-10% of the MC_ROLLRATE_P gain value.**
How can I see the roll and pitch rates? what about setpoint? Are we still in the situation where I keep the drone in my hand, titled, with 50% throttle ?

**Feed Forward Tuning
Typical value is 0.8…0.9. (For aerial video optimal value may be much smaller to get smooth response.)**
If typical value is 0.8 - 0.8, why the default is 0.5 ?

@hamishwillee @dagar

Thanks,
Nicolae

hamishwillee:
@nicolaerosia 

>> "Set all MC_XXX_P to zero (ROLL, PITCH, YAW)" if these parameters do what they advertise, 

> I don't understand how this will work since they will output 0.

This is a *preparation step* ("Step 1: Preparation") - so essentially you're setting everything to a nothing state before you start tuning each of the individual values. So yes, you're right, but it is intentional.

The rest of the questions seem reasonably and I certainly can't answer. While I hope @dagar will answer them, can you please raise them here: http://discuss.px4.io/c/multicopter
Then we can update the docs as we get answers.

dagar:
> "Keep the multi rotor in your hand and increase the thrust to about 50%, so that the weight is virtually zero. "
> In which mode? Manual Stabilized?
Yes manual/stabilized

[...]

PR #78, opened 2017-10-16
Tidy up MC PID tuning doc

hamishwillee:
This attempts to improve the PID tuning doc in line with  #75 .

This tries to add more structure to the doc - separating actual tuning instructions from "this is how tuning/architecture works". I've moved some of the warnings around to where they are more relevant. It also makes it clear that the guide should not be "necessary" if you're using a supported configuration - it is more for modified vehicles and new airframes.


hamishwillee:
There is some outstanding stuff to be done - specifically it is not clear whether vehicle is held for all steps or not.

hamishwillee:
@dagar Decided merge this, it is IMO an improvement.

Open question - "is vehicle is held for all steps or not. If not held for some of the testing, in which steps is it held in some other way?"

PR #79, opened 2017-10-16
Move VTOL to advanced config. Make clear tuning is for new airframes

hamishwillee:
@sanderux @stifael FYI, this is the result of [recent discussion](https://px4.slack.com/archives/C581ST17V/p1507761377000126) we had about whether tuning is "advanced config". 

Essentially the baseline user is someone using a canned/standard airframe config. For this user tuning is not  *required*. The change is to make it clear when you need to tune, when you don't need to tune, and move VTOL into that "advanced" section.


hamishwillee:
@sanderux Ping

hamishwillee:
Screw it, I'm happy with this. Merging :-)

PR #80, opened 2017-10-17
update airframe reference and add Quad H

bkueng:
Firmware PR: https://github.com/PX4/Firmware/pull/8143
@hamishwillee

### DIFF
--- a/en/airframes/airframe_reference.md
+++ b/en/airframes/airframe_reference.md
@@ -182,6 +182,11 @@ This page lists all supported airframes and types including
  <td style="vertical-align: top;"><p>Maintainer: Lorenz Meier <lorenz@px4.io></p><p><code>SYS_AUTOSTART</code> = 6001</p></td>
 
 </tr>
+<tr id="copter_hexarotor_x_generic_hexarotor_x_geometry">
+ <td style="vertical-align: top;">Generic Hexarotor x geometry</td>
+ <td style="vertical-align: top;"><p>Maintainer: Lorenz Meier <lorenz@px4.io></p><p><code>SYS_AUTOSTART</code> = 6001</p></td>
+
+</tr>
 </tbody></table>
 
 ### Octo Coax Wide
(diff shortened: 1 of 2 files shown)


en/airframes/airframe_reference.md
```diff
@@ -401,6 +406,25 @@ This page lists all supported airframes and types including
 </tr>
 </tbody></table>
 
+### Quadrotor h
```
hamishwillee: This is consistent, but should it be capital "H" reflecting the shape?

en/airframes/airframe_reference.md
```diff
@@ -401,6 +406,25 @@ This page lists all supported airframes and types including
...
+<img src="../../assets/airframes/types/QuadRotorH.svg" width="29%" style="max-height: 180px;"/>
+</div>
+
+<table style="width: 100%; table-layout:fixed; font-size:1.5rem;">
+ <colgroup><col style="width: 30%"><col style="width: 70%"></colgroup>
+ <thead>
+   <tr><th>Name</th><th></th></tr>
+ </thead>
+<tbody>
+<tr id="copter_quadrotor_h_reaper_500_quad">
```
hamishwillee: There are no "common" outputs here. That is correct based on the [config file](https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/init.d/4040_reaper), but given this is the only case, should we perhaps update that init file and regenerate?

en/airframes/airframe_reference.md
bkueng: I was going for consistency, but you're right, H makes more sense.

en/airframes/airframe_reference.md
bkueng: Yep, I added the outputs

hamishwillee:
Great. I have approved the other request. Will check this again once you've merged in the updates.

hamishwillee:
Thanks @bkueng !

PR #95, opened 2017-11-14
Remove LNDMC_POS_UPTHR parameter from docs

hamishwillee:


hamishwillee:
FYI The removal of the parameter and that the throttle position is hard coded to 62.5% was confirmed by @julianoes

PR #98, opened 2017-11-17
added documentation of precision land with IRLock

ndepal:
as per https://github.com/PX4/Devguide/pull/348#issuecomment-345101317


hamishwillee:
This is excellent. I will merge then improve the language a bit in a separate PR.

en/advanced_features/precland.md
```diff
@@ -0,0 +1,114 @@
+# Precision Landing
+
+PX4 supports precision landing for multicopters with the [IRLock](https://irlock.com/products/ir-lock-sensor-precision-landing-kit) sensor, an IRLock beacon, and a downward facing range sensor.
+This enables landing with a precision of roughly 10 cm, as opposed to the precision given by GPS, which may be several meters.
+
+A precision landing can be initiated by entering the Precision Land flight mode, or as part of a mission uploaded via MAVLink.
```
hamishwillee: Is this actually a separate flight mode? And if you query the mode over mavlink, will it return PRECLAND?

ndepal: Precision land is a separate flight mode. It is entered only when doing an explicit mode switch. When doing a precision landing during a mission, it stays in mission mode.

I'm not sure what you mean by querying over mavlink. `commander_state.main_state` reflects the precision land state and that's what's used in some mavlink messages, right?

I also see that QGC does display a different flight mode, but it does not know what to call it: It states "Unknown 129:0".

hamishwillee: You have answered my question thanks. 
I'm not sure it should be separate though. When selecting a flight mode switch I expect people would just want to select LAND, and precision land to be used by default if some parameter is set.

PR #101, opened 2017-11-20
Improve introduction page. Move flight reporting into body of docs

hamishwillee:


en/README.md
```diff
@@ -1,15 +1,13 @@
+<div style="float:right; padding:10px; margin-right:20px;"><a href="http://px4.io/"><img src="../assets/site/logo_pro_small.png" title="PX4 Logo" width="180px"/></a></div>
```
hamishwillee: This adds PX4 logo and link to site to front page. Provides context.  The remainder of the changes aim to make it look just a bit more professional. 

Ultimately I hope to create a more detailed use-case based entry point here, but the rest of the organisation still needs too much work for that.

en/flying/flight_reporting.md
```diff
@@ -0,0 +1,16 @@
+# Flight Reporting
```
hamishwillee: @bkueng So I moved flight reporting under flying and updated it to current QGC. I think this is not a top level topic, and that this is the right place - OK?

One question, should we better highlight on front page "how to raise issues"? We could then link to this document?

bkueng: I'm ok with that.

> One question, should we better highlight on front page "how to raise issues"? We could then link to this document?

Sure.

hamishwillee: What is our position on that then? Just raise an issue if you see a problem, or do we push them to ask on slack or whatever first for some pre-triaging?

bkueng: If it's a general usability issue, I'd say it should first go through discuss/slack.
If someone has a log with a specific issue, it can go to github. But I'm not quite sure if there's a better way. Sometimes it's just a setup problems (e.g. vibrations), then it does not need to go to github.

hamishwillee: Thanks very much. I added the note in https://github.com/PX4/px4_user_guide/pull/102

PR #104, opened 2017-11-23
Move sfxx lidar and uLanding radar sensors from DevGuide

hamishwillee:
User guide is for "supported hardware" while devguide is for integrating new hardware - so this is right place for these docs. I have updated them to be readable, but there are open questions inline.

@bkueng  I've put you as reviewer, but if there is someone else you'd recommend, then please hook them in to help answer the questions.


en/sensor/sfxx_lidar.md
```diff
@@ -0,0 +1,43 @@
+# LightWare SF1XX Lidar
+
+LightWare develops a range of light-weight, general purpose, laser altimeters ("Lidar") suitable for mounting on UAVs. These are useful for applications including terrain following, precision hovering (e.g. for photography), warning of regulatory height limits, anti-collision sensing etc.
+
+![LightWare SF11/C Lidar](../../assets/hardware/sensors/sf11c_120_m.jpg)
+
+PX4 supports the following LightWare Lidar rangefinders through the I2C bus (only):
+* [SF10/A](http://lightware.co.za/shop2017/drone-altimeters/26-sf10a-25-m.html) (25 m)
+* [SF10/B](http://lightware.co.za/shop2017/drone-altimeters/25-sf10b-50-m.html) (50 m)
+* SF10/C (100m) (Discontinued)
```
hamishwillee: FYI, I can't find any mention of this on their site , so I believe it is discontinued.

en/sensor/sfxx_lidar.md
```diff
@@ -0,0 +1,43 @@
...
+
+The only configuration required is to set the type of the rangefinder. You can do this in *QGroundControl* [Settings > Parameters](https://docs.qgroundcontrol.com/en/SetupView/Parameters.html).
+
+Set the [SENS_EN_SF1XX](../advanced_config/parameter_reference.md#SENS_EN_SF1XX) parameter to match the Lidar model and then reboot:
+* `0` lidar disabled
+* `1` SF10/a
+* `2` SF10/b
+* `3` SF10/c
+* `4` SF11/c
+* `5` SF/LW20
```
hamishwillee: Don't know why this  has SF prefix - the product name is LW20

hamishwillee:
General rangefinder questions:
- How do you test that an integration is working - ie where can you go to get the range from the rangefinder and check that it matches what you think it should be?
- What other parameters affect all rangefinders? Specifically, I know that you probably need to set the offset from wherever the rangefinder is mounted to the bottom of the landing gear.

en/sensor/sfxx_lidar.md
```diff
@@ -0,0 +1,43 @@
...
+* [LW20](http://lightware.co.za/shop2017/drone-altimeters/51-lw20-100-m.html) (100 m) - Waterproofed (IP67) with servo for sense-and-avoid applications
+
+
+## Configuring Lidar
+
+Newer sensors/versions require no additional configuration.
+
+> **Tip** Older sensors may be miss-configured to have an I2C address equal to `0x55` (this conflicts with `rgbled` module).
+> You can check the address by connecting the sensor via USB (it has an internal USB to serial converter), run terminal, press **space**.
+> If the I2C address is equal to `0x66` the sensor can be used with PX4.
```
hamishwillee: 1. I assume this only works on Linux? What do you do on Windows?
2. What do you do if it has the  wrong address - are Lidar user-writable?

en/sensor/ulanding_radar.md
```diff
@@ -0,0 +1,46 @@
+# uLanding Radar
+
+The *Aerotenna* [uLanding](https://aerotenna.com/sensors/#ulanding) altimeter is compact microwave rangefinder that has been optimised for use on UAVs. With a sensing range of 45m, it is useful for applications including terrain following, precision hovering (e.g. for photography), anti-collision sensing etc. Particular advantages of this product are that it can operate effectively in all weather conditions and over all terrain types (including water).
+
+<img src="../../assets/hardware/sensors/uLanding_lite_1.jpg" alt="Aerotenna uLanding" width="300px" />
+
+> **Note** Due to flash space limitations on some flight controllers the driver may not be built into all firmware.
```
hamishwillee: This is in nuttx_px4fmu-v2_default.cmake - does that mean it is in all FMUv2 boards - e.g. Pixhawk1 ?

hamishwillee:
The doc states that you have to put the driver in the firmware and set it to start on system boot,  pointing to the serial port that you've attached the rangefinder to.  
1. If it is already in firmware I assume it is still started on boot?
2. If it is started on boot, what port does it require/expect - or is something going on that makes it robust enough to use any port?

Upshot, for other devices you just set a parameter and things work. You don't need to set a specific port or startup script. What needs to happen here to make this more "turnkey"?

en/sensor/ulanding_radar.md
bkueng: Yes but it's more complicated: some newer boards that support the 2MB flash can use the nuttx_px4fmu-v3_default.cmake target (e.g. pixhawk mini)

bkueng:
To test whether a lidar is working you can check the `distance_sensor` topic using the listener (https://dev.px4.io/en/debug/sensor_uorb_topic_debugging.html). I think it's also possible to see the distance using the QGC analyze view.

Generally the problem with UART drivers is that you cannot configure easily the used port and baudrate (see also https://github.com/PX4/Firmware/issues/8302).

I think @ChristophTobler can help with lidar questions.

ChristophTobler:
@hamishwillee 
>How do you test that an integration is working - ie where can you go to get the range from the rangefinder and check that it matches what you think it should be?

@bkueng is right -> e.g. `listener distance_sensor 5` or use the QGC Analyse tool.

>What other parameters affect all rangefinders? Specifically, I know that you probably need to set the offset from wherever the rangefinder is mounted to the bottom of the landing gear.

Estimator specific parameters. For EKF2 all `EKF2_RNG_*`, e.g. delay, offset, etc. 

>Do we support any other types of rangefinders? Any links to that info? (I think we probably need to have a top level topic about options if you want to use a rangefinder)
We support all these distance sensors -> https://github.com/PX4/Firmware/pull/7984#issue-258816928. I don't thing we have any further info/links on that...

en/sensor/ulanding_radar.md
hamishwillee: So basically it would appear in FMUv2 or FMUv3 boards that use the FMUv2 firmware? Otherwise you need to add it yourself?

en/sensor/sfxx_lidar.md
hamishwillee: @ChristophTobler "Apparentl" you can check the I2C address by connecting a sensor to your computer via USB and running terminal. Pressing space should display the address.
That sounds quite odd
- It does mean connect to a computer right?
- Is this just Ubuntu?
If you don't know, is there anyone else you would recommend we ask?

en/sensor/sfxx_lidar.md
```diff
@@ -0,0 +1,43 @@
...
+
+
+## Hardware Setup
+
+Connect the Lidar the autopilot I2C port as shown below (in this case, for the [Pixhawk 1](../flight_controller/mro_pixhawk.md)).
+
+![SF1XX LIDAR to I2C connection](../../assets/hardware/sensors/sf1xx_i2c.jpg)
+
+> **Note** The PX4 driver for this rangefinder only supports I2C connections. You cannot use a UART etc.
```
hamishwillee: @bkueng @ChristophTobler Is it true that you can only connect to I2C? Looking at the "enabling parameters) it looks like all these are serial [SENS_EN_SF0X](https://docs.px4.io/en/advanced_config/parameter_reference.html#SENS_EN_SF0X) and these must be connected by I2C: [SENS_EN_SF1XX](http://localhost:4000/en/advanced_config/parameter_reference.html#SENS_EN_SF1XX).

If these can be connected by serial, would the connections be as described here: https://pixhawk.org/peripherals/rangefinder#lightware_sf02f_laser_rangefinder_40_m ?

hamishwillee:
Thanks guys.

@ChristophTobler I have a few more questions for you :-)

>> How do you test that an integration is working - ie where can you go to get the range from the rangefinder and check that it matches what you think it should be?

> @bkueng is right -> e.g. listener distance_sensor 5 or use the QGC Analyse tool.

1. Is there any information about  QGC Analyse tool  anywhere - cant find it in the devguide or docs for QGC. Essentially "how" you would test?
2. I'm after foolproof step-by-step instructions  (i.e. me) that can ideally be done by a non-developer on a normal QGC build. Essentially where I tell them to put vehicle on bench and see it says zero (m) then move to different levels and confirm results is expected. What approach from about do you recommend?
3. I tried using QGC MAVLink Console but I get `"ERROR [mavlink] Failed to start shell (-1)"` from vehicle shell when I try any commands from QGC (simulator running in VM). If I run (in the simulator px4 shell) `commander takeoff` and then `listener distance_sensor 5` I can see the vehicle take off in QGC. However I don't get messages from the vehicle:
```
pxh> listener distance_sensor 5

TOPIC: distance_sensor instance 0 #1
timestamp: 0
min_distance:   0.0000
max_distance:   0.0000
current_distance:   0.0000
covariance:   0.0000
type: 0
id: 0
orientation: 0
Waited for 2 seconds without a message. Giving up.
```
Thoughts?

>> Do we support any other types of rangefinders? Any links to that info? (I think we probably need to have a top level topic about options if you want to use a rangefinder)
> We support all these distance sensors -> PX4/Firmware#7984 (comment). I don't thing we have any further info/links on that...

----
Do all rangefinders we support require separate power / that the aux bus be powered? (I'm asking because this seems to have been the cause of [#6919](https://github.com/PX4/Firmware/issues/6919).
Is it worth a general comment on powering rangefinders?

hamishwillee:
@bkueng FYI, I also added a parent grouping topic with some of the information from @ChristophTobler . This is not quite ready for review - may modify in line with feedback.

en/sensor/rangefinders.md
```diff
@@ -0,0 +1,96 @@
...
+```
+
+For more information see: [Sensor/Topic Debugging using the Listener Command](https://dev.px4.io/en/debug/sensor_uorb_topic_debugging.html) (PX4 Development Guide).
+
+
+## Simulation
+
+Lidar and sonar rangefinders can be used in the [Gazebo Simulator](https://dev.px4.io/en/simulation/gazebo.html) (PX4 Development Guide).
+
+<!--
```
hamishwillee: @ChristophTobler How do you enable lidar/sonar in Gazebo?

en/sensor/sfxx_lidar.md
ChristophTobler: Never heard of this, sorry... Where did you get this information from? I haven't seen it on the Devguide. Just check who wrote it, I guess.

en/sensor/rangefinders.md
ChristophTobler: You have to use the iris model that includes the rangefinder -> `make posix_sitl_default gazebo_iris_opt_flow` (or include the sensor in another model, see https://github.com/PX4/sitl_gazebo/blob/master/models/iris_opt_flow/iris_opt_flow.sdf#L8-L22). If you don't want to use the rangefinder with GPS change `EKF2_AID_MASK` to `1`.

en/sensor/ulanding_radar.md
bkueng: It would appear in FMUv2 or FMUv3 boards that use the FMUv**3** firmware

Yes otherwise you have to add it yourself and most likely remove some other unused driver.

bkueng:
>  Is there any information about QGC Analyse tool anywhere - cant find it in the devguide or docs for QGC. Essentially "how" you would test?

That should be in the QGC docs then (I don't know of any existing docs, you have a better overview on that). It's generally useful, e.g to check the mag orientation.

>  I'm after foolproof step-by-step instructions (i.e. me) that can ideally be done by a non-developer on a normal QGC build. Essentially where I tell them to put vehicle on bench and see it says zero (m) then move to different levels and confirm results is expected. What approach from about do you recommend?

QGC analyze is the easiest way

>  I tried using QGC MAVLink Console but I get "ERROR [mavlink] Failed to start shell (-1)" from vehicle shell when I try any commands from QGC (simulator running in VM). If I run (in the simulator px4 shell) commander takeoff and then listener distance_sensor 5 I can see the vehicle take off in QGC.

The shell is only supported on NuttX -> you should try with a pixhawk board.

ChristophTobler:
@hamishwillee 

1. I guess not... But it should be in https://docs.qgroundcontrol.com/en/,  I think.
2. TODO
3. You can't run the shell and QGC at the same time. If you want to try in the simulation with `listener distance_sensor 5`, you have to use the iris model that includes the rangefinder -> `make posix_sitl_default gazebo_iris_opt_flow` (and change `EKF2_AID_MASK` to `1` to use GPS instead of optical flow).
4. Same as 3.
5. No. Optical flow just depends on a rangefinder but not vice versa. In the end, optical flow estimates the horizontal velocities.
6. Yes. We use a terraranger on the Snapdragon -> see https://docs.px4.io/en/flight_controller/snapdragon_flight.html. And on the Aero we use a Lidar Lite V3 -> see https://docs.px4.io/en/flight_controller/intel_aero.html
7. Not sure... Looks like the sonar being used on the PX4Flow module -> https://www.google.ch/search?q=srf02&oq=srf02&aqs=chrome..69i57j0l5.347j1j7&sourceid=chrome&ie=UTF-8
8. This is specific to the ll40ls (usually Lidar Lite) connected via PWM, I think. When running `ll40ls test pwm`, there will be an output. You can also use it for testing if you connect it via I2C. https://github.com/PX4/Firmware/blob/master/src/drivers/ll40ls/ll40ls.cpp#L336-L345 Just replace `pwm` with `i2c`
9. I don't think so.
10. I don't know...

en/sensor/rangefinders.md
hamishwillee: Thanks very much. I will update appropriately!

en/sensor/sfxx_lidar.md
hamishwillee: FYI< just leaving as is.

en/sensor/sfxx_lidar.md
hamishwillee: Leaving marked as discontinued. It certainly isn't available.

en/sensor/sfxx_lidar.md
hamishwillee: There is no information. So I will replace this with a "hint" to check out http://manpages.ubuntu.com/manpages/zesty/man8/i2cdetect.8.html

en/sensor/sfxx_lidar.md
hamishwillee: @dagar The text here indicates that the SF1XX only supports I2C but as per my comment https://github.com/PX4/px4_user_guide/pull/104/files/c0db40938219d8480ad7f11be04f6853fac349d1#r153391498 it looks like you can use serial. Do you know if this limitation is true or some sort of legacy rubbish? If not, can you suggest someone who might know?

en/sensor/ulanding_radar.md
hamishwillee: That is a little odd, because the driver is in "This is in nuttx_px4fmu-v2_default.cmake", not the FMuv3 config !

At this point I won't specify - below I've explained how they can add/check whether it is present in a configuration.

hamishwillee:
Thanks guys.

> QGC analyze is the easiest way

@bkueng @ChristophTobler Finally found it - in the Widgets menu - not where I expected! The bad news is that this does not show the altitude. The good news is that the MAVLink Inspector in the widgets does. So I will update appropriately.

hamishwillee:
@bkueng @ChristophTobler 
.... althought Although I guess will need to turn off GPS as main source of height measurement to ensure altitude is what I see from the inspector. How do I set rangefinder as the main height source/turn off GPS?

en/sensor/rangefinders.md
```diff
@@ -0,0 +1,171 @@
...
+
+## Testing
+
+The easiest way to test the rangefinder is to vary the range and compare to the values detected by PX4. The sections below show some approaches to getting the measured range.
+
+### QGC MAVLink Inspector
+
+The *QGroundControl MAVLink Inspector* tool lets you check the altitude sent from the vehicle. If you set the vehicle to use a rangefinder as the main height source, then this should reflect the sensor values:
+
+1. Set the rangefinder as the main height source (TBD)
```
hamishwillee: @bkueng I think this PR is ready for review. The only outstanding bit AFAIK is this setting.

ChristophTobler:
@hamishwillee 
You have to set `EKF2_HGT_MODE` to range finder (2), to enable height estimation using just the range finder. The default height sensor is the barometer.
Another option is enabling range aid -> `EKF2_RNG_AID`. This uses the range finder when your vehicle meets some checks (velocity < `EKF2_RNG_A_VMAX`, distance to ground < `EKF2_RNG_A_HMAX`) and the sensor specified in `EKF2_HGT_MODE` otherwise.

hamishwillee:
@ChristophTobler Awesome. Updated (and thanks for all the other information too.

@bkueng (and Christoph) this is ready for review when you are.

en/sensor/rangefinders.md
```diff
@@ -0,0 +1,180 @@
...
+
+### QGC MAVLink Inspector
+
+The *QGroundControl MAVLink Inspector* tool lets you check the altitude sent from the vehicle. If you set the vehicle to use a rangefinder as the main height source, then this should reflect the sensor values:
+
+1. Make the rangefinder the main height source (set parameter `EKF2_HGT_MODE` to Range finder (`2`))
+1. Open the menu **Widgets > MAVLink Inspector**:
+
+   ![Menu for QGC MAVLink Inspector](../../assets/qgc/menu_mavlink_inspector.png)
+1. Select the target vehicle and then the `ALTITUDE` message. Observe the value as shown:
```
bkueng: In the inspector you should also see a `DISTANCE_SENSOR` message (without having to change the height source).
Or even better with `Widgets -> Analyze`, you can see a graph in real-time:
screenshot_20171130_100139

hamishwillee: Nope, that message isn't visible in either view. I get the small subset below. Perhaps the set is different in a daily build?
Actually more likely I just don't have a distance sensor on this simulated vehicle ?

image

ChristophTobler: That's the Analyse tool Beat is talking about and not the Mavlink Inspector

hamishwillee: The subset of options in Analyse is even smaller.

ChristophTobler: You're running `make posix_sitl_default gazebo_iris_opt_flow`?

bkueng: @hamishwillee the set of sent/displayed messages is defined by the Firmware (the configured mavlink streams), and only visible if there's a distance sensor attached (i.e. not with jMAVSim).

en/SUMMARY.md
```diff
@@ -76,6 +76,10 @@
   * [FrSky Telemetry](peripherals/frsky_telemetry.md)
   * [RTK GPS - Here+](peripherals/rtk_gps_hex_hereplus.md)
   * [RTK GPS - Drotek XL](peripherals/rtk_gps_drotek_xl.md)
+  * [RTK GPS - Drotek XL](peripherals/rtk_gps_drotek_xl.md)
```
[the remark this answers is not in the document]
hamishwillee: Thanks. Good catch.

en/sensor/rangefinders.md
```diff
@@ -0,0 +1,183 @@
...
+   * Specifically, the rangefinder is enabled when: 
+     * velocity < [EKF2_RNG_A_VMAX](../advanced_config/parameter_reference.md#EKF2_RNG_A_VMAX)
+     * distance to ground < [EKF2_RNG_A_HMAX](../advanced_config/parameter_reference.md#EKF2_RNG_A_HMAX)
+   * Other parameters affecting "Range aid" are prefixed with `EKF2\_RNG\_A\_`.
+
+## Testing
+
+The easiest way to test the rangefinder is to vary the range and compare to the values detected by PX4. The sections below show some approaches to getting the measured range.
+
+### QGroundControl Analyze Tool
```
hamishwillee: Thanks @bkueng @ChristophTobler  - using wrong vehicle was the cause. Tested against the gazebo simulation and things worked well.  This section now updated.

Anything else needed? I think this is a much more complete rangefinder story than we had.

en/sensor/rangefinders.md
```diff
@@ -0,0 +1,183 @@
...
+
+
+### QGroundControl MAVLink Console
+
+You can also use the *QGroundControl MAVLink Console* to observe the `distance_sensor` uORB topic:
+```sh
+listener distance_sensor 5
+```
+
+> **Note** This works when connected to Pixhawk or other NuttX targets, but not the Simulator.
```
ChristophTobler: Maybe we can mention that it does work in the terminal where you start the simulation

hamishwillee: Good catch. Done!

hamishwillee:
Thanks @bkueng @ChristophTobler - really appreciate all your help on this one. Merging now.

en/sensor/sfxx_lidar.md
dagar: Sorry not sure. git blame?

hamishwillee: Me either. I've asked [on slack](https://px4.slack.com/archives/C3GFT2QS3/p1512534416000033). If no response, maybe create a firmware "question" bug?

dagar: Oh, are you talking about the PX4 driver by that name or the physical device? Most of them are both i2c and serial. The drivers could probably be merged in PX4 with the interface differences handled internally.

hamishwillee: I'm talking about the device - the doc is about the device, not the driver. What I would guess has happened is that at some point there was just one driver so this was a true statement. It is also possible that this doc was written just for the I2C driver - which makes little sense. 

Think I can remove the comment safely?

dagar: Yes I think so.

hamishwillee: Cheers. Updated in https://github.com/PX4/px4_user_guide/pull/115

PR #106, opened 2017-12-04
Add updated telemetry docs from the devguide

hamishwillee:
This ports the docs from the devguide. Still needs a little more work .


en/telemetry/sik_radio.md
```diff
@@ -0,0 +1,71 @@
+# SiK Radio
+
+[SiK radio](https://github.com/LorenzMeier/SiK) is a collection of firmware and tools for telemetry radios.
+
+Hardware for the SiK radio can be obtained from various manufacturers/stores in variants that support different range and form factors. Typically you will need a pair of devices - one for the vehicle and one for the ground station. The ground station-based radio is connected via USB (essentially plug-n-play). The vehicle is connected to the flight-controller's `TELEM` port, and typically requires no further configuration.
```
hamishwillee: @bkueng I think this is correct - can you please confirm? Especially that you connect to TELEM and that you don't need to do further configuration in PX4.

en/telemetry/sik_radio.md
```diff
@@ -0,0 +1,71 @@
...
+### Vendors
+
+* 3DR Radio \(small\) - [US (915MHz)](https://store.3dr.com/products/915-mhz-telemetry-radio) / [European (433MHz)](https://store.3dr.com/products/433-mhz-telemetry-radio)
+* [HK Radio](http://www.hobbyking.com/hobbyking/store/uh_viewitem.asp?idproduct=55559) \(small\)
+* [RFD900u](http://rfdesign.com.au/products/rfd900u-modem/) \(small\)
+* [RFD900](http://rfdesign.com.au/products/rfd900-modem/) \(long range\)
+
+![SiK Radio](../../assets/hardware/telemetry/sik_radio.jpg)
+
+## Build Instructions
```
hamishwillee: @bkueng So I think that generally speaking if you purchase SiK hardware it will already contain firmware that is built and configured appropriately. Is that correct?
So perhaps best thing here is to add a note that you can build the source and configure the software, but generally speaking you won't need to?

en/telemetry/telemetry_wifi.md
```diff
@@ -0,0 +1,20 @@
+# WiFi Telemetry Radio
+
+WiFi telemetry enables MAVLink communication between a WiFi radio on a vehicle and a GCS.  WiFi typically offers shorter range than a normal telemetry radio, but supports higher data rates, and makes it easier to support FPV/video feeds. Usually only a single radio unit for the vehicle is needed (assuming the ground station already has WiFi).
+
+PX4 supports telemetry via UDP and Wifi. It broadcasts a heartbeat to port 14550 on 255.255.255.255 until it receives the first heartbeat from a ground control station, at which point it will only send data to this ground control station.
+
+
+## 3DR WiFi Telemetry
+
+The [3DR WiFi Telemetry Radio](https://store.3dr.com/products/wifi-telemetry-radio) is supported by PX4. Simply connect it to the flight controller's `TELEM` port. Once connected to the WiFi, it should automatically connect to *QGroundControl*.
```
hamishwillee: @bkueng So I just want to confirm the semantics here. You've plugged the wifi into TELEM. I think that what happens is that it creates a wifi hotspot with the hotspot and password below. So on your GCS you just connect to that hotspot. Is that correct? (if so this will need minor update).

en/SUMMARY.md
```diff
@@ -76,6 +76,9 @@
   * [FrSky Telemetry](peripherals/frsky_telemetry.md)
   * [RTK GPS - Here+](peripherals/rtk_gps_hex_hereplus.md)
   * [RTK GPS - Drotek XL](peripherals/rtk_gps_drotek_xl.md)
+  * [Telemetry](telemetry/README.md)
```
bkueng: Should we rename this to `MAVLink Telemetry`, to make it clear it's separate from `FrSky Telemetry`?

en/telemetry/sik_radio.md
bkueng: This is correct. I think I'd write `TELEM1`, since most boards provide several telemetry outputs, and the first one is used for the ground station.

en/telemetry/sik_radio.md
bkueng: > So I think that generally speaking if you purchase SiK hardware it will already contain firmware that is built and configured appropriately

Yes. It's however possible that it does not yet support mavlink 2, and then you need to update.

> So perhaps best thing here is to add a note that you can build the source and configure the software, but generally speaking you won't need to?

Yes, and I've never even done that myself :) I don't think it needs to be part of the user-guide, maybe in the dev-guide. And more importantly: it's even possible to update the radio firmware via QGC (in the same way as updating the autopilot FW via QGC works: change to the FW upgrade tab, then attach the radio).

en/telemetry/telemetry_wifi.md
bkueng: I think so but I've never used that myself, so I cannot say for sure.

en/telemetry/sik_radio.md
hamishwillee: >  I don't think it needs to be part of the user-guide

It belongs in the user guide because a user (someone who wants to build vehicles using existing supported components) might reasonably need to do this.  Agree it is not a "main use case", hence need note to make that clear. Thanks for info. Will update.

hamishwillee: Actually, I'm changing my mind on this. No user should have to see this :-)

en/SUMMARY.md
hamishwillee: The complication is that the radios might also be used for other protocols and data. How about Telemetry Radios - differentiates a little by making it clear that these are radios that are just for telemetry.

en/telemetry/sik_radio.md
```diff
@@ -0,0 +1,32 @@
...
+![SiK Radio](../../assets/hardware/telemetry/sik_radio.jpg)
+
+## Setup/Configuration
+
+The ground station-based radio is connected via USB (essentially plug-n-play). 
+
+The vehicle is connected to the flight-controller's `TELEM1` port, and typically requires no further configuration.
+
+
+## Firmware Update
```
hamishwillee: @bkueng I've updated this section in line with your comments. Can you please check it?

I tried this with 3DR radio, but it is not detected by QGC for firmware update. Can you confirm that if this *did* work the latest version, including support for MAVLink2, would be installed?

hamishwillee:
@bkueng So module new comment on firmware update I think this is nearly good to go. BUT, I found this page: https://pixhawk.org/peripherals/radio-modems/start

Do we support some of these other hardwares with PX4? E.g. xbee? If so, can you identify which radios we know work?

en/telemetry/sik_radio.md
bkueng: Yes, I can confirm that at least it used to work, because I've done that.

hamishwillee:
Hi @bkueng Thanks for the review. Re the Xbee etc, I have created #108 to track, but will merge this now.
Cheers!

PR #109, opened 2017-12-05
rc_transmitter_receiver: Amend remote control recommendations

MaEtUgR:
Polished text, corrected to account for X4R receivers some of which support S-BUS, added premium long range sets

### DIFF
--- a/en/getting_started/rc_transmitter_receiver.md
+++ b/en/getting_started/rc_transmitter_receiver.md
@@ -69,11 +69,13 @@ The appropriate connection for each type of receiver is described below:
 
 ### Transmitters
 
-One of the most popular radio systems is the open source *FrSky Taranis PPM-Sum Compatible Transmitter*. This is compatible with many high quality FrSky PPM-Sum compatible receivers.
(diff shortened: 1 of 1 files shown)


hamishwillee:
Thanks - looks great. Merged.

Looking at this doc with fresh eyes, I like that you've separated the Tx/Rx concept from the "Remote" (Physical unit held by the user). I think it might be better to do this across the doc. 

Further, is there a good collective name for the Tx/Rx - the problem with this is that often they are bidirectional, so might be better of referring to them as "Radio System" or similar and referring to the units as land-based and vehicle based. Thoughts?

MaEtUgR:
@hamishwillee Sorry I didn't go though everything. I just had a quick shot.
Yes I think pyhsical joystick device and radio communication system is confused a lot all over the internet.

My personal opinion:
I would name specialized radio solutions collectively after what your primary goal is when using them and in this case it would be remote controlling the vehicle (with a **remote control**). If the system supports displaying telemetry that's only an additional (still very useful) feature. Also to call the vehicle end piece receiver is totally common because of the same reason. The name is not to confuse with any general communication solution which allows you to configure (parameters), program (send mission) and stream (live video). That I would call **radio link**.

hamishwillee:
Thanks. Agreed. I've attempted to improve the wording a bit in: https://github.com/PX4/px4_user_guide/pull/114

PR #111, opened 2017-12-05
Update LeddarOne text

hamishwillee:


en/sensor/leddar_one.md
```diff
@@ -17,9 +25,11 @@ You can simply add the following line to an [extras.txt](https://dev.px4.io/en/a
 leddar_one start -d /dev/serial_port
 ```
 
-In the above command you will have to replace the last argument with the serial port you have connected the hardware to. Make sure that there is no MAVLink stream in the serial port choosen, if there is one find the parameter to disable it.
+In the above command you will have to replace the last argument with the id of the serial port used to connect the hardware. Make sure that there is no MAVLink stream in the selected serial port (if there is find the parameter to disable it).
```
hamishwillee: @zehortigoza Is there any guidance we can point users to in order to find out if a MAVLink stream will be using the selected port? And just say we use TELEM2, how would we disable the stream to that port?

zehortigoza: About disable MAVLink stream in a Pixhawk would as the same as for ULanding instructions(set SYS_COMPANION to 0) https://github.com/PX4/px4_user_guide/blob/master/en/sensor/ulanding_radar.md

zehortigoza: A way to check if serial port is in use is run:
mavlink status
in the nuttx terminal and check what is the port of every stream.

hamishwillee: Thank you. I have updated the text to match ulanding in https://github.com/PX4/px4_user_guide/pull/112

This makes sense because it is a specific warning about a specific port. Your text would be even better - but only if we could find/link to general instructions to help users work out whether "any" port is being used, what by, and how to disable it.

hamishwillee: Ok that is useful to know.  I think I'll add that info - but later on: https://github.com/PX4/px4_user_guide/issues/113

Still worth having the specific note I added though :-)

PR #121, opened 2017-12-22
pixhawk series move pixhawk mini and mindpx to FMUv3

dagar:


dagar:
@hamishwillee MindPX is a separate config. https://github.com/PX4/Firmware/tree/master/nuttx-configs

image

The mRo/AUAV x2.1 is nearly px4fmu-v2/v3 compatible, but has some minor differences and a separate build.

hamishwillee:
Thanks @dagar  - as you say : https://docs.px4.io/en/flight_controller/mindpx.html#how-to-build :-)

PR #130, opened 2018-01-09
Improve/update thermal calibration guide

hamishwillee:
Changes:
- Cross link to parameters
- Improve rendering by using code blocks appropriately
- Fix a few typos, add consistent grammar.

This is rendered [here](https://hamishwillee.gitbooks.io/ham_px4userguide/content/v/thermal_calib/en/advanced_config/sensor_thermal_calibration.html)

### DIFF
--- a/en/advanced_config/sensor_thermal_calibration.md
+++ b/en/advanced_config/sensor_thermal_calibration.md
@@ -2,22 +2,38 @@
 
 PX4 contains functionality to calibrate and compensate rate gyro, accelerometer and barometric pressure sensors for the effect of changing sensor temperature on sensor bias.
 
-This topic details the [calibration procedures](#procedures) (including best practises). At the end there is a description of the [implementation](#implementation). 
+This topic details the [test environment](#test_setup) and [calibration procedures](#calibration_procedures). At the end there is a description of the [implementation](#implementation). 
 
 
-## Calibration Procedures {#procedures}
+## Test Setup/Best Practice {#test_setup}
+
+The [calibration procedures](#calibration_procedures) described in the following sections are ideally run in an *environment chamber* (a temperature and humidity controlled environment) as the board is heated from the lowest to the highest operating/calibration temperature. Before starting the calibration, the board is first *cold soaked* (cooled to the minimum temperature and allowed to reach equilibrium).
+
+For the cold soak you can use a regular home freezer to achieve -20C, and commercial freezers can achieve of the order of -40C. 
+The board should be placed in a ziplock/anti-static bag containing a silica packet, 
+with a power lead coming out through a sealed hole. 
+After the cold soak the bag can be moved to the test environment and the test continued in the same bag. 
+
+> **Note** The bag/silica is to prevent condensation from forming on the board.
+
+It possible to perform the calibration without a commercial-grade environment chamber.
+A simple environment container can be created using a styrofoam box with a very small internal volume of air. 
+This allows the autopilot to self-heat the air relatively quickly (be sure that the box has a small hole to equalize to ambient room pressure, but still be able to heat up inside). 
+
+Using this sort of setup it is possible to heat a board to ~70C. 
+Anecdotal evidence suggests that many common boards can be heated to this temperature without adverse side effects. 
+If in doubt, check the safe operating range with your manufacturer.
+
+> **Tip** To check the status of the onboard thermal calibration use the MAVlink console (or NuttX console) to check the reported internal temp from the sensor. 
+
+
+## Calibration Procedures {#calibration_procedures}
 
 PX4 supports two calibration procedures: 
 * [onboard](#onboard_calibration) - calibration is run on the board itself. This method requires knowledge of the amount of temperature rise that is achievable with the test setup.
 * [offboard](#offboard_calibration) - compensation parameters are calculated on a development computer based on log information collected during the calibration procedure. This method allows users to visually check the quality of the data and curve-fit.
 
 The offboard approach is more complex and slower, but requires less knowledge of the test setup and is easier to validate. 
-
-
-### Best Practices
-
-
-TBD
 
 
 ### Onboard Calibration Procedure {#onboard_calibration}
--- a/en/advanced_config/sensor_thermal_calibration.md
+++ b/en/advanced_config/sensor_thermal_calibration.md
@@ -2,7 +2,9 @@
 
 ## Overview
 
-PX4 conains functionality to calibrate and compensate rate gyro, accelerometer and barometric pressure sensors for the effect of changing sensor temperature on sensor bias. Calibration refers to the process of measuring the change in sensor value across a range of internal temperatures and performing a polynomial fit on the data to calculate a set of coefficients stored as parameters that can be used to correct the sensor data. Compensation refers to the process of using the internal temperature to calculate an offset which is subtracted from the sensor reading to correct for changing offset with temperature
+PX4 contains functionality to calibrate and compensate rate gyro, accelerometer and barometric pressure sensors for the effect of changing sensor temperature on sensor bias. 
+
+Calibration refers to the process of measuring the change in sensor value across a range of internal temperatures, and performing a polynomial fit on the data to calculate a set of coefficients (stored as parameters) that can be used to correct the sensor data. Compensation refers to the process of using the internal temperature to calculate an offset that is subtracted from the sensor reading to correct for changing offset with temperature
 
 The inertial rate gyro and accelerometer sensor offsets are calculated using a 3rd order polynomial, whereas the barometric pressure sensor offset is calculated using a 5th order polynomial. Examples fits are show below:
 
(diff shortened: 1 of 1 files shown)


en/advanced_config/sensor_thermal_calibration.md
```diff
@@ -12,94 +14,104 @@ The inertial rate gyro and accelerometer sensor offsets are calculated using a 3
...
+Correction of the accelerometer, barometers or rate gyroscope data is enabled by setting [TC_A_ENABLE](../advanced_config/parameter_reference.md#TC_A_ENABLE), [TC_B_ENABLE](../advanced_config/parameter_reference.md#TC_B_ENABLE) or [TC_G_ENABLE](../advanced_config/parameter_reference.md#TC_G_ENABLE) parameters to 1 respectively.
 
-Correction os the accelerometer, barometers or rate gyroscope data is enabled by setting TC\_A\_ENABLE\_, \_TC\_B\_ENABLE or TC\_G\_ENABLE parameters to 1 respectively.
+### Compatibility with legacy `CAL_*` parameters and commander controlled calibration
 
-## Compatibility with legacy CAL\_\* parameters and commander controlled calibration
+The legacy temperature-agnostic PX4 rate gyro and accelerometer sensor calibration is performed by the commander module and involves adjusting offset, and in the case of accelerometer calibration, scale factor calibration parameters. The offset and scale factor parameters are applied within the driver for each sensor. These parameters are found in the [CAL parameter group](../advanced_config/parameter_reference.md#sensor-calibration).
```
hamishwillee: I think what you're saying is that the sensor module applies both legacy and thermal compensation, so if you want to use thermal compensation you should set the "legacy parameters" to zero all the time (both during calibration and compensation). Presumably if you don't want to use thermal compensation you'd set the thermal parameters to zero. Is that correct?

If that is correct, could we use a title like "Disable Legacy CAL_* Parameters" ?

hamishwillee:
@priseborough 

I am also tempted to put the current Overview topic under "Implementation" and move it to the end. So basically the user first sees instructions on how to do the calibration and only gets the explanation of how everything works afterwards. Does this seem reasonable, or do you think it is important that they fully understand how things work before they start the calibration process?

hamishwillee:
@priseborough How often/when does thermal calibration normally need to be done? Is this a one off?

Antiheavy:
@M-Skelton let's compile some thoughts about thermal calibration best practices and contribute to this thread.  

One I will add:
- if you are not fortunate enough to be using a real environmental chamber, one easy method is to place the autopilot inside a styrofoam box with a very small internal volume of air so that the autopilot can self-heat the air relatively quickly.  Be sure the box has a small hole to equalize to ambient room pressure, but still be able to heat up inside.

ryanjAA:
Prob should list some max temps do people don’t go overboard too.

en/advanced_config/sensor_thermal_calibration.md
```diff
@@ -12,94 +14,104 @@ The inertial rate gyro and accelerometer sensor offsets are calculated using a 3
 
 ![Thermal calibration barometer](../../assets/calibration/thermal_calibration_baro.png)
 
-## Calibration Parameter Storage
-
-With the existing parameter system implementation we are limited to storing each value in the struct as a separate entry. To work around this limitation the following logical naming convention for the parameters is used:
+### Calibration Parameter Storage
 
-TC\_[type][instance]\_[cal\_name]\_[axis] , where
+With the existing parameter system implementation we are limited to storing each value in the struct as a separate entry. To work around this limitation the following logical naming convention is used for the [thermal compensation parameters](../advanced_config/parameter_reference.md#thermal-compensation):
```
hamishwillee: "we are limited to storing each value in the struct as a separate entry." - What struct are we talking about here - has not previously been mentioned.

hamishwillee:
The difference between onboard and offboard calibration is not articulated. Would I be correct in saying the difference is that with offboard we do measurements and then run tools on dev computer to generate the required parameters, while with onboard the parameters are generated internally. The value offboard being that you can view more easily how well things fit?

hamishwillee:
> Prob should list some max temps do people don’t go overboard too.

@ryanjAA You mean like value for SYS_CAL_TMAX ? 

Looking at the instructions there are some good comments about what happens if you set particular values too high or too low (e.g.  SYS_CAL_TDEL) but nothing to suggest what sort of range might be reasonable. So would be good to have initial values for everything.

ryanjAA:
I meant in even more basic terms. Like disclaimer type since people may heat them up much too high. 

As far as some specifics, definitely all the Parama but we should address maximum ramp rate (the silica and bag is a great idea to deal with condensation), ideal min temp to start at so people actually do useful calibration. Maybe some example of before and after to show users what they can expect.

hamishwillee:
Hi All, 
I've restructured to put the calibration procedures first in the document  (and explain the difference between the onboard and offboard). This makes sense because much of the "overview" was actually background implementation information that most people will not need. I can reorder if there is strong disagreement. 

As part of this I have created a **Calibration Procedures > Best Practises** placeholder section. 

With respect to that, possibly we need to go even further back - for example I don't know what the _cold soak_ referred to in the docs means and what an ideal vs a workaround setup looks like. I infer from the above that something called an environment chamber is the best approach. I think that @mhkabir is suggesting using a ziplock bag instead, and that @Antiheavy is suggesting a styrofoam box for the same purpose (but it could be that @mhkabir might put his ziplock inside the styrofoam box).

Upshot, can you assume (correctly) that I know nothing and provide some background?

en/advanced_config/sensor_thermal_calibration.md
```diff
@@ -1,105 +1,140 @@
...
+### Calibration Parameter Storage
 
-If gyro thermal compensation has been enabled by setting the TC\_\_G\_\_ENABLE parameter to 1, then the commander controlled gyro calibration can still be perfomred, however it will be used to shift the compensation curve up or down by the amount required to zero the angular rate offset. It achieves this by adjusting the X0 coefficients.
+With the existing parameter system implementation we are limited to storing each value in the struct as a separate entry. To work around this limitation the following logical naming convention is used for the [thermal compensation parameters](../advanced_config/parameter_reference.md#thermal-compensation):
```
hamishwillee: What is the struct referred to here?

priseborough:
'Cold Soak' refers to the process of waiting until the temperature of the test article has reached equilibrium.

M-Skelton:
> I infer from the above that something called an environment chamber is the best approach. I think that @mhkabir is suggesting using a ziplock bag instead, and that @Antiheavy is suggesting a styrofoam box for the same purpose (but it could be that @mhkabir might put his ziplock inside the styrofoam box).

@hamishwillee I can clear this up. The Ziplock bag with silica is to prevent condensation from forming on the board. I like to use a resealable anti-static bag with a silica packet in it. 

The styrofoam box is more like an environmental chamber. It allows the cold soaked board to self-heat in an insulated environment. I have found that this is a good method to get the board to heat to warmer temperatures without having to add a heat source. 

I think the best practice would be to cold soak the board in a bag with a silica packet in it. Then when the soak is complete, remove the board from the bag and place it into the environmental chamber as quickly as possible. However, it is important to seal the environmental chamber (although it should be ported so that the pressure in the chamber is the same as the ambient pressure) before starting the calibration. Otherwise, you can create some bogus baro readings. 



> Prob should list some max temps do people don’t go overboard too.

@ryanjAA I have had my custom FMUv4 boards and pixracers up to ~70C (~160F) during calibrations and haven't seen any adverse side effects. What do you think a good upper limit is? How would this limit be enforced? Even if TMAX is set to the highest allowed value there currently isn't a good way to get feedback from the board that it has reached that temp. From my experience, the only way(s) to know the status of the onboard thermal cal is through the NuttX console, or by using the MAVlink console to check the reported internal temp from the sensor. Both of those methods are sort of tedious. thoughts?

M-Skelton:
@priseborough Correct. I forgot to mention there is also a silica packet in my environmental chamber (styrofoam box). I add the silica to the antistatic bag when cold soaking so that when I transfer the board from the freezer into room temperature air then to the box I don't get any condensation forming. 

I could probably streamline the process by just leaving the board in the anti-static bag and putting the whole thing in the box. I was removing the board from the bag in an attempt to not have the board heat up too quickly. The bag I use is just big enough to get a board into, so it's _sort _of a blanket.

en/advanced_config/sensor_thermal_calibration.md
hamishwillee: @priseborough ?

hamishwillee:
@mhkabir @M-Skelton Et al, is my summation in https://github.com/PX4/px4_user_guide/pull/130#issuecomment-382946449 above correct, and can any of you confirm the upper/lower bounds? Then I can create basic docs.

hamishwillee:
@priseborough Yes, here https://hamishwillee.gitbooks.io/ham_px4userguide/content/v/2785badeee91e3996b0f762b7573d507ddf2178f_1/en/advanced_config/sensor_thermal_calibration.html

hamishwillee:
Thanks Paul and everyone else. Merging now.

PR #133, opened 2018-01-19
Update pixhawk-2.md

proficnc:
Pixhawk 2 has always had two perfectly working CAN busses with internal trancievers. I have no idea where this story about the transceiver came from...


hamishwillee:
Thanks. Merging!

PR #138, opened 2018-01-31
Clarify how to set up tfmini on different boards

hamishwillee:
Fixes docs as discussed in https://github.com/PX4/Firmware/pull/8726/files#r164661606

### DIFF
--- a/en/sensor/rangefinders.md
+++ b/en/sensor/rangefinders.md
@@ -78,15 +78,31 @@ The uLanding Radar is not present in "most" firmware by default and must be star
 The *LeddarOne* must typically be added to firmware before it can be used. For setup/usage information see: [LeddarOne](../sensor/leddar_one.md).
 
 
-### TFmini 
+### TFmini
 
 The [Benewake TFmini LiDAR](http://www.benewake.com/en/tfmini.html) is a tiny, low cost, and low power LIDAR with 12m range. 
 
-The sensor should be connected to the SERIAL 4/5 port on Pixhawk devices. It is enabled by setting parameter [SENS_EN_TFMINI](../advanced_config/parameter_reference.md#SENS_EN_TFMINI) > 0.
+> **Note** The lidar driver is currently in master, and should appear in releases from PX4 v1.8.
 
-> **Note** This lidar is currently in master, and should appear in releases from PX4 v1.8.
+For boards based on [PX4FMU_V3](../flight_controller/pixhawk_series.md#fmu-versions) firmware (only) the sensor can be connected to the SERIAL 4/5 port and enabled by setting parameter [SENS_EN_TFMINI](../advanced_config/parameter_reference.md#SENS_EN_TFMINI) > 0.
 
-<!-- driver: drivers/tfmini -->
+For other boards you may need to add the driver to the firmware and ensure it is started on boot:
+1. The driver is present by default in most firmware. If not, add the following line to the *cmake* [config file](https://github.com/PX4/Firmware/tree/master/cmake/configs) that corresponds to the target board:
+   ```
+   drivers/tfmini
+   ```
+1. Update the system to start the driver during system startup on your required port.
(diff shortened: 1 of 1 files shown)


en/sensor/rangefinders.md
```diff
@@ -78,15 +78,31 @@ The uLanding Radar is not present in "most" firmware by default and must be star
 The *LeddarOne* must typically be added to firmware before it can be used. For setup/usage information see: [LeddarOne](../sensor/leddar_one.md).
 
 
-### TFmini 
+### TFmini
 
 The [Benewake TFmini LiDAR](http://www.benewake.com/en/tfmini.html) is a tiny, low cost, and low power LIDAR with 12m range. 
 
-The sensor should be connected to the SERIAL 4/5 port on Pixhawk devices. It is enabled by setting parameter [SENS_EN_TFMINI](../advanced_config/parameter_reference.md#SENS_EN_TFMINI) > 0.
+> **Note** The lidar driver is currently in master, and should appear in releases from PX4 v1.8.
```
hamishwillee: @ChristophTobler Is this correct? ie. is this already "Live" in PX4 v1.7? If so I would delete this note.

en/sensor/rangefinders.md
```diff
@@ -78,15 +78,31 @@ The uLanding Radar is not present in "most" firmware by default and must be star
...
+   ```
+   #start the driver on serial 4/5 (default)
+   tfmini start
+   ```
+   
+   You can also start the driver on other free serial ports (e.g. TELEM2) by specifying your desired serial port id.
+   ```
+   tfmini start -d serial_port
+   ```
+1. You may also need to disable access to the selected serial port from other services. For example, if you're using TELEM2 then set the parameter [SYS_COMPANION](../advanced_config/parameter_reference.md#SYS_COMPANION) to 0 in order to disable MAVLink communication with a companion computer on this port.
```
hamishwillee: @ChristophTobler You made a comment "you need to manually disable mavlink on the port (rcS) and start the driver.". For other rangefinders we advise to set the param SYS_COMPANION to zero. Is this the mavlink access you were talking about? IF so, I think this is a better approach than messing with the rcS file (OK?). If not, then can you expand on the modification required?

en/sensor/rangefinders.md
```diff
@@ -78,15 +78,31 @@ The uLanding Radar is not present in "most" firmware by default and must be star
...
+For boards based on [PX4FMU_V3](../flight_controller/pixhawk_series.md#fmu-versions) firmware (only) the sensor can be connected to the SERIAL 4/5 port and enabled by setting parameter [SENS_EN_TFMINI](../advanced_config/parameter_reference.md#SENS_EN_TFMINI) > 0.
 
-<!-- driver: drivers/tfmini -->
+For other boards you may need to add the driver to the firmware and ensure it is started on boot:
+1. The driver is present by default in most firmware. If not, add the following line to the *cmake* [config file](https://github.com/PX4/Firmware/tree/master/cmake/configs) that corresponds to the target board:
+   ```
+   drivers/tfmini
+   ```
+1. Update the system to start the driver during system startup on your required port.
+   You can simply add the following line to an [extras.txt](https://dev.px4.io/en/advanced/system_startup.html) file located on your SD card. For example to start the driver on serial 4/5:
```
hamishwillee: @dagar Using extras.txt for adding stuff to system startup seems to be the normal recommended way to configure a driver to start on boot. I imagine this would be a pain if you were a drone maker and wanted this in a ROM. In that case, what would be the "right" way to add to system startup?

dagar: I'm glad we have the flexibility to do this, but I think it's pretty horrible. We need to think of a good way to configure the vast majority of cases via parameter and I'll find time to implement it.

hamishwillee: Agree. So no other slightly less horrible option now - other than perhaps forking and adding direct to the normal board startup config?

hamishwillee:
Superceded

en/sensor/rangefinders.md
dagar: Right.

PR #145, opened 2018-02-09
Fixed minor typos and made minor changes for ease of understanding

barzanisar:


en/advanced_config/pid_tuning_guide_fixedwing.md
```diff
@@ -131,7 +131,7 @@ fixed wing setups and usually does not require adjustment.
 All L1 parameters are described [here](../advanced_config/parameter_reference.md#fw-l1-control).
 
 - [FW_L1_PERIOD](../advanced_config/parameter_reference.md#FW_L1_PERIOD) - This is the L1 distance and defines the tracking
-  point ahead of the aircraft its following. A value of 25 meters
+  point ahead of the aircraft it is following . A value of 25 meters
```
hamishwillee: Additional space before full stop is not needed (otherwise good fix, though I would have simply changed to "it's")

en/flight_modes/README.md
```diff
@@ -51,15 +51,15 @@ Throttle Command is mapped direct to Motor Speed.
 
 "Assisted" flight modes are also user controlled but offer some level of "automatic" assistance to gain or restore controlled flight.
 
-- [Altitude](../flight_modes/altitude.md): More easily control vehicle altitude, and in particular reach and maintain a fixed altitude. The mode does not use GPS, and hence will not attempt to hold position/heading against wind.
+- [Altitude](../flight_modes/altitude.md): More easily control vehicle altitude, and in particular reach and maintain a fixed altitude. The mode does not use GPS, and hence will not attempt to hold the x and y position/heading against wind.
```
hamishwillee: Revert this. This clarification is less correct because if you expand it out to "hold the x and y position/hold the x and y heading" you can see the second case doesn't work. Also it won't hold the z position either :-)

en/flight_modes/README.md
```diff
@@ -51,15 +51,15 @@ Throttle Command is mapped direct to Motor Speed.
...
+  - **Fixed wing aircraft:** Neutral inputs (meaning when roll, pitch and yaw sticks are centered) provide a level flight and
     will crab against the wind if needed to maintain a straight line.
   - **Multirotors** Roll controls left-right speed, pitch controls
     front-back speed. When roll and pitch are all centered (inside
     deadzone) the multirotor will hold position. Yaw controls yaw rate
-    as in MANUAL mode. Throttle controls climb/descent rate as in ALTCTL
+    as in MANUAL mode. Throttle controls climb/descent rate as in Altitude
     mode. 
     
   > **WARNING** Prior to Landing in *Position mode* be sure that landing is
```
hamishwillee: Can you please change this also to `**Warning**` (lower case will create the warning "note" box).

en/getting_started/px4_basic_concepts.md
```diff
@@ -30,7 +30,7 @@ In order to set up, control and interact with your PX4 drone you need to connect
 
 * [Remote Control \(RC\) Connection](../getting_started/rc_transmitter_receiver.md) - Connection between the RC handset and vehicle-based receiver that you use to manually direct the vehicle.
 * [Data/Telemetry Connection](TBD) - Connection between QGroundControl and your drone by Data Radio, Wifi or USB cable.
-* Off-board Connection - A data connection between PX4 and external microcomputer that can control PX4.
+* Off-board Connection - A data connection via serial cable or wifi between PX4 and external microcomputer that can control PX4.
```
hamishwillee: Strictly speaking the data connection can be any mechanism at all. I think this is a helpful change though.

hamishwillee:
Thanks @barzanisar - Lots of good fixes there. I put a few comments in on more questionable changes.

en/flight_modes/README.md
barzanisar: should I replace it with:
"will not attempt to hold the x, y position against wind."   ?

en/flight_modes/README.md
```diff
@@ -51,18 +51,18 @@ Throttle Command is mapped direct to Motor Speed.
...
     will crab against the wind if needed to maintain a straight line.
   - **Multirotors** Roll controls left-right speed, pitch controls
     front-back speed. When roll and pitch are all centered (inside
     deadzone) the multirotor will hold position. Yaw controls yaw rate
-    as in MANUAL mode. Throttle controls climb/descent rate as in ALTCTL
+    as in MANUAL mode. Throttle controls climb/descent rate as in Altitude
     mode. 
     
-  > **WARNING** Prior to Landing in *Position mode* be sure that landing is 
+  > **Warning** Prior to Landing in *Position mode* be sure that landing is
```
barzanisar: @hamishwillee did you mean lowercase all the letters of "Warning"?

hamishwillee: No, I meant just like this. We have a gitbook plugin that turns that syntax into a special type of note. Similarly, you can add `> **Note**` and `> **Tip**` for special note rendering.

en/flight_modes/README.md
hamishwillee: No, I'd fully revert to "will not attempt to hold position/heading against wind."

hamishwillee:
Looking good. You'll also have to resolve conflicts before we can merge.

en/flight_modes/README.md
barzanisar: In altitude mode, as I understand it, it holds z position against wind but not x and y position. Don't you think it will be better for understanding if we mention which axes position this mode holds steady, in order to make this mode distinct from position control mode? 

I also don't understand the significance of just putting "heading" here, since heading means yaw, whereas in altitude mode all angles (roll, pitch, yaw) are treated like in manual mode so they can drift with the wind.

hamishwillee: > In altitude mode, as I understand it, it holds z position against wind but not x and y position. Don't you think it will be better for understanding if we mention which axes position this mode holds steady, in order to make this mode distinct from position control mode?

Not really. I think the text was clear enough. 

> I also don't understand the significance of just putting "heading" here, since heading means yaw, whereas in altitude mode all angles (roll, pitch, yaw) are treated like in manual mode so they can drift with the wind.

Essentially heading is part of the differentiation with position control mode. In altitude mode if you're heading in a particular direction wind can blow you off course. In position control mode the vehicle will attempt to maintain its course against wind.

barzanisar: How about this? 

"More easily control vehicle altitude, and in particular reach and maintain a fixed altitude when the RC sticks are centered. The mode does not use GPS, and hence will not attempt to hold the horizontal position (position in the x and y plane of the world frame) against wind. If you're heading the vehicle in a particular direction wind can blow you off course, where as in position control mode the vehicle will attempt to maintain its course against wind."

To me, position means x, y, z position. In Altitude control mode, when the sticks are centered the current z position/altitude is held constant against the wind. This is what I confirmed from the px4 team. But then I got confused when I read that the position is not maintained against the wind. Here, shouldn't we specify that the "horizontal" position is not held against the wind?

hamishwillee: Too wordy. I'll compromise: 

> More easily control vehicle altitude, and in particular reach and maintain a fixed altitude. The mode does not use GPS, and hence will not attempt to hold the vehicle x and y position or course against wind.

hamishwillee:
Thanks @barzanisar . I'll merge this now and fix up the mode as discussed above.

hamishwillee:
FYI I did some tidy up in #164.

PR #156, opened 2018-02-19
Explained RTL_RETURN_ALT clearly

barzanisar:


hamishwillee:
Merged. Thanks for this. Never completely sure if we should work to update the firmware version: https://docs.px4.io/en/advanced_config/parameter_reference.html#RTL_RETURN_ALT . Thoughts?

PR #171, opened 2018-03-12
Pixhawk mini: known issue with MPU9250 and PX4

pietrodn:
In [this issue](https://github.com/PX4/Firmware/issues/7327), it is mentioned that the main internal IMU of the Pixhawk mini is disabled by default on PX4 because of a hardware defect on the Mini which makes it unreliable.

I think it is worth mentioning in the PX4 user guide, since this makes it impossible to fly indoor (e.g., with mocap) without an external magnetometer or an attached GPS.

### DIFF
--- a/en/flight_controller/pixhawk_mini.md
+++ b/en/flight_controller/pixhawk_mini.md
@@ -17,12 +17,13 @@ Store](https://store.3dr.com/products/3dr-pixhawk). Wiring information is availa
 
 **Processors:**
 
-- **Main Processor​:** STM32F427 Rev 3
-- **IO Processor​:** STM32F103
+- **Main Processor:** STM32F427 Rev 3
+- **IO Processor:** STM32F103
 
 **Sensors:**
 
 - **Accel/Gyro/Mag:** MPU9250
+  - [deprecated](https://github.com/PX4/Firmware/pull/7618) by the PX4 firmware
 - **Accel/Gyro:** ICM20608
 - **Barometer:** MS5611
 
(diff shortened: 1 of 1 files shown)


pietrodn:
I also removed some invisible characters which messed up the Markdown rendering.

en/flight_controller/pixhawk_mini.md
```diff
@@ -170,6 +171,13 @@ ST Microelectronics® and a NuttX real-time operating system, delivering
 excellent performance, flexibility, and reliability for controlling any
 autonomous vehicle.
 
+## Known issues
+
+- The Pixhawk mini has a [hardware defect](https://github.com/PX4/Firmware/issues/7327#issuecomment-317132917) that makes the internal MPU9250 IMU unreliable.
```
mrpollo: The issue is only affecting earlier revisions of the board that released by the manufacturer but has since corrected the issue on a later revision.

https://github.com/PX4/Firmware/issues/7327#issuecomment-372393609
> Hi this is a hardware issue there's a pin that was not routed it was left open and is causing a problem with the mpu9250. I believe this was fixed on later revisions of the board

hamishwillee: Yes. We should also say what you do if you have the bad version, and how to detect the bad version.

hamishwillee:
So David suggested this approach for working out if you have problem hardware:

> I would imagine from being power off for a long time, one can try start the mpu9250 on the command line. If the board does not have the correct wiring, the driver will fail."

See https://github.com/PX4/Firmware/issues/7327#issuecomment-372763656

@pietrodn Can you update this with the information provided?

pietrodn:
We tried starting the MPU9250 from the MAVLink console and it did not start.
It would be nice to determine which boards are affected *before* buying them, though...

hamishwillee:
@pietrodn 
> It would be nice to determine which boards are affected before buying them, though...

Yes it would. We don't have that information though. Probably best you contact manufacturer directly.

hamishwillee:
@pietrodn Can you add the info in https://github.com/PX4/px4_user_guide/pull/171#discussion_r173963171 . If not I will merge and add as a post-process.

pietrodn:
Done!

hamishwillee:
Thanks very much @pietrodn ! Now merged

PR #176, opened 2018-03-19
Reorg flight modes page

hamishwillee:
FYI @barzanisar This updates the flight modes table:
1. Rendering of columns  wasn't great because columns were fixed-width:
  image
  This makes them scale better. It expands the table wider than it could normally, and also forces a scroll bar at lower widths.
1. Moved the key down the bottom and linked from top. That way people who already know the key don't have to read it. (many can guess it). On this note, I used "M" and "S" in the mode column as well as the mode to hint at what M and S might be in the other columns.
1. Made the first column link to detail pages, where possible.
1. Made the key into a table. 

I'm going to merge now. This works :-)

### DIFF
--- a/en/flight_modes/flight_modes_summary.md
+++ b/en/flight_modes/flight_modes_summary.md
@@ -1,46 +1,86 @@
 # Flight Modes Summary
 
-The tables below summarizes flight modes for fixed wing and multicopter. Key for understanding the table is as follows:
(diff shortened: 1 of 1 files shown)


hamishwillee:
PS @barzanisar Do you think it might be worth making the "Y" into a "tick" image and the Position into a target or GPS or similar image? That tells people more quickly what we're talking about that having to read the key for "Y" or whatever ...

PR #177, opened 2018-03-20
Tidy of advanced tuning docs etc

hamishwillee:
@Stifael This is update of advanced tuning docs and some associated docs. It _needs_ your review (there are some questions and questionable changes). I will add some specific questions around those changes inline.


en/flight_modes/manual_stabilized.md
```diff
@@ -0,0 +1,21 @@
+# Manual/Stabilized Flight Mode (Multicopter)
```
hamishwillee: FYI This is a new stand along doc for Manual MC mode. This is referred to as "Manual/Stabilized" across our docset. In QGC it is either Manual or Stabilized - selecting either gives you the same mode.

en/flight_modes/manual_stabilized.md
```diff
@@ -0,0 +1,21 @@
...
+The pilot's inputs are passed as roll and pitch angle commands and a yaw rate command. Throttle is passed directly to the output mixer. The autopilot controls the attitude, meaning it regulates the roll and pitch angles to zero when the RC sticks are centered inside the controller deadzone (consequently leveling-out the attitude). The autopilot does not compensate for drift due to wind (or other sources).
+
+> **Note**
+>  * Manual input is required (RC controller, or gamepad/thumbsticks through MAVLink).
+
+## Parameters
+
+Parameter | Description
+--- | ---
+<span id="MPC_HOLD_DZ"></span>[MPC_HOLD_DZ](../advanced_config/parameter_reference.md#MPC_HOLD_DZ) | Deadzone of sticks where position hold is enabled. Default: 0.1.
```
hamishwillee: @Stifael Does 0.1 for the default value indicate 10% of the total range of the sticks bottom to top? (no units specified in the parameters).

book.json
```diff
@@ -15,6 +15,7 @@
         "bulk-redirect@git+https://github.com/Dronecode/gitbook-plugin-bulk-redirect.git",
         "toolbar@git+https://github.com/hamishwillee/gitbook-plugin-toolbar.git",
         "validate-links",
+        "mathjax@git+https://github.com/Dronecode/plugin-mathjax.git#update_cdn",
```
hamishwillee: FYI this adds the fixed mathjax plugin that renders our symbols.

en/advanced_config/advanced_mc_position_tuning.md
```diff
@@ -1,96 +1,103 @@
-<!-- Fixes page load bug in notation, courtesy https://github.com/GitbookIO/plugin-mathjax/issues/34#issuecomment-349453673 -->
-<script>
```
hamishwillee: FYI, not needed any more.

en/advanced_config/advanced_mc_position_tuning.md
```diff
@@ -1,96 +1,103 @@
...
+<span></span>
+> **Tip** Follow the instructions in the [Multicopter PID Tuning Guide](https://docs.px4.io/en/advanced_config/pid_tuning_guide_multicopter.html) *before* doing any of the higher-level related control tuning described here. Do not use the advanced position control tuning parameters to fix bad tracking or vibration!
+
+## Overview
 
-## Short Background
-The position controller consists of an outer P-position control loop and an inner PID-velocity control loop. Depending on the control mode either
-the velocity control loop only is active or both. For the remaining part of this tutorial the term **position-control** represents the position and velocity control loop
-activated and **velocity-control** when just the velocity control loop is in use.
+Higher-level tuning applies to any parameter that has an effect on the desired position/velocity setpoint, in contrast to [lower-level tuning](pid_tuning_guide_multicopter.md) that influences the tracking.
```
hamishwillee: @Stifael I don't understand the difference. Lower level tuning influences the tracking of "what"? I "think" you might be saying that lower level tuning influences how well the autopilot tracks setpoints, but I'm not then sure how the higher level tuning differs.

en/advanced_config/advanced_mc_position_tuning.md
```diff
@@ -1,96 +1,103 @@
...
 
-## Precondition
-Before doing any higher-level related control tuning, please follow first the instruction [PID-tuning](https://docs.px4.io/en/advanced_config/pid_tuning_guide_multicopter.html).
-It is not recommended to use the advanced position control tuning parameters as a fix for bad tracking or vibration.
+The set of higher-level tuning parameters can be split into two sections: tuning parameters for [Manual/Stabilized](#manual_mode) mode and tuning parameters for [Mission](#mission_mode) mode. Some parameters will have an effect on both modes.
```
hamishwillee: @Stifael 
1. FYI Auto mode is now called "Mission". In the user guide we always refer to the modes by the name that they have in QGC selection.
2. Similarly, I THINK that when you refer to manual position control mode you mean "Manual/Stabilized" mode.

en/advanced_config/advanced_mc_position_tuning.md
```diff
@@ -1,96 +1,103 @@
...
-be mapped directly to the physical quantity. 
+> **Note** The position controller ([diagram here](https://dev.px4.io/en/flight_stack/controller_diagrams.html#multicopter-position-controller)) consists of an outer **P** position-control loop and an inner **PID** velocity-control loop. 
+Depending on the control (flight) mode either both loops are active or just the velocity control loop. 
+For the remainder of this topic the term **position-control** represents the case where both loops are active while **velocity-control** refers to the case when only the velocity control loop is in use.
+
+## Manual/Stabilized Mode {#manual_mode}
+
+In [Manual/Stabilized](../flight_modes/manual_stabilized.md) mode the stick inputs are mapped either to **position-control** or **velocity-control**.
+ 
+Position-control is active when the stick inputs are within the deadzone [MPC_HOLD_DZ](../advanced_config/parameter_reference.md#MPC_HOLD_DZ), and velocity-control otherwise. All the parameters below are tuning parameters and cannot be mapped directly to the physical quantity.
```
hamishwillee: The position-control and velocity-control terminology are a bit annoying in the context of describing flight modes, because we tend to think of position control as being a mode that controls the position of the vehicle - like altitude mode ... and manual mode is "not position controlled" in this context - it can drift with wind. 

I understand that you're saying that in the deadzone both the velocity and position cotnrol loops are active, while outside the zone just the velocity loop is active (and I guess that is because inside we're using the position loop to level attitude?) 

But it would be great if we could use a different terminology that didn't clash.

en/advanced_config/advanced_mc_position_tuning.md
```diff
@@ -1,96 +1,103 @@
...
 
 #### MPC_ACC_HOR_MAX
 
-This parameter is used for position-control in the horizontal direction, where the vehicle is supposed to stay at the current location. The limit for the rate of change of the velocity setpoint is defined by 
- `MPC_ACC_HOR_MAX`. This parameter should be set larger than any of the other acceleration related parameters in the horizontal direction.
+This parameter is used for position-control in the horizontal direction, where the vehicle is supposed to stay at the current location.
```
hamishwillee: @Stifael So now I'm confused, because this does sound like you mean "position control to keep location fixed in 3d co-ordinate system". Perhaps it is the "Manual/Stabilized" mode that I have wrong - do you actually mean "any mode that has a remote control component"?

hamishwillee:
OK, @Stifael That's, all my comments. A [rendered view is here](https://hamishwillee.gitbooks.io/ham_px4userguide/content/v/adv_mctuning/en/advanced_config/advanced_mc_position_tuning.html).

en/flight_modes/manual_stabilized.md
Stifael: yes

en/advanced_config/advanced_mc_position_tuning.md
Stifael: > lower level tuning influences how well the autopilot tracks setpoints

That is correct. The goal of the P/PID controller is to keep the error between the desired setpoint and estimate of the vehicle state as small as possible. Wrong P/PID Gains can lead to instability. 
The higher level tuning, on the other hand, has no influence on the tracking at all. It just adjusts the setpoint. For instance, instead of step input, you will have a ramp input. Furthermore, it is not possible to make the system instable by adjusting the higher level tuning parameter. 
By the way, the name "higher level" is something that I made up. The point is that there has to be a clear difference between the two kind of parameters: One is for keeping the error small, the other set only effects the setpoint

en/advanced_config/advanced_mc_position_tuning.md
Stifael: 1. Ok
2. No. Manual/Stabilized is the mode where you have no position/velocity control active. The user "controls" the attitude and thrust via RC stick inputs. What I am referring to is Manual for RC stick inputs together with position/velocity control active. In QGC it is called Position.

en/advanced_config/advanced_mc_position_tuning.md
Stifael: Ok, but where is the definition? To me the naming are already messed up anyway, and also not correct in QGC.  I will quickly try to leverage everything and maybe the naming in general is something to review. There are terminology that I cannot change, even if a mode has a name that has a similar name as the control loop.
I first write down the mode (how it is called in PX4). In parenthesis I write down the stage at which the control pipeline is entered. 
The entire control-pipeline is:
position-velocity-attitude-rate-control (only P/PID loops are considered): 

If I write position control, then the entire pipeline is used. If I write velocity control, then the pipeline from velocity up to rate control is active. Attitude control means that attitude + rate is active, and rate refers to the control part where only rate is active. 

Manual (everything that has RC stick inputs)
-- Acro (rate-control)
-- Rattitude (attitude + rate control): depending on stick (deadzone threshold) attitude or just rate control is active
-- Stabilized (throttle + attitude-control: throttle + roll/pitch/yawt)
-- Altitude (position/velocity and attitude control: Altitude is controlled using position/velocity control                    depending on stick (deadzone threshold); roll/pitch/yaw used for xy
-- Position (position/velocity control in 3D: depending on stick (deadzone threshold) either position control or just velocity is active).

Auto:
-- Mission/RTL/Reposition and so on: they all use position control

Offboard:
Each stage of the control pipeline can be accessed trough offboard 

It is important to note that I cannot call position/velocity/atttiude/rate-control differently, because that is the correct name and represents the kind of control. In contrast, a mode name is something that we made up and has different naming convention based on the autopilot.

hamishwillee: Thanks for taking the time to put that explanation together. I will think about it tomorrow. If the naming convention is correct then I will need to find another way around this or live with it.

en/advanced_config/advanced_mc_position_tuning.md
hamishwillee: OK, so I assume this would be right if I just change to heading/link to Position mode.

Stifael:
@hamishwillee another option would be to split this advanced tuning section between user and dev guide. In the user guide it would just be about what parameters can be changed and what affects it will have. More controller specific details will be in the devguide

en/advanced_config/advanced_mc_position_tuning.md
```diff
@@ -1,96 +1,111 @@
...
+This document provides an overview of the multicopter position-control tuning parameters that affect the value of a desired setpoint (as contrasted with those parameters that affect how well the vehicle tracks the setpoint). 
+
+> **Warning** This guide is for advanced users/experts.
+
+<span></span>
+> **Tip** Follow the instructions in the [Multicopter PID Tuning Guide](../advanced_config/pid_tuning_guide_multicopter.md) *before* doing any of the higher-level related control tuning described here. Do not use the advanced position control tuning parameters to fix bad tracking or vibration!
+
+## Overview
+
+The input to the P/PID controller is a *desired setpoint* that the vehicle should attempt to track. [PID Tuning](../advanced_config/pid_tuning_guide_multicopter.md) ("Lower level" tuning) aims reduce the error between the desired setpoint and the estimate of the vehicle state. Poor P/PID Gains can lead to instability.
```
hamishwillee: @Stifael Thanks for explanation. I have updated the overview to provide a definition of the different types of tuning. To make this work I did an "implicit" renaming:
- Lower:  PID tuning, 
- Higher: Setpoint value tuning
I think that is more intuitive. 

Anyway, can you please read this introduction section and confirm it is OK. In particular, are there any other impacts to poor tuning of the setpoint value?

hamishwillee:
@Stifael 

> @hamishwillee another option would be to split this advanced tuning section between user and dev guide. In the user guide it would just be about what parameters can be changed and what affects it will have. More controller specific details will be in the devguide

I think this is OK in the guide. The information you put in the short background and [here](https://github.com/PX4/px4_user_guide/pull/177#discussion_r175686355) about the control naming might be useful in the devguide though. 

I think this is OK (modulo your comment to my other post). Can you confirm one thing for me though.  For position control in MC, this only corrects position in the deadzone when not moving - it isn't like FW where the heading is also maintained against wind when the vehicle is moving?

en/advanced_config/advanced_mc_position_tuning.md
Stifael: yes, correct. But is it not "aims at ..."?

The only impact for poor setpoint tuning are the extremes: very jerky or very slow/smooth (with large delay in direction) flight motion

Stifael:
>  For position control in MC, this only corrects position in the deadzone when not moving - it isn't like FW where the heading is also maintained against wind when the vehicle is moving

For MC, heading is always controlled, independent of position/velocity control. At least for now, you can consider heading as part of the attitude control.

hamishwillee:
>> For position control in MC, this only corrects position in the deadzone when not moving - it isn't like FW where the heading is also maintained against wind when the vehicle is moving

> For MC, heading is always controlled, independent of position/velocity control. At least for now, you can consider heading as part of the attitude control.

Sorry to harp on this, but I want to make sure I'm not misunderstanding you. If I'm in position control mode  and I'm  using just the pitch stick to move forward (other sticks centred) then on FW the theory is that the vehicle will hold the horizontal line forward - it won't move sideways  even if wind blows it sideways.  Are you saying it is exactly the same for MC - the vehicle will hold the indicated path?

en/advanced_config/advanced_mc_position_tuning.md
hamishwillee: "aims to". I'll fix this.

hamishwillee:
>> For position control in MC, this only corrects position in the deadzone when not moving - it isn't like FW where the heading is also maintained against wind when the vehicle is moving

> For MC, heading is always controlled, independent of position/velocity control. At least for now, you can consider heading as part of the attitude control.

@Stifael 
Sorry to harp on this, but I want to make sure I'm not misunderstanding you. If I'm in position control mode  and I'm  using just the pitch stick to move forward (other sticks centred) then on FW the theory is that the vehicle will hold the horizontal line forward - it won't move sideways  even if wind tries to blow it sideways.  

Are you saying it is exactly the same for MC - the vehicle will hold the indicated path?

I think you are also saying that this will be true in other modes - presumably it is dependent on GPS though - so not in Manual/Stabilised?

Stifael:
ah, now I get what you mean. I thought you are talking about heading = yaw.
No, in MC the vehicle will not keep the exact path. As soon the user moves the stick outside the deadzone, the vehicle will exit position control and only tracks velocity. If there is a strong wind, then the vehicle will drift.

hamishwillee:
Thanks. Sorry for being imprecise in first case! I'll merge this now - I think with the new introduction it is a little more accessible.

PR #180, opened 2018-03-23
Update Flight modes table

hamishwillee:
A number of changes around Flight modes:
- Renamed "rtl.md" to return
- Ordered TOC to put auto modes after manual modes

For table ordered similar modes together to show flow - ie manual -> altitude -> position, and acro -> rattitude. Updated descriptions to relate them.


hamishwillee:
@barzanisar @bkueng FYI only, this updates the [Flight mode summary](https://docs.px4.io/en/flight_modes/flight_modes_summary.html#multicopter), in particular for MC.

THe changes are described mostly above. I guess the one I want you to notice are:
1. The descriptions of manual, altitude, position modes now reference each other - ie related to each other. Same for acro and rattitude. I think that helps understand very quickly what is different
2. Previously we referred to angle and angle rate commands. For a user an angle command translates to a speed in a particular direction, while an angle rate translates to an angular rotation. So in the body of the text I now refer to how the vehicle moves, not the type of command. 

Still lots to do, this is just to give you an idea of where I am heading with this.

en/flight_modes/flight_modes_summary.md
```diff
@@ -293,21 +316,26 @@ Multicopter hovers at the current GPS position and altitude.
...
  <td colspan="3">Auto</td>
  <td>Y</td>
- <td>The vehicle obeys a position, velocity or attitude setpoint provided over MAVLink (often from a companion computer connected via serial cable or wifi).</td>
+ <td>Vehicle obeys a position, velocity or attitude setpoint provided over MAVLink (often from a companion computer connected via serial cable or wifi).</td>
 </tr>
  
 </tbody></table>
 
 
+## VTOL
```
hamishwillee: @bkueng I added this section. I've pretty much said VTOL does the flight modes for MC/FW based on what config it is in. The exception being offboard, which is true for both configs, even though it is only in MC normally.

en/flight_modes/flight_modes_summary.md
```diff
@@ -191,93 +181,126 @@ th {
...
  <td></td>
- <td><p>This mode allows manual control with assistance from autopilot to stabilize attitude.</p>
-<p>Centered RC sticks level-out the attitude.</p>
+ <td><p>RC/manual mode where centered sticks level vehicle (position is not stabilized).</p>
+   <p>
+   <ul>
+    <li>Centered RP sticks level vehicle.</li>
+    <li>Outside center:
+      <ul>
+       <li>RP sticks control left/right and forward/back speed (respectively).</li>
```
bkueng: I'm not sure about this. The stick controls the tilt angle, which then affects the vehicle's speed. To control the speed directly, you would need to have position/velocity information.

hamishwillee: That was intentional - the user doesn't really care about angles, just the way the vehicle moves (ie left /right). But I thought about this a bit and this is more a "techical" version, so now try say both (added inhttps://github.com/PX4/px4_user_guide/pull/185)

Basically it now says:


```
       <li>Roll/Pitch sticks control tilt angle in those orientations, resulting in corresponding left-right and forward-back movement.</li>
       <li>Throttle stick controls up/down speed (and movement speed in other axes).</li>
       <li>Yaw stick controls rate of angular rotation above the horizontal plane.</li>
```

bkueng: Thanks

PR #186, opened 2018-03-27
Update position mode (MC)

hamishwillee:
This makes position mode docs consistent (MC only).

FYI, now going through each mode and making sure that content in "overview", summary table, and detail page for mode are consistent.

The detail page basically has the same content for mode as the overview, a technical overview from the summary table, and a listing of parameters that affect it. 

I plan to merge this tomorrow.

### DIFF
--- a/en/flight_modes/flight_modes_summary.md
+++ b/en/flight_modes/flight_modes_summary.md
@@ -200,13 +200,13 @@ th {
  <td>S<sub>rate</sub></td>
  <td>S<sup>+</sup></td>
  <td><a href="#key_position_fixed"><img src="../../assets/site/position_fixed.svg" title="Position fix required (e.g. GPS)" width="20px" /></a></td>
- <td><p>RC/manual mode where centered sticks level vehicle and hold it to fixed position and altitude against wind.
+ <td><p>RC/manual mode where RPT sticks control <em>speed</em> in corresponding directions. Centered sticks level vehicle and hold it to fixed position and altitude against wind.
   <ul>
     <li>Centered RPT sticks hold x, y, z position steady against any wind and levels attitude.</li>
     <li>Outside center:
       <ul>
-       <li>Roll/Pitch sticks control tilt angle in those orientations, resulting in corresponding left-right and forward-back movement.</li>
-       <li>Throttle stick controls up/down speed (and movement speed in other axes).</li>
+       <li>Roll/Pitch sticks control speed over ground in left-right and forward-back directions (respectively) relative to the "front" of the vehicle.</li>
+       <li>Throttle stick controls speed of ascent-descent.</li>
        <li>Yaw stick controls rate of angular rotation above the horizontal plane.</li>
       </ul>
     </li>
--- a/en/flight_modes/flight_modes_summary.md
+++ b/en/flight_modes/flight_modes_summary.md
@@ -200,7 +200,7 @@ th {
  <td>S<sub>rate</sub></td>
  <td>S<sup>+</sup></td>
  <td><a href="#key_position_fixed"><img src="../../assets/site/position_fixed.svg" title="Position fix required (e.g. GPS)" width="20px" /></a></td>
- <td><p><a href="#manual_stabilized_mc">Manual/Stabilized</a> mode with <em>3D position stabilization</em> (centered sticks level vehicle and hold it to fixed position and altitude against wind).
+ <td><p>RC/manual mode where centered sticks level vehicle and hold it to fixed position and altitude against wind.
   <ul>
     <li>Centered RPT sticks hold x, y, z position steady against any wind and levels attitude.</li>
     <li>Outside center:
@@ -225,7 +225,7 @@ th {
  <td>S<sub>rate</sub></td>
  <td>S<sup>+</sup></td>
  <td><a href="#key_position_fixed"><img src="../../assets/site/position_fixed.svg" title="Altitude required (e.g. Baro, Rangefinder)" width="20px" /></a><sup><a href="#baro_only">++</a></sup></td>
- <td><p><a href="#manual_stabilized_mc">Manual/Stabilized</a> mode with <em>altitude stabilization</em> (centered sticks level vehicle and hold it to fixed altitude).
+ <td><p>RC/manual mode where centered sticks level vehicle and hold it to a fixed altitude (but not horizontal position).
   <ul>
     <li>Centered sticks:
       <ul>
@@ -253,7 +253,7 @@ th {
  <td>S<sub>rate</sub></td>
  <td>M</td>
  <td></td>
- <td><p>RC/manual mode where centered sticks level vehicle (position is not stabilized).</p>
+ <td><p>RC/manual mode where centered sticks level vehicle (only - position is not stabilized).</p>
    <p>
    <ul>
     <li>Centered RP sticks level vehicle.</li>
--- a/en/flight_modes/position_mc.md
+++ b/en/flight_modes/position_mc.md
@@ -1,17 +1,14 @@
 # Position Flight Mode (Multicopter)
 
-![GPS](../../images/flight_modes/GPS_s.png)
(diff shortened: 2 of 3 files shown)


en/flight_modes/flight_modes_summary.md
```diff
@@ -200,7 +200,7 @@ th {
  <td>S<sub>rate</sub></td>
  <td>S<sup>+</sup></td>
  <td><a href="#key_position_fixed"><img src="../../assets/site/position_fixed.svg" title="Position fix required (e.g. GPS)" width="20px" /></a></td>
- <td><p><a href="#manual_stabilized_mc">Manual/Stabilized</a> mode with <em>3D position stabilization</em> (centered sticks level vehicle and hold it to fixed position and altitude against wind).
+ <td><p>RC/manual mode where centered sticks level vehicle and hold it to fixed position and altitude against wind.
   <ul>
     <li>Centered RPT sticks hold x, y, z position steady against any wind and levels attitude.</li>
     <li>Outside center:
```
bkueng: Not part of the PR, but outside center controls the velocity

en/flight_modes/position_mc.md
```diff
@@ -1,17 +1,14 @@
...
 
-![GPS](../../images/flight_modes/GPS_s.png)
+[<img src="../../assets/site/difficulty_1.svg" title="Difficulty (Easiest)" width="30px" />](../getting_started/flight_modes.md#key_difficulty)&nbsp;[<img src="../../assets/site/remote_control.svg" title="Manual/Remote control required" width="30px" />](../getting_started/flight_modes.md#key_manual)&nbsp;[<img src="../../assets/site/position_fixed.svg" title="Position fix required (e.g. GPS)" width="30px" />](../getting_started/flight_modes.md#key_position_fixed)
 
-*Position* mode is an assisted RC mode, in which the vehicle can easily be brought to a stop with a level attitude (irrespective of wind and other forces), and in which the flight controls are intuitive and straightforward. 
+*Position* mode is an easy-to-fly RC mode in which the vehicle can be "locked" to a position in 3D space (irrespective of wind and other forces) by releasing (centering) the RC sticks.
 
-> **Tip** Position mode is the safest manual mode for new fliers.
+> **Tip** Position mode is the safest manual mode for new fliers. It is much like [Altitude](../flight_modes/altitude.md) mode, but additionally stabilizes horizontal position.
```
bkueng: Hmm, I would not say that. I would rather say that Altitude is much like Stabilized/Manual, except that the altitude is stabilized as well (the roll/stick input is treated the same way in both modes).
Whereas in Position mode the velocity is controlled, which from a flight experience is quite a bit different from the direct tilt angle in Altitude (for example letting go of the stick in position mode with a high velocity, the vehicle will actively brake. In Altitude mode, it continues and just slows down through air resistance).

en/flight_modes/position_mc.md
```diff
@@ -1,17 +1,14 @@
...
+*Position* mode is an easy-to-fly RC mode in which the vehicle can be "locked" to a position in 3D space (irrespective of wind and other forces) by releasing (centering) the RC sticks.
 
-> **Tip** Position mode is the safest manual mode for new fliers.
+> **Tip** Position mode is the safest manual mode for new fliers. It is much like [Altitude](../flight_modes/altitude.md) mode, but additionally stabilizes horizontal position.
 
-<span></span>
-> **Tip** Position mode is much like [Altitude](../flight_modes/altitude.md) mode, but additionally stabilizes horizontal position.
+Vehicle movement is intuitive. The roll and pitch sticks move the vehicle left-right and front-back over ground, with speed increasing proportionally with distance from the center (the sticks actually control the tilt angle, which in turn alters the speed in each direction). The throttle controls climb/descent speed. The yaw input controls angular rate of rotation over the horizontal plane.
```
bkueng: Same as above: `the sticks actually control the tilt angle` -> velocity is controlled.

bkueng: You might want to mention that roll/pitch sticks control the direction of motion with respect to where the vehicle points to. For example if it points towards yourself, moving the roll stick to the right will make the vehicle move to the left.

en/flight_modes/position_mc.md
```diff
@@ -27,7 +24,13 @@ The diagram below shows the mode behaviour visually (for a mode 2 transmitter).
 
 ## Technical Summary
 
-Roll controls left-right speed over ground, pitch controls front-back speed over ground, and throttle controls climb/descent speed (as in [Altitude](../flight_modes/altitude.md) mode). Yaw controls yaw *rate* over horizontal plane (as in [Manual/Stabilized](../flight_modes/manual_stabilized_mc.md) mode).  When roll, pitch and throttle inputs are all centered (inside deadzone) the multirotor will hold position (in x, y, z frame), compensating for drift due to wind (or other sources).
+Manual/Stabilized mode with <em>3D position stabilization</em> (centered sticks level vehicle and hold it to fixed position and altitude against wind).
+
+* Centered RPT sticks (in RC deadzone) hold x, y, z position steady against any wind and levels attitude.
+* Outside center:
+  * Roll/Pitch sticks control tilt angle in those orientations, resulting in corresponding left-right and forward-back movement.
+  * Throttle stick controls up/down speed (and movement speed in other axes).
```
bkueng: `(and movement speed in other axes)` If I understand this correct, it's not right: throttle really only controls up/down speed, horizontal is not affected.
You could mention that centering the throttle stick will make the vehicle keep the altitude.

hamishwillee:
> Can I ask why you changed 'speed over ground' to use tilt angle? Speed over ground is more accurate for the position mode.

Because that is what I understood from your previous comments :-( All very frustrating. 
So from https://github.com/PX4/px4_user_guide/pull/186#discussion_r177424956 ..
I think you are saying that stabilized and altitude control the yaw angle - which leads to a control over velocity *indirectly* (essentially what I currently wrote for position). While in position mode the sticks control vehicle velocity directly?

hamishwillee:
Thanks for the review @bkueng ! I have modified as you have suggested. Can you please recheck.

In particular are the parameters [listed in the detail document](https://github.com/PX4/px4_user_guide/pull/186/files#diff-dc6c3bac6acb8d72273b11539e157c63R36) correct? By "correct" I mean, "are they a good set to talk about" for people who want to "tune" this mode. If not, what is missing?

bkueng:
You might want to mention the different speed limits (takeoff, landing, MPC_VEL_MANUAL, ... I'm not sure myself which ones are interesting). Also Stick Expo is worth mentioning (MPC_XY_MAN_EXPO).

hamishwillee:
> You might want to mention the different speed limits (takeoff, landing, MPC_VEL_MANUAL, ... I'm not sure myself which ones are interesting). Also Stick Expo is worth mentioning (MPC_XY_MAN_EXPO).

Docs for MPC_VEL_MANUAL say "for manual controlled mode" - does this mean for _all_ manual controlled modes?

Same for MPC_XY_MAN_EXPO - does it apply on Acro, Manual, stabilized etc

bkueng:
No, all of these are only for Position control - the description could be better there

hamishwillee:
Perhaps just better to add a note saying "Everything in https://docs.px4.io/en/advanced_config/parameter_reference.html#multicopter-position-control is relevant"?

IF there are specific ones you think need to be highlighted I am very happy to add them, but I need you to tell me exactly which ones you would want to be mentioned here.

bkueng:
> Perhaps just better to add a note saying "Everything in https://docs.px4.io/en/advanced_config/parameter_reference.html#multicopter-position-control is relevant"?

I agree. The ones that are relevant/interesting depend a bit on the use-case and whether you like a specific behavior with the default param values.

hamishwillee:
Thanks @bkueng . I added link to the MC position control parameters in https://github.com/PX4/px4_user_guide/commit/6ab388d2e4717881642556e232977f11d2a717d3

PR #197, opened 2018-04-13
adding registered trademarks according to brand guide

barzanisar:
Added trademark next to registered company names the first time they are used in each page of pixhawk series section and wherever DJI is used


hamishwillee:
@barzanisar Never really liked trademark "markup". Did someone ask you to do this? 

If it is necessary for whatever reason we probably should create a list of all things we know might be registered and do the change globally.

barzanisar:
Yes, Kevin Sartori asked me to do it. 
Is it possible to automate the process of doing it globally? We need to add it on every page the first time the company's name is used.

hamishwillee:
There is no way to automate this. I've asked Kevin what the relative cost of not doing this is, since to do it right and keep it right is quite a lot of effort for zero gain for our readership.

barzanisar:
Kevin did say to me that ideally it should be on every page but for now we can just add it in the Pixhawk series section and for DJI. 

Is he now saying that we don't need them at all?

hamishwillee:
Hi @barzanisar  

From Kevin: 

> just add the trademark symbol to the new pages that you are working on for Pixhawk 4. We can establish this a best practice with the new documentation pages. No need for me to reverse-change all the documents.

I still don't know why we need to do this, so am not proposing to add it to a best practices section. If a good explanation of why this is useful or needed gets provided then we could do that. 

How about we just merge this and you continue any additional work just limited to the new Pixhawk 4 pages you are creating? (No need to undo any "extra" work you have done).

Cheers
H

hamishwillee:
P.S. Do you think that you could create a Pixhawk 4 quickstart modelled on this one: https://docs.px4.io/en/assembly/quick_start_pixhawk.html ?

I think it is really good to have a dummies guide to basic assembly for each board. The Pixhawk 1 is a good model. All it lacks is infographic/setup for platforms other than MC - ie Plane.

barzanisar:
> How about we just merge this and you continue any additional work just limited to the new Pixhawk 4 pages you are creating? (No need to undo any "extra" work you have done).

Ok, thanks!

> P.S. Do you think that you could create a Pixhawk 4 quickstart modelled on this one: https://docs.px4.io/en/assembly/quick_start_pixhawk.html ?

Yes, I already have that ready, except the infographic setup.

hamishwillee:
Thanks. Merged.

>> P.S. Do you think that you could create a Pixhawk 4 quickstart modelled on this one: https://docs.px4.io/en/assembly/quick_start_pixhawk.html ?

> Yes, I already have that ready, except the infographic setup.

Awesome. Infographic will be very cool. Especially if we can get one for plane and VTOL too. That has been a hole in all our docs

PR #198, opened 2018-04-13
Added an overview of all sections

barzanisar:


en/README.md
```diff
@@ -5,7 +5,18 @@
...
+- [Airframe Builds](airframes/README.md) - gives examples of how to setup different flight controller hardware on common airframes/vehicles.
+- [Flying](flying/README.md) - covers flying basics and explanation of different flight modes, including manual, autopilot-assisted and fully autonomous mission modes.
+- [Advanced Features](advanced_features/README.md) - contains topics related to some of the more advanced features of the PX4 autopilot such as RTK GPS and precision landing.
+- [Advanced Configuration](advanced_config/README.md) - how to tune parameters of the vehicle that can affect its flight (e.g. PID gains) and topics about less-common sensors and peripherals.
+- [Peripheral Hardware](peripherals/README.md) - topics about external sensors and optional peripheral hardware that you can use with the autopilot.
+- [Autopilot Hardware](flight_controller/README.md) - describes different autopilot hardware that can be used to run the PX4 flight stack.
```
hamishwillee: Maybe "describes different flight controller hardware"

en/README.md
```diff
@@ -5,7 +5,18 @@
 
 PX4 is the *Professional Autopilot*. Developed by world-class developers from industry and academia, and supported by an active world wide community, it powers all kinds of vehicles from racing and cargo drones through to ground vehicles and submersibles.
 
-This guide contains everything you need to assemble, configure, and safely fly a PX4-based vehicle.
+This guide contains everything you need to assemble, configure, and safely fly a PX4-based vehicle, including:
+
+- [Getting Started](getting_started/README.md) - basic concepts for beginners needed to build and fly an unmanned vehicle using PX4.
+- [Basic Assembly](assembly/README.md) - how to mount flight controller hardware and wire/connect most important peripherals.
+- [Basic Configuration](config/README.md) - how to install PX4 firmware onto the flight controller hardware, select an airframe, configure/calibrate the core sensors that PX4 needs (e.g. compass, GPS, gyro etc.), setup the battery, setup the radio control transmitter and map flight modes to its switches.
+- [Airframe Builds](airframes/README.md) - gives examples of how to setup different flight controller hardware on common airframes/vehicles.
```
hamishwillee: You could generally remove the intro word. So this might be "Examples of how to ..."

en/README.md
```diff
@@ -5,7 +5,18 @@
...
+- [Basic Assembly](assembly/README.md) - how to mount flight controller hardware and wire/connect most important peripherals.
+- [Basic Configuration](config/README.md) - how to install PX4 firmware onto the flight controller hardware, select an airframe, configure/calibrate the core sensors that PX4 needs (e.g. compass, GPS, gyro etc.), setup the battery, setup the radio control transmitter and map flight modes to its switches.
+- [Airframe Builds](airframes/README.md) - gives examples of how to setup different flight controller hardware on common airframes/vehicles.
+- [Flying](flying/README.md) - covers flying basics and explanation of different flight modes, including manual, autopilot-assisted and fully autonomous mission modes.
+- [Advanced Features](advanced_features/README.md) - contains topics related to some of the more advanced features of the PX4 autopilot such as RTK GPS and precision landing.
+- [Advanced Configuration](advanced_config/README.md) - how to tune parameters of the vehicle that can affect its flight (e.g. PID gains) and topics about less-common sensors and peripherals.
```
hamishwillee: Perhaps 

> Advanced tuning and configuration for non-standard airframes and less common sensors and peripheral hardware.

The theory is that out of the box the basic calibration should be enough. This is just for when you're working on a new airframe or have significantly modified a standard airframe or want to do fine tuning.

hamishwillee:
Thanks @barzanisar . I am in two minds about this section. I chose not to add something like this because
* the structure is still something that might change, resulting in maintenance
* generally you can infer the content from the sidebar heading. 

However I don't feel strongly about this. Just thing perhaps make it a little more "terse".

barzanisar:
I think something like this would be useful for users to find exactly what they need rather than going through each section separately.

hamishwillee:
> I think something like this would be useful for users to find exactly what they need rather than going through each section separately.

Not really - what is there in this text that people can't infer from the TOC title? I don't want people to have to read stuff that they can easily guess, especially if it pushes down information like how to get help.

That said, you make an excellent point that we might better help users find out exactly what they need. Instead of a simple index, how about we provide **signposts** for the _different types_ of users we have (a bit like [this](https://www.dronecode.org/contribute/)). We have users who are building airframes, users who are buying RTF and just want to fly and users who want to create new airframes or add peripherals. Making it clear what sections each type use might be is more clearly of value.

What do you think?

barzanisar:
Yes, this is actually a very good idea. I'll start working on that and create another PR soon.

hamishwillee:
Great. Thanks!

barzanisar:
I reopened this pull request to confirm a few things.

What are the examples of RTF frames that come mounted with Pixhawk/Pixracer and have important peripherals already wired? I am only aware of DIY frames.

I was thinking of reusing this overview guide and assigning user levels to it, just like we did for (flight modes)[https://docs.px4.io/en/flight_modes/]. 

Here's my proposal:

We can have a table with three columns, titled "User level", "Section", "Description". The Section and Description will be taken from this PR. 

There will be a key in the beginning for the 3 user levels you mentioned above:
- Green: I want to fly a Ready-To-Fly airframe.
- Orange: I want to build an (airframe)[https://docs.px4.io/en/airframes/] already supported by PX4. 
- Red: I am creating a new airframe and want to add extra peripherals.

Table:
- Getting Started: G, O, R
- Basic Assembly: O, R
- Basic Configuration: G, O, R
- Airframe Builds: O
- Flying: G, O, R
- Advanced Features: R 
- Advanced Configuration: R 
- Peripheral Hardware: R
- Autopilot Hardware: O, R
- Development: R

I think the descriptions in the table will keep things transparent among different users so for example a Orange user may also be interested in some of Red user's content.

What do you think?

barzanisar:
@hamishwillee I have added guidelines for different users like you mentioned. Can you please review again? Thanks!

hamishwillee:
@barzanisar Thanks for this. I have made some modifications to make it (IMO) more readable. Can you please review the new version.

hamishwillee:
Merged! Thanks @barzanisar

PR #202, opened 2018-04-17
Frsky Telemetry doc update

hamishwillee:
This updates FrSky Telemetry docs. 

@bkueng Initial update is "minimal response" to https://github.com/PX4/Firmware/issues/9240#issuecomment-380526217 . Essentially it adds diagrams for pixracer connections, links to additional information, and some info about other boards. I have some more questions about this that I will ask inline.

Quite a lot more information required to make this actually useful:
1. What receivers are supported
2. What transmitters are supported
3. What information can be transmitted
4. Info about support for smartport sensors.

Should probably do some of that in this update.


en/peripherals/frsky_telemetry.md
```diff
@@ -1,27 +1,65 @@
...
-## LuaPilot: Telemetry Screen on the Taranis
-If you have a Taranis that can receive the telemetry stream (e.g. a Taranis X9D
-Plus), you can install a script that shows you a nice overview of the received
-telemetry:
-![](../../images/taranis_telemetry.jpg)
+There is no need to configure anything in PX4 itself. Once connected (as discussed below) FrSky telemetry auto-starts on Pixracer and detects D or S mode. 
+
+#### Connecting Receivers with D-port
+
+Connect the Pixracer FrSky TX line to the receiver's RX line. Also connect GND (This applies to D4R-II and similar).
```
hamishwillee: @bkueng I think the image below is not associated with this text. Ie it looks to me like this is how you connect the transmitter for normal use (ie not FrSky telemetry). Is that correct? 
For FrSky telemetry in this D-port case I THINK this is saying that you connect the port labeled FSout (TX line) to the Receiver RX line -but we have no image of that. Is that right?

en/peripherals/frsky_telemetry.md
```diff
@@ -1,27 +1,65 @@
...
+#### Connecting Receivers with S-port
+
+Connect the Pixracer FrSky TX and RX lines together (solder the wires together) to the X series receiver's S.port pin. 
+Also connect GND. The S-port connection is shown below (using the provided I/O Connector).
+
+![Grau b Pixracer FrSkyS.Port Connection](../../assets/flight_controller/pixracer/grau_b_pixracer_frskys.port_connection.jpg)
+
+![Pixracer FrSkyS.Port Connection](../../assets/flight_controller/pixracer/pixracer_FrSkyTelemetry.png)
+
+
```
hamishwillee: FYI, I guess I should also link here to the existence of custom cables.

en/peripherals/frsky_telemetry.md
```diff
@@ -1,27 +1,65 @@
...
+
+### Pixhawk v2
+
+#### Configuration
+
+After connecting the board (see below), you also need to configure the port for use with FrSky, by setting the [SYS_COMPANION](../advanced_config/parameter_reference.md#SYS_COMPANION) parameter to 10 (see [Parameters](../advanced_config/parameters.md) for information on how).
```
hamishwillee: Note that I didn't talk here about modifying the startup script. I see that as Developer land configuration.

en/peripherals/frsky_telemetry.md
```diff
@@ -1,27 +1,65 @@
...
+
+You will need to connect via a UART to S.PORT adapter board, or using a [ready-made cable](#ready_made_cable).
+
+<!-- ideally add diagram here -->
+
+### Pixhawk Pro
+
+[Pixhawk 3 Pro](../flight_controller/pixhawk3_pro.md) can be connected to TELEM4 (no additional software configuration is needed).
+
+You will need to connect via a UART to S.PORT adapter board, or using a [ready-made cable](#ready_made_cable).
```
hamishwillee: @bkueng Is this correct? More specifically, which FMU version adds the electronics so you no longer need the adaptor?

en/peripherals/frsky_telemetry.md
```diff
@@ -1,27 +1,65 @@
...
+
+
+## RC Transmitter Setup
+
+### Taranis - LuaPilot Setup
```
hamishwillee: I moved this down a level because we should one day allow extension to have other transmitters.

en/peripherals/frsky_telemetry.md
```diff
@@ -35,16 +73,15 @@ You can see:
...
 local BattLevelmAh = -1  -- use the battery level calculation from the vehicle
 local SayFlightMode = 0  -- there are no WAV files for the PX4 flight modes
 ```
 
 
-
-## Cable/adapter sources
+## Cable/adapter sources {#ready_made_cable}
 
 Ready-made cables (which include the required adapters) are available from:
```
hamishwillee: @bkueng Do we know of other suppliers of ready made cables?

en/peripherals/frsky_telemetry.md
```diff
@@ -54,3 +91,10 @@ Ready-made cables (which include the required adapters) are available from:
 UART to S.PORT adapters can be sourced from:
 * SPC: [getfpv.com](http://www.getfpv.com/frsky-smart-port-converter-cable.html), [unmannedtechshop.co.uk](https://www.unmannedtechshop.co.uk/frsky-smart-port-converter-spc/) 
 * FUL-1: [unmannedtech.co.uk](https://www.unmannedtechshop.co.uk/frsky-transmitter-receiver-upgrade-adapter-ful-1/)
+
+
+## Additional Information
+
+For additional information, see the following links:
```
hamishwillee: @bkueng FYI I just pasted your links here. They may not be best, but I'm not ready to provide better yet.

hamishwillee:
There is a great chart [here](http://www.redsilico.com/frsky-receiver-chart) showing all FrSky receivers. I think we should incorporate (with attribution!) some of this information. 
1. Are all the receivers marked with "yes" for supporting telemetry ALSO good for use with PX4? If so, which ones are/are not OK? I know we support the X4Rs, X8Rs,  D4R-II, but not sure about the various R and G variants.
   - I guess that we should support in theory any with Smart Port and Hub Protocol as the Digital Telemetry option?
2. Which fields in this are most relevant/not relevant to users making a selection? 
   - I think that receiver name, range, combined output and digital telemetry input are useful. What about the rest? 
   - Is there any missing field someone might want. I was thinking "weight" or "size".

en/peripherals/frsky_telemetry.md
```diff
@@ -1,27 +1,65 @@
...
+FrSky telemetry allows you to access vehicle telemetry and error information on a compatible RC transmitter.
 
+This information includes: flight mode, battery level, RC signal strength, speed, altitude etc. Some transmitters can additionally provide audible and vibration feedback, which is particularly useful for low battery and other failsafe warnings.
 
-## Pixracer
+## Flight Controller Hardware
 
-> **Success** FrSky telemetry auto-starts on Pixracer and
-  detects D or S mode, there is no need to configure anything.
+Pixhawk/PX4 supports D (old) and S (new) FrSky telemetry. FMUv4 based boards (Pixracer) include the required electronics for the interface, older boards (Pixhawk) need a UART to S.PORT adapter board.
```
hamishwillee: What else does S offer PX4 users - or is it irrelevant other than for connections?

en/peripherals/frsky_telemetry.md
bkueng: yes the image is only for the radio.
I don't think that we have an image. This is how you need to connect for D type:
RX - FS out
TX - FS in
GND: not needed, sincee you already have that from the RC/SBUS connector.
The vast majority today uses S.PORT as far as I can tell.

en/peripherals/frsky_telemetry.md
```diff
@@ -1,27 +1,65 @@
...
+
+Connect the Pixracer FrSky TX line to the receiver's RX line. Also connect GND (This applies to D4R-II and similar).
+
+![Radio Connection](../../assets/flight_controller/pixracer/grau_setup_pixracer_radio.jpg)
+
+
+#### Connecting Receivers with S-port
+
+Connect the Pixracer FrSky TX and RX lines together (solder the wires together) to the X series receiver's S.port pin. 
+Also connect GND. The S-port connection is shown below (using the provided I/O Connector).
```
bkueng: Same: GND is not required

en/peripherals/frsky_telemetry.md
```diff
@@ -1,27 +1,65 @@
...
+
+![Grau b Pixracer FrSkyS.Port Connection](../../assets/flight_controller/pixracer/grau_b_pixracer_frskys.port_connection.jpg)
+
+![Pixracer FrSkyS.Port Connection](../../assets/flight_controller/pixracer/pixracer_FrSkyTelemetry.png)
+
+
+
+### Pixhawk v2
```
bkueng: This applies to any other Pixhawk (i.e. v5 as well)

en/peripherals/frsky_telemetry.md
bkueng: > I see that as Developer land configuration.

agreed

en/peripherals/frsky_telemetry.md
bkueng: Yes. Only pixracer comes without requiring an adapter.

en/peripherals/frsky_telemetry.md
bkueng: Hmm, no

en/peripherals/frsky_telemetry.md
bkueng: These are fine. Chances are high if you have another transmitter/receiver and you do a specific google search you find a blog post on how to set it up.

hamishwillee:
Thanks @bkueng !

>> I think that receiver name, range, combined output and digital telemetry input are useful. What about the rest?

> Yes these plus the weight and size.

So I'll omit redundancy etc.

> FYI I just found out that there's a new protocol F.PORT: https://www.frsky-rc.com/frsky-betaflight-introduction-of-f-port-protocol/

That is cool!

en/peripherals/frsky_telemetry.md
hamishwillee: OK. Doesn't seem all that useful to us since we can already send out this info (without having to add sensors to a bus).

en/peripherals/frsky_telemetry.md
hamishwillee: Only Pixracer or only FMUv4? (and not FMUv5?)

en/peripherals/frsky_telemetry.md
hamishwillee: What is "this" are you saying that you always need to set? (ie is it the configuration below - in all cases you need to set SYS_COMPANION to 10) or are you refering to the "Same: GND is not required" comment?

en/peripherals/frsky_telemetry.md
```diff
@@ -35,16 +76,42 @@ You can see:
...
-If you open the `LuaPil.lua` script with a text editor, you can edit the
-configuration. Suggested modifications:
+If you open the `LuaPil.lua` script with a text editor, you can edit the configuration. 
+Suggested modifications include:
 ```
 local BattLevelmAh = -1  -- use the battery level calculation from the vehicle
 local SayFlightMode = 0  -- there are no WAV files for the PX4 flight modes
 ```
 
+## FrSky Telemetry Receivers
```
hamishwillee: @bkueng I added section table as shown below. I think I've captured that these should work, but that the X series ones are the ones that we _know_ work. 

I changed the table telemetry input name to D.Port from HubProtocol. Is there anything that should be added or that is missing from the table (ie is it correct and useful?)

hamishwillee:
>> What information can be transmitted

> This is slightly different for D vs S.PORT. See for example here: https://github.com/iNavFlight/inav/blob/master/docs/Telemetry.md#available-smartport-sport-sensors

Thanks, 
1. I look at that link and it all looks like s.port data. I can't see sensors supported for Dport - https://github.com/iNavFlight/inav/blob/master/docs/Telemetry.md#frsky-telemetry - am I blind?
2. Would I be correct in thinking that if the S.PORT is configured for telemetry there is a module in PX4 pumping out the sensor values according to the protocol listed at your link?

en/peripherals/frsky_telemetry.md
hamishwillee: But presumably not Pixhawk Pro since that uses TELEM4 and explicitly states that it doesn't need further configuration?
So I think you are saying that normally the S.PORT on Pixhawk is TELEM2?

en/peripherals/frsky_telemetry.md
bkueng: The D4R-II works for sure as well.
The table looks good.

bkueng:
> I look at that link and it all looks like s.port data. I can't see sensors supported for Dport - https://github.com/iNavFlight/inav/blob/master/docs/Telemetry.md#frsky-telemetry - am I blind?

No, you're right it's only S PORT. Sorry I can't find anything about D type at the moment.

> Would I be correct in thinking that if the S.PORT is configured for telemetry there is a module in PX4 pumping out the sensor values according to the protocol listed at your link?

Yes, it's `frsky_telemetry`: if correctly setup `frsky_telemetry status` should show you that it is running in D typ/S PORT mode.

en/peripherals/frsky_telemetry.md
bkueng: On the Pro, TELEM4 is just the default for frsky telemetry. You can set SYS_COMANION even there (and on Pixracer as well) to run it on TELEM2 instead.

> Does that mean the port is dedicated to S.PORT and always activated for that purpose?

Yes. But this was just because it had no other purpose.

> So I think you are saying that normally the S.PORT on Pixhawk is TELEM2?

Yes

hamishwillee:
@bkueng I think this set looks OK for DPort - http://shipow.github.io/cleanflight-web/docs/telemetry.html

hamishwillee:
Hi Beat,

Just FYI, gave this an update to add supported messages, and in particular to improve the introduction. It is nearly there. 

I'm not completely happy with the pixhawk/pixracer sections. They are correct but it doesn't quite flow. I'll have a think over weekend and let you know whether it is worth reviewing or not.

bkueng:
Ok, just tell me when you're ready. FYI there's an ongoing discussion for building a DIY cable here: http://discuss.px4.io/t/px4-frsky-telemetry-setup-with-pixhawk2-and-x8r-receiver/6362

hamishwillee:
> building a DIY cable here: http://discuss.px4.io/t/px4-frsky-telemetry-setup-with-pixhawk2-and-x8r-receiver/6362

Sounds like madness given the cost of the cable to buy premade

hamishwillee:
In fact, I am thinking of reorging to be "buy cable first" and all the rest of our instructions follow on after that.

hamishwillee:
@bkueng Hi Beat, 
Now reorganised to focus on ready-made cable first. It could be expanded further (and should be in the area of diye cables) but is much better for now.
Ready for your review thank you.

en/peripherals/frsky_telemetry.md
```diff
@@ -1,56 +1,209 @@
...
+
+  Number is sent as ABCDE detailed below. The numbers are additive (for example: if digit C is 6, it means both position hold and altitude hold are active).
    
-UART to S.PORT adapters can be sourced from:
+  - **A:** 1 = Flaperon mode, 2 = Auto tune mode, 4 = Failsafe mode.
+  - **B:** 1 = Return to home, 2 = waypoint mode, 4 = Headfree mode.
+  - **C:** 1 = Heading hold, 2 = Altitude hold, 4 = Position hold.
+  - **D:** 1 = Angle mode, 2 = Horizon mode, 4 = Passthrough mode.
+  - **E:** 1 = OK to arm, 2 = arming is prevented, 4 = armed.
+- **Tmp2:** GPS lock status, accuracy, home reset trigger, and number of satellites. Number is sent as ABCD detailed below. Typical minimum GPS 3D lock value is 3906 (GPS locked and home fixed, HDOP highest accuracy, 6 satellites).
```
bkueng: Tmp1 and Tmp2 are not standardized and have a different meaning in PX4:
- Tmp1 is the flight mode: https://github.com/PX4/Firmware/blob/master/src/drivers/telemetry/frsky_telemetry/frsky_telemetry.cpp#L85
- Tmp2: https://github.com/PX4/Firmware/blob/master/src/drivers/telemetry/frsky_telemetry/frsky_data.cpp#L236

en/peripherals/frsky_telemetry.md
```diff
@@ -1,56 +1,209 @@
...
+FrySky Telemetry can transmit most of the more useful status information from PX4. S-Port and D-Port receivers transmit different sets of messages, as listed the following sections.
+
+### S-Port {#s_port}
+
+S-Port receivers transmit the following messages (from [here](https://github.com/iNavFlight/inav/blob/master/docs/Telemetry.md#available-smartport-sport-sensors)):
+
+- **AccX, AccY, AccZ:** Accelerometer values.
+- **Alt:** Barometer based altitude, relative to home location.
+- **ASpd:** True air speed (from pitot sensor).
+- **A4:** Average cell value.
```
bkueng: Not supported in PX4

en/peripherals/frsky_telemetry.md
```diff
@@ -1,56 +1,209 @@
...
+
+FrySky Telemetry can transmit most of the more useful status information from PX4. S-Port and D-Port receivers transmit different sets of messages, as listed the following sections.
+
+### S-Port {#s_port}
+
+S-Port receivers transmit the following messages (from [here](https://github.com/iNavFlight/inav/blob/master/docs/Telemetry.md#available-smartport-sport-sensors)):
+
+- **AccX, AccY, AccZ:** Accelerometer values.
+- **Alt:** Barometer based altitude, relative to home location.
+- **ASpd:** True air speed (from pitot sensor).
```
bkueng: Not supported in PX4

en/assembly/quick_start_pixracer.md
```diff
@@ -2,32 +2,46 @@
 
 > **Warning** Under construction.
```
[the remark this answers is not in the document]
hamishwillee: Probably. 

The ideal Quick Start here is more like this one for [Pixhawk](https://docs.px4.io/en/assembly/quick_start_pixhawk.html) essentially covering all the main connections with some detailed information - aimed at beginners. This Pixracer one is just a bunch of wiring diagrams and misses a whole bunch of useful detail. 

So the argument for removing is that this is unlikely to be updated. The argument for keeping is that it is a reminder that more could/should be done.

PR #209, opened 2018-04-26
Altitude updates

hamishwillee:
This adds new icon, fixes up text to link to it, and also fixes up the Land mode docs to clarify when (only) altitude info required.


hamishwillee:
Thanks @bkueng

PR #210, opened 2018-04-27
Import preflight checks from Devguide

hamishwillee:
This imports the preflight checks from devguide, makes them a little bit more "end user facing" (ie less developer focused) and cross links to other relevant docs. 

After this is merged, we need to delete the version in Devguide.

### DIFF
--- a/en/flying/led_meanings.md
+++ b/en/flying/led_meanings.md
@@ -1,15 +1,16 @@
 # LED Meanings (Pixhawk Series)
 
-[Pixhawk-series flight controllers](../flight_controller/pixhawk_series.md) have an RGB LED to indicate the current status of the vehicle. The diagram below shows the relationship between LED and vehicle status.
+[Pixhawk-series flight controllers](../flight_controller/pixhawk_series.md) have an RGB LED to indicate the current status of the vehicle. The image below shows the relationship between LED and vehicle status.
 
-![LED meanings](../../images/led_meanings.gif)
-
-> **Note** **A valid global position estimate is required to takeoff:** 
-  The LED will only turn green (ready to execute guided mission) if PX4 is able to validate the global position. If the LED does not change from blue to green, verify that the GPS module is properly attached, Pixhawk is reading your GPS properly, and that the GPS is sending a proper GPS position.  
+> **Warning** It is possible to have a GPS lock (Green LED) and still not be able to arm the vehicle because PX4 has not yet [passed preflight checks](../flying/pre_flight_checks.md). **A valid global position estimate is required to takeoff!**
 
 <span></span>
-> **Tip** In the event of an error (blinking red), or if the vehicle can't achieve GPS lock, check for more detailed status information in *QGroundControl*.
-> Useful information includes both calibration status, and any error messages from the [Preflight Checks (Internal)](../flying/pre_flight_checks.md) messages.
+> **Tip** In the event of an error (blinking red), or if the vehicle can't achieve GPS lock (change from blue to green), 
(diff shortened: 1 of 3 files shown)


en/flying/led_meanings.md
```diff
@@ -4,8 +4,13 @@
 
 ![LED meanings](../../images/led_meanings.gif)
 
-> **Note** **GPS Lock is required to takeoff: ** 
-  The LED will only turn green (ready to execute guided mission) if PX4 is able to validate the global position. If the LED does not change from blue to green, verify that the GPS module is properly attached, Pixhawk is reading your GPS properly, and that the GPS is sending a proper GPS position.
+> **Note** **A valid global position estimate is required to takeoff:**
```
hamishwillee: @priseborough 
1. Do the LEDs reflect GPS or position estimator status? If it is the former, is it therefore possible to have the light effectively saying "ready to arm" because you have GPS lock but the vehicle is not actually ready to take off because the estimator isn't reporting sufficient quality?

en/flying/led_meanings.md
```diff
@@ -4,8 +4,13 @@
 
 ![LED meanings](../../images/led_meanings.gif)
 
-> **Note** **GPS Lock is required to takeoff: ** 
-  The LED will only turn green (ready to execute guided mission) if PX4 is able to validate the global position. If the LED does not change from blue to green, verify that the GPS module is properly attached, Pixhawk is reading your GPS properly, and that the GPS is sending a proper GPS position.
+> **Note** **A valid global position estimate is required to takeoff:** 
+  The LED will only turn green (ready to execute guided mission) if PX4 is able to validate the global position. If the LED does not change from blue to green, verify that the GPS module is properly attached, Pixhawk is reading your GPS properly, and that the GPS is sending a proper GPS position.  
+
+<span></span>
+> **Tip** In the event of an error (blinking red), or if the vehicle can't achieve GPS lock, check for more detailed status information in *QGroundControl*.
```
hamishwillee: @priseborough Would I be correct in thinking that if you get a preflight check it will display the blinking red LED?

en/flying/pre_flight_checks.md
```diff
@@ -0,0 +1,73 @@
+# Preflight Sensor/Estimator Checks
```
hamishwillee: This is essentially the same as merged to DevGuide in https://github.com/PX4/Devguide/pull/510#issuecomment-384519648 but I have changed the introduction to explain were you might see the messages and link to EKF info - as end users might not know that level of detail about our estimators.

en/flying/led_meanings.md
priseborough: That is correct. The GPS can have a 3D lock, but the EKF will not start using it ad report a valid global position until the output quality from the GPS has passed the EKF's internal checks. The estimator_status.gps_check_fail_flags message in the logs shows which GPS quality checks are failing

en/flying/led_meanings.md
priseborough: My understanding is that it should flash red if any preflight checks fail, but I have not tested that.

en/flying/led_meanings.md
hamishwillee: Thanks - I have added the fail case to LED meanings doc and the gps_check_fail_flags message info to the Preflight checks doc.

hamishwillee:
Thanks @bkueng . I'm going to merge. Errors should be good - checked by Paul. In any case, this has all been in the Devguide for some time, and it is significantly better than before.

PR #211, opened 2018-04-27
Add information about acro mode stick mapping algorithm

hamishwillee:
This is for current implementation. Rendered version shown [here](https://hamishwillee.gitbooks.io/ham_px4userguide/content/v/acro_stick/en/flight_modes/acro_mc.html#stick-input-mapping)


hamishwillee:
Yes @bkueng - added the yaw parameters, and I'll merge now.

MaEtUgR:
@hamishwillee Super nice! Maybe you want to mention that the controls in acro correspond to a model helicopter just without negative thrust (which anyways not all helicopters have).

hamishwillee:
Hi @MaEtUgR 

> Maybe you want to mention that the controls in acro correspond to a model helicopter just without negative thrust (which anyways not all helicopters have).

I'm not sure. Generally it is worth adding comparisons where readers already understand something and you want them to say "aha, this is just like that thing I already know". But I don't think most of our users know that much about helicopters. 

If you think it is an important point to make, ping me offline!

PR #212, opened 2018-05-01
Failsafe/Safety docs

hamishwillee:
This adds failsafe/safety docs as discussed in https://github.com/PX4/px4_user_guide/issues/205

This adds an introduction in the PX4 concepts and detailed information about the different settings in a configuration page. The config page has both the QGC page and associated params linked.

There are some open questions inline


en/config/safety.md
```diff
@@ -0,0 +1,225 @@
+# Safety Configuration (Failsafes) 
+
+PX4 has configurable failsafe systems to protect and recover your vehicle if something goes wrong! These allow you to specify areas and conditions under which you can safely fly, and the [action](#failsafe_actions) that will be performed if a failsafe is triggered (for example, landing, holding position, or returning to a specified point).
+
+The most important failsafe settings are configured in the *QGroundControl* [Safety Setup](#qgc_safety_setup) page. Others must be configured via [parameters](#failsafe_other).
+
+## Failsafe Actions {#failsafe_actions}
```
hamishwillee: The failsafe actions are often things like Return to home, Land, Loiter.
1. My assumption is that these are **exactly** like the modes, Return, Land, Hold - except presumably the system already knows that they are handled in a failsafe, and can offer different actions if a further failsafe occurs:
   * Is this correct?
   * If so, should we not mark up the parameters to use the new names for the modes - Return, Land, Hold etc? (or Return mode, Land mode, Hold mode). This is especially true as return to home is no longer going to be necessarily the destination.

en/config/safety.md
```diff
@@ -0,0 +1,225 @@
...
+Each failsafe defines its own set of actions. Some of the more common failsafe actions are:
+
+Action | Description
+--- | --- 
+None/Disabled | No action (the failsafe will be ignored).
+Warning | A warning message will be sent to *QGroundControl*.
+Loiter | The vehicle will enter [Hold mode](../flight_modes/hold.md). For multicopters this means the vehicle will hover, while for fixed/wing the vehicle will circle.
+Return to Land | The vehicle will enter [Return mode](../flight_modes/return.md). Return behaviour can be set in the [Return Home Settings](#return_settings) (below).
+Land at current position |  The vehicle enters [Land mode](../flight_modes/land.md), and lands immediately.
+RC Auto Recovery (CASA Outback Challenge rules) | ?
```
hamishwillee: What are the rules for "RC Auto Recovery"? No information about this around.

en/config/safety.md
```diff
@@ -0,0 +1,225 @@
...
+Terminate/Flight termination | ?
```
hamishwillee: What happens in Terminate/Flight termination. Is this just "Kill -drop out of the sky?"

en/config/safety.md
```diff
@@ -0,0 +1,225 @@
...
+Lockdown | ?
```
hamishwillee: What is lockdown action?

en/config/safety.md
```diff
@@ -0,0 +1,225 @@
...
+
+The low battery failsafe is triggered when the battery capacity drops below one (or more warning) level values. 
+
+![Safety - Battery (QGC)](../../images/qgc/setup/safety_battery.png)
+
+The most common configuration is to set the values and action as above (with `Warn > Failsafe > Emergency`).
+With this configuration the failsafe will trigger warning, then return, and finally landing if capacity drops below the respective levels.
+ 
+It is also possible to set the *Failsafe Action* to warn, return, or land. 
+In this case the vehicle will perform the specified action when the *Warn* failsafe level is reached.
```
hamishwillee: Is this correct? Specifically, is it the Warn level that triggers the failsafe that is specified int eh failsafe action - or is it one of the other thresholds?

en/config/safety.md
```diff
@@ -0,0 +1,225 @@
...
+
+![Safety - RC Loss (QGC)](../../images/qgc/setup/safety_rc_loss.png)
+
+The settings and underlying parameters are shown below.
+
+Setting | Parameter | Description
+--- | --- | ---
+RC Loss Timeout | [COM_RC_LOSS_T](../advanced_config/parameter_reference.md#COM_RC_LOSS_T) | Amount of time after losing the RC connection before the failsafe will trigger.
+Failsafe Action | [NAV_RCL_ACT](../advanced_config/parameter_reference.md#NAV_RCL_ACT) | Disabled, Loiter, Return, Land, RC Auto Recovery, Terminate, Lockdown.
+RC Loss Loiter Time | [NAV_RCL_LT](../advanced_config/parameter_reference.md#NAV_RCL_LT) | If the *Failsafe Action* (`NAV_RCL_ACT`) is set to *CASA Outback Challenge rules* this sets the loiter time after RC loss.
```
hamishwillee: FYI, there is no way to set this in QGC (ie if you set the action to Casa Rules it does not appear). Worth raising a bug report over?

en/config/safety.md
```diff
@@ -0,0 +1,225 @@
...
+![Safety - Data Link Loss (QGC)](../../images/qgc/setup/safety_data_link_loss.png)
+
+The settings and underlying parameters are shown below.
+
+Setting | Parameter | Description
+--- | --- | ---
+Data Link Loss Timeout | [COM_DL_LOSS_T](../advanced_config/parameter_reference.md#COM_DL_LOSS_T) | Amount of time after losing the data connection before the failsafe will trigger.
+Failsafe Action | [NAV_DLL_ACT](../advanced_config/parameter_reference.md#NAV_DLL_ACT) | Disabled, Loiter, Return, Land, RC Auto Recovery, Terminate, Lockdown.
+
+
```
hamishwillee: There are a bunch of other parameters that appear to be related here: https://docs.px4.io/en/advanced_config/parameter_reference.html#data-link-loss

When do these come into play? - what vehicles?

en/config/safety.md
hamishwillee: PS There were notes in geofence for this:

> Note: Setting this value to 4 enables flight termination, which will kill the vehicle on violation of the fence. Due to the inherent danger of this, this function is disabled using a software circuit breaker, which needs to be reset to 0 to really shut down the system. 

So it does sound like this is a kill switch. If so, 
- what is the "Breaker" (ie another parameter). 
- What happens instead of killing if you don't have the breaker enabled?
- Does this same breaker apply to any termination, or just due to Geofence?

en/config/safety.md
```diff
@@ -0,0 +1,225 @@
...
+### Position (GPS) Loss Failsafe
+
+The Position Loss failsafe is triggered if the quality of the PX4 position estimate falls below acceptable levels (this might be caused by GPS loss).
+
+The relevant parameters are shown below:
+
+Parameter | Description
+--- | ---
+[COM_POS_FS_DELAY](../advanced_config/parameter_reference.md#COM_POS_FS_DELAY) | Delay after loss of position before the failsafe is triggered.
+[CBRK_GPSFAIL](../advanced_config/parameter_reference.md#CBRK_GPSFAIL) | Circuit breaker that can be used to disable GPS failure detection.
```
hamishwillee: Is this correct - ie should it be mentioned here. Is it really GPS or it kills all of position estimation failure warnings?

en/config/safety.md
```diff
@@ -0,0 +1,227 @@
...
+--- | --- | ---
+Action on breach | [GF_ACTION](../advanced_config/parameter_reference.md#GF_ACTION) | None, Warning, Loiter, Return, Flight Terminate.
+Max Radius | [GF_MAX_HOR_DIST](../advanced_config/parameter_reference.md#GF_MAX_HOR_DIST) | Horizontal radius of geofence cylinder. Geofence disabled if 0.
+Max Altitude | [GF_MAX_VER_DIST](../advanced_config/parameter_reference.md#GF_MAX_VER_DIST) | Height of geofence cylinder. Geofence disabled if 0.
+
+
+<!-- flight termination notes: Note: Setting this value to 4 enables flight termination, which will kill the vehicle on violation of the fence. Due to the inherent danger of this, this function is disabled using a software circuit breaker, which needs to be reset to 0 to really shut down the system. WHAT is the breaker, and what happens instead -->
+
+<!--  What are these? Not in QGC
+* GF_ALTMODE (INT32)	- Geofence altitude mode. Comment: Select which altitude reference should be used 0 = WGS84, 1 = AMSL. Values:
```
hamishwillee: These params are not in QGC:
- Are they used?
- If so, what is GF_SOURCE - ie what does source mean?

en/config/safety.md
```diff
@@ -0,0 +1,227 @@
...
+
+Parameter | Description
+--- | ---
+[COM_POS_FS_DELAY](../advanced_config/parameter_reference.md#COM_POS_FS_DELAY) | Delay after loss of position before the failsafe is triggered.
+[CBRK_GPSFAIL](../advanced_config/parameter_reference.md#CBRK_GPSFAIL) | Circuit breaker that can be used to disable GPS failure detection.
+[NAV_GPSF_LT](../advanced_config/parameter_reference.md#NAV_GPSF_LT) | Loiter time (waiting for GPS recovery before it goes into flight termination). Set to 0 to disable.
+
+Parameters that only affect Fixed Wing vehicles:
+
+Parameter | Description
```
hamishwillee: Is this correct? I assume these can only apply to FW

en/config/safety.md
```diff
@@ -0,0 +1,227 @@
...
+The relevant parameters are shown below:
+
+Parameter | Description
+--- | ---
+* [MIS_DIST_1WP](../advanced_config/parameter_reference.md#MIS_DIST_1WP) | The mission will not be started if the current waypoint is more distant than this value from the home position. Disabled if value is 0 or less.
+* [MIS_DIST_WPS](../advanced_config/parameter_reference.md#MIS_DIST_WPS) | The mission will not be started if any distance between two subsequent waypoints is greater than this value.
+
+
+### Traffic Avoidance Failsafe
+
```
hamishwillee: Is this correct?

en/config/safety.md
```diff
@@ -0,0 +1,227 @@
...
+The Traffic Avoidance Failsafe allows PX4 to respond to transponder data (e.g. from ADSB transponders).
+
+The relevant parameters are shown below:
+
+Parameter | Description
+--- | ---
+[NAV_TRAFF_AVOID](../advanced_config/parameter_reference.md#NAV_TRAFF_AVOID) | Set the failsafe action: Disabled, Warn, Return, Land.
+
+
+### Adaptive QuadChute Failsafe
```
hamishwillee: Is this worth including? 

I think that it is VTOL only, and that it is a failsafe for when the altitude estimate diverges too far from the setpoint - ie vehicle can no longer achieve desired height in FW mode. Correct?

en/getting_started/px4_basic_concepts.md
```diff
@@ -28,23 +28,70 @@ You can read in depth about heading in [Flight Controller Orientation](../config
```
hamishwillee: FYI, the PX4 concepts is a bit of a mishmash of general concepts and overview/features list. This updates it more towards being an overview doc that references other concepts docs. As such it includes new section on Flight Modes and this update for Failsafes. I could have added a separate doc for failsafe at this point if you thought worth highlighting as a standalone concept (I thought overkill) ?

en/config/safety.md
barzanisar: I noticed different failsafe action in Position mode than in Mission mode for GPS failure, as described [here](https://github.com/PX4/Firmware/issues/9336). The description you wrote holds for "Global position loss" failsafe which is triggered upon GPS failure in mission mode. However in position mode, the GPS (or any other position source) failsafe action is defined by  COM_POSCTL_NAVL.

barzanisar: Maybe it is worth separating "Global Position Loss Failsafe" from "Local Position Loss Failsafe"

barzanisar: > for a time while trying to regain GPS, and then enter flight termination.

When I simulated GPS fail with multicopter in mission mode, it "landed on spot" rather than "terminating flight". I learned from your other comments that "terminating flight" could mean kill switch behaviour/disarming motors immediately so it is better to confirm and distinguish between the two.

en/config/safety.md
barzanisar: Even NAV_GPSF_LT does not apply to multicopters according to @RomanBapst and I checked this in simulation. @bkueng can you please confirm this as well?

en/config/safety.md
```diff
@@ -0,0 +1,227 @@
...
+
+The relevant parameters are shown below:
+
+Parameter | Description
+--- | ---
+* [MIS_DIST_1WP](../advanced_config/parameter_reference.md#MIS_DIST_1WP) | The mission will not be started if the current waypoint is more distant than this value from the home position. Disabled if value is 0 or less.
+* [MIS_DIST_WPS](../advanced_config/parameter_reference.md#MIS_DIST_WPS) | The mission will not be started if any distance between two subsequent waypoints is greater than this value.
+
+
+### Traffic Avoidance Failsafe
```
barzanisar: What is "Traffic Avoidance Failsafe"?

en/config/safety.md
```diff
@@ -0,0 +1,227 @@
...
+The relevant parameters are shown below:
+
+Parameter | Description
+--- | ---
+[NAV_TRAFF_AVOID](../advanced_config/parameter_reference.md#NAV_TRAFF_AVOID) | Set the failsafe action: Disabled, Warn, Return, Land.
+
+
+### Adaptive QuadChute Failsafe
+
+Maximum negative altitude error for fixed wing flight. If the altitude drops below this value below the altitude setpoint the vehicle will transition back to MC mode and enter failsafe RTL.
```
barzanisar: > below this value below the altitude setpoint
 I didn't understand this part.

en/getting_started/px4_basic_concepts.md
barzanisar: I like this update very much. I think flight modes and failsafes are both important basic concepts, every new user would be interested in.

en/config/safety.md
```diff
@@ -0,0 +1,235 @@
...
+
+
+## Other Failsafe Settings {#failsafe_other}
+
+This section contains information about failsafe settings that cannot be configured through the *QGroundControl* [Safety Setup](#qgc_safety_setup) page.
+
+### Position (GPS) Loss Failsafe
+
+The *Position Loss Failsafe* is triggered if the quality of the PX4 position estimate falls below acceptable levels (this might be caused by GPS loss) while in a mode that requires an acceptable position estimate.
+
```
hamishwillee: @bkueng This is a bit of a guess based on my testing and @barzanisar testing in https://github.com/PX4/Firmware/issues/9336#issuecomment-386274299

It may not be correct because COM_POSCTL_NAVL is only listed in mission section, but it does appear to explain observed behaviour.

The main question is when the Fixed wing loiter happens (see line 162 below). It could happen immediately the failsafe is engaged, but before the action is triggered (ie you go to loiter then a manual mode after that) or it could be only in certain modes. Thoughts?

en/config/safety.md
hamishwillee: Thanks @barzanisar . I have re-written this section. I think that behaviour is somewhere between what you and I wrote - essentially COM_POSCTL_NAVL applies everywhere. 

AFAIK little evidence there is a difference between local and global position handling, so not worth differentiating. From what Paul has told me the source is always EKF.

en/config/safety.md
hamishwillee: Good point. But , when does it apply - immediately after failsafe engaged (before doing the failsafe action) or at some other point?

en/config/safety.md
hamishwillee: What this text is saying is that there is a setpoint that the VTOL is trying to reach, but it can't - possibly the vehicle is moving too slow. If the actual altitude falls by this much below the desired alt then the failsafe will engage, changing to from FW to MC mode. 

Based on the name "Adaptive QuadChute Failsafe" I GUESS that what happens is that the user engages a parachute to slow down in FW mode. This trigger automatically switches to MC when the speed drops sufficiently. Could be completely wrong.

en/config/safety.md
bkueng: - Terminate: the effect is a mode switch in commander, which turns off all controllers, and more importantly for a user, all outputs are set to a failsafe pwm value. The failsafe value can be defined in an airframe config (`FAILSAFE ` variable).
- Lockdown: bypasses everything, and just kills the motors (sets them to disarmed). This is the same as using the kill switch.

For geofence: the circuit breaker param is CBRK_FLIGHTTERM, and it's only used when flight termination is requested from navigator (i.e. geofence violation).

en/config/safety.md
```diff
@@ -0,0 +1,235 @@
...
+Warning | A warning message will be sent to *QGroundControl*.
+Loiter | The vehicle will enter [Hold mode](../flight_modes/hold.md). For multicopters this means the vehicle will hover, while for fixed/wing the vehicle will circle.
+Return to Land | The vehicle will enter [Return mode](../flight_modes/return.md). Return behaviour can be set in the [Return Home Settings](#return_settings) (below).
+Land at current position |  The vehicle enters [Land mode](../flight_modes/land.md), and lands immediately.
+RC Auto Recovery (CASA Outback Challenge rules) | ?
+Terminate/Flight termination | ?
+Lockdown | ?
+
+
+> **Note** You can only specify the action for the *first* failsafe event. Once a failsafe occurs the system will enter special handling code, such that subsequent failsafe triggers are managed by separate system level and vehicle specific code. This might result in the vehicle being changed to a manual mode so the user can directly manage recovery.
```
bkueng: That is new to me. From where did you get this?

en/config/safety.md
bkueng: No, it's at the "Battery Failsafe Level"

en/config/safety.md
bkueng: Yes I think so.

en/config/safety.md
bkueng: These are used in combination with 'Data Link Auto Recovery (CASA Outback Challenge rules)'

en/config/safety.md
bkueng: Yes they are used. With GF_SOURCE you can say whether you want to use the estimated global position or the one directly from the GPS device. Not sure how relevant this is though.

en/config/safety.md
bkueng: I don't know about FW

en/config/safety.md
bkueng: Yes

en/config/safety.md
bkueng: Yes it is important. But I'm not sure in which cases people use it.

en/config/safety.md
bkueng: @RomanBapst can you expand here?

en/config/safety.md
hamishwillee: I can't/shouldn't remove stuff from docs unless they are also removed from the actual parameters. For these ones I'll just put TBD for now.

en/config/safety.md
hamishwillee: It comes from discussion with PaulR [here](https://github.com/PX4/Firmware/issues/9336). I haven't got any tests to show it is true or not true. What I do know is that we do handle the first failsafe event, and if this is true then I don't have to try and document what happens in the event of failsafes occuring in failsafes.

en/config/safety.md
hamishwillee: Great. Done in https://github.com/PX4/Firmware/pull/9429

en/config/safety.md
hamishwillee: Thanks. Done https://github.com/mavlink/qgroundcontrol/issues/6428

en/config/safety.md
bkueng: Ok. I think it depends on the failure case: from what I know if you lose RC, and the vehicle goes into RTL, and then you regain RC, you can simply switch modes and regain control. Subsequent RC losses will be handled the same way. But this should quickly be verified.

en/config/safety.md
hamishwillee: OK. Who would know?

en/config/safety.md
hamishwillee: Here we aren't actually talking about recovery cases - this is more what would happen say if you had RC loss followed immediately by battery loss.

The confusion I have is more from cases like RC loss followed by position loss. This looks like 2 failsafes (so should be higher level system handling), but the actual handling seems more like it is defined by what happens in COM_POSCTL_NAVL

What you're saying above is almost certainly true - ie that if you can recover from failsafe (ie you get RC back)  then everything can be reset by the user changing mode. Worth saying?

en/config/safety.md
```diff
@@ -0,0 +1,234 @@
...
+
+Action | Description
+--- | --- 
+None/Disabled | No action (the failsafe will be ignored).
+Warning | A warning message will be sent to *QGroundControl*.
+[Hold mode](../flight_modes/hold.md) | The vehicle will enter *Hold mode*. For multicopters this means the vehicle will hover, while for fixed/wing the vehicle will circle.
+[Return mode](../flight_modes/return.md) | The vehicle will enter *Return mode*. Return behaviour can be set in the [Return Home Settings](#return_settings) (below).
+[Land mode](../flight_modes/land.md) | The vehicle will enter *Land mode*, and lands immediately.
+RC Auto Recovery (CASA Outback Challenge rules) | TBD
+Terminate/Flight termination | Turns off all controllers and sets all PWM outputs to a failsafe value (defined in airframe configuration using `FAILSAFE` variable).
```
hamishwillee: Thanks for previous info - just copied in here.
But I'm not sure how this really differs from Lockdown. Presumably you could set failsafe to some value for motors to still spin (or whatever) but wouldn't this be pretty pointless for a quadcopter since you wouldn;t have any controllers running?
Upshot, in practise, how does it differ from Kill switch :-)

hamishwillee:
Thanks @bkueng @barzanisar  - fixed up those points. I think this is good to merge now. I am aware there are a couple of rough edges, but this is far better than nothing.

hamishwillee:
Thanks @bkueng !

PR #214, opened 2018-05-08
Pixhawk 4 guide

barzanisar:
This guide contains user guide content for Pixhawk 4, based on the content and structure already existing for old Pixhawk. I will merge this PR for now to make it available to public, but we have to review it. @hamishwillee


hamishwillee:
All the images, in particular the hero image on the assembly guide are beautiful. But these are huge - 2.6MB takes about 4 seconds after the page to load on my computer in Australia. Imagine how long that will take in China! Also, from now until the end of time, every time someone syncs this repo they will have to download that huge file. 

So as a guideline, unless there is a good reason, try keep file sizes for a "large" image less than 200Kb - ideally less than 100Kb. Typically I use gimp to convert to 90% jpg and that gets the needed sizes. Sometimes I drop the width. 

It is also generally a good idea to have these all in one folder on their own - makes it easy to find them and edit all images associated with a topic.

Anyway, I'll start reviewing and updating. First set of changes in: https://github.com/PX4/px4_user_guide/pull/218

ndepal:
> Also, from now until the end of time, every time someone syncs this repo they will have to download that huge file.

Just in case you're not aware, you can amend the commit in question and (force) push that changed history . That will remove any trace of that large file.

hamishwillee:
Good point @ndepal . I have never done this. In case we do want to, the PR which reduces the size of the images is https://github.com/PX4/px4_user_guide/pull/220

PR #217, opened 2018-05-09
Update LeddarOne docs

hamishwillee:
The old docs assume LeddarOne is not in firmware and that driver must be started manually. Currently the driver is often in firmware and can be started using a parameter.

This update explains the "out of the box" case, and what to do in other cases.


hamishwillee:
Thanks for review @zehortigoza ! Merged.

PR #219, opened 2018-05-09
Modified according to Jinger's review.

barzanisar:
Changed PMB info and merged Pinouts subsection with Autopilot Hardware > Pixhawk 4


hamishwillee:
Good. I was going to propose this approach.

PR #222, opened 2018-05-11
Add yaw stick info for fw position and alt. modes

hamishwillee:
Fixes https://github.com/PX4/px4_user_guide/issues/213


hamishwillee:
FYI @RomanBapst - this is the change we discussed recently about yaw stick in FW position and altitude modes.

PR #224, opened 2018-05-11
SD card is included in the kit

barzanisar:
Jinger requested to mention that the SD card is included in the kit.

### DIFF
--- a/en/assembly/quick_start_pixhawk4.md
+++ b/en/assembly/quick_start_pixhawk4.md
@@ -136,7 +136,7 @@ The vehicle-based radio should be connected to the **TELEM1** port as shown belo
 
 ## SD Card (Optional)
 
-SD cards are most commonly used to [log and analyse flight details](../flying/flight_reporting.md). Insert the card into *Pixhawk 4* as shown below.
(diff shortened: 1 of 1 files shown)


en/assembly/quick_start_pixhawk4.md
```diff
@@ -136,7 +136,7 @@ The vehicle-based radio should be connected to the **TELEM1** port as shown belo
 
 ## SD Card (Optional)
 
-SD cards are most commonly used to [log and analyse flight details](../flying/flight_reporting.md). Insert the card into *Pixhawk 4* as shown below.
+SD cards are most commonly used to [log and analyse flight details](../flying/flight_reporting.md). Insert the card (included in Pixhawk 4 kit) into *Pixhawk 4* as shown below.
 
 ![Pixhawk 4/SD Card](../../assets/flight_controller/pixhawk4/pixhawk4_sd_card.png)
```
hamishwillee: Probably you should delete the following Tip for good SD cards. We will assume that the included SD card is fine.

barzanisar: The SD card included is only 4 GB. I think we should keep the tip as an extra information.

hamishwillee: Good point! Agree

PR #246, opened 2018-05-28
pixhawk 4: add note about voltage divider when another PM board is used

bkueng:


en/assembly/quick_start_pixhawk4.md
```diff
@@ -102,6 +102,8 @@ Pin | Signal | Volt
 5(black) | GND | GND
 6(black) | GND | GND
 
+> **Note** If another Power Module is used, for example one that comes with a
+> Pixracer, the voltage divider needs to be updated as documented [here](https://docs.qgroundcontrol.com/en/SetupView/Power.html)
```
hamishwillee: 1. Terminate with full stop.

Why don't you need to do the voltage divider setting for the standard power module? Do you still need to do the other power calibration settings (e.g. number of cells)?

Depending on what you say, you might revise this slightly (below is example, not final text) to explain what you do and do not need to change. 

>**Note** Using the Power Module that comes with the kit you will need to configure the [Power Settings](https://docs.qgroundcontrol.com/en/SetupView/Power.html) for cell number xxxx but you won't need to calibrate the voltage divider. You will (may?) have to update the voltage divider if you are using any other power module (e.g. the one from the Pixracer).

LorenzMeier: @bkueng Good suggestions here.

bkueng: Indeed! Updated, thanks @hamishwillee

hamishwillee:
Thanks @bkueng - merged!

PR #256, opened 2018-06-12
Add a racer setup page & MC PID tuning page rewrite

bkueng:
Preview:
- https://bkueng.gitbooks.io/px4-user-guide/content/en/advanced_config/pid_tuning_guide_multicopter.html
- https://bkueng.gitbooks.io/px4-user-guide/content/en/advanced_config/racer_setup.html

Remarks:
- @Stifael, @MaEtUgR, @RomanBapst I'll need some actual control experts go over this and correct me where I'm wrong :)
- the control latency & filtering might be better moved to another place
- there are still some TODO's left, for example if we need YAW_FF, or documenting Airmode.

@hamishwillee I'm sure you have some good input as well.
Feel free to push directly to the branch.


en/advanced_config/pid_tuning_guide_multicopter.md
```diff
@@ -83,143 +215,24 @@ might not climb or lose altitude a bit, but it will never flip over. The
...
-so that the weight is virtually zero. Rotate it around yaw, and observe
-the response. It should go slowly back to the initial heading. If it
-oscillates, tune down P. Once the control response is slow but correct,
-increase P until the response is firm, but it does not oscillate.
-Typical value is around 2…3.
-
-Look at `ATTITUDE.yaw` in *QGroundControl*. Yaw overshoot should be not more
-than 2-5% (which is less than the overshoot for roll and pitch angles).
 
+<!-- TODO
```
bkueng: @Stifael, @MaEtUgR, @RomanBapst can one of you write the Position controller tuning part?

en/advanced_config/pid_tuning_guide_multicopter.md
```diff
@@ -1,74 +1,206 @@
...
+  never stop** when the vehicle is armed. This can be tested in acro
+  or in manual mode, without props:
+  - arm the vehicle and lower the throttle to the minimum,
+  - tilt the vehicle to all directions, about 60 degrees,
+  - check that no motor turns off.
+- Optionally enable the high-rate logging profile with the `SDLOG_PROFILE`
+  parameter. It will help evaluating the rate and attitude tracking performance
+  from the log. The option can be disabled afterwards.
+
+> **Warning** Always disable `MC_AIRMODE` when tuning a vehile.
```
hamishwillee: vehicle spelling.

en/advanced_config/pid_tuning_guide_multicopter.md
```diff
@@ -1,74 +1,206 @@
...
 
-**P**roportional, **I**ntegral, **D**erivative controllers are the most widespread control technique. 
-There are substantially better performing control techniques (LQR/LQG) from the model predictive control (MPC), 
-but since these techniques require a more or less accurate model of the system, they are not as widely used. 
-The goal of all PX4 control infrastructure is to move as soon as possible on MPC, 
-but since not all supported systems models are available, 
-PID tuning is very relevant (and PID control is sufficient for many cases).
+PX4 uses **P**roportional, **I**ntegral, **D**erivative (PID) controllers, which are the most widespread control technique. 
+
+The controllers are layered, which means a higher-level controller passes its
```
hamishwillee: I know what the attitude, rate and position controller do, but you can't infer that from the name. I think it is worth explaining what each controller "tracks" and what modes it is used in. 
It also isn't clear how the cascading levels work ie how are the results passed and why rate needs to feed into attitude - though it is clear why you are saying this - so they you need to do rate controller first.

en/advanced_config/racer_setup.md
```diff
@@ -0,0 +1,212 @@
...
+
+### GPS with external Mag
+A GPS adds some weight and it should be placed away from high currents to avoid
+mag interferences (on a GPS mast), which unfortunately means it will easily
+break during crashes. On the other hand you get some benefits, that are
+interesting especially for beginners:
+- You can go into position hold and the vehicle will just stay in one place.
+  This is handy if you lose the orientation or need a brake. It can also be used
+  to land safely.
+- Return-To-Land (RTL) can be used, either on a switch or as RC loss/low
```
hamishwillee: Call it "Return mode" and link to the document on this.

en/advanced_config/pid_tuning_guide_multicopter.md
```diff
@@ -1,74 +1,206 @@
...
+## Tuning steps
+
+> **Note** For safety reasons, the default gains are set to a small value.  
+> You must increase the gains before you can expect good control responses. 
+
+Here are some general points to follow when tuning:
+- All gains should be increased very slowly, by 20-30% per iteration, and
+  even 5-10% for final fine tuning.
+  Note that too large gains may cause dangerous oscillations!
+- It is advised to always quickly land when changing a parameter, and then
```
hamishwillee: I think you are saying that all parameters should be changed when landed in order to check for oscillations? Will everyone know what you mean by oscillations - wobbling, rotating?

en/advanced_config/pid_tuning_guide_multicopter.md
```diff
@@ -1,74 +1,206 @@
...
+  slowly increase the throttle and check for oscillations.
+- Tune the vehicle around the hovering thrust point, and
+  use the thrust curve parameter or TPA to account for thrust
```
hamishwillee: What is TPA?

en/advanced_config/esc_calibration.md
```diff
@@ -1,32 +1,30 @@
-# ESC calibration
+# ESC Calibration
```
hamishwillee: @bkueng OK, I redid this page to make it more clear. 

My understanding is that all calibration does is determine and the correct PWM MIN and MAX for the ESCs. For high quality ESCs you can trust the info on their specs for these values, but for low quality ESC you need to calibrate to find these values. 
1. Is that correct?
2. Is it correct that there is only one stored ESC PWM min/max value? The implication being that you must the same model of ESC for all motors and all will have same max/min values?
3. How does the calibration work? Ie I assume that the calibration ramps up PWM until motor movement starts to get PWM_MIN and then continues until increasing PWM does not result in greater speed to get PWM_MAX ... but what is it actually measuring to get that speed range?

Further more, we don't say anything much about ESC selection anywhere else. We should probably have a section on that in our assembly guide. This PR might not be the right place but:
1. Do we support all types of ESCs? 
   * Are there types we do/don't support
   * Are there types we do/don't recommend
1.  I understand that some ESCs use different firmware. What types of firmware do we support recommend?
1. Are all ESCs controlled as PWM ie via MAIN or are some off other bus?
1. Do all ESC require calibration - ie if off another bus?

en/advanced_config/pid_tuning_guide_multicopter.md
```diff
@@ -1,217 +1,198 @@
...
 
-This tutorial explains how to tune the PID loops on PX4 for all [multi rotor setups](../airframes/airframe_reference.md#copter) (Quads, Hexa, Octo etc).
+This tutorial explains how to tune the PID loops on PX4 for all [multicopter setups](../airframes/airframe_reference.md#copter) (Quads, Hexa, Octo etc).
 
-Tuning is required when creating a new airframe type or significantly modifying an existing supported frame. Generally if you're using a [supported configuration](../airframes/airframe_reference.md#copter) (e.g. using an airframe in [QGroundControl > Airframe](../config/airframe.md)) the default tuning should be acceptable (particularly for larger frames). 
+Tuning is required when creating a new airframe type or modifying an existing supported frame. Generally if you're using a [supported specific configuration](../airframes/airframe_reference.md#copter) (e.g. using an airframe in [QGroundControl > Airframe](../config/airframe.md)), where all components are fixed, the default tuning should work.
+To get the best performance it is usually good to tune a new vehicle, as for example using other ESCs or motors will require different tuning gains.
```
hamishwillee: @bkueng The paragraph here is a little confusing because 
- We don't define what might constitute "modifying an existing supported frame"
- "new vehicle" here is confusing because our airframe configurations do not specify a particular ESC or motor.  

I think we can address this with a slight refocus .... thoughts?

> Generally if you're using a [supported specific configuration](../airframes/airframe_reference.md#copter) (e.g. using an airframe in [QGroundControl > Airframe](../config/airframe.md)) the default tuning should be sufficient to fly the vehicle safely.
> To get the very best performance it is usually good to tune a new vehicle. For example, different ESCs or motors  require different tuning gains for optimal flight.

en/advanced_config/pid_tuning_guide_multicopter.md
```diff
@@ -1,217 +1,185 @@
...
+  ![TPA](../../images/mc_pid_tuning/MC_PID_tuning-TPA.svg)
+<!-- The drawing is on draw.io: https://drive.google.com/file/d/1N0qjbiJX6JuEk2I1-xFvigLEPKJRIjBP/view?usp=sharing
+     On the second Tab
+-->
 
-Look at `ATTITUDE.yaw` in *QGroundControl*. Yaw overshoot should be not more
-than 2-5% (which is less than the overshoot for roll and pitch angles).
 
 
+<!-- TODO: needed?
```
hamishwillee: I don't know :-)

en/advanced_config/pid_tuning_guide_multicopter.md
```diff
@@ -223,3 +191,63 @@ aerial video optimal value may be much smaller to get smooth response.)
...
+With Airmode enabled, the commanded thrust is increased by
+<span style="color:#B85450">b</span>. When it is disabled,
+<span style="color:#9673A6">r</span> is reduced.
+
+  ![Airmode](../../images/mc_pid_tuning/MC_PID_tuning-Airmode.svg)
+<!-- The drawing is on draw.io: https://drive.google.com/file/d/1N0qjbiJX6JuEk2I1-xFvigLEPKJRIjBP/view?usp=sharing
+     On the first Tab
+-->
+
+If mixing becomes saturated towards the upper bound, the Airmode logic is always used (the commanded thrust is reduced).
```
hamishwillee: @bkueng In airmode isn't the commanded thrust increased to meet a required roll? So for upper bound I would expect to **not** use airmode so that we don't saturate.

en/advanced_config/racer_setup.md
```diff
@@ -0,0 +1,200 @@
+# Racer Setup
```
hamishwillee: @bkueng This is awesome! I've changed some of the heading levels. You might want to check they make sense still.

en/advanced_config/racer_setup.md
```diff
@@ -0,0 +1,200 @@
...
+
+### Failsafe
+
+Configure [RC loss and low battery failsafe](../config/safety.md). 
+If you do not use a GPS, set the failsafe to **Lockdown**, which turns off the motors. 
+Test RC loss on the bench without props attached by turning off the remote when the vehicle is armed.
+
+Make sure to assign a [kill switch](../config/flight_mode.md#kill-switch) or an [arming switch](#arming_switch).Test it and train to use it!
+
+#### Arm Switch {#arming_switch}
```
hamishwillee: @bkueng When would you use a kill switch and when an arming switch - what is the effective difference for a user? There is a practical difference that there is an option to select kill switch as a flight mode in QGC setup but not arming switch.

So I think this might be more useful generically than racers, but it is not clear where it should go. I have the same problem with kill switch, which currently gets referenced as a flight mode you can select [here](https://docs.px4.io/en/config/flight_mode.html#kill-switch). I am aware it is not a flight mode, but it made sense to put it there as it was set as a flight mode.
Proposal is to move docs for kill switch and arming switch into the Safety page here: https://docs.px4.io/en/config/safety.html - then link to that from everywhere.  Thoughts?

hamishwillee:
This all looks so much better than before. I made some modes, added some questions. Built version https://hamishwillee.gitbooks.io/ham_px4userguide/content/v/racer_setup_page/en/advanced_config/racer_setup.html

en/advanced_config/racer_setup.md
dagar: I've found there's a fair amount of user confusion over the kill switch and arming switch.

hamishwillee: @dagar Me too. Can you resolve it - ie explain the effective difference?

en/advanced_config/pid_tuning_guide_multicopter.md
bkueng: It depends on the explanation. The way I meant it is that with airmode, the mixer is allowed to change the commanded thrust. This can be towards positive or negative direction. Which is why the logic for the upper bound is the same as for airmode, just with different signs.

en/advanced_config/racer_setup.md
```diff
@@ -1,212 +1,200 @@
...
+
+A racer usually omits some sensors. 
+
+The minimal configuration is to use only a gyro and accelerometer sensor.
+
+> **Note** If the board has an internal magnetometer, it should not be used (small racers are particularly prone to strong electromagnetic interference).
+
+Racers typically do not have a GPS as it adds some weight and is prone to damage during crashes
+(a GPS + external magenetometer must be placed on a GPS mast away from high currents to avoid
+mag interferences, which unfortunately means it easy to break).
```
bkueng: it is easy to?

en/advanced_config/racer_setup.md
```diff
@@ -1,212 +1,200 @@
...
 
 ### Vibration Setup
-There are various ways how to do the mounting to reduce vibrations, and there is
-probably not a single best method. You will however have much less problems with
-vibrations if you use high-quality components (frame, motors, props) as for
-example used in [this build](../frames_multicopter/qav_r_5_kiss_esc_racer.md).
-Make sure to use **balanced props**. The FC can be mounted with
-[O-rings](../frames_multicopter/qav_r_5_kiss_esc_racer.md#mounting).
+
+There are various mounting approaches to reduce vibrations. For example, the flight controller can be mounted with vibration dampening foam, or using [O-rings](../frames_multicopter/qav_r_5_kiss_esc_racer.md#mounting).
```
bkueng: The o-rings are not foam, but silicone or EPDM (this is important).

en/advanced_config/racer_setup.md
bkueng: Kill-switch: it is a safety switch that you can always trigger and it will always immediately kill the motors. Whereas the arming switch will have the same effect in manual or acro, but for example in position mode, it will not allow you to disarm while the land detector did not detect landing yet (while the vehicle is still in-air).

Also the armings logic is more complex (depends on flight mode, etc.), and thus if you are doing low-level FC software changes a kill-switch just gives you more confidence in making sure the system turns off once you trigger it.

> There is a practical difference that there is an option to select kill switch as a flight mode in QGC setup but not arming switch.

Yes the arming switch should be exposed as well in QGC. But I first wanted to change its logic a bit (https://github.com/PX4/Firmware/pull/9622). 

> Proposal is to move docs for kill switch and arming switch into the Safety page here: https://docs.px4.io/en/config/safety.html - then link to that from everywhere. Thoughts?

Sounds good. It is not limited to racers, but they are particularly interested in this.

en/advanced_config/racer_setup.md
```diff
@@ -1,212 +1,200 @@
...
+> **Tip** Many things described here can also be applied to improve the flight performance of other types of multicopters.
 
-> **Note** A racer usually leaves away certain sensors, such as the GPS, which means that less failsafe options are available.
+<span></span>
+> **Note** A racer usually omits some sensors (e.g. GPS). As a result, fewer failsafe options are available.
 
 
 ## Build Options
-Racers typically do not have a GPS, but you will get some benefits if you add one.
-### Minimal
```
bkueng: I somewhat liked the clearer distinction between minimal vs. GPS, because PX4 has the reputation that it requires a mag and GPS in order to work. Not sure how to best highlight that though.

en/advanced_config/pid_tuning_guide_multicopter.md
bkueng: I wanted to keep the level of explaining these to a minimum, as mostly only the tuning order is relevant. But I agree, a bit more is needed here.

en/advanced_config/pid_tuning_guide_multicopter.md
```diff
@@ -1,217 +1,185 @@
...
-### Step 2: Stabilize Roll and Pitch Rates
-
-#### P Gain Tuning
+- Roll rate control ([MC_ROLLRATE_P](../advanced_config/parameter_reference.md#MC_ROLLRATE_P), [MC_ROLLRATE_I](../advanced_config/parameter_reference.md#MC_ROLLRATE_I), [MC_ROLLRATE_D](../advanced_config/parameter_reference.md#MC_ROLLRATE_D))
+- Pitch rate control ([MC_PITCHRATE_P](../advanced_config/parameter_reference.md#MC_PITCHRATE_P), [MC_PITCHRATE_I](../advanced_config/parameter_reference.md#MC_PITCHRATE_I), [MC_PITCHRATE_D](../advanced_config/parameter_reference.md#MC_PITCHRATE_D))
+- Yaw rate control ([MC_YAWRATE_P](../advanced_config/parameter_reference.md#MC_YAWRATE_P), [MC_YAWRATE_I](../advanced_config/parameter_reference.md#MC_YAWRATE_I), [MC_YAWRATE_D](../advanced_config/parameter_reference.md#MC_YAWRATE_D))
 
-Parameters: [MC_ROLLRATE_P](../advanced_config/parameter_reference.md#MC_ROLLRATE_P), [MC_PITCHRATE_P](../advanced_config/parameter_reference.md#MC_PITCHRATE_P).
+A well-tuned rate controller is very important as it affects all flight modes.
+The difference between a badly and a well tuned rate controller is, for example, easily noticeable in [Position mode](../flight_modes/position_mc.md).
```
Stifael: This comment can lead to more confusion and it does not say much. Bad attitude and velocity control tuning will  be visible during position mode as well. It would be better to describe how bad rate control tuning is visible to the user during rate control, because bad rate control tuning should be visible in all modes.

en/advanced_config/pid_tuning_guide_multicopter.md
```diff
@@ -1,217 +1,185 @@
...
-attitude (\~30 degrees) with 6 times the angular speed, i.e. 3 radians/s
-or \~170 degrees/s. Then if gain for the inner loop `MC_ROLLRATE_P` is
-0.1 then thrust control output for roll will be 3 \* 0.1 = 0.3. This
-means that it will lower the speed of rotors on one side by 30% and
-increase the speed on the other one to induce angular momentum in order
-to go back to level.
+The controllers are layered, which means a higher-level controller passes its
+results to a lower-level controller. The lowest-level controller is the the **rate
+controller**, then there is the **attitude contoller**, and then the **position
+controller**.
```
Stifael: velocity controller is missing

LorenzMeier:
Please do any follow-up fixes in a new PR. Thanks!

en/advanced_config/pid_tuning_guide_multicopter.md
bkueng: The intention is to highlight the importance of the rate controller for all higher modes. You are right that the other controllers are important as well, but I often see logs where the rate controller is very badly tuned, and I want to improve that.

en/advanced_config/pid_tuning_guide_multicopter.md
```diff
@@ -1,217 +1,185 @@
...
-- Set `MC_YAW_P` to a low value, e.g. 1
+There are two ways to counteract that:
+- Adjust the **thrust curve** with the [THR_MDL_FAC](../advanced_config/parameter_reference.md#THR_MDL_FAC) parameter (preferred method).
+  The thrust to PWM mapping is linear by default — setting `THR_MDL_FAC` to 1 makes it quadratic. 
+  Values in between use a linear interpolation of the two.
+  Typical values are between 0.3 and 0.5.
+  You can start off with 0.3 and then increase it by 0.1 at a time. 
+  If it is too high, you will start to notice oscillations at lower throttle values. 
+  
+  > **Note** The rate controller must be re-tuned if you change this parameter.
```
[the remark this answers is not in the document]
bkueng: Yes. The gains are not far off but still need readjustments, because a fixed difference in thrust will lead to different motor commands with a different factor.

en/advanced_config/pid_tuning_guide_multicopter.md
```diff
@@ -223,3 +191,63 @@ aerial video optimal value may be much smaller to get smooth response.)
...
+- [MPC_ACC_UP_MAX](parameter_reference.md#MPC_ACC_UP_MAX) : 1000 
+- [MPC_ACC_DOWN_MAX](parameter_reference.md#MPC_ACC_DOWN_MAX) : 1000 
+- [MPC_JERK_MAX](parameter_reference.md#MPC_JERK_MAX) : 0 
+- [MPC_JERK_MIN](parameter_reference.md#MPC_JERK_MIN) : 1
+ -->
+
+
+### Airmode & Mixer Saturation {#airmode}
+
+The rate controller outputs roll, pitch, yaw and thrust commands, which need to be converted into individual motor thrust commands.
```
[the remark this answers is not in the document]
bkueng: Thanks, continued in #272

en/advanced_config/pid_tuning_guide_multicopter.md
hamishwillee: >  It would be better to describe how bad rate control tuning is visible to the user during rate control, because bad rate control tuning should be visible in all modes.

@bkueng I think that what @Stifael is saying is that it would be better to explain the *symptoms* for bad rate control, and I agree. If this is important then it should also be in a callout - ie warning, tip or note. 

So I would change to something like:

> **Note** A well-tuned rate controller is very important as it affects *all flight modes*. A badly tuned rate controller will be visible in [Position mode](../flight_modes/position_mc.md) (for example) as [Insert symptoms here].

en/advanced_config/pid_tuning_guide_multicopter.md
hamishwillee: @bkueng What I think you are saying is that whether airmode is turned on or off, if the total commanded thrust for any motor is higher than 100% the commanded thrust will be reduced until no motor is being requested to deliver more than 100%. 

This is confusing because 
- we only talk in airmode about increasing the commanded thrust, and we only talk about decreasing the commanded thrust when not in airmode
- This actually sounds more like the "not airmode" type of behaviour.

Perhaps change to 

> If mixing becomes saturated towards the upper bound the commanded thrust is reduced to ensure that no motor is commanded to deliver more than 100% thrust. This behaviour is true irrespective of whether Airmode is enabled or disabled.

A diagram would also be nice for this case.

en/advanced_config/pid_tuning_guide_multicopter.md
bkueng: Yep that's better. Symtoms are: twitches, the vehicle will not stand perfectly still in the air.

en/advanced_config/pid_tuning_guide_multicopter.md
bkueng: > and we only talk about decreasing the commanded thrust when not in airmode

This is not correct, we decrease commanded roll when not in airmode, not commanded thrust. This is why the logic is the same as in Airmode: it changes the commanded thrust, rather than changing the commanded roll.
If it's too confusing, we can add a diagram. But it will look like the Airmode case.

The changed text is good, but I'd like to mention that it's like the Airmode logic.

en/advanced_config/pid_tuning_guide_multicopter.md
Stifael: @hamishwillee yep thats what I meant.

en/advanced_config/esc_calibration.md
```diff
@@ -1,33 +1,30 @@
...
-Electronic Speed Controllers (ESC) translate the input from the RC or the
-flight controller to the motors. ESCs do not always have to be
-calibrated, but in most cases it's good to do it anyway. High-quality
-controllers come with a factory calibration and
-can be just configured using the `PWM_MIN` and `PWM_MAX` parameters. This
-ensures that all the ESCs in a multicopter will scale the same way the
-input from the Flight Controller (FC) to the motors.  However, many low-cost models do
-require calibration and so if unsure calibration is recommended. Here is
-how you do it.
+Electronic Speed Controllers (ESC) regulate the speed (and direction) of motors based on a PWM input from the flight controller (FC).
```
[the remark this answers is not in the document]
hamishwillee: Made clear in #273

en/advanced_config/racer_setup.md
hamishwillee: :-) Fixed in #272

en/advanced_config/pid_tuning_guide_multicopter.md
hamishwillee: I updated as suggested in https://github.com/PX4/px4_user_guide/pull/272/commits/7cabfa6cd1ea91f1b582c671f2c3dde6bb9fbefb

en/advanced_config/pid_tuning_guide_multicopter.md
hamishwillee: Great. Fixed inhttps://github.com/PX4/px4_user_guide/pull/272/commits/994d95c14305ce771d704cae8c650d19378e92e9

PR #259, opened 2018-06-13
update Snapdragon documentation

RomanBapst:
- the repositoy ``` snap-vislam-ros``` was renamed to ros-examples so I had to correct that in the docs
- added some comments into rc.local
- improved startup sequence to ensure px4 and mavros live in harmony
- added instructions how to install geographiclib which is required by mavros


LorenzMeier:
Thanks!

PR #264, opened 2018-06-14
Jenkins don't delete old generated content first

dagar:


LorenzMeier:
I'm concerned about regressions if we don't start with a clean slate.

dagar:
Yes so am I, but this was a quick fix to prevent your CNAME from being deleted in https://github.com/PX4/docs.px4.io.

How about we copy the CNAME file over to this repository so that everything in https://github.com/PX4/docs.px4.io is a complete throwaway. Same for PX4/Devguide.

dagar:
New PR - https://github.com/PX4/px4_user_guide/pull/265

PR #274, opened 2018-06-21
Safety switch docs

hamishwillee:
Hi @bkueng , This centralises the arming/kill switch changes we discussed. Can you please review?

### DIFF
--- a/en/config/safety.md
+++ b/en/config/safety.md
@@ -1,8 +1,10 @@
 # Safety Configuration (Failsafes) 
 
-PX4 has configurable failsafe systems to protect and recover your vehicle if something goes wrong! These allow you to specify areas and conditions under which you can safely fly, and the [action](#failsafe_actions) that will be performed if a failsafe is triggered (for example, landing, holding position, or returning to a specified point).
+PX4 has a number of safety features to protect and recover your vehicle if something goes wrong:
 
+* *Failsafes* allow you to specify areas and conditions under which you can safely fly, and the [action](#failsafe_actions) that will be performed if a failsafe is triggered (for example, landing, holding position, or returning to a specified point).
 The most important failsafe settings are configured in the *QGroundControl* [Safety Setup](#qgc_safety_setup) page. Others must be configured via [parameters](#failsafe_other).
+* [Safety switches](#safety_switch) on the remote control can be used to immediately crash or return the vehicle in the event of a problem. 
 
 ## Failsafe Actions {#failsafe_actions}
 
@@ -17,7 +19,7 @@ Warning | A warning message will be sent to *QGroundControl*.
 [Land mode](../flight_modes/land.md) | The vehicle will enter *Land mode*, and lands immediately.
 RC Auto Recovery (CASA Outback Challenge rules) | TBD
 Terminate/Flight termination | Turns off all controllers and sets all PWM outputs to a failsafe value (defined in airframe configuration using `FAILSAFE` variable). For a fixed-wing vehicle this might allow you to glide the vehicle to safety.
-Lockdown | Kills the motors (sets them to disarmed). This is the same as using the [kill switch](../config/flight_mode.md#kill-switch).
+Lockdown | Kills the motors (sets them to disarmed). This is the same as using the [kill switch](#kill_switch).
 
 > **Note** It is possible to recover from a failsafe action (if the cause is fixed) by switching modes. For example, in the case where RC Loss failsafe causes the vehicle to enter *Return mode*, if RC is recovered you can change to *Position mode* and continue flying. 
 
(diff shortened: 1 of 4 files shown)


en/config/safety.md
```diff
@@ -227,6 +229,44 @@ Parameter | Description
...
+
+The kill switch is enabled as part of *QGroundControl* [Flight Mode Setup](../config/flight_mode.md#safety_switch).
+
+
+### Arm/Disarm Switch {#arming_switch}
+
+> **Tip** The arming switch is similar to the [kill switch](#kill_switch), but cannot be enabled in all modes during flight (i.e. those modes where it less likely to be needed).
+
+An arm/disarm switch immediately disarms (stop) motors for those [flight modes](../getting_started/flight_modes.md) that *support disarming in flight* (i.e. for these modes it is the same as the [kill switch](#kill_switch)). This includes the modes: 
+- *Manual mode*
```
hamishwillee: Can you please tell me all the modes in each group? Makes it easier for users.

en/config/safety.md
```diff
@@ -1,8 +1,10 @@
 # Safety Configuration (Failsafes) 
 
-PX4 has configurable failsafe systems to protect and recover your vehicle if something goes wrong! These allow you to specify areas and conditions under which you can safely fly, and the [action](#failsafe_actions) that will be performed if a failsafe is triggered (for example, landing, holding position, or returning to a specified point).
+PX4 has a number of safety features to protect and recover your vehicle if something goes wrong:
 
+* *Failsafes* allow you to specify areas and conditions under which you can safely fly, and the [action](#failsafe_actions) that will be performed if a failsafe is triggered (for example, landing, holding position, or returning to a specified point).
 The most important failsafe settings are configured in the *QGroundControl* [Safety Setup](#qgc_safety_setup) page. Others must be configured via [parameters](#failsafe_other).
+* [Safety switches](#safety_switch) on the remote control can be used to immediately crash or return the vehicle in the event of a problem.
```
bkueng: Instead of crash I'd write 'stop all motors'. Because there are situations when you are on the ground (or really close to it) and you trigger it.

en/config/safety.md
bkueng: They are: Manual, Acro, Stabilized & Rattitude

en/config/safety.md
```diff
@@ -227,6 +229,44 @@ Parameter | Description
...
+A kill switch immediately stops all motor outputs, causing the vehicle to crash.
+
+The kill switch is enabled as part of *QGroundControl* [Flight Mode Setup](../config/flight_mode.md#safety_switch).
+
+
+### Arm/Disarm Switch {#arming_switch}
+
+> **Tip** The arming switch is similar to the [kill switch](#kill_switch), but cannot be enabled in all modes during flight (i.e. those modes where it less likely to be needed).
+
+An arm/disarm switch immediately disarms (stop) motors for those [flight modes](../getting_started/flight_modes.md) that *support disarming in flight* (i.e. for these modes it is the same as the [kill switch](#kill_switch)). This includes the modes:
```
bkueng: I would phrase this whole section differently: An arming switch is not really an alternative to a kill switch. It is a replacement for the default stick (dis)arming (and thus also used for arming). And it serves the exact same purpose: making sure there is an intentional step involved before the motors start. The same for disarming.

I think most people would use an arming switch because:
- preference to use a switch for arming instead of a stick motion (they might be used to it because they used another autopilot)
- it avoids accidentially triggering it in-air with a certain stick motion
- there is no delay, it reacts immediately

en/config/safety.md
```diff
@@ -227,6 +229,44 @@ Parameter | Description
...
+
+### Arm/Disarm Switch {#arming_switch}
+
+> **Tip** The arming switch is similar to the [kill switch](#kill_switch), but cannot be enabled in all modes during flight (i.e. those modes where it less likely to be needed).
+
+An arm/disarm switch immediately disarms (stop) motors for those [flight modes](../getting_started/flight_modes.md) that *support disarming in flight* (i.e. for these modes it is the same as the [kill switch](#kill_switch)). This includes the modes: 
+- *Manual mode*
+- *Acro mode*
+
+For modes that do not support disarming in flight, the command is ignored until after landing has been detected. This includes the modes:
```
bkueng: > until after landing has been detected

You actually have to trigger it again after you landed for it to take effect. We might want to refer to the auto disarming as well here.

en/config/safety.md
```diff
@@ -227,6 +229,44 @@ Parameter | Description
...
+## Safety Switches {#safety_switch}
+
+A safety switch allows you to immediately crash or return the vehicle from the remote control transmitter
+(if you lose control of the vehicle, this may be better than allowing it to continue flying).
+
+This section lists the available safety switches.
+
+### Kill Switch {#kill_switch}
+
+A kill switch immediately stops all motor outputs, causing the vehicle to crash.
```
bkueng: We should also note that the kill switch will not disarm the vehicle.

bkueng: Yes

hamishwillee:
THanks @bkueng . I fixed in line with your suggestions - can you please recheck?

I did NOT add info about auto disarming. Did not think it was the right place. Though if you think this is a "General safety tip" we could put this in its own section?

en/config/safety.md
```diff
@@ -227,6 +229,49 @@ Parameter | Description
...
+- It avoids accidentally triggering arming/disarming in-air with a certain stick motion.
+- There is no delay (it reacts immediately).
+
+The arm/disarm switch immediately disarms (stop) motors for those [flight modes](../getting_started/flight_modes.md) that *support disarming in flight*. This includes: 
+- *Manual mode*
+- *Acro mode*
+- *Stabilized*
+- *Rattitude*
+
+For modes that do not support disarming in flight, the switch is ignored during flight, but may be used after landing is detected. This includes *Position mode*.
```
bkueng: We might want to mention autonomous modes (such as mission) as well?

hamishwillee: Yes. I'll just change to "This includes *Position mode* and autonomous modes (e.g. *Mission*, *Land* etc.)". Thanks for review!

PR #280, opened 2018-06-25
Deprecate 3DR Pixhawk Mini

hamishwillee:
3DR have stopped selling hardware. This flight controller and associated components - e.g. telemetry radios are no longer in production. This either removes links or marks text as deprecated. For example it merges the quick start into the autopilot page to reduce the overall content to manage if this eventually gets archived. 

Archiving is a further step when this controller is no longer "in the wild".


en/telemetry/telemetry_wifi.md
```diff
@@ -1,20 +1,10 @@
...
-essid: APM_PIX
-password: 12345678
-```
-
-Connect your ground control station to the above WiFi SSID. After connecting the vehicle should automatically be detected and connect to *QGroundControl*.
-
-![3DR Wifi Telemetry Radio 1](../../assets/hardware/telemetry/3dr_telemetry_wifi_1.jpg)
-![3DR Wifi Telemetry Radio 2](../../assets/hardware/telemetry/3dr_telemetry_wifi_2.png)
+Available Wifi Telemetry modules include:
+* [3DR Telemetry Wifi](../telemetry/3dr_telemetry_wifi.md) (Discontinued)
```
hamishwillee: @bkueng This is no longer available so I split out the doc into sidebar as "discontinued". However it would be good to add other alternatives. Do you think the ESP8266 - WiFi module is reasonable? Anything else?

hamishwillee: PS This is merged, but would still like to address above.

bkueng: > Do you think the ESP8266 - WiFi module is reasonable?

Definitely. I don't know any other in that size class.

hamishwillee: Thanks. Now done.

PR #285, opened 2018-06-28
Add KSP to introduction that QGC can be used to download firmware

hamishwillee:
I think it is worth highlighting this benefit because it really makes Pixhawk a lot easier to use than alternatives.

### DIFF
--- a/en/flight_controller/pixhawk_series.md
+++ b/en/flight_controller/pixhawk_series.md
@@ -5,6 +5,9 @@
 
 Manufacturers have created many different boards based on the open designs, with form factors that are optimised for applications from cargo carrying though to first person view (FPV) racers.
 
+> **Tip** A main benefit of using a *Pixhawk series* controller is that *QGroundControl* automatically downloads the correct firmware for a connected autopilot. This is possible because all boards within a *version* are software-compatible.
+
+<span></span>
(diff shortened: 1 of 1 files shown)


en/flight_controller/pixhawk_series.md
```diff
@@ -5,6 +5,9 @@
 
 Manufacturers have created many different boards based on the open designs, with form factors that are optimised for applications from cargo carrying though to first person view (FPV) racers.
 
+> **Tip** A main benefit of using a *Pixhawk series* controller is that *QGroundControl* automatically downloads the correct firmware for a connected autopilot. This is possible because all boards within a *version* are software-compatible.
```
bkueng: Do you really mean to say it's the main benefit?

hamishwillee: "A" main benefit - from a usability perspective. We should probably have more information about what we see as the benefits. I guess things like flexiblity for manufacturers to customise the end result etc.

What would you suggest?

bkueng: Right, it's certainly a benefit, but I see others that are just as important, such as:
- software support (these are the best-maintained boards for use with PX4)
- lots of flexibility in terms of what you can attach
- high quality
- customizability, as you write
- widely used and thus well-tested, stability

hamishwillee: THanks! I've added all these in a section as: https://github.com/PX4/px4_user_guide/pull/285/commits/d4d37fd04ef2a7af0734ce135e205a69a7cc6bb5

OK?

PR #286, opened 2018-06-29
Add merged px4flow guide and devguide

hamishwillee:
This adds the px4flow guide and devguide from the old pixhawk.org site. I have merged the doc, because it is more useful to do so than have a separate devguide (which would sit oddly in this doc)

@bkueng Some duplication has been removed, but I think some of the user-ishcontent needs to be moved to devguide. I'll add specific notes for your attention.


en/sensor/px4flow.md
```diff
@@ -0,0 +1,456 @@
...
+The default orientation (meaning: zero rotation) is defined as Y on flow board pointing towards **front of vehicle** as shown in the following picture.
+
+![PX4Flow align with Pixhawk](../../assets/hardware/sensors/px4flow/px4flowalignwithpixhawk.jpg)
+
+
+The orientation of PX4Flow can be set using the parameter [MAV_SENSOR_ORIENTATION](https://mavlink.io/en/messages/common.html#MAV_SENSOR_ORIENTATION)
+
+> **Warning** This parameter can only be applied to PX4 flight stack and is **not** part of the sensor itself settings
+
+<!-- Remove below if orientation  param above is correct -->
```
hamishwillee: I commented this section because the SENS_FLOW_ROT no longer exists. I imagine it has been replaced by MAV_SENSOR_ORIENTATION but can't see where that happens. Can you confirm what is correct orientation param here?

en/sensor/px4flow.md
```diff
@@ -0,0 +1,456 @@
...
+It is highly recommended to use twisted / shielded I2C wires to minimize radiated noise ([I2C spec](http://www.nxp.com/documents/user_manual/UM10204.pdf)). From the spec:
+
+If the bus lines are twisted-pairs, each bus line must be twisted with a VSS return. Alternatively, the SCL line can be twisted with a VSS return, and the SDA line twisted with a VDD return. In the latter case, capacitors must be used to decouple the VDD line to the VSS line at both ends of the twisted pairs.
+If the bus lines are shielded (shield connected to VSS), interference is minimized. However, the shielded cable must have low capacitive coupling between the SDA and SCL lines to minimize crosstalk.
+
+
+
+## I2C
+
+The 7 Bit **I2C Address** of the Flow Module is user selectable. Using the solder jumpers on the back of the flow board you can add an offset to the baseaddress 0x42. The address range for the 8 possible choices is: 0x42 - 0x49.
```
hamishwillee: I think most of this section actually belongs in devguide right? My understanding is that all I2C thingies are connected in the same way - if not, can we get instructions on how you would connect it to pixhawk?

en/sensor/px4flow.md
```diff
@@ -0,0 +1,456 @@
...
+###  Test Pattern Mode
+
+If the parameter IMAGE_TEST_PAT is set to 1, the sensor inserts a vertical gray-shaded test pattern in the signal chain. 
+* Test Pattern 64x64 (VIDEO ONLY Mode is OFF)
+  ![test_pattern_64x64](../../assets/hardware/sensors/px4flow/test_pattern_64x64.png)
+* Test Pattern 376x240 (VIDEO ONLY Mode is ON)
+  ![test_pattern_376x240](../../assets/hardware/sensors/px4flow/test_pattern_376x240.png)
+
+
+## General Troubleshooting
```
hamishwillee: Should this go up in the user guide section?

en/sensor/px4flow.md
bkueng: The parameter still exists: https://github.com/PX4/Firmware/blob/master/src/modules/sensors/sensor_params_flow.c#L54

MAV_SENSOR_ORIENTATION is for the distance sensor, used by the distance message: https://mavlink.io/en/messages/common.html#DISTANCE_SENSOR.
Since the device works via I2C, we don't need to mention MAV_SENSOR_ORIENTATION here.

en/sensor/px4flow.md
bkueng: Right, they are connected the same way. Each device has an address and if you have multiple devices on the same bus, all addresses need to be different. This is why the px4flow module allows to change the address (which happens typically via solder jumpers), and it's important to mention it.
It's one of those things I don't know if it belongs to the dev- or user-guide. A user connecting different devices to a single I2C bus should know about addresses.

en/sensor/px4flow.md
bkueng: Yes it can be in the user guide

en/sensor/px4flow.md
hamishwillee: Thanks! Actually I was being mad - that's a message, not a parameter. I have fixed: https://github.com/PX4/px4_user_guide/pull/286/commits/15392f80e79fbfbf4dbe7cd749240627cfcc9b13

en/sensor/px4flow.md
hamishwillee: @bkueng Thank you. So how about this fix. 
1. We remove all the stuff immediately under Wiring (because any connection to I2C is the same - according to above). Specifically, take away this bit:
   > It is highly recommended to use twisted / shielded I2C wires to minimize radiated noise ([I2C spec](http://www.nxp.com/documents/user_manual/UM10204.pdf)). From the spec:
   >
   > If the bus lines are twisted-pairs, each bus line must be twisted with a VSS return. Alternatively, the SCL line can be twisted with a VSS return, and the SDA line twisted with a VDD return. In the latter case, capacitors must be used to decouple the VDD line to the VSS line at both ends of the twisted pairs.
   > If the bus lines are shielded (shield connected to VSS), interference is minimized. However, the shielded cable must have low capacitive coupling between the SDA and SCL lines to minimize crosstalk.
2. Instead we add a statement. 
   > The PX4flow is connected to the I2C bus. If you are connecting multiple devices to the same bus you will need to set each with a unique address. The next section explains how.
3. We rename this section from `## I2C` to `Set PX4FLOW I2C Device address`
4. What/who is the bit below :`The **I2C frame** contains 22 bytes of data in the following order:`. Doesn't look like is part of setting the I2C address. I am guessing it is driver implementation information. What do you think. It isn't generic or I would suggest having it in http://dev.px4.io/en/sensor_bus/i2c.html .

en/sensor/px4flow.md
hamishwillee: Thanks! Moved in https://github.com/PX4/px4_user_guide/pull/286/commits/4307920f6c371efb71cdd05c1e65be0a509ed738

en/sensor/px4flow.md
```diff
@@ -0,0 +1,456 @@
...
+https://youtu.be/0Jpq6DU_HVg
+{% endyoutube %}
+
+
+## Where to Buy
+
+Order this module from:
+
+* [Unmanned Tech](http://www.unmannedtechshop.co.uk/px4flow-smart-camera-optical-flow-sensor/) (UK)
+* [UAV store](http://www.uav-store.de/sensoren/px4-flow-kit/) (Germany & EU)
```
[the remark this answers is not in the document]
hamishwillee: Thank you. Done in: https://github.com/PX4/px4_user_guide/pull/286/files#diff-706bfeacf8ee11db84fdedc9cf9f5fd9

hamishwillee:
Thanks @bkueng . Just one open issue: https://github.com/PX4/px4_user_guide/pull/286/files#r200073746

en/sensor/px4flow.md
bkueng: 1-3: sounds good
4: It describes the communication protocol and should IMO be documented in the driver: https://github.com/PX4/Firmware/blob/master/src/drivers/px4flow/i2c_frame.h. I can move it there if you wish.

hamishwillee: Thanks Beat! I'll do 1-3 tomorrow, and delete "4". If you could move the comms part of the protocol that would be much appreciated.

hamishwillee: Fixed in https://github.com/PX4/px4_user_guide/pull/286/commits/7e89e5265be6dc02868266ee096d38e97ec811f8

Thank @bkueng , merging now.

PR #288, opened 2018-07-02
Update FW landing slope image

hamishwillee:
This update follows on from discussion here: https://github.com/PX4/Firmware/issues/9752#issuecomment-401380598


thomasgubler:
lgtm

PR #291, opened 2018-07-02
Add power module as esc calibration precondition

hamishwillee:
This originates from https://github.com/PX4/Firmware/pull/9347#issuecomment-401738727


RomanBapst:
@hamishwillee great, thanks for that!

PR #295, opened 2018-07-18
Update lidar_lite connection/wiring info

### LINKED ISSUE #294: Lidar-Lite misleading note for the Intel Aero
Hi Hamish,

The [Lidar-Lite page](https://github.com/PX4/px4_user_guide/blob/67745a094608bcaaca9c4452f3b078270d4083ad/en/sensor/lidar_lite.md#lidar-lite) has a little bit misleading note for the Intel Aero. The note links you to the Intel Aero page just to find out a [waring](https://github.com/PX4/px4_user_guide/blob/df32443e5fbb14240270a26e93526f7ddeccfd0b/en/flight_controller/intel_aero.md#connecting-lidar-lite-range-finder-lidar_lite) actually not recommending to use it with the Intel Aero. If the warning for the Intel Aero is correct, then it's probably better to add it to the Lidar-Lite page as well. Unless that waring only applies to the previous v2 version of the sensor.

Also, notice that the first picture at the top of the The [Lidar-Lite page](https://github.com/PX4/px4_user_guide/blob/67745a094608bcaaca9c4452f3b078270d4083ad/en/sensor/lidar_lite.md#lidar-lite) shows the latest v3 version of the sensor; which now has [a different connector and pin layout sequence](http://static.garmin.com/pumac/LIDAR_Lite_v3_Operation_Manual_and_Technical_Specifications.pdf). However, the [wiring section](https://github.com/PX4/px4_user_guide/blob/67745a094608bcaaca9c4452f3b078270d4083ad/en/sensor/lidar_lite.md#wiring) shows the previous version of the sensor; which users should not follow to connect the latest v3 model since as shown below the pin sequence is different.  Therefore, I suggest to add a waring in that section.

screen shot 2018-07-17 at 9 34 09 am

hamishwillee:
- Fixes #294
- Fixes up docs that suggest you connect to Aux "row 6", when means "column"
- Adds info about LidarLite pinout - so can be used for I2C connections by people who know how to wire I2c. Also adds the garmin spec for v3 and the standard wiring images from spec.
- Clarifies that v3/2 pinout is the same.

Open question 
- Pinout v2/3 is same, but physical layout of what is pin 1 and pin 6 may have flipped. Confirm. 
- v3 advice is 1KOhm pulldown resister while v2 advice is 470Ohms. Does it matter? What should we recommend?


bkueng:
Re pinout: I don't know the numbering, but red is 5V. And the cable seems to be colored, see https://www.jetsonhacks.com/2015/11/02/lidar-lite-v2-installation-and-demo-nvidia-jetson-tk1/ for example.

Looks good, but I don't know about the resister. @DanielePettenuzzo might know.

DanielePettenuzzo:
@hamishwillee @bkueng both values are good. We just don't want to go too low because then currents start to get high. We also don't want to go too high otherwise we create an RC circuit and risk to have bad waveforms (not squared waves anymore). Looked online and seems that anything between 200 ohms and 1 kohm will be fine. Let's just go for 1kohm to keep it simple.

hamishwillee:
Thanks @bkueng @DanielePettenuzzo  - changed resister to 1KOhm as suggested and merged.

@bkueng Decided to "ignore" confusing pinouts. I think it is just the case that the Garmin images don't show the laser, so it looks like you're looking at the image from the bottom.

PR #296, opened 2018-07-19
Update pixracer.md

ocularmagic:
The link to buy the pixracer at the AUAV store goes to the web site, but it is now just text that says "AUAV Expect good things."  There is no store, or links to buy the hardware.  I changed the link to go to the mRobotics version which is currently working.


hamishwillee:
Thanks @ocularmagic !

PR #303, opened 2018-07-31
omnibus: add CRSF crossfire telemetry section

bkueng:
As discussed @hamishwillee. I wanted to add the section close to the UARTs documentation, though it might not be the best ordering.

### DIFF
--- a/en/flight_controller/omnibus_f4_sd.md
+++ b/en/flight_controller/omnibus_f4_sd.md
@@ -135,6 +135,27 @@ Here is an example implementation. I used a Spektrum plug to get 3.3v from the D
 ![Omnibus F4 SD Pullup Implementation](../../assets/flight_controller/omnibus_f4_sd/pullup.jpg)
 
 
+## CRSF Crossfire Telemetry
+In addition to the [FrSky Telemetry](../peripherals/frsky_telemetry.md), the Omnibus also supports TBS CRSF Crossfire telemetry.
+It is used to send telemetry data from the flight controller (the vehicles's attitude, battery, flight mode and GPS data) to the RC transmitter (Taranis).
+
+Benefits are:
+- Only a single UART is needed for RC and telemetry.
+- The CRSF protocol is optimized for low latency.
+- 150 Hz RC update rate.
+- The signals are uninverted and thus no (external) inverter logic is required.
+
+It is best to use the [TBS Crossfire Nano RX](http://team-blacksheep.com/products/prod:crossfire_nano_rx), since it is specifically designed for small Quads
+(the referenced link also contains the documentation).
+
+On the radio you will also need a [Transmitter Module](http://team-blacksheep.com/shop/cat:rc_transmitters#product_listing).
+You can plug this to the back of the Taranis for example.
+
+### Setup
+To setup the Nano RX on the Omnibus, connect the TX pin of UART1 to Ch2 and the RX pin to Ch1 of the Nano.
+Nothing else needs to be configured on the flight controller side, the RC protocol is auto-detected.
+The next steps include enabling the CRSF protocol and setting up the telemetry.
+This is documented in the [TBS Crossfire Manual](https://www.team-blacksheep.com/tbs-crossfire-manual.pdf), search for 'Setting up radio for CRSF'.
 
 ## Schematics
 


en/flight_controller/omnibus_f4_sd.md
```diff
@@ -135,27 +135,41 @@ Here is an example implementation. I used a Spektrum plug to get 3.3v from the D
 ![Omnibus F4 SD Pullup Implementation](../../assets/flight_controller/omnibus_f4_sd/pullup.jpg)
 
 
-## CRSF Crossfire Telemetry
-In addition to the [FrSky Telemetry](../peripherals/frsky_telemetry.md), the Omnibus also supports TBS CRSF Crossfire telemetry.
-It is used to send telemetry data from the flight controller (the vehicles's attitude, battery, flight mode and GPS data) to the RC transmitter (Taranis).
+## RC Telemetry
```
hamishwillee: FYI pushed it down a level because 
- this is RC telemetry, not Wifi telemetry (ie wanted to clarify at heading level)
- We may well want to add a section on FrSky telemetry and now there is a place for that. Reason, I have no idea how you would connect an FrSky receiver for telemetry to this board (?)
- This makes the setup heading a lower level, which in turn differentiates that from the next section on schematics (ie before they looked to be associated).

en/flight_controller/omnibus_f4_sd.md
```diff
@@ -135,27 +135,41 @@ Here is an example implementation. I used a Spektrum plug to get 3.3v from the D
...
 
-Benefits are:
+The Omnibus supports telemetry to the RC Transmitter using [FrSky Telemetry](../peripherals/frsky_telemetry.md) or [CRSF Crossfire Telemetry](#crsf_telemetry).
+
+
+### CRSF Crossfire Telemetry {#crsf_telemetry}
+
+TBS CRSF Crossfire telemetry is used to send telemetry data from the flight controller (the vehicle's attitude, battery, flight mode and GPS data) to the RC transmitter (Taranis).
+
+Benefits over FrSky telemetry include:
```
hamishwillee: Note that these are benefits over FrSky telemetry , not benefits "in abstract" (except for the update rate) so I added that.

en/flight_controller/omnibus_f4_sd.md
```diff
@@ -135,27 +135,41 @@ Here is an example implementation. I used a Spektrum plug to get 3.3v from the D
...
+On the handheld controller (e.g. Taranis) you will also need a [Transmitter Module](http://team-blacksheep.com/shop/cat:rc_transmitters#product_listing).
+This can be plugged into the back of the RC controller.
+
+> **Note** The referenced links above contains the documentation for the TX/RX modules.
+
+#### Setup
+
+Connect the Nano RX and Omnibus pins as shown:
+
+Omnibus UART1 | Nano RX
```
hamishwillee: FYI This is just a layout choice. 

@bkueng I was hoping here to link to information about the UART1 pins. However if you look above, we don't say what the pinout is (we do cover UART6). So I suggest that perhaps you add information on the pinout for UART1 above and link to it.

en/flight_controller/omnibus_f4_sd.md
```diff
@@ -135,27 +135,41 @@ Here is an example implementation. I used a Spektrum plug to get 3.3v from the D
...
 - The CRSF protocol is optimized for low latency.
 - 150 Hz RC update rate.
 - The signals are uninverted and thus no (external) inverter logic is required.
 
-It is best to use the [TBS Crossfire Nano RX](http://team-blacksheep.com/products/prod:crossfire_nano_rx), since it is specifically designed for small Quads
-(the referenced link also contains the documentation).
+For Omnibus we recommend the [TBS Crossfire Nano RX](http://team-blacksheep.com/products/prod:crossfire_nano_rx), since it is specifically designed for small Quads.
+
+On the handheld controller (e.g. Taranis) you will also need a [Transmitter Module](http://team-blacksheep.com/shop/cat:rc_transmitters#product_listing).
+This can be plugged into the back of the RC controller.
```
hamishwillee: So we talk about taranis here, and plugging into the back of the controller. However if you go to the TX link it doesn't explain what modules it can plug into - ie it is not clear what it is compatible with in terms of RC controllers either in terms of protocol OR physical connectors.

Perhaps all RC controllers have a standardized physical format for modules? 

If not, we need to confirm the supported set of RC controllers, ideally via link to docs.

hamishwillee:
@bkueng Looks good. I did some minor restructure and there are a couple of questions.

en/flight_controller/omnibus_f4_sd.md
bkueng: > I have no idea how you would connect an FrSky receiver for telemetry to this board (?)

We're going to need configurable UARTs to make this user-friendly.

en/flight_controller/omnibus_f4_sd.md
bkueng: > So I suggest that perhaps you add information on the pinout for UART1 above and link to it.

I don't know what more there could be added. UART1 is already mentioned for RC and the pictures contain the pins as well. Note that on that board you directly solder pins, there are no connectors (it's a different world than Pixhawk's).

en/flight_controller/omnibus_f4_sd.md
bkueng: > Perhaps all RC controllers have a standardized physical format for modules?

Yes the standard is called JR.

en/flight_controller/omnibus_f4_sd.md
hamishwillee: Will your audience know that TX1/RX1 are the TX/RX for the UART1? I'm more used to seeing UARTs labeled like "GPS", "TELEM1". If so, then fine! Often UARTs have other pins - for CTS etc. Looks like some UARTs do here and some don't -s that worth noting?

en/flight_controller/omnibus_f4_sd.md
hamishwillee: Thanks. I should add that in the RC documentation - we just hand-wave over this by suggesting people buy a setup RC Transmitter/Receiver bundle.

hamishwillee: I.e. this is good to go then from my side.

PR #306, opened 2018-08-07
Initial Quick Start document for BeagleBone Blue board

UAV-Pilot:
BeagleBone Blue board is similar (alternative) to Raspberry Pi 2/3 + Navio2 Autopilot kit.

PR #308, opened 2018-08-08
Updates to beagle_bone blue structure

hamishwillee:
@UAV-Pilot This tidies up rendering from #306. I have merged it to fix a few bugs, but there are still structural issues. I will add comments inline - can you please respond.

### DIFF
--- a/en/flight_controller/beaglebone_blue.md
+++ b/en/flight_controller/beaglebone_blue.md
@@ -1,38 +1,31 @@
 # BeagleBone Blue
 
-![](../../assets/hardware/BeagleBone_Blue_balloons.png)
+[BeagleBone Blue](https://beagleboard.org/blue) is an all-in-one Linux-based computer. 
+Although it is optimized for robotics, this compact and inexpensive board has all necessary sensors and peripherals needed by a flight controller. 
+This topic shows how to set up the board to run PX4 with [librobotcontrol](https://github.com/StrawsonDesign/librobotcontrol) robotics package.
 
-## Quick Start
-
-[BeagleBone Blue](https://beagleboard.org/blue) is an all-in-one Linux-based 
-computer. Although it is optimized for robotics, this compact inexpensive 
-board has all necessary sensors and peripherals needed by a flight controller. 
-This topic shows how to set up the board to run PX4 with 
-[librobotcontrol](https://github.com/StrawsonDesign/librobotcontrol) robotics 
-package.
-
-### OS Image
-
-The latest official OS image for [BeagleBone Blue](https://beagleboard.org/blue) 
-is available [here](https://beagleboard.org/latest-images), and test OS images 
-are frequently released on [this site](https://rcn-ee.net/rootfs/bb.org/testing/). 
-Refer to [this page](https://github.com/beagleboard/beaglebone-blue/wiki/Flashing-firmware) 
-when flashing OS image. Other useful information can be found in 
-[this FAQ](https://github.com/beagleboard/beaglebone-blue/wiki/Frequently-Asked-Questions-&lpar;FAQ&rpar;).
+![BeagleBone - labelled diagram](../../assets/hardware/BeagleBone_Blue_balloons.png)
 
 
-### Robot Control Library
+## OS Image
 
-On [BeagleBone Blue](https://beagleboard.org/blue), PX4 uses 
-[librobotcontrol](https://github.com/StrawsonDesign/librobotcontrol) of 
-version 1.0.0 or higher. BeagleBoard OS images come with librobotcontrol 
-preinstalled. We need choose OS images in which librobotcontrol works 
-properly. Optionally you can update to a realtime kernel, and if you do, 
-re-check if librobotcontrol works properly with the realtime kernel.
+*BeagleBone Blue* images can be found here:
+- [Latest stable OS image](https://beagleboard.org/latest-images).
+- [Test OS images](https://rcn-ee.net/rootfs/bb.org/testing/) (updated frequently).
 
-One way to check if librobotcontrol works properly is to run rc_test_drivers 
-which comes with librobotcontrol. As shown in the following example, all tests
-should pass. Optionally run other tests such as rc_test_bmp, rc_test_mpu, etc.
+Information about flashing OS images can be found on [this page](https://github.com/beagleboard/beaglebone-blue/wiki/Flashing-firmware).
+Other useful information can be found in the [FAQ](https://github.com/beagleboard/beaglebone-blue/wiki/Frequently-Asked-Questions-&lpar;FAQ&rpar;).
+
+
+## Robot Control Library
+
+On [BeagleBone Blue](https://beagleboard.org/blue), PX4 uses [librobotcontrol](https://github.com/StrawsonDesign/librobotcontrol) version 1.0.0 or higher.
+
+BeagleBoard OS images come with *librobotcontrol* preinstalled, but it may not work properly in all OS images.
+
+One way to check if *librobotcontrol* works properly is to run *rc_test_drivers* which comes with *librobotcontrol*. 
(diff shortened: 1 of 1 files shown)


en/flight_controller/beaglebone_blue.md
```diff
@@ -65,102 +58,88 @@ Currently running on a:
...
 Robot Control library Version:
 1.0.0
-
 ```
 
-As of this writing, the latest OS images in which librobotcontrol works properly
-is [bone-debian-9.5-iot-armhf-2018-07-22-4gb.img.xz](https://rcn-ee.net/rootfs/bb.org/testing/2018-07-22/stretch-iot/bone-debian-9.5-iot-armhf-2018-07-22-4gb.img.xz).
+> **Tip** Optionally you can update to a realtime kernel, and if you do, re-check if *librobotcontrol* works properly with the realtime kernel.
 
-#### Setup Robot Control Library
```
hamishwillee: @UAV-Pilot it isn't clear which parts need to be done on dev computer vs Beaglebone Blue or both.

en/flight_controller/beaglebone_blue.md
```diff
@@ -65,102 +58,88 @@ Currently running on a:
...
 
-1. use the one pre-installed in BeagleBoard images
+If you want to build PX4, there are additional setup steps for this library.
 
-2. install from debian package or repository 
-```sh
-    sudo apt update && sudo apt install librobotcontrol
-```
+At time of writing the *librobotcontrol* debian package is only available on BeagleBoard products (not BeagleBone Blue). 
+Here are the ways to obtain the *librobotcontrol* on BeagleBone Blue:
```
hamishwillee: So just to confirm, you only need to do these on beaglebone blue?
Do you also need to get this library/do this if you are doing a cross-compiler build?

en/flight_controller/beaglebone_blue.md
```diff
@@ -65,102 +58,88 @@ Currently running on a:
...
 
-3. install from source
-```sh
-    git clone https://github.com/StrawsonDesign/librobotcontrol.git
-    cd librobotcontrol
-    sudo make install
-```
+1. Use the one pre-installed in BeagleBoard images.
+1. Install from debian package or repository:
+   ```sh
```
hamishwillee: @UAV-Pilot Didn't you just say that this debian repo isn't available for BBB?

en/flight_controller/beaglebone_blue.md
```diff
@@ -65,102 +58,88 @@ Currently running on a:
...
+   ```
 
 After acquiring the pre-built library,
 
-* select librobotcontrol installation directory, and define it as LIBROBOTCONTROL_INSTALL_DIR environment variable so that other unwanted headers will not be included
-
-* install robotcontrol.h and rc/* into $LIBROBOTCONTROL_INSTALL_DIR/include
-
- install pre-built native (ARM) version of librobotcontrol.* into $LIBROBOTCONTROL_INSTALL_DIR/lib
+1. Select the *librobotcontrol* installation directory, and set it in the `LIBROBOTCONTROL_INSTALL_DIR` environment variable so that other unwanted headers will not be included
```
hamishwillee: @UAV-Pilot Is this just on the BBB for native builds or something we have to do on dev computer too?

My guess is that you probably need to do it on the computer you are building on.

en/flight_controller/beaglebone_blue.md
```diff
@@ -65,102 +58,88 @@ Currently running on a:
...
 
-```sh
-cd /home/debian/px4 
-sudo ./bin/px4 -s px4.config 
-```
-Currently librobotcontrol requires root access.
-
-Note that if you use a PX4 version before [Posix Shell](https://github.com/PX4/Firmware/pull/10173) 
-is available, run the following commands instead:
+1. First set up *rsync* (this is is used to transfer files from the development computer to the target board over a network - WiFi or Ethernet).
```
hamishwillee: @UAV-Pilot YOu will note that I removed this from a top level topic because it read as something you only needed to do for cross compiler build. 

Is that correct? Presumably you need to set up SSH for BOTH build because otherwise you can't talk to the BBB to set up autostart etc?

en/flight_controller/beaglebone_blue.md
```diff
@@ -65,102 +58,88 @@ Currently running on a:
...
+   sudo apt-get update
+   sudo apt-get install cmake python-empy
+   ```
+1. Clone the PX4 Firmware directly onto the BeagleBone Blue.
+1. Continue with the [standard build system installation](https://dev.px4.io/en/setup/dev_env_linux.html).
 
-Continue with our [standard build system installation](https://dev.px4.io/en/setup/dev_env_linux.html).
 
-### Autostart during Boot
+## Autostart During Boot
```
hamishwillee: @UAV-Pilot Note, moved up to top level, because presumably this is something you want to do however you build.

Some concerns:
1. This is something done on the BBB right? Should say that. Also how do we modify these - via nano or similar over SSH?
2. There are two files described [/etc/rc.local] and *systemd*. Are these alternative different methods for autostart or two things you need to do to get systemd autostart to work?

en/flight_controller/beaglebone_blue.md
UAV-Pilot: If user chooses native build, then he/she does not need setup a separate dev computer, so librobotcontrol setup always need be setup in Beaglebone Blue as the executable need load the library at run time, and it only need be setup on the dev computer if the user chooses cross compile.

en/flight_controller/beaglebone_blue.md
UAV-Pilot: This debian repo is available for BBBlue.

en/flight_controller/beaglebone_blue.md
UAV-Pilot: BeagleBone Blue is part of BeagleBoard products, so librobotcontrol debian package is available on BeagleBone Blue. Currently, the build (make) system in librobotcontrol is setup in a way that does not allow cross compile (e.g., no configure script, C compiler is fixed to gcc, etc) so you cannot easily cross compile on other machine (e/g/, a machine with Intel CPU) without changing its build system. Therefore normal users can only get the library binary on BeagleBoard products (in our case, it's BeagleBone Blue). For cross compile dev computer, the user need copy the library binary and header files from BeagleBone Blue (our in rare case another BeagleBoard product, e.g., BeagleBone Black) to the dev computer.

en/flight_controller/beaglebone_blue.md
UAV-Pilot: If it's cross compile then these steps only need be done on cross compile dev computer, otherwise only on native board.  And yes, user need do it on the machine that compiles PX4.

en/flight_controller/beaglebone_blue.md
UAV-Pilot: Sorry for the confusions.
1. Yes, on BBB. Yes, any editor like nano, vi over SSH.
2. They alternative methods, and only one method should be used. These 2 methods should not be applied at the same time.

UAV-Pilot: BTW, autostart is only needed when you actually fly a multicopter or a fixed wing airplane, etc. For most part of my porting work, I used manual start to test code change. Yes, for normal users who just use PX4, autostart is needed otherwise it's not convenient.

en/flight_controller/beaglebone_blue.md
UAV-Pilot: Yes, it's only needed for cross compile build. 

Without any setup after OS image installation,  tools like Putty already can talk to BBB over SSH. The SSH setup mentioned here is only needed if user wants to skip typing password in rsync operation, in other words, without the special SSH setup, rsync still works, but requires user to input password each time. So the SSH setup for rsync is optional, and it's for convenience.

PR #309, opened 2018-08-08
update pixhawk 4 datasheet

pfjare:
Update link to Pixhawk 4 technical data sheet


hamishwillee:
I think moving the datasheet here is a good idea - but do "we" own the master copy or Holybro? I just want to understand whether there is another version that we should be tracking.

pfjare:
I remade holybros basically so I could make sure the Pixhawk 4 Mini datasheet and the Pixhawk 4 datasheet had a consistent style.

hamishwillee:
Thanks for clarification. Makes perfect sense.

pfjare:
On the live site the link is the same. Do you push changes on a schedule? Jinger is asking me why it hasnt changed

dagar:
> On the live site the link is the same. Do you push changes on a schedule? Jinger is asking me why it hasnt changed

The credentials for PX4BuildBot changed and haven't been updated everywhere yet. I'll have it fixed soon.

pfjare:
Ok thank you

PR #310, opened 2018-08-09
Removed outdated info after posix shell feature was merged in, and ma…

UAV-Pilot:
Posix shell feature was merged in master today, so removed outdated info; also made one correction

### DIFF
--- a/en/flight_controller/beaglebone_blue.md
+++ b/en/flight_controller/beaglebone_blue.md
@@ -19,7 +19,7 @@ Other useful information can be found in the [FAQ](https://github.com/beagleboar
 
 ## Robot Control Library
 
-On [BeagleBone Blue](https://beagleboard.org/blue), PX4 uses [librobotcontrol](https://github.com/StrawsonDesign/librobotcontrol) version 1.0.0 or higher.
+On [BeagleBone Blue](https://beagleboard.org/blue), PX4 requires [librobotcontrol](https://github.com/StrawsonDesign/librobotcontrol) version 1.0.0 or higher.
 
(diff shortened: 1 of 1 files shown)


UAV-Pilot:
BTW, when will the update appear on https://docs.px4.io/en/ ? I'd like inform the creator/maintainer of librobotcontrol so that he might add a link to this document page on his web sites, and he might take PX4 usage of the library into considerations when he makes changes. He also suggested to me to write document about PX4 on BeagleBone Blue.

hamishwillee:
Thanks @UAV-Pilot 

1. What if people try to build for PX4 v1.7? Can they? If so, won't they still need that information you removed? I don't have a problem with saying this is only supported from a particular version, but we should make that clear.
2. This is built by jenkins. I'm not sure exactly how long it takes, but I think the cycle is about 12 hours. If you don't see it by then we chase Danial Agar to find out why.

UAV-Pilot:
Sorry I'm relatively new to PX4, and not quite clear with its releases. In fact this porting work is my first time to work with PX4. Since the initial pull request for the BeagleBone Blue porting was just merged into master branch this Monday (Aug 6), I thought removing those info should be fine. I just checked that the porting code is not in 1.8 release branch. I'm not sure how new features (not bug fixes) on master branch get ported back to a already released branch. I guess it should be fine and it's not difficult for them to figure out this kind of small details even they use the porting code in 1.7 release.

hamishwillee:
@UAV-Pilot You've done good work and have nothing to worry about. The guides reflect master, because we don't have resource to version them effectively. So what you have done is fine.

That said, sometimes I will add a note that something was introduced/removed in a particular version. Don't worry about it for this.

PR #320, opened 2018-08-15
optical flow: fix doc on how to enable flow with ekf2

ChristophTobler:
Just came across this: '1' means GPS only. The second bit needs to be enabled for flow


en/sensor/optical_flow.md
```diff
@@ -36,21 +36,20 @@ The easiest way to calculate the optical flow is to use the [PX4Flow](../sensor/
...
 You can use any supported [distance sensor](../sensor/rangefinders.md).
-However we recommend using a LIDAR over a Sonar, because of robustness and accuracy. 
+However we recommend using a LIDAR over a Sonar, because of robustness and accuracy.
 
 ## Estimators
 
 ### Extended Kalman Filter (EKF2)
 
-For Optical Flow fusion using EKF2, the parameter [EKF2_AID_MASK](../advanced_config/parameter_reference.md#EKF2_AID_MASK) must be set to `1`.
+For Optical Flow fusion using EKF2, the second bit of the parameter [EKF2_AID_MASK](../advanced_config/parameter_reference.md#EKF2_AID_MASK) must be enabled (e.g. flow only -> `2`).
```
hamishwillee: @ChristophTobler Thanks. It is definitely more correct to refer to "bits enabled" rather than a numerical value. Do people tend to index bitmaps from 1 (or zero)? 

Perhaps the last bit might be better as (i.e. to enable just flow the decimal value would be 2).

mhkabir: I think it's a bad idea to say bits/numbers in user-facing docs, since QGC exposes either check-boxes or a drop-down list with strings, which are in no way shown to be correlated with these numbers.

hamishwillee: @mhkabir Good point. I was going to argue that the information about bits is also useful for people working with the parameter via a developer API, but then they will know to follow the link to the parameter. 

How about we add a screenshot. So something like:

> For optical flow fusion using EKF2, set the *use optical flow* flag in the [EKF2_AID_MASK](../advanced_config/parameter_reference.md#EKF2_AID_MASK) parameter, as shown using *QGroundControl* below:
>  image

ChristophTobler: Sure, makes sense!

hamishwillee: Great :-). Are you set up with gitbook enough to do this - or do you want me to look at it?

ChristophTobler: Mh why gitbook? can I not just add the pic and link it to the md file?

hamishwillee: I only ever use git to add images. If you're able via some other tool - ie the gitbook editor, be my guest :-)

ChristophTobler:
@hamishwillee does that work for you?

hamishwillee:
Yes it does. Thanks! merged.

PR #337, opened 2018-09-17
Add pwm escs and servo from old pixhawk.org site

hamishwillee:
@bkueng I made a few minor revisions to this since Devguide PR. Specifically I think this would be improved by connection diagrams. While I don't have time for those now, I've split the doc up a little into sections so that it would be obvious how to add them if we choose to do so in future.

I think it is good to merge.


en/peripherals/pwm_escs_and_servo.md
```diff
@@ -0,0 +1,112 @@
...
+
+Always connect Power VBAT and GND to the battery, and the PWM signal and GND from the servo plug to the motor.
+
+> **Tip** There is **no setup** that does not require signal ground to be connected!
+
+
+The connection to the +5V wire (if present) depends on the ESC/Vehicle.
+
+
+### Fixed Wing
```
bkueng: / VTOL?

hamishwillee:
Thanks @bkueng - I took your suggestion to make the FW section also apply to VTOL and merged.

PR #341, opened 2018-09-18
Clarify what body frame is

hamishwillee:
This doc is the place where body frame is effectively defined, but that term was not used. This change at least means the page will show up when the term is searched. 

@bkueng I have been avoiding creating a glossary but this seems like it might be a good idea? 

FYI If so I will avoid the gitbook glossary infrastructure as this automatically cross links glossary terms (slows build time to a crawl).

### DIFF
--- a/en/config/flight_controller_orientation.md
+++ b/en/config/flight_controller_orientation.md
@@ -4,7 +4,7 @@ By default the flight controller (and external compass(es), if present) should b
 
 ## Calculating Orientation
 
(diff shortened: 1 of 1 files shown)


mhkabir:
I actually want to create a "frames of reference" page for the Devguide describing all the possible and implemented frames in Firmware.

FYI @Tsc21

hamishwillee:
> I actually want to create a "frames of reference" page for the Devguide describing all the possible and implemented frames in Firmware.

@mhkabir Yes please :-). Depending on what you end up with, it might belong better in the user guide. This is a quick fix for a particular problem.

PR #346, opened 2018-09-20
Update sensor_selection.md

ocularmagic:
PX4Flow image link is broken.  Link format looks right, so the image may not be in the correct directory?  Deleted the reference in case there was a problem with the image.


hamishwillee:
Thanks - very much appreciated! I patched this with the right image and merged.

PR #349, opened 2018-09-25
Update vtol_quadplane_falcon_vertigo_hybrid_rtf_dropix.md

ocularmagic:
Looks like flyingwings is going out of business.  The link no longer works and their website says they're closing soon.  I couldn't find any other links to purchase the RTF kit so I just labeled it as discontinued.


hamishwillee:
@ocularmagic Thanks very much. I'll also move it down the list as a post-process

PR #352, opened 2018-09-26
add a page to document the serial port configuration

bkueng:
@hamishwillee this is my first draft for the new serial configuration parameters. Feel free to edit/take over.

In response to https://github.com/PX4/Firmware/pull/10337.

Still missing:
- remove the SENS_EN_* params
- remove SYS_COMPANION param and add references to this page as needed (also on the dev-guide)


hamishwillee:
@bkueng I did a bit of a play with structure - it is very similar. I am not yet ready to take over. 

Some thoughts:
- Is my section on "Deconfliction" correct? What do we advise here?
- The ports are UART 6, TELEM 1, TELEM 2, TELEM 3, TELEM/SERIAL 4, GPS 1, GPS 2. Are these "standard"? I remember seeing SERIAL4/5 on many devices but that does not appear to be mentioned here?
- GPS port is often labelled GPS/I2C. Is this port always I2C/what port is normally I2C (basically I'm wondering whether there is some conflict we need to talk about here. 
- MAV_0_CONFIG is telemetry stream for GCS. What is MAV_1_CONFIG? What if I want another stream?
- What serial ports other than RCIN can't be assigned? Can SPKT/DSM be used. More specifically, is it worth listing the ports that definitively can't be used?
- Is there any advice for something that can be connected to both I2C and a UART? Ie how it is configured?

en/sensor/leddar_one.md
```diff
@@ -1,44 +1,15 @@
 # LeddarOne Lidar
```
hamishwillee: @bkueng This is updated LeddarOne page. I've just done this one as a tester for you to review. My assumption is that all you need to do is set up the params as in your serial config doc - there is no longer any specific ports that need to use on each board. 

However I assume you STILL need to check the driver is present. Would we be better off saying it isn't in FMUv2 but can be expected in other boards?

hamishwillee:
OK, so most of the remainder of the docs should be OK - they are much like the LeddarOne doc. 
However I am not sure how to approach [MAVLink (OSD/Telemetry)](https://docs.px4.io/en/peripherals/mavlink_peripherals.html). 

Is this MAV_1_CONFIG ? So OSD simply becomes map this to the port you want and set the baud rate you want.

Actually, that is another question for OSD and for LeddarOne and other rangefinders - what baud rate should we set for those?

hamishwillee:
Had a brief look at [devguide use of SYS_COMPANION](https://dev.px4.io/en/?q=SYS_COMPANION). Basically same questions I don't "get" 
- what mavlink "service" config parameter intended for Companion computer and what baud rate (we only have two mavlink instances, so I then wonder what for OSD vs Companion vs GCS)
- What to do in [HITL doc](https://dev.px4.io/en/simulation/hitl.html)

en/peripherals/serial_configuration.md
```diff
@@ -0,0 +1,49 @@
...
+1. Reboot the vehicle in order for the other parameters to show up.
+1. Select the **Serial** tab, and open the [SER_TEL2_BAUD](../advanced_config/parameter_reference.md#SER_TEL2_BAUD) parameter (`TELEM 2` port baud rate): set it to *Auto*.
+  ![QGC Serial Baudrate Example](../../assets/peripherals/qgc_serial_baudrate_example.png)
+
+
+## Deconficting Ports
+
+At time of writing there is no way to de-conflict ports.
+
+> **Caution** If you assign the same port to multiple services neither service may work properly.
```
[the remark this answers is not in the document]
hamishwillee: I'll update. Is there is an issue in place for getting user feedback added?

hamishwillee:
Very useful and I have updated the docs again. The questions around ports showed my ignorance and the confusion in terminology with respect to ports and connectors - i.e. I have been using GPS port when I should be using GPS connector (to I2C and GPS serial port)

> Does this answer everything?

Almost. See:

>> MAV_0_CONFIG is telemetry stream for GCS. What is MAV_1_CONFIG? What if I want another stream?

> There are different mavlink instances that you can configure, so that you can run MAVLink on several UART ports simultanously. You can use any of the instances MAV_x_CONFIG to run on a specific port.
However by convention MAV_0_CONFIG is for telemetry on TELEM1, and MAV_1_CONFIG is for TELEM2 (companion). But this is not fixed/enforced, and a user can freely change that if he/she wants to.

OK, so essentially 
1. MAV_x_CONFIG is a representation of a mavlink channel that you are mapping to a service? 
2. I assume that the channel for a GCS and for a companion might contain different information and that the channel mapping is in source? 
   - Doesn't that imply that whatever the output port is, MAV_0_CONFIG should actually represent the GCS channel?
   - If so, why isn't this documented, and even better this named something like TELEM_GCS_CONFIG?
   - Similarly, why not MAV_1_CONFIG -> TELEM_COMPANION_CONFIG ?
3. MAV_x_CONFIG is only defined for x =0, x=1. What do I use for OSD?

en/sensor/sfxx_lidar.md
```diff
@@ -1,9 +1,11 @@
-# LightWare SF1XX Lidar
+# LightWare SF1X/SF02/LW20 Lidar
```
hamishwillee: @bkueng I've also updated this (lightware) lidar - for your review. Main point of concern is that I assume we still need to set SENS_SF0X_CFG for the lidar so that the system knows which particular device the driver is connected to?

en/sensor/tfmini.md
```diff
@@ -0,0 +1,30 @@
+# Benewake TFmini LiDAR
```
hamishwillee: @bkueng FYI Here I added a new TFmini page from info originally in parent doc. This follows the leddarone doc pattern.

en/peripherals/frsky_telemetry.md
```diff
@@ -34,14 +34,19 @@ Ready-made cables (which include the required adapters) are available from:
...
 
-The typical port used for FrSky telemetry is TELEM2. This can be enabled by [setting](../advanced_config/parameters.md) the [SYS_COMPANION](../advanced_config/parameter_reference.md#SYS_COMPANION) parameter to 10.
+[Configure the serial port](../peripherals/serial_configuration.md) on which FrSky will run using [TEL_FRSKY_CONFIG](../advanced_config/parameter_reference.md#TEL_FRSKY_CONFIG) (typically `TELEM 2` is used for FrSky telemetry). 
+There is no need to set the baud rate for the port, as this is configured by the driver.
 
+> **Note** If `TEL_FRSKY_CONFIG` is not available in *QGroundControl* then you will need to [add the driver to the firmware](#firmware).
+
+No further configuration is required; FrSky telemetry auto-starts when connected and detects D or S mode.
+
+<!--
```
hamishwillee: @bkueng This is update to the FrSky telmetry. 
I have commented this bit (pending deletion) to check - do Pixhawk 3 Pro and Pixracer set these ports up by default still (as it says)? If so we should probably add that info here and to the serial config page.

en/peripherals/frsky_telemetry.md
```diff
@@ -198,6 +203,20 @@ You will need to connect via a UART to S.PORT adapter board, or a [ready-made ca
 <!-- ideally add diagram here -->
 
 
+
+## Firmware Setup {#firmware}
```
hamishwillee: @bkueng FYI you will note that in a number of places I have this section duplicated. I would love to have this information generically, but AFAIK there is no generic way to add all rangefinder drivers, just the serial ones.

Do you think we might be better off being specific and replacing this with a specific note about just uncommenting the driver for the FMUv2 target (everywhere)?

en/advanced_features/satcom_roadblock.md
```diff
@@ -65,29 +65,25 @@ The module is now ready to be used with PX4.
...
-
-	# Sleep a bit to give Mavlink time to set up
-	usleep 100000
-
-else
-	echo "WARN: Iridiumsbd driver not started, reboot"
-	tune_control play -m "ML<<CP4CP4CP4CP4CP4"
-fi
+[Configure the serial port](../peripherals/serial_configuration.md) on which the RockBlock module will run using [ISBD_CONFIG](../advanced_config/parameter_reference.md#ISBD_CONFIG).
+<!-- There is no need to set the baud rate for the port, as this is configured by the driver. -->
```
hamishwillee: @bkueng This is iridium update. Does the driver correctly set the baud rate (as commented) or must it be set by the user (as per next line)

en/advanced_features/satcom_roadblock.md
```diff
@@ -65,29 +65,25 @@ The module is now ready to be used with PX4.
...
+The baud rate for the connected port should be set to: 115200.
+
+> **Note** If `ISBD_CONFIG` is not available in *QGroundControl* then you will need to [add the driver to the firmware](#firmware).
+
+### Firmware Setup {#firmware}
+
+Flash the Pixhawk with the latest PX4 master.
```
hamishwillee: @bkueng When will this go into a release (specifically the serial port changes)?

hamishwillee:
@bkueng The config parameter for https://docs.px4.io/en/sensor/ulanding_radar.html appears to be missing?

hamishwillee:
@bkueng OK, bunch more updates here, and some questions. It is close to done though for this library.

Next we can look at the dev guide. We will need instructions on how you should set up  drivers/driver parameters for a **new** module. I presume that is on your list too :-)

en/advanced_features/satcom_roadblock.md
bkueng: Yes, so that it cannot be misconfigured.

en/advanced_features/satcom_roadblock.md
bkueng: Next release but I guess that does not help you. I don't know more.

en/advanced_features/satcom_roadblock.md
```diff
@@ -65,29 +65,25 @@ The module is now ready to be used with PX4.
...
+
+> **Note** PX4 firmware includes the required driver by default on most [Pixhawk-series](../flight_controller/pixhawk_series.md) boards.
+  This step is only required for the few boards that don't include it in firmware.
+
+You can include the driver in firmware by adding the following line to the [cmake config file](https://github.com/PX4/Firmware/tree/master/cmake/configs) that corresponds to the target you want to build for:
+```
+drivers/iridiumsbd
```
bkueng: This got moved to `drivers/telemetry/iridiumsbd`

en/flight_controller/omnibus_f4_sd.md
```diff
@@ -106,7 +106,6 @@ RC is connected to one of the following ports:
   - TX: MCU pin PA0
   - RX: MCU pin PA1
   - 57600 baud
-  - It can be configured via [SYS_COMPANION](../advanced_config/parameter_reference.md#SYS_COMPANION) parameter.
```
bkueng: You can write that this can be configured as TELEM2 port.

en/peripherals/frsky_telemetry.md
bkueng: For Pixracer, yes. For Pixhawk 3 Pro, TELEM4 is now a configurable UART (so FrSky does not run by default).

en/peripherals/frsky_telemetry.md
```diff
@@ -198,6 +203,20 @@ You will need to connect via a UART to S.PORT adapter board, or a [ready-made ca
...
 
+
+## Firmware Setup {#firmware}
+
+> **Note** PX4 firmware includes the FrSky driver by default on most [Pixhawk-series](../flight_controller/pixhawk_series.md) boards.
+  This step is only required for the few boards that don't include it in firmware.
+
+You can include the driver in firmware by adding the following line to the [cmake config file](https://github.com/PX4/Firmware/tree/master/cmake/configs) that corresponds to the target you want to build for:
+```
+drivers/frsky_telemetry
```
bkueng: Moved to `drivers/telemetry/frsky_telemetry`

en/peripherals/frsky_telemetry.md
bkueng: Hmm, not sure what's best. One one hand we might have more flash-limited boards/targets in future that we want to support, and on the other hand we have the Bootloader update for FMUv2, so that most people are not affected anymore.
So maybe a minimal note here with the driver that needs to be added (`drivers/telemetry/frsky_telemetry`) and a link to a more extensive & general page would do it?

en/sensor/sfxx_lidar.md
bkueng: Yes this is correct, we still need that parameter (unfortunately)

en/advanced_features/satcom_roadblock.md
hamishwillee: Thanks. We'll just update this next time I arrive if there has been a release.

en/peripherals/frsky_telemetry.md
hamishwillee: OK, I've fixed this with a troubleshooting section at the end of the the *Serial Port Configuration* page. The links to this are not very thin in the peripheral pages.

hamishwillee:
@bkueng 

>> Next we can look at the dev guide. We will need instructions on how you should set up drivers/driver parameters for a new module. I presume that is on your list too :-)

> All you need to do is in this PR PX4/Firmware#10591. Should I still write a draft?

Yes please. FYI I am away for next few days, so won't look at it until end of week/next week.

hamishwillee:
@bkueng OK, so re MAVLink I get the big picture - you set an instance to a port, and then use the mode to to say what MAVLink stream that instance actually represents. Still lots of questions though !

1. For the mode we have these options: 0: Normal, Custom, Onboard, OSD, Magic, Config, Iridium,  Minimal
   - What are they for?
     - ie what is "Normal" - should that be "GCS"?
     - Iridium - we already set Iridium via the iridium config param. When/why would we set it again here?
     - What are custom, magic, config for?
     - Is onboard the companion computer?
   - What if I want another stream for my mavlink camera or magic gimbal. How do I define that?
1. The mode doc says " and their sending rates." BUT we also have MAV_0_RATE which is supposed to define max rate. What is the difference?
   - Also on "rate" it says " If this is set to 0, a value of /20 is used, which corresponds to half of the theoretical maximum bandwidth." - how does that work - ie how is /20 = half?

bkueng:
> What are they for? 

To configure which messages are streamed and at which rate. See https://github.com/PX4/Firmware/blob/master/src/modules/mavlink/mavlink_main.cpp#L1739.

> Iridium - we already set Iridium via the iridium config param. When/why would we set it again here?

You would not. It is just there for completness reasons, but could be removed.

> What if I want another stream for my mavlink camera or magic gimbal. How do I define that?

2 options:
- add the stream command to the startup script on the SD card:
  - https://dev.px4.io/en/middleware/modules_communication.html#mavlink e.g. `mavlink stream -d /dev/ttyS2 -s HIGHRES_IMU -r 50`
  - Add that command to the `extras.txt`: https://dev.px4.io/en/concept/system_startup.html#customizing-the-system-startup
- change the source directly (see above)

> The mode doc says " and their sending rates." BUT we also have MAV_0_RATE which is supposed to define max rate. What is the difference? 

MAV_0_RATE is to define the maximum rate for the link in bytes per second. This includes all streams, and the individual streaming rates are reduced if the total sending rate exceeds MAV_0_RATE.

> how does that work - ie how is /20 = half?

Well this goes deep into UARTs now, but you asked for it :) The baudrate is defined in bits/second, and you have 1 or 2 bits overhead per sent byte (baudrate != bitrate). So baudrate / 10 (for 2 bits overhead) == bytes / second.

Is this clear?

en/peripherals/mavlink_peripherals.md
```diff
@@ -1,24 +1,46 @@
...
+In order to assign a particular peripheral to a serial port we use the (abstract) concept of a *MAVLink instance*. 
+Each instance can represent a particular set of streamed messages (see "mode" below); parameters are used to define the set of messages, the port used, data rate, etc.
+
+> **Note** At time of writing three instances are defined, which correspond to the 0, 1, 2 in the parameters listed below.
+
+The parameters for each instance are:
+- [MAV_X_CONFIG](../advanced_config/parameter_reference.md#MAV_0_CONFIG) - Set the serial port (UART) for this instance "X", where X is 0, 1, 2. 
+  It can be any unused port, e.g.: TELEM2, TELEM3, GPS2 etc. For more information see [Serial Port Configuration](../peripherals/serial_configuration.md).
+- [MAV_X_MODE](../advanced_config/parameter_reference.md#MAV_0_MODE) - Specify the telemetry mode/target (the set of messages to stream for the current instance and their rate).
+  The values are:
```
hamishwillee: >> What are they for?

> To configure which messages are streamed and at which rate. See https://github.com/PX4/Firmware/blob/master/src/modules/mavlink/mavlink_main.cpp#L1739.

No, I mean "specifically" - someone using this needs to be able to choose the right option. I have put some guesses in below but some of them aren't helpful. ie Normal and magic are empty, so why bother? Same with "Minimal" and "Config" - can you confirm WHEN you should set these?

hamishwillee:
Hi @bkueng 

Thanks. I have updated the docs a bit more and added some more questions. Regarding 

>> What if I want another stream for my mavlink camera or magic gimbal. How do I define that?
> 2 options:  ...

So essentially out of the box you only have the mavlink modes that are preconfigured. If you want to have a completely new telemetry "thing" you have to create the stream definition in code. Presumably you would then have to somehow create a new option for the mode setting?

FWIW in practise you could probably add your streams in "Custom" mode (maybe that is what it is for)? Alternatively, what is really done for the case of a mavlink camera is to add the necessary messages to another stream and then turn on forwarding for the instance so that whole stream is sent to the camera interface (which then only handles the appropriate messages). 

ANyway, I guess the question I am trying to answer for end users here is how easy it is to add support for another mode. Now I see that this would actually be part of the devguide.

bkueng:
> Presumably you would then have to somehow create a new option for the mode setting?

Yes but this is rarely necessary, as there are already standard messages for camera and gimbals.

> FWIW in practise you could probably add your streams in "Custom" mode (maybe that is what it is for)? 

Exactly. You could start them in the `extras.txt` file on the SD card or let the communication partner request the streams.

> ANyway, I guess the question I am trying to answer for end users here is how easy it is to add support for another mode. Now I see that this would actually be part of the devguide.

That definitely belongs to the dev-guide. If manufacturers/users stick to the standard, they will not have to addd another mode (which is why you don't find any camera mode for example). Iridium would be an example where it's needed, but for that special use-case, the integration goes even deeper (for instance we don't send missions via Iridium).

en/peripherals/mavlink_peripherals.md
```diff
@@ -1,24 +1,46 @@
...
+  - *Custom*: Nothing (in the default PX4 implementation).
+  - *Onboard*: Standard set of messages for a companion computer.
+  - *OSD*: Standard set of messages for an OSD system.
+  - *Magic*: Nothing (in the default PX4 implementation).
+  - *Config*: Set of messages that may be useful for configuration.
+  - *Iridium*: Standard set of messages for an Iridium satellite system. This should not be set (use [ISBD_CONFIG](../advanced_features/satcom_roadblock.md) to configure a satellite system).
+  - *Minimal*: Minimal set of messages.
+- [MAV_X_RATE](../advanced_config/parameter_reference.md#MAV_0_MODE) - Set the maximum data rate for this instance (bytes/second). 
+  This is the combined rate for all streams of individual message (the rates for individual messages are reduced if the total rate exceeds this value).
+- [MAV_X_FORWARD](../advanced_config/parameter_reference.md#MAV_0_FORWARD) - Enable forwarding of MAVLink packets received by the current instance onto other interfaces.
```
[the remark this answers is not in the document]
bkueng: You would enable it for example if you have a mavlink-enabled camera connected, so that a GCS (connected to another port on the pixhawk) can directly talk to the camera.

en/peripherals/mavlink_peripherals.md
```diff
@@ -1,24 +1,46 @@
...
+  The values are:
+  - *Normal*: Standard set of messages for a GCS.
+  - *Custom*: Nothing (in the default PX4 implementation).
+  - *Onboard*: Standard set of messages for a companion computer.
+  - *OSD*: Standard set of messages for an OSD system.
+  - *Magic*: Nothing (in the default PX4 implementation).
+  - *Config*: Set of messages that may be useful for configuration.
+  - *Iridium*: Standard set of messages for an Iridium satellite system. This should not be set (use [ISBD_CONFIG](../advanced_features/satcom_roadblock.md) to configure a satellite system).
+  - *Minimal*: Minimal set of messages.
+- [MAV_X_RATE](../advanced_config/parameter_reference.md#MAV_0_MODE) - Set the maximum data rate for this instance (bytes/second).
```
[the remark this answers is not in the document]
bkueng: Yes. For example telemetry links can often not reach the maximum rate. Especially w/o flow control. What then happens is that the link starts to get congested and lots of messages get dropped. The same might happen with long/bad FTDI cables.
So you can take this as indicator: reduce the rate when you see increased amount of message drops.
In many cases the default should be good however.

en/peripherals/mavlink_peripherals.md
```diff
@@ -1,24 +1,46 @@
...
+  It can be any unused port, e.g.: TELEM2, TELEM3, GPS2 etc. For more information see [Serial Port Configuration](../peripherals/serial_configuration.md).
+- [MAV_X_MODE](../advanced_config/parameter_reference.md#MAV_0_MODE) - Specify the telemetry mode/target (the set of messages to stream for the current instance and their rate).
+  The values are:
+  - *Normal*: Standard set of messages for a GCS.
+  - *Custom*: Nothing (in the default PX4 implementation).
+  - *Onboard*: Standard set of messages for a companion computer.
+  - *OSD*: Standard set of messages for an OSD system.
+  - *Magic*: Nothing (in the default PX4 implementation).
+  - *Config*: Set of messages that may be useful for configuration.
+  - *Iridium*: Standard set of messages for an Iridium satellite system. This should not be set (use [ISBD_CONFIG](../advanced_features/satcom_roadblock.md) to configure a satellite system).
```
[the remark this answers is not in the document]
bkueng: Ok, I'll remove it.

en/peripherals/mavlink_peripherals.md
```diff
@@ -1,24 +1,47 @@
...
--   Normal telemetry mode at 57600 baud
-
-> **Note** OSD mode defines the MAVLink stream required by OSD (set of activated messages & their rates). The other modes are for other specific hardware (e.g. ESP8266 for WiFi), and use the optimal baud rate settings for their link.
-
-![QGC Telemetry Setup](../../images/qgc_telemetry_setup.png)
-
+# MAVLink Peripherals (GCS/OSD/Companion)
+
+Ground Control Stations (GCS), On-Screen Displays (OSD), Companion Computers, and other MAVLink peripherals interact with PX4 using separate MAVLink streams, sent via different serial ports.
+These communication channels are are configured using the [MAVLink parameters](../advanced_config/parameter_reference.md#mavlink).
```
bkueng: typo: `are are`

en/peripherals/mavlink_peripherals.md
```diff
@@ -1,24 +1,47 @@
...
+> **Note** At time of writing three instances are defined, which correspond to the 0, 1, 2 in the parameters listed below.
+
+The parameters for each instance are:
+- [MAV_X_CONFIG](../advanced_config/parameter_reference.md#MAV_0_CONFIG) - Set the serial port (UART) for this instance "X", where X is 0, 1, 2. 
+  It can be any unused port, e.g.: TELEM2, TELEM3, GPS2 etc. For more information see [Serial Port Configuration](../peripherals/serial_configuration.md).
+- [MAV_X_MODE](../advanced_config/parameter_reference.md#MAV_0_MODE) - Specify the telemetry mode/target (the set of messages to stream for the current instance and their rate).
+  The values are:
+  - *Normal*: Standard set of messages for a GCS.
+  - *Custom* or *Magic*: Nothing (in the default PX4 implementation).
+    These might be used for testing when developing a new mode.
```
bkueng: You can extend: or the other mavlink endpoint requests the streams that it wants (with `MAV_CMD_SET_MESSAGE_INTERVAL`)

hamishwillee: Thanks @bkueng - I don't know what this means. I think you are suggesting a sentence, but if so, it makes no sense :-0:

> These might be used for testing when developing a new mode *or the other mavlink endpoint requests the streams that it wants (with `MAV_CMD_SET_MESSAGE_INTERVAL`)*.

Or are you saying, unrelated to the mode information here, that you can request a new rate for a particular message using the `MAV_CMD_SET_MESSAGE_INTERVAL` message?

bkueng: Both is actually true, but in case your endpoint requests all required streams, you can turn off all of the default ones.

> These might be used for testing when developing a new mode or the other mavlink endpoint requests the streams that it wants (with MAV_CMD_SET_MESSAGE_INTERVAL).

This sounds good.

hamishwillee: > This sounds good.

Sorry, no it doesn't. I've read it heaps of times and it doesn't mean anything to me "as a sentence".

hamishwillee: Ignore that. I get it now. We can word it better.

hamishwillee: @bkueng OK, so I "think" I have captured your point in https://github.com/PX4/px4_user_guide/pull/352/commits/1a8466626199d0809761e0c8d960f468163c3b3d . 
- Essentially you are saying that the mode/rate sets default messages and rates but a connected mavlink system can use `MAV_CMD_SET_MESSAGE_INTERVAL` to request what it wants instead (right?). 
- You had put this against "custom/magic". I am guessing because these are empty - so rather than setting desired rates in code you're saying request what you want from the other endpoint.

Given that, I decided to make the option a note at the end of the modes section. I would also be happy with this at the end of the whole section if you thought it better?

I'll merge now and we can do as a post process :-)

bkueng: You got it! :)
This is good as it is.

PR #358, opened 2018-10-09
MC manual stabilized: add MPC_THR_CURVE

bkueng:
From https://github.com/PX4/Firmware/pull/10665.

- MPC_HOLD_DZ is not used for Stabilized, so remove it
- MPC_HOVER_THR is also important for Stabilized, so add it

@hamishwillee is that understandable?

Partially related: MPC_HOVER_THR is generally an important parameter for the landing detector to work correctly. Do we have that documented somewhere?


hamishwillee:
Hi Beat,

Do you mean [MPC_THR_HOVER](https://dev.px4.io/en/advanced/parameter_reference.html#MPC_THR_HOVER)? This refers to `MPC_HOVER_THR` throughout - was there a rename?

I did a minor clarification to make it clearer for myself (ie what the distortion actually means in terms of stick mapping): https://github.com/PX4/px4_user_guide/pull/358/commits/e9d565f5f48b6844cd5caec6875cf848f52e7ef3

1. The mode is discussed in two other places. IMO these do not need to be updated:
* https://docs.px4.io/en/getting_started/flight_modes.html#manual_stabilized_mc 
* https://docs.px4.io/en/flight_modes/#manual_stabilized_mc

> Partially related: MPC_HOVER_THR is generally an important parameter for the landing detector to work correctly. Do we have that documented somewhere?

Yes. Is this sufficient? https://dev.px4.io/en/tutorials/land_detector.html#multicopter-land-detector-configuration

hamishwillee:
>> Yes. Is this sufficient? https://dev.px4.io/en/tutorials/land_detector.html#multicopter-land-detector-configuration

> I was thinking more of something in the user-guide, as part of the setup process.

So no we don't and yes we should. 

Any reason not to move this to [Advanced Configuration](https://docs.px4.io/en/advanced_config/) (user guide)? Might even be worth splitting into FW and MC versions.

bkueng:
> Any reason not to move this to Advanced Configuration (user guide)? Might even be worth splitting into FW and MC versions.

We can do that. I was just wondering if explaining the land detector states was a bit too technical?

hamishwillee:
>> Any reason not to move this to Advanced Configuration (user guide)? Might even be worth splitting into FW and MC versions.

> We can do that. I was just wondering if explaining the land detector states was a bit too technical?

Thats why I suggested "advanced" configuration :-). Seriously though, I would move that section to the end of the the document kind of "supporting information", and ideally add state machine info for FW as well if we can. I can do this move if you want?

hamishwillee:
@bkueng OK, I merged this because it is OK as is. We have open question on guide doc above https://github.com/PX4/px4_user_guide/pull/358#issuecomment-428500481 - ie do you want me to move that doc?

bkueng:
>  Thats why I suggested "advanced" configuration :-). Seriously though, I would move that section to the end of the the document kind of "supporting information", and ideally add state machine info for FW as well if we can. I can do this move if you want?

Ok, convinced :) Yes please go ahead. Thanks!

hamishwillee:
Done!

PR #360, opened 2018-10-12
Land detector configuration 

hamishwillee:
Port from devguide as discussed in #358

### DIFF
--- a/en/SUMMARY.md
+++ b/en/SUMMARY.md
@@ -82,6 +82,7 @@
   * [Parameters](advanced_config/parameters.md)
   * [ESC Calibration](advanced_config/esc_calibration.md)
   * [Bootloader Update (FMUv2)](advanced_config/bootloader_update.md)
+  * [Land Detector Configuration (MC / FW)](advanced_config/land_detector.md)
   * [Multicopter PID Tuning Guide](advanced_config/pid_tuning_guide_multicopter.md)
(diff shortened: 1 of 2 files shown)


hamishwillee:
@bkueng I did some minor updates on porting for structure and content. What else do you think a user might need to know? 
I actually think the state machine for FW might be useful because the MC one is what explains the detailed behaviour of the parameters. BUT imo this could be merged now and we proceed later with any additional updates. Thoughts?

hamishwillee:
@bkueng Thanks. 

> One more thing that comes to my mind: One of the frequent issues with land detector is that if you don't configure MPC_THR_HOVER correctly, you can have in-air ground-contact detected (or even maybe-landed). This happens especially when trying to descend in pos ctrl or alt ctrl. The result is that the vehicle twitches (turns down the motors, and then immediately turns them back up). If you see that you might not immediately connect it to the land detector.

Hmmm. I have addressed this by putting a note in [Basic Flying > Takeoff and Landing](https://docs.px4.io/en/flying/basic_flying.html#takeoff-and-landing). NOt hugely visible, but this is the only place I have a specific section on landing that isn't the mode itself or FW landing path docs. I also added a note in the MPC_THR_HOVER part of land detector. Please review: https://github.com/PX4/px4_user_guide/pull/365

PR #372, opened 2018-10-26
[WIP] Add information about logging configuration

hamishwillee:
Adds information to flight reporting topic about additional configuration.

If is WIP because the information about the mission log is dependent on https://github.com/PX4/Firmware/pull/10738

Note that we might later have a separate topic about mission log, but that would depend on something existing that used it in a "concrete" way - e.g. if we had a topic on geotagging for example.


en/getting_started/flight_reporting.md
```diff
@@ -6,13 +6,34 @@ PX4 logs detailed aircraft performance data once the system has been armed until
...
+
+Logging may further be configured using the [SD Logging](../advanced_config/parameter_reference.md#sd-logging) parameters.
+The parameters you are most likely to change are listed below.
+
+Parameter | Description
+--- | ---
+[SDLOG_MODE](../advanced_config/parameter_reference.md#SDLOG_MODE) | Logging Mode defines when logging starts and stops.<br />- `0`: log when armed until disarm (default)<br />- `1`: log from boot until disarm<br />- `2`: log from boot until shutdown.
+[SDLOG_PROFILE](../advanced_config/parameter_reference.md#SDLOG_PROFILE) | Logging profile. Use this to enable less common logging/analysis (e.g. for EKF2 replay, high rate logging for PID & filter tuning, thermal temperature calibration).
+[SDLOG_MISSION](../advanced_config/parameter_reference.md#SDLOG_MISSION) | Create very small additional "Mission Log". 
+  This log can *not* be used with *Flight Review*, but is useful when you need a small log for geotagging or regulatory compliance.
```
bkueng: I might add some upload protection for flight review to avoid mission logs from being uploaded. On the other hand, it might be useful as you can still see the flown trajectory and metadata.

Antiheavy:
It would be awesome if flight review could parse and display the mission log info. Even more awesome if it could also export the mission log data as .CSV file so folks could manually manipulate things like the camera trigger metadata.

Antiheavy:
Also, since the Mission log is similar file format to the full .ulog, does that mean the QGC geotagger will “just work” with Mission Logs too?

en/getting_started/flight_reporting.md
hamishwillee: OK. I'm going to merge this now though, because it is true "for now".

hamishwillee:
@Antiheavy 

> It would be awesome if flight review could parse and display the mission log info. Even more awesome if it could also export the mission log data as .CSV and KmL files so folks could manually manipulate things like the camera trigger metadata.

Can you post as a feature request: https://github.com/PX4/flight_review

> Also, since the Mission log is similar file format to the full .ulog, does that mean the QGC geotagger will “just work” with Mission Logs too?

I don't know or have capacity to test this. @DonLakeFlyer can you comment about QGC geotagger and the Mission log format?

PR #375, opened 2018-11-05
Collision Prevention docs

hamishwillee:
This is not complete, and depends on https://github.com/PX4/Firmware/pull/10785 being merged.

Note, that it is linked from advanced features. In future will move to "Computer vision" sub page, but only when we have a few more computer vision topics.

@baumanta I've put this in place now so you can see what is missing. Specifically, the PX4 side is very clear but still have no idea what hardware, software and configuration required in the companion - or even what companion we recommend.
We need to provide enough info so that someone else can set this up as a turnkey solution.

### DIFF
--- a/en/computer_vision/collision_avoidance.md
+++ b/en/computer_vision/collision_avoidance.md
@@ -1,48 +1,54 @@
 # Collision Avoidance
 
-<!-- Initial PR: https://github.com/PX4/Firmware/pull/10785 -->
-
 *Collision Avoidance* may be used to automatically slow and stop a vehicle before it can crash into an obstacle.
 
-This feature is intended for manual modes, where the endpoint destination is not known. 
-It is currently only enabled for multicopter vehicles in [Position mode](../flight_modes/position_mc.md).
+It can be enabled for multicopter vehicles in [Position mode](../flight_modes/position_mc.md), and at time of writing requires a companion computer.
 
-> **Tip** Automatic modes instead use *Obstacle Avoidance*, which navigates around obstacles in order to follow a planned path.
-
-
-## How it Works
-
-Collision is enabled by setting [MPC_COL_AVOID](../advanced_config/parameter_reference.md#MPC_COL_AVOID) to 1 (and disabled by setting it to zero).
-
-The system uses information from the MAVLink [OBSTACLE_DISTANCE](https://mavlink.io/en/messages/common.html#OBSTACLE_DISTANCE) message for obstacle detection.
-This message is typically sent from a companion computer, and may be generated from a laser scanner/rangefinder or from an obstacle avoidance/navigation system.
+> **Tip** This feature is intended for manual modes. 
+  Automatic modes instead use *Obstacle Avoidance*, which navigates around obstacles in order to follow a planned path.
 
 
-> **Note** At time of writing this does not support collision avoidance from a rangefinder connected directly to the flight controller.
+## Overview
+
+*Collision Avoidance* is enabled on PX4 using the parameter [MPC_COL_AVOID](../advanced_config/parameter_reference.md#MPC_COL_AVOID).
+The closest allowed distance to an obstacle is set using [MPC_COL_AVOID_D](../advanced_config/parameter_reference.md#MPC_COL_AVOID_D).
+
+The feature requires obstacle information from either an external system (sent using the [OBSTACLE_DISTANCE](https://mavlink.io/en/messages/common.html#OBSTACLE_DISTANCE) message) or a [distance sensor](../sensor/rangefinders.md) connected to the flight controller.
+
+> **Warning** *Collision Avoidance* currently only works with a companion computer! 
+  Support for distance sensors on the flight controller is coming soon.
 
 The vehicle starts braking as soon as it detects an obstacle.
-The velocity setpoint towards the obstacle is reduced linearly such that it is set to zero at the point when the vehicle reaches the minimum allowed distance ([MPC_COL_AVOID_D](../advanced_config/parameter_reference.md#MPC_COL_AVOID_D)).
+The velocity setpoint towards the obstacle is reduced linearly such that it is set to zero at the point when the vehicle reaches the minimum allowed distance.
 If the vehicle approaches any closer (i.e. it overshoots or is pushed) negative thrust is applied to repel it from the obstacle.
 
 Only the velocity components *towards* the obstacle are affected.
 RC inputs that cause the vehicle to move tangentially to the obstacle are still obeyed. 
-So if a vehicle approaches an obstacle from an angle the vehicle will slow until it reaches the minimum distance and then drift along the surface until the forward component can be executed again.
+So if a vehicle approaches an obstacle from an angle, the vehicle will slow until it reaches the minimum distance and then "slide" along the surface until it is no longer blocked.
 
-While the collision detection engaged, the user is notified through *QGroundControl*.
+The user is notified through *QGroundControl* while *collision avoidance* is actively controlling velocity setpoints.
 
 
 ## PX4 (Software) Setup
 
-Use *QGroundControl* to [set the following parameters](../advanced_config/parameters.md):
+Set the following [parameters](../advanced_config/parameters.md) in *QGroundControl*:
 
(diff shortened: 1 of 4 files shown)


en/computer_vision/collision_avoidance.md
```diff
@@ -0,0 +1,54 @@
...
+## PX4 (Software) Setup
+
+Use *QGroundControl* to [set the following parameters](../advanced_config/parameters.md):
+
+* [MPC_COL_AVOID](../advanced_config/parameter_reference.md#MPC_COL_AVOID) - set to 1 to enable collision avoidance.
+* [MPC_COL_AVOID_D](../advanced_config/parameter_reference.md#MPC_COL_AVOID_D) - The minimum distance to allow the vehicle to approach.
+  This should be tuned based on available space and likely speed on the vehicle. 
+
+
+### Companion Setup
```
hamishwillee: So here we need to explain:
1. Requirements of companion
2. What hardware to attach with link to where it can be obtained. If we have instructions for setting up the companion we should link to them.
3. What software is needed to connect each hardware and create the obstacle messages. What rate they should be sent at.

baumanta: First of all: The collision avoidance does not depend on a companion computer. I should work by just attaching a range sensor to the pikhawk. But as it has not been tested that way yet, we would have to figure out how to fit the output of the range sensor properly in the distance sensor message.

Alternatively, if no range sensor is available but a companion system running the obstacle avoidance is installed (highly unlikely except for us developing the stuff), the local_planner can be used to emulate range sensor measurements. So if the collision avoidance is used that way, the companion has to be set up such that it can run the local planner algorithm. This documentation should go in the documentation of the obstacle avoidance.

hamishwillee:
Hi @baumanta 

> The collision avoidance does not depend on a companion computer. I should work by just attaching a range sensor to the pikhawk. But as it has not been tested that way yet, we would have to figure out how to fit the output of the range sensor properly in the distance sensor message.

Using a rangefinder makes perfect sense. We really only covered the external case in previous discussion. Perhaps I should have asked "tell me everything about how this vehicle can be connected and used". 

You're going to to implement and test sensor support before releasing the software?

More questions that will need to be answered:
- Minimum sensor requirements
- minimum range/capabilities for sensor needed
- What if vehicle has FC based sensors and a companion - what will the system do?

I think PX4 only supports one rangefinder on vehicle at a time of a particular type right? ie there is no way to index multiple http://dev.px4.io/en/advanced/parameter_reference.html#SENS_EN_SF1XX say?

baumanta:
hi @hamishwillee,
I'm quite sure that we will test that quite soon, as there is high interest in this feature. But we do not have the capacity to do the sensor case in the next 2 weeks. Therefore, the code will be released before testing that. If someone wants to use it that way he will have to figure out how to read the sensor inputs.

en/computer_vision/collision_avoidance.md
```diff
@@ -0,0 +1,62 @@
...
+* [MPC_COL_AVOID_D](../advanced_config/parameter_reference.md#MPC_COL_AVOID_D) - Set the minimum allowed distance (the closest distance that the vehicle can approach the obstacle).
+  This should be tuned for both the *desired* minimal distance and likely speed of the vehicle.
+
+
+## Companion Setup
+
+The companion computer needs to supply [OBSTACLE_DISTANCE](https://mavlink.io/en/messages/common.html#OBSTACLE_DISTANCE) messages when an obstacle is detected.
+The messages must be sent at a rate of TBD Hz.
+
+The tested hardware/software platform is [Intel Aero](../flight_controller/intel_aero.md) running the *local_planner* avoidance software (setup as per [Intel Aero > Obstacle Avoidance](../flight_controller/intel_aero.md#obstacle-avoidance) and in the [PX4/avoidance](https://github.com/PX4/avoidance#obstacle-detection-and-avoidance) Github repo).
```
hamishwillee: @baumanta Can we recommend an update rate for Obstacle information? I guess it depends on speed, but we must have some recommendations on minimum for a fast quad?

en/computer_vision/collision_avoidance.md
```diff
@@ -0,0 +1,62 @@
+# Collision Avoidance
+
+*Collision Avoidance* may be used to automatically slow and stop a vehicle before it can crash into an obstacle.
+
+It can be enabled for multicopter vehicles in [Position mode](../flight_modes/position_mc.md), and at time of writing requires a companion computer.
```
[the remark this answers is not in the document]
baumanta: So as far as discussions go, we will probably also enable it for further modes in the future. The solution regarding VTOL might be more involved as this feature as is can only be applied to vehicles that can hover.

en/computer_vision/collision_avoidance.md
baumanta: Actually the local_planner just sens out the information as fast as it can. As the camera images come in at 30Hz this will for sure be the upper limit. But I would have to run some tests to determine the actual rate we get on the obstacle_distance topic. Of course this is not tested in all corner cases. So I would recommend putting a warning in the doc to not fly too fast. Actually, the feature has only been tested for flights with a maximum velocity of 4 m/s

baumanta: Oh and the comment about Intel Aero is actually incorrect. We tested it on our IF750A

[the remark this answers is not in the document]
baumanta: 1. the local planner works only with the information it currently gets from the sensors and plans a path that is aroud a meter long. When the next set of information come in it plans a new piece of the path. The global planner, on the other hand, build a map of the environment. It gathers all the information ever received and then plans a path that reaches to the goal. This happens at a much lower frequency than in the local_planner. Also the global planner is much more costly on the computational side.
And yes, the collision avoidance works only with the local planner. The local planner internally build a histogram  where the free and blocked cells are marked. This is then compressed into 1D to serve as a laser scan emulation. The global planner does not use such a histogram.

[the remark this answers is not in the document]
baumanta: 2. In rqt_reconfigure there is a checkbox where you can set the parameter "send_obstacles_fcu" to true. Once this is done, the local_planner will output the emulated range sensor data

hamishwillee:
@baumanta Thanks for all that information. I have given this one more tidy up:
- Added information about test platform
- Added information about tested rates of messages.
- Removed comparison/differentiator note up top about obstacle avoidance - specific things this implementation covers already mentioned.

This links to PX4/avoidance for software and hardware setup info. That is the *right thing to do* as the software/hardware are dependent. Unfortunately that repo isn't well enough documented. I created https://github.com/PX4/avoidance/issues/138 for suggestions on possible improvements. 

I think this is as good as it needs to be when the firmware changes are merged. The only thing I might still do for release would be change the name of the component.

hamishwillee:
@baumanta FYI I've updated this to be "Collision Prevention" as we discussed. 

1. Do you plan on renaming MPC_COL_AVOID and friends to match (e.g. MPC_COL_PREVENT)?
2. Does gazebo simulation support this for the local planner case? 
3. Is there a plan to support this in simulation for the PX4 rangefinder case. I assume it would work out of the box if you can set up virtual rangefinder forward facing.

baumanta:
hi @hamishwillee 
1. In the PR the params are now called MPC_COL_PREV and MPC_COL_PRV_D
2. yes, if you I also tested it in sitl
3. not yet, but I think this will be one of the next steps towards rangefinder integration
4. there is only one message comming from the planner, this is the range data on ROS topic /mavros/obstacle/send. To the planner flows the usual information about state etc.

hamishwillee:
Thanks @baumanta ! I've added https://github.com/PX4/px4_user_guide/pull/375/commits/85fe0d62b00530246d799957ea309a1af7a0a8dd with new param names and links to info about gazebo setup.

We can merge this when you put the PR in.

baumanta:
Nice @hamishwillee! I hope it won't take too long now until we can merge the collision prevention library. I'll keep you in the loop :)

hamishwillee:
Hi @baumanta . FYI, this is now merged following acceptance of https://github.com/PX4/Firmware/pull/10785. I updated the parameter information in line with the changes made on that PR.

You can check out live version here: https://docs.px4.io/en/computer_vision/collision_prevention.html

PR #379, opened 2018-11-15
Flight Review and analysis

hamishwillee:
This continues https://github.com/PX4/Devguide/pull/645

- Adds the new topic: `Log Analysis using Flight Review`
- Moves `Flight Analysis` topic from Devguide.
- All links fixed up (ie back to devguide if needed, and from user guide). 
- Update the Flight reporting to link to this.

I think it is good to go.

After this we need to:
1. Clean up moved topic in Devguide.
2. Consider whether we should remove some of the other content. In particular, the list of good SD cards. What do you think?


bkueng:
Nice, thanks! Looks good. I cherry-picked the 2 commits, as the branch also contains collision-avoidance commits.

hamishwillee:
Thanks very much @bkueng ! Devguide tidied up in https://github.com/PX4/Devguide/pull/653.

PR #388, opened 2018-11-27
Update docs to use new make configs

hamishwillee:
User guide updates for https://github.com/PX4/Firmware/pull/10875


hamishwillee:
I am self-merging. This does reflect associated changes in Firmware.
Note, I took a tag release just before for v1.8.2 so we can create a versioned user guide before this change if we like.

PR #389, opened 2018-11-27
Add note about the need for global position

ndepal:
This is causing some confusion.
The limitation in question is discussed here: https://github.com/PX4/Firmware/pull/7432


hamishwillee:
This is good @ndepal - do you think you will remember to remove it when the issue is fixed? IF not, can we add a link to an issue or PR (in the note) that users can click to find out that it was fixed.

ndepal:
Fair point @hamishwillee, I probably won't. There is the PR that tracks the root cause that I linked at the top. But there is currently no PR/issue specifically for precision landing. I suspect once the position controller is compatible with local positions several other flight modes will also have to be re-written.

PR #393, opened 2018-11-29
LL change

ryanjAA:
Changes the resistor value to the correct one (tested with 1kohm various ways and it does not work. Need to revert to 470 ohm.)


hamishwillee:
Thanks @ryanjAA ! I'll also update the image above, which still shows 1KOhm

hamishwillee:
Done https://github.com/PX4/px4_user_guide/commit/96aac40d11984097daa5b04196de24c0aa38dac2

PR #396, opened 2018-12-04
change a port label on 3d view of pixhawk 4 mini

pfjare:


pfjare:
Hmm. I am not seeing that. From what i can tell it deletes and adds new image

hamishwillee:
@pfjare Thanks very much. FWIW I find the replace image notification in Github a little confusing too.

PR #404, opened 2018-12-11
Add Docs for Companion Computer Peripherals

bryanmonti:
Here are docs that discuss potential hardware add-ons for developers to use with companion computers.

@hamishwillee


en/SUMMARY.md
```diff
@@ -70,6 +70,7 @@
   * [Advanced Controller Orientation](advanced_config/advanced_flight_controller_orientation_leveling.md)
   * [Full Parameter Reference](advanced_config/parameter_reference.md)
 * [Peripheral Hardware](peripherals/README.md)
+  * [Companion Computer Peripherals](peripherals/companion_computer_peripherals.md)
```
hamishwillee: Something very weird is going on - the left side "master" is nothing like the PX4 master branch TOC, so this addition will fail. Can you resolve the conflict (you can do this from the front page of this PR via the Github GUI) basically just use what is in master, and add this new page to the end.

hamishwillee:
@bryanmonti Other than the conflicting summary this looks pretty good to me (though I might tidy a few sentences to my taste, all the information I suggested is there, and in a good structure). You *might* add an image showing the physical connection of some of the parts to remove any ambiguity at all - but that would be up to you.

@bkueng Can you sanity check this page? Are there any other peripherals you'd add?

hamishwillee:
Thanks @bkueng . Any idea when that 4-in-1 FTDI will be "confirmed"?

@bryanmonti Can you please add the FTDI cable from Beat and fix the summary?

Then we are  good to merge.

en/SUMMARY.md
```diff
@@ -179,4 +180,4 @@
 * [QGroundControl Developer Guide](https://dev.qgroundcontrol.com/en/)
 * [MAVLink Guide](https://mavlink.io/en/)
 * [Dronecode SDK](https://sdk.dronecode.org/en/)
-* [Dronecode Camera Manager](https://camera-manager.dronecode.org/en/)
+* [Dronecode Camera Manager](https://came
```
hamishwillee: Yoiks, this is truncated! Can you please fix?

en/SUMMARY.md
```diff
@@ -103,6 +103,7 @@
   * [Advanced Controller Orientation](advanced_config/advanced_flight_controller_orientation_leveling.md)
   * [Full Parameter Reference](advanced_config/parameter_reference.md)
 * [Peripheral Hardware](peripherals/README.md)
+  * [Companion Computer Peripherals](peripherals/companion_computer_peripherals.md)
```
hamishwillee: This is now sane, but can you please move it to the bottom after `[Camera](peripherals/camera.md)` ? 
My logic is that companion computers are not the first peripheral people need. If we end up with a huge list then I'd move to alphabetic, and this would be the right place.

hamishwillee:
Thanks very much @bryanmonti ! Merged :-)

PR #407, opened 2018-12-14
Fix make command for Pixhawk4 and Pixhawk4Mini

bys1123:
I think maybe you typed one more "make".


hamishwillee:
Yes we did. Thanks very much @bys1123 !

PR #409, opened 2018-12-18
Correct the link to the mavlink message reference

AlexisTM:
`MAV_CMD_NAV_LAND` is not a PX4 parameter but a MAVLink message.


hamishwillee:
So it is. Thank you. MErged.

PR #413, opened 2019-01-09
Create autopilot hardware section for complete vehicles

hamishwillee:
As discussed in https://github.com/PX4/px4_user_guide/issues/412#issuecomment-452220732


en/SUMMARY.md
```diff
@@ -158,18 +158,19 @@
...
-    * [Configuration](flight_controller/snapdragon_flight_configuration.md)
-    * [Snapdragon Advanced](flight_controller/snapdragon_flight_advanced.md)
-  * [Intel® Aero Ready to Fly Drone](flight_controller/intel_aero.md)
   * [Raspberry Pi 2/3 Navio2](flight_controller/raspberry_pi_navio2.md)
   * [OcPoC-Zynq Mini](flight_controller/ocpoc_zynq.md)
-  * [Crazyflie 2.0](flight_controller/crazyflie2.md)
-  * [Parrot Bebop](flight_controller/bebop.md)
   * [BeagleBone Blue](flight_controller/beaglebone_blue.md)
+  * [Complete Vehicles](flight_controller/complete_vehicles.md)
+    * [Snapdragon Flight](flight_controller/snapdragon_flight.md)
```
bkueng: Any reason why you put Snapdragon under complete vehicles?

en/flight_controller/complete_vehicles.md
```diff
@@ -0,0 +1,13 @@
+# Complete Vehicles.
+
+This section contains information about complete vehicles that run PX4.
+
+It includes
+- Consumer drones that run PX4 and have explicit support in project code (many other vehicles can run PX4 but do not require custom code/build instructions and/or are supported outside the PX4 project):
```
bkueng: Do you mean `do require`?

hamishwillee: No. I mean this list only includes the ones that have explicit code support in the project. The ones that do not require explicit support are not in the list - basically anything that uses pixhawk say, or is supported by another project. I think this is correct. I will try tidy to reduce confusion though.

en/SUMMARY.md
hamishwillee: Incompetence?

All fixed now. Can you please re-review.

en/flight_controller/complete_vehicles.md
bkueng: Ok I got it now, thanks. But if I read `custom code` (also in the updated version), I associate that with vendor-specific downstream changes, and not board/vehicle-specific code in upstream.
Also, for a complete vehicle you will always need something custom/extra, at the minimum an airframe config with the (tuning, etc.) parameters.

hamishwillee: Hmm. I'm trying to capture that these are supported by the project and need their own build instructions vs things some other vehicles that perhaps use a pixhawk series board (so no "special code to support them") or the special code is outside of our project. 

Would you be happy with:

> Consumer drones running PX4 that are supported and maintained by the project (many other vehicles don't need "explicit" support from the project because they are compatible with the FMU standard used by Pixhawk series boards, or are supported outside the PX4 project):

bkueng: I think it comes down to that: to me the category 'other vehicles that perhaps use a pixhawk series board (so no "special code to support them")' also belongs to the 'Complete Vehicle' list if it's indeed a complete vehicle that you can buy.
But if you want to define it differently it's ok with me as well.

bkueng: I see. Yes "every complete vehicle that can run PX4", was what I was going for. Essentially a list for hackers who want to buy a finished vehicle and change the software (or have the possibility to do so).
But I agree there's some further implications in terms of content, so I leave the decision up to you.

hamishwillee: We don't need to make the decision now. If you or someone else wants to give me a significant list of vehicles that can run PX4 without modification + others that might need special steps and those steps then we can look at this again. I even like the idea, but I don't have time to drive it. 

If we did this, I'd want to split the vehicles into their own top level folder, out of flight controllers. I'm very tempted to do that anyway, but maybe after you have put in whoop.
Seem a good idea.

bkueng: Ok, let's leave the list for now, and we can come back at a later point.

> Seem a good idea?

Agreed, makes them more visible.

PR #427, opened 2019-01-31
MountingFC: Link to Vibration analysis docs

hamishwillee:
This provides better cross linking to information about vibration testing from mounting your controller as discussed with @MariliG 

@bkueng If they use the recommended anti-vibration mechanism and they the verify that vibration is still a problem then we don't explain what person can do next. I assume that at that point there isn't much else to say?

### DIFF
--- a/en/assembly/mount_and_orient_controller.md
+++ b/en/assembly/mount_and_orient_controller.md
@@ -2,7 +2,7 @@
 
 ## Orientation
 
-Almost all Flight Controllers have a *heading mark arrow* (shown below). 
+Almost all Flight Controllers have a *heading mark arrow* (shown below).
 The controller should be placed on the frame top-side up, oriented so that the arrow points towards the front of the vehicle (on all aircraft frames - airplane, multirotor, VTOL, ground vehicles etc.).
 
 ![FC Heading Mark](../../images/fc_heading_mark_1.png)
@@ -13,7 +13,7 @@ The controller should be placed on the frame top-side up, oriented so that the a
 
 ## Vibration Damping
 
-Flight Control boards with in-built accelerometers are sensitive to vibrations. 
+Flight Control boards with in-built accelerometers or gyros are sensitive to vibrations.
 Some boards include in-built vibration-damping, while others come with *vibration damping foam* that you can use to isolate the controller from the vehicle.
 
 ![Pixhawk Mounting foam](../../images/3dr_anti_vibration_mounting_foam.png) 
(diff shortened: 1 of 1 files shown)


en/assembly/mount_and_orient_controller.md
```diff
@@ -1,25 +1,24 @@
...
 
 ![FC Orientation](../../images/fc_orientation_1.png)
 
 > **Note** If the controller cannot be mounted in the recommended/default orientation (e.g. due to physical constraints) you will need to configure the autopilot software with the orientation that you actually used: [Flight Controller Orientation](../config/flight_controller_orientation.md).
 
-## Vibration damping
+## Vibration Damping
 
-Flight Control boards with in-built accelerometers are sensitive to vibrations. Some boards include in-built vibration-damping, while in other cases the boards come with vibration damping foam that you can use to isolate the controller from the vehicle.
+Flight Control boards with in-built accelerometers are sensitive to vibrations.
```
bkueng: This affects gyros too.

hamishwillee:
Thanks very much @bkueng . I've made some minor fixes in line with your feedback and merged this for now.

With respect to vibration damping it would be great to have more guidance. Ideally a detailed page or two, but if not, links to external resources:
- Are there any good resources external we should link to?
- Do we have anyone here expert enough to write these for our systems?
- Minimally, would it be useful to link the vibration isolation text to the wikipedia article: https://en.wikipedia.org/wiki/Vibration_isolation ?

bkueng:
> Do we have anyone here expert enough to write these for our systems?

@bresch @priseborough @RomanBapst

hamishwillee:
> @bresch @priseborough @RomanBapst

Specifically Beat wrote some high level guidance on spotting vibration issues using logs, and included this section on what you can do if you have vibration: https://docs.px4.io/en/log/flight_review.html#solutions
We also suggest "following manufacturer instructions for mounting" in our mounting the flight controller page.

It is a little thin, so I would like to either have more guidance on what you can do, or link to more detail guidance. Can you advise and** can you help?**

If the case is that this is not useful and/or not important, then let me know that too.

RomanBapst:
@hamishwillee @bkueng maybe we can change the formulation from Vibration "damping" to vibration "isolation". Damping reduces the peak at the resonance but it also reduces roll-off, which means frequency above the natural frequency are amplified more. What we are trying to do is isolate, not just dampen.

Maybe for the advanced user we could add suggestions how to measure the natural frequencies of a Vibration isolation system?

hamishwillee:
I've moved discussion to https://github.com/PX4/px4_user_guide/issues/434

PR #431, opened 2019-02-04
Add tune examples and docs

### LINKED ISSUE #426: Add info about tunes
In the guide add information about the tunes - including sound files to play. Probably just below: https://docs.px4.io/en/flying/led_meanings.html

For info see: https://github.com/PX4/px4_user_guide/pull/423#issuecomment-458886972

hamishwillee:
Fixes #426

@davids5 As far as possible we need to clarify when each of these are played, what they mean, and what the user can do.


en/flying/tunes.md
```diff
@@ -0,0 +1,156 @@
+# Tune Meanings (Pixhawk Series)
+
+[Pixhawk-series flight controllers](../flight_controller/pixhawk_series.md) use audible tones/tunes to indicate vehicle state and events (e.g. arming success and failure, low battery warnings).
+
+The current set of tones are listed/can be played below.
+
+Startup tone - played on completion of startup, before arming.
```
hamishwillee: So, when played? What does it mean - basically just a notification that we're now searching for satellites?

en/flying/tunes.md
```diff
@@ -0,0 +1,156 @@
...
+<audio controls>
+  <source src="../../assets/tunes/1_startup_tone.mp3" type="audio/mpeg">
+Your browser does not support the audio element.
+</audio>
+
+Error tune -
```
hamishwillee: When played? What should user do?

en/flying/tunes.md
```diff
@@ -0,0 +1,156 @@
...
+<audio controls>
+  <source src="../../assets/tunes/2_error_tune.mp3" type="audio/mpeg">
+Your browser does not support the audio element.
+</audio>
+
+
+Notify positive tone
+
+<audio controls>
+  <source src="../../assets/tunes/3_notify_positive_tone.mp3" type="audio/mpeg">
```
hamishwillee: Ditto - when generally is this used. What should user do. Same question for everything ....

davids5: Should we break this into modes/phases of boot and operational?

For boot see https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/init.d/rcS

it describes the logic and what tones are played when.

hamishwillee:
Hi @davids5 ,

Thanks. Yes, I have tried to split into operational and boot sequence. Error tune lives in both. 

There are a number of tunes I can't see in use in code - e.g. baro warning, EKF. Would appreciate your confirmation that the ones I haven't go info for aren't used.

You will note I also moved the LED, Tones and Preflight checks from section **Flying** into *Getting Started > Vehicle Status Notifications**. Not completely sure about this:
- Argument for is that not all warnings are about flying - they include things like calibration, and startup, so knowing about them early is useful
- Argument against is that many people will first encounter them when they try and fly and can't. 
-  Would appreciate your and @bkueng thoughts on that? We could keep the section, but have it in flying instead.

davids5:
@hamishwillee 

I am not sure if this is an artifact of rendering using view file but the \<audio controls\> tags are not rendering in chrome "Your browser does not support the audio element."

hamishwillee:
Thanks guys, this is merged.

@bkueng Thanks for your advice - I did something similar (moved arming and takeoff into flying 101 and put the link as part of arming).

> I am not sure if this is an artifact of rendering using view file but the <audio controls> tags are not rendering in chrome "Your browser does not support the audio element."

@davids5 This works in gitbook. Not sure why not in github, but we can survive that.

You should see it all working here: https://docs.px4.io/en/getting_started/tunes.html

davids5:
> @davids5 This works in gitbook. Not sure why not in github, but we can survive that.

Glad it works.

Now that I hear them.......

Please change:

`Hard fault when mounting microSD card or no microSD card`  to `Hard fault has caused a system reboot.`

move: "no microSD card"  under Mounting failed

hamishwillee:
Thanks @davids5 - Fixed in https://github.com/PX4/px4_user_guide/commit/66c404f62c39eff0ecc4f2a6621edb299902578a

junwoo091400:
Hi @hamishwillee, I was interested in how you converted the tunes into the audio files you have added in this PR. Could I learn how? :wink:

hamishwillee:
I played them locally and recorded them on my phone. Yes, I'm a luddite.

PR #436, opened 2019-02-12
Move ecl/ekf doc from devguide

hamishwillee:
@priseborough This moves the ecl/ekf doc from the devguide to the user guide - as previously discussed. If accepted, next step would be to remove from Devguide and create a redirect.

I have put it under Advanced Config as shown. Are you happy with this?

image

PReview here: https://hamishwillee.gitbooks.io/ham_px4userguide/content/v/ekf/en/advanced_config/tuning_the_ecl_ekf.html


hamishwillee:
OK, merged then. I'll delete the version in the devguide.

PR #438, opened 2019-02-17
Updated native build of librobotcontrol for BeagleBone Blue

UAV-Pilot:
To fix an reported here: issue https://github.com/PX4/Firmware/issues/11150


UAV-Pilot:
@jadonk FYI, PX4 autopilot was available on BeagleBone Blue, and it's documented here: https://docs.px4.io/en/flight_controller/beaglebone_blue.html

LorenzMeier:
Thanks!

PR #443, opened 2019-02-20
Add LTE information

### LINKED ISSUE #439: Telemetry Radios - Need to add something on LTE
[Telemetry Radios](https://docs.px4.io/en/telemetry/) refers to Sik radio and Telemetry Wiki, but not LTE (or more generically "telephony module"?).

Can you point me to info (or people) that explain how this works with PX4? Ie 
- Are there any "protocols" for the connection
- What drivers are used
- Any specific models or versions that we know work.

Basically enough for people to be able to find something as "turnkey" AND enough info for them to find a particular module and be able to work out if it is likely to be useful.

Last of all, are there any more types of telemetry we're omitting to mention that we support? 

This was triggered by http://discuss.px4.io/t/4g-lte-pixhawk-integration/9615

hamishwillee:
Fixes #439

### DIFF
--- a/en/peripherals/companion_computer_peripherals.md
+++ b/en/peripherals/companion_computer_peripherals.md
@@ -58,17 +58,22 @@ They are in no way guaranteed to be plug and play with your companion computer.
 
 An LTE USB module can be attached to a companion computer and used to route MAVLink traffic between the flight controller and the Internet.
 
-There is no "standard method" for a ground station and companion to connect over the Internet (i.e. for the GCS to discover the vehicle IP address, and visa versa).
-Depending on your mobile network, it is likely that you won't be able to just directly connect to the remote QGC computer IP over the Internet.
+There is no "standard method" for a ground station and companion to connect over the Internet.
+Generally you can't connect them directly because neither of them will have a public/static IP on the Internet.
 
-A common approach is to set up a virtual private network between the companion and GCS computer (i.e. install a VPN system on both computers).
-The companion then uses [mavlink-router](https://github.com/intel/mavlink-router) to route traffic between the serial interface (flight controller) and the VPN UDP broadcast address/standard QGC listener port (i.e. `x.x.x.255:14550`, where 'x' depends on the VPN system).
+> **Note** Typically your router (or the mobile network) has a public IP address, and your GCS computer/vehicle are on a *local* network.
+  The router uses network address translation (NAT) to map *outgoing* requests from your local network to the Internet, and can use the map to route the *responses* back to requesting system.
+  However NAT has no way to know where to direct the traffic from an arbitrary external system, so there is no way to *initiate* a connection to a GCS or vehicle running in the local network.
 
-> **Note** Using the VPN broadcast address means that you do not need to know the IP address of the QGC computer - all computers in your VPN will get the message
-   (although if you do know the address you can route to that computer).
-   Either way, this method has the benefit that the QGC computer address can be static within the VPN, so the configuration of the *mavlink router* does not need to change over time.
+A common approach is to set up a virtual private network between the companion and GCS computer (i.e. install a VPN system like [zerotier](https://www.zerotier.com/) on both computers).
+The companion then uses [mavlink-router](https://github.com/intel/mavlink-router) to route traffic between the serial interface (flight controller) and GCS computer on the VPN network.
+
+This method has the benefit that the GCS computer address can be static within the VPN, so the configuration of the *mavlink router* does not need to change over time.
+
+> **Note** You can also choose to route to the VPN broadcast address (i.e. `x.x.x.255:14550`, where 'x' depends on the VPN system).
+  This approach means that you do not need to know the IP address of the QGC computer, but may result in more traffic than desired (since packets are broadcast to every computer on the VPN network).
 
 
 Some USB modules that are known to work include:
 - [Huawei E8372](https://consumer.huawei.com/en/mobile-broadband/e8372/) and [Huawei E3372](https://consumer.huawei.com/en/mobile-broadband/e3372/) 
-  > **Tip** The *E8372* includes WiFi which you can use to configure the SIM while it is plugged into the companion (making the development workflow a little easier). The *E3372* lacks WiFi, so you have to configure it by plugging the stick into a laptop.
+  - The *E8372* includes WiFi which you can use to configure the SIM while it is plugged into the companion (making the development workflow a little easier). The *E3372* lacks WiFi, so you have to configure it by plugging the stick into a laptop.
--- a/en/peripherals/companion_computer_peripherals.md
+++ b/en/peripherals/companion_computer_peripherals.md
@@ -2,15 +2,15 @@
 
 This section contains information about peripherals for connecting Pixhawk (PX4) to a companion computer, and peripherals that might be connected to such a companion computer (and potentially triggered/accessed by PX4).
 
-## Communication
 
-Typical companion computer work with Pixhawk requires a companion link to transmit/receive the data between
-the companion computer and Pixhawk hardware (eg. Intel NUC and Pixhawk 4).
+## Companion/Pixhawk Communication
+
+Typical companion computer work with Pixhawk requires a companion link to transmit/receive the data between the companion computer and Pixhawk hardware (e.g. Intel NUC and Pixhawk 4).
 
 There are a few devices that allow this communication bridge such as FTDI USB breakouts and level shifters (see below).
 
 > **Note** PX4 configuration for communicating with a companion computer over MAVLink configuration is covered in [MAVLink \(OSD / Telemetry\)](../peripherals/mavlink_peripherals.md#TELEM2).
-  Relevent topics/sections in the developer guide include: [Companion Computer for Pixhawk Series](http://dev.px4.io/en/companion_computer/pixhawk_companion.html), [Robotics](http://dev.px4.io/en/robotics/) and [RTPS/ROS2 Interface: PX4-FastRTPS Bridge](http://dev.px4.io/en/middleware/micrortps.html).
+  Relevant topics/sections in the developer guide include: [Companion Computer for Pixhawk Series](http://dev.px4.io/en/companion_computer/pixhawk_companion.html), [Robotics](http://dev.px4.io/en/robotics/) and [RTPS/ROS2 Interface: PX4-FastRTPS Bridge](http://dev.px4.io/en/middleware/micrortps.html).
 
 ### FTDI Devices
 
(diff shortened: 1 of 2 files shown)


en/peripherals/companion_computer_peripherals.md
```diff
@@ -44,10 +44,23 @@ Cameras are among the most common form of data collection on a drone whether it
...
 * [Intel® RealSense™ Depth Camera D415](https://click.intel.com/intelr-realsensetm-depth-camera-d415.html)
 * [DUO MLX ](https://duo3d.com/product/duo-minilx-lv1)
 * [Ironsides](https://www.perceptin.io/products)
+
+
+## Data Telephony (LTE) {#data_telephony}
+
+An LTE USB module can be attached to a companion computer and used to route MAVLink traffic between the flight controller and the Internet.
+
+There is no standard approach for a ground station and companion to connect (i.e. for the GCS to discover the vehicle IP address, and visa versa).
```
hamishwillee: @bkueng I know this section is "weak" because I don't fully understand the approach discussed in https://github.com/PX4/px4_user_guide/issues/439#issuecomment-465037494

As I understand it there are two ways you can connect to a ground station:
1. UDP - QGC listens for UDP traffic on its IP address at port 14550. 
   - So if the vehicle knows the ground station IP address it can output UDP to that address and port. Is that correct?
   - Presumably you would have to change the code in PX4 to specify this address?
2. TCP- the GCS can connect via TCP to a known IP address/port. However AFAIK PX4 doesn't support TCPIP for this connection - so this wouldn't work?

And you talked about VPN. So as I understand you are suggesting you would set up VPN on the vehicle and connect to some VPN server "somewhere". Then presumably you still have to do "1" above to connect to the GCS?

Sorry if I'm being dumb.

bkueng: >     * So if the vehicle knows the ground station IP address it can output UDP to that address and port. Is that correct?

Correct.

>     * Presumably you would have to change the code in PX4 to specify this address? Or could you use MAVLink router?

Yes you have to use some kind of routing on the companion, mavlink-router is the canonical approach.

>     * Doesn't this imply you have to know the address of the GCS or service at configuration time?

Correct, for that approach. This is another reason why a VPN (with fixed IP's) helps you. If you use a VPN you can also use broadcasting, then the GCS IP does not need to be known.

>     2\. TCP- the GCS can connect via TCP to a known IP address/port. However AFAIK PX4 doesn't support TCPIP for this connection - so this wouldn't work?

It does (also mavlink-router), but it won't auto-connect AFAIK. So we generally use UDP.

> And you talked about VPN. So as I understand you are suggesting you would set up VPN on the vehicle and connect to some VPN server "somewhere". Then presumably you still have to do "1" above to connect to the GCS?

There does not need to be a VPN server, depending on the VPN. But for sure you will have to setup the VPN on the vehicle and the GCS machine. The VPN does 2 things:
- provide a fixed IP addresses for the GCS
- provide visibility: allows traffic to be sent directly to a GCS. If you want to route directly through the internet instead, and even if you know/have a fixed IP for the GCS, you cannot simply send MAVLink data to it, because of [NAT](https://en.wikipedia.org/wiki/Network_address_translation). If you want to do that you have to use a server in between (or some more advanced networking which we don't support).

The above approaches are for a direct vehicle-to-GCS connection, but it's possible to route MAVLink though a server if you want to (using mavlink-router on the server as well).

hamishwillee: Thanks @bkueng 

> You are touching a bigger topic here and I'm not sure how we should limit the scope to avoid having to document every possible option in detail.
Though I think it makes sense to document one or two options in more detail.

Indeed! More questions of course. 

I know only high level "consumer" stuff about VPNs - ie you set up a VPN client on your computer, it takes over your network connection and routes all your IP traffic encrypted to their server which then handles the request. To you it just looks like a normal internet connection. 
Given that, you'll understand my "how" questions below.

> There does not need to be a VPN server, depending on the VPN. But for sure you will have to setup the VPN on the vehicle and the GCS machine. The VPN does 2 things:
- provide a fixed IP addresses for the GCS
- provide visibility: allows traffic to be sent directly to a GCS. 

On the companion end, 
- I assume that the VPN automatically just routes all traffic - ie you don't have to do anything in mavlink router to get traffic to go via it?
- What I think you are saying is that:
  -if you do a UDP broadcast then this will be sent via the VPN tunnel and broadcast on the network running the GCS computer. That means every computer on the network including the GCS will get the message on port you set (presumably 14550 or whatever the default GCS port is). 
    - Is that right? 
    - Is it a problem that every computer on that same remote network also gets the UDP traffic, or is the "network" here restricted to just the remote computer running VPN/GCS?
  - You might otherwise send UDP to specific IP address of remote computer, but then intel router would need to know it in advance - correct? 
    - YOu say `provide fixed IP address for the GCS` - not quite sure how VPN helps here?

> If you want to route directly through the Internet instead, and even if you know/have a fixed IP for the GCS, you cannot simply send MAVLink data to it, because of [NAT](https://en.wikipedia.org/wiki/Network_address_translation). 

How does NAT hurt you? 

Anyway, once I see how VPN helps in big picture, would be good to know:
1. How to config intel router to forward traffic to UDP broadcast via VPN
2. What VPN solutions we have tried
3. What their setup looks like - ie what you need to know to connect the two ends.

hamishwillee: PS Perhaps we should also add this option to http://dev.px4.io/en/ros/offboard_control.html - removing the LTE part of course and just calling the channel "Internet".

bkueng: >     * I assume that the VPN automatically just routes all traffic - ie you don't have to do anything in mavlink router to get traffic to go via it?

Correct. You only need to configure to use the VPN IP of the GCS.

>     * Is that right?

No, it's broadcast within the VPN network. You can look at the VPN as a virtual local network. It bypasses the 'real' local network of a GCS. However a broadcast within the VPN will reach all clients connected to the VPN.

>     * or is the "network" here restricted to just the remote computer running VPN/GCS?

Exactly.

>     * You might otherwise send UDP to specific IP address of remote computer, but then intel router would need to know it in advance - correct?

Correct.

>     * YOu say `provide fixed IP address for the GCS` - not quite sure how VPN helps here?

Because your IP within the VPN usually stays fixed. It has to be fixed, because you use it in the mavlink router config.

>     1. How to config intel router to forward traffic to UDP broadcast via VPN

Just set it to the broadcast IP of the VPN. It's usually x.x.x.255.

> PS Perhaps we should also add this option to http://dev.px4.io/en/ros/offboard_control.html - removing the LTE part of course and just calling the channel "Internet".

I don't think so. You typically want to run ROS via local network, because you usually have large amounts of data that needs to sent in real-time (low-latency).

en/peripherals/companion_computer_peripherals.md
```diff
@@ -44,10 +44,31 @@ Cameras are among the most common form of data collection on a drone whether it
...
 * [DUO MLX ](https://duo3d.com/product/duo-minilx-lv1)
 * [Ironsides](https://www.perceptin.io/products)
+
+
+## Data Telephony (LTE) {#data_telephony}
+
+An LTE USB module can be attached to a companion computer and used to route MAVLink traffic between the flight controller and the Internet.
+
+There is no "standard method" for a ground station and companion to connect over the Internet (i.e. for the GCS to discover the vehicle IP address, and visa versa).
+Depending on your mobile network, it is likely that you won't be able to just directly connect to the remote QGC computer IP over the Internet.
```
hamishwillee: You said this was due to NAT, but can you explain why? Looking on various forums it looks like mobile networks don't expose a public IP or a fixed/static IP, which I take to mean that you couldn't initiate connection TO the drone from QGC, but I don't see why (if you know the QGC IP) you couldn't connect to it from the vehicle?

Or is this just a matter that you'd have to change the configuration of your intel router if the QGC computer ever moved IP? Not sure that is relevant because again, your VPN is going to have to know the QGC computer IP presumably ...?

en/peripherals/companion_computer_peripherals.md
```diff
@@ -44,10 +44,31 @@ Cameras are among the most common form of data collection on a drone whether it
...
+
+A common approach is to set up a virtual private network between the companion and GCS computer (i.e. install a VPN system on both computers).
```
hamishwillee: -Are these two new paragraphs correct/OK?
- Is there a vpn system/client we recommend or have used that I could use here as an example?

hamishwillee:
Thanks @bkueng - I've updated following your feedback. A couple more questions to confirm this is OK and possibly provide an example. If this is correct I think it is about the right balance that people can see in theory how it works and know how to start going about getting the bits and pieces. 

It might be worth its own topic/tutorial?

en/peripherals/companion_computer_peripherals.md
bkueng: I would not suggest the broadcasting by default, as it can create more traffic than required (if more machines are in the VPN). We don't use broadcasting for example.

We use zerotier as VPN.

en/peripherals/companion_computer_peripherals.md
bkueng: >  Looking on various forums it looks like mobile networks don't expose a public IP or a fixed/static IP, which I take to mean that you couldn't initiate connection TO the drone from QGC

Exactly

> but I don't see why (if you know the QGC IP) you couldn't connect to it from the vehicle?

Ok, let's dive into networking :) The standard setup of a network (for IPv4) is: you have a local network and within that your GCS machine has a local IP. Between the local network and the internet, there is NAT, which is handled by the router - the router has only one public IP facing to the internet. So there are 2 cases for incoming traffic:
- the target is the public IP of the network, which means the router does the adress translation from public to local IP. Which requires a mapping that is initiated by outgoing traffic. But if you send traffic from the vehicle to the public IP, there is no mapping, so the router will discard the data.
- you might think the vehicle could directly use the local IP of the GCS, but this is not possible - local IP's are reserved for that purpose and not routed through the internet.

Is that clear?

A VPN hides all of this so you don't have to worry.

bkueng:
> If this is correct I think it is about the right balance that people can see in theory how it works and know how to start going about getting the bits and pieces.

Yes I think so for now. We can probably add another in-depth tutorial of what we do later on.

en/peripherals/companion_computer_peripherals.md
hamishwillee: @bkueng . This stuff comes under the set of things "I used to know". Got it again now - thanks very much.

en/peripherals/companion_computer_peripherals.md
hamishwillee: Thanks!

hamishwillee:
@bkueng Thanks for your help and patience. This is now ready for your re-review, and I am (finally) happy with it.

en/peripherals/companion_computer_peripherals.md
```diff
@@ -44,10 +44,36 @@ Cameras are among the most common form of data collection on a drone whether it
...
+  However NAT has no way to know where to direct the traffic from an arbitrary external system, so there is no way to *initiate* a connection to a GCS or vehicle running in the local network.
+
+A common approach is to set up a virtual private network between the companion and GCS computer (i.e. install a VPN system like [zerotier](https://www.zerotier.com/) on both computers).
+The companion then uses [mavlink-router](https://github.com/intel/mavlink-router) to route traffic between the serial interface (flight controller) and GCS computer on the VPN network.
+
+This method has the benefit that the GCS computer address can be static within the VPN, so the configuration of the *mavlink router* does not need to change over time.
+
+> **Note** You can also choose to route to the VPN broadcast address (i.e. `x.x.x.255:14550`, where 'x' depends on the VPN system).
+  This approach means that you do not need to know the IP address of the QGC computer, but may result in more traffic than desired (since packets are broadcast to every computer on the VPN network).
```
bkueng: Should we use `GCS computer` for consistency?

en/peripherals/companion_computer_peripherals.md
```diff
@@ -44,10 +44,36 @@ Cameras are among the most common form of data collection on a drone whether it
...
+Generally you can't connect them directly because neither of them will have a public/static IP on the Internet.
+
+> **Note** Typically your router (or the mobile network) has a public IP address, and your GCS computer/vehicle are on a *local* network.
+  The router uses network address translation (NAT) to map *outgoing* requests from your local network to the Internet, and can use the map to route the *responses* back to requesting system.
+  However NAT has no way to know where to direct the traffic from an arbitrary external system, so there is no way to *initiate* a connection to a GCS or vehicle running in the local network.
+
+A common approach is to set up a virtual private network between the companion and GCS computer (i.e. install a VPN system like [zerotier](https://www.zerotier.com/) on both computers).
+The companion then uses [mavlink-router](https://github.com/intel/mavlink-router) to route traffic between the serial interface (flight controller) and GCS computer on the VPN network.
+
+This method has the benefit that the GCS computer address can be static within the VPN, so the configuration of the *mavlink router* does not need to change over time.
```
bkueng: Since we are talking about sending data over the internet here, I find it important to mention security as well. I.e. MAVLink does not use encryption, but the VPN gives you a secure link.

hamishwillee: OK, added

hamishwillee:
Thanks @bkueng . I made those last few changes and will now merge.

PR #445, opened 2019-02-21
flight_log_analysis: replace outdated FlightPlot link

MaEtUgR:
Even though DrTon is the original creator behind the beloved FlightPlot tool PX4 uses AFAIK the fork https://github.com/PX4/FlightPlot which is linked under flight plot. I think it's best to link to the topic inside the guide about FlightPlot instead of to a different repository which could generate confusion.

On the other hand we could add credit to the original creator and repository under the FlightPlot chapter...

### DIFF
--- a/en/log/flight_log_analysis.md
+++ b/en/log/flight_log_analysis.md
@@ -99,7 +99,7 @@ Name (multi id, message size in bytes)    number of data points, total bytes
 
 ### pyFlightAnalysis
 
-[pyFlightAnalysis](https://github.com/Marxlp/pyFlightAnalysis) is a cross-platform PX4 flight log (ULog) visual analysis tool, inspired by [FlightPlot](https://github.com/DrTon/FlightPlot).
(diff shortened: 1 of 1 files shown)


en/log/flight_log_analysis.md
```diff
@@ -110,7 +110,7 @@ Key features:
 ![pyFlightAnalysis 1.0.1b1](../../assets/flight_log_analysis/pyflightanalysis.png)
 
 
-### FlightPlot (Desktop)
+### FlightPlot {#flightplot}
```
hamishwillee: FYI, did you mean to get rid of (Desktop) in the heading? Note that the good thing about the anchor syntax `{#flightplot}` is that it no longer matters what the preceding text in the heading is - the anchor stays the same.

Anyway, I don't think this breaks anything. Merging.

MaEtUgR: Yes I deliberately got rid of `(Desktop)` because it looks ugly and most other tools listed are of the same type "for desktop computers" which was probably the meaning.

PR #446, opened 2019-02-21
Add PlotJuggler to list of logs

facontidavide:
... and save few Kbytes of bandwidth too


hamishwillee:
@facontidavide Thanks very much. Looks good to me - merged.

Just FYI, thanks for reducing the images
- generally for new images I try get them under 100Kb - depends on the content, but using jpgs is usually better than PNG, and often you don't need great detail, so a screenshot can be reduced to 900 pixels (or whatever).
- I'm wary of compressing _existing_ images unless there is a significant improvement. The new image ends up being downloaded for every person who syncs the repo and this has to be balanced against reduction in page load times for users. I'm sure you already considered this, because I would probably have made the same changes you did.


 - generally I'm aiming for under 100kb.

mhkabir:
@hamishwilee https://github.com/marketplace/imgbot

PR #450, opened 2019-02-27
Typo 60A vs 60V?

jacobschloss:
It sounds from context like this should be 60V.


hamishwillee:
Sounds reasonable. Thank you

PR #451, opened 2019-02-28
BootloaderUpdate: Make generic

hamishwillee:
This makes the bootloader doc "generic", adding the option for using dronecode probe to it. 

The info on FMUv2 update is now directly included in the page on Firmware update. There is some duplication in docs from this, but I am OK with that.

### DIFF
--- a/en/SUMMARY.md
+++ b/en/SUMMARY.md
@@ -101,7 +101,7 @@
     * [VTOL Weather Vane](config_vtol/vtol_weathervane.md)
   * [ESC Calibration](advanced_config/esc_calibration.md)
   * [ECL/EKF Overview & Tuning](advanced_config/tuning_the_ecl_ekf.md)
-  * [Bootloader Update (FMUv2)](advanced_config/bootloader_update.md)
+  * [Bootloader Update](advanced_config/bootloader_update.md)
   * [Land Detector Configuration](advanced_config/land_detector.md)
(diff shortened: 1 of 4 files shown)


bkueng:
@hamishwillee we don't have `SYS_BL_UPDATE` generically available - only for v2. It's easy to do but we did not have a reason for an update so far on other boards.
The Pixhawk 4 is a special case for us as well, because we flash a BL that supports updating the Firmware through Telem2. Which we'll probably want to have upstream as well at some point.
Other than that the changes look good.

hamishwillee:
Thanks @bkueng . Bugger.

So to clarify `SYS_BL_UPDATE` is only generically in FMUv2, and in our custom build for Pixhawk 4?

How about I simply add a note that firmware that supports upload will have this parameter? Then if it appears in Pixhawk 4 or any other firmware people can use this. In the meantime it does not particular harm?

bkueng:
> So to clarify `SYS_BL_UPDATE` is only generically in FMUv2, and in our custom build for Pixhawk 4?

For now yes.

> How about I simply add a note that firmware that supports upload will have this parameter? Then if it appears in Pixhawk 4 or any other firmware people can use this. In the meantime it does not particular harm?

Exactly, this is the easiest way to avoid having to update docs.

hamishwillee:
Thanks @bkueng ! I've added the note and merged.

PR #457, opened 2019-03-19
Clarify rangefinder usage options

hamishwillee:
This PR adds all the options for using rangefinder on PX4. I.e. specifically it adds `MPC_ALT_MODE` which was not previously mentioned.

The idea is that it should make it easier for people to find relevant docs if looking for information about terrain following while searching.

### DIFF
--- a/en/flying/terrain_following_holding.md
+++ b/en/flying/terrain_following_holding.md
@@ -1,40 +1,81 @@
-# Terrain Following/Holding
+# Terrain Following/Holding & Range Aid
 
-PX4 supports *Terrain Following* and *Terrain Hold* on **multicopters** that have a [distance sensor](../sensor/rangefinders.md).
+PX4 supports [Terrain Following](#terrain_following) and [Terrain Hold](#terrain_hold) in [Position](../flight_modes/position_mc.md) and [Altitude modes](../flight_modes/altitude_mc.md), on *multicopters* and *VTOL vehicles in MC mode* that have a [distance sensor](../sensor/rangefinders.md).
 
-The features are available in manual altitude-control modes: [Position](../flight_modes/position_mc.md), [Altitude](../flight_modes/altitude_mc.md).
-They are not available in acrobatic modes or automatic modes (e.g. Return, Land, Mission, etc.).
+PX4 also supports [using a distance sensor as the primary source of altitude data](#distance_sensor_primary_altitude_source) in any mode, either all the time, or just when flying at low altitudes ([Range Aid](#range_aid)).
 
-## Terrain Following
 
-Terrain-following allows a low-flying vehicle to automatically maintain a (relatively) constant altitude above ground level.
-It is useful for avoiding obstacles when traveling over varied terrain, and for maintaining constant height for aerial photography.
+## Terrain Following {#terrain_following}
 
-In this mode:
-- At low altitude the vehicle moves up and down with terrain height variation (using the distance sensor for altitude data while within its effective range).
-- At higher altitudes the vehicle will typically fly at a near-constant height above mean sea level (AMSL) using the barometer for altitude data.
-  > **Note** More precisely, the vehicle will use the *primary source of altitude data* as defined in [EKF2_HGT_MODE](../advanced_config/parameter_reference.md#EKF2_HGT_MODE).
+When *terrain following* is enabled, a vehicle will automatically maintain a relatively constant height above *ground level* if traveling at low altitudes.
+This is useful for avoiding obstacles and for maintaining constant height (e.g. for aerial photography) when flying over varied terrain.
+
+*Terrain following* uses the output of the EKF estimator to provide the altitude estimate, and the estimated terrain altitude (calculated from distance sensor measurements using another estimator) to provide the altitude setpoint.
+As the distance to ground changes, the altitude setpoint adjusts to keep the height above ground constant.
+
+At higher altitudes (when the estimator reports that the distance sensor data is invalid) the vehicle switches to *altitude following*, and will typically fly at a near-constant height above mean sea level (AMSL) using the barometer for altitude data.
+
+> **Note** More precisely, the vehicle will use the *primary source of altitude data* as defined in [EKF2_HGT_MODE](../advanced_config/parameter_reference.md#EKF2_HGT_MODE).
     This is, by default, the barometer.
 
-## Terrain Hold
-
-Terrain-holding helps a vehicle to better hold/track altitude in [Position](../flight_modes/position_mc.md) or [Altitude](../flight_modes/altitude_mc.md) modes when horizontally stationary at low altitude.
-This allows a vehicle to avoid altitude changes due to barometer drift or excessive baro interference from rotor wash.
-
-In this mode:
-- When stationary (horizontally) the vehicle uses the distance sensor for altitude data (while within its effective range).
-- When moving (`speed > MPC_HOLD_MAX_XY`) or at higher altitudes the vehicle gets altitude data from the *primary source of altitude data* ( [EKF2_HGT_MODE](../advanced_config/parameter_reference.md#EKF2_HGT_MODE)).
+Terrain following is enabled by setting [MPC_ALT_MODE](../advanced_config/parameter_reference.md#MPC_ALT_MODE) to `1`.
 
 
-> **Tip** *Range Aid* ([EKF2_RNG_AID=1](../advanced_config/parameter_reference.md#EKF2_RNG_AID)) has similar behaviour to terrain hold, but operates in all flight modes.
+## Terrain Hold {#terrain_hold}
+
+Terrain holding uses a distance sensor to help a vehicle to better maintain a constant height above ground in [position](../flight_modes/position_mc.md) or [altitude](../flight_modes/altitude_mc.md) modes, when horizontally stationary at low altitude.
+This allows a vehicle to avoid altitude changes due to barometer drift or excessive barometer interference from rotor wash.
+
+> **Note** Terrain hold is enabled only in *multicopters* and *VTOL vehicles in MC mode*, and can be used in [position](../flight_modes/position_mc.md) and [altitude modes](../flight_modes/altitude_mc.md).
+
+When moving horizontally (`speed >` [MPC_HOLD_MAX_XY](../advanced_config/parameter_reference.md#MPC_HOLD_MAX_XY), or above the altitude where the distance sensor is providing valid data, the vehicle will instead use the *primary source of altitude data* ([EKF2_HGT_MODE](../advanced_config/parameter_reference.md#EKF2_HGT_MODE)).
+Typically this is the barometer.
+
+Terrain holding is enabled by setting [MPC_ALT_MODE](../advanced_config/parameter_reference.md#MPC_ALT_MODE) to `2`.
+
+> **Note** *Terrain hold* is implemented similarly to *terrain following.
+  It uses the output of the EKF estimator to provide the altitude estimate, and the estimated terrain altitude (calculated from distance sensor measurements using another estimator) to provide the altitude setpoint.
+  As the distance to ground changes due to external forces, the altitude setpoint adjusts to keep the height above ground constant.
 
 
-## Setup
+## Distance Sensor as Primary Source of Height {#distance_sensor_primary_altitude_source}
 
-Both terrain following and hold features require a [distance sensor](../sensor/rangefinders.md).
-Set up and activation/enabling of the sensor depends on the specific device (see docs for each sensor).
-The generic configuration that is common to all rangefinders (e.g. setting the position offset relative to the frame) is covered in [Distance Sensor > Configuration/Setup](../sensor/rangefinders.md#configuration).
+PX4 allows you to make a distance sensor the *primary source of altitude data*.
+This may be useful when no barometer is available, or for applications when the vehicle is *guaranteed* to only fly over a near-flat surface (e.g. indoors).
 
-Terrain following and holding are then enabled using [MPC_ALT_MODE](../advanced_config/parameter_reference.md#MPC_ALT_MODE):
-- `MPC_ALT_MODE=1` - Terrain Following enabled
-- `MPC_ALT_MODE=2` - Terrain Holding enabled
+> **Tip** The default and preferred altitude sensor for most use cases is the barometer (when available).
+
+When using a distance sensor as the primary source of height, fliers should be aware:
+- Flying over obstacles can lead to the estimator rejecting data (due to internal data consistency checks).
+  This will result in TBD.
+  > **Note** A scenario where this can occur is when the vehicle is ascending a slope at a near-constant height above ground;
+    The measured/estimated altitude does not change, but consistency checks show an error against the barometer measurements.
+- The local NED origin will move up and down with ground level.
+- Rangefinder performance over uneven surfaces (e.g. trees) can be very poor, resulting in noisy and inconsistent data.
+
+The feature is enabled by setting: [EKF2_HGT_MODE=2](../advanced_config/parameter_reference.md#EKF2_HGT_MODE).
+
+
+## Range Aid {#range_aid}
+
+*Range Aid* uses a distance sensor as the primary source of height estimation during low speed/low altitude operation, but will otherwise use the primary source of altitude data defined in [EKF2_HGT_MODE](../advanced_config/parameter_reference.md#EKF2_HGT_MODE) - typically the barometer.
+It is primarily intended for *takeoff and landing*, in cases where the barometer setup is such that interference from rotor wash is excessive and can corrupt EKF state estimates.
+
+Range aid may also be used to improve altitude hold when the vehicle is stationary.
+
+> **Tip** [Terrain Hold](#terrain_hold) is recommended over *Range Aid* for terrain holding.
+  This is because terrain hold uses the normal ECL/EKF estimator for determining height, and this is generally more reliable than a distance sensor in most conditions.
+
+*Range Aid* is enabled by setting [EKF2_RNG_AID=1](../advanced_config/parameter_reference.md#EKF2_RNG_AID) (when the primary source of altitude data ([EKF2_HGT_MODE](../advanced_config/parameter_reference.md#EKF2_HGT_MODE)) is *not* the rangefinder).
+
+Range aid is further configured using the `EKF2_RNG_A_` parameters:
+- [EKF2_RNG_A_VMAX](../advanced_config/parameter_reference.md#EKF2_RNG_A_VMAX): Maximum horizonatal speed, above which range aid is disabled.
+- [EKF2_RNG_A_HMAX](../advanced_config/parameter_reference.md#EKF2_RNG_A_HMAX): Maximum height, above which range aid is disabled.
+- [EKF2_RNG_A_IGATE](../advanced_config/parameter_reference.md#EKF2_RNG_A_IGATE): Range aid consistency checks "gate" (a measure of the error before range aid is disabled).
+
+
+   <!-- 
+If vehicle motion causes repeated switching between the primary height sensor and range finder, an offset in the local position origin can accumulate. 
+Also range finder measurements are less reliable and can experience unexpected errors.
+
+-->
--- a/en/flying/terrain_following_holding.md
+++ b/en/flying/terrain_following_holding.md
@@ -1,0 +1,40 @@
+# Terrain Following/Holding
+
+PX4 supports *Terrain Following* and *Terrain Hold* on **multicopters** that have a [distance sensor](../sensor/rangefinders.md).
+
+The features are available in manual altitude-control modes: [Position](../flight_modes/position_mc.md), [Altitude](../flight_modes/altitude_mc.md).
+They are not available in acrobatic modes or automatic modes (e.g. Return, Land, Mission, etc.).
+
+## Terrain Following
+
+Terrain-following allows a low-flying vehicle to automatically maintain a (relatively) constant altitude above ground level.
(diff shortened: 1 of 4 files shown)


en/sensor/rangefinders.md
```diff
@@ -68,31 +68,43 @@ PX4 also supports the Bebop rangefinder.
...
+   - Specifically, the rangefinder is enabled when:
+     - velocity < [EKF2_RNG_A_VMAX](../advanced_config/parameter_reference.md#EKF2_RNG_A_VMAX)
+     - distance to ground < [EKF2_RNG_A_HMAX](../advanced_config/parameter_reference.md#EKF2_RNG_A_HMAX)
+   - Other parameters affecting "Range aid" are prefixed with `EKF2_RNG_A_`.
+
+1. Enable rangefinder-supported *terrain following* by setting [MPC_ALT_MODE=1](../advanced_config/parameter_reference.md#MPC_ALT_MODE).
+   - In this mode the vehicle altitude moves up and down with terrain height variation.
+   - At altitudes above the distance sensor effective range the vehicle reverts to using height above the home position (as provided by the primary altitude source defined in `EKF2_HGT_MODE`).
+
+1. Enable rangefinder-supported *terrain holding* by setting [MPC_ALT_MODE=2](../advanced_config/parameter_reference.md#MPC_ALT_MODE).
```
hamishwillee: Hi @bkueng,
Just want to confirm I get terrain following and hold correctly. 

My assumption for terrain following is that if I were in any manual mode where sticks control position, yaw, altitude, and move forward (while range finder able to act) I would follow terrain at that altitude. 

For terrain hold, I think this might only apply when in an altitude hold or position mode. But basically the altitude is locked based on terrain height from sensor rather than baro. Correct?

hamishwillee: Also, how does this differ (effectively) to setting the range aid or EKF2_HGT_MODE to the range sensor (within the distance sensor active range to terrain)?

hamishwillee:
@bkueng All of this was kicked off by there not being any terrain following or range aid docs (ie I was trying to find out what effect having a broken rangefinder might have on behaviour). 
This change means that at least some info will be available. However not sure this is enough/where people will look.

Specifically, do you think we need to have a Terrain Following specific topic, and if so, where might it live. My feeling is that we *might* benefit from a topic on this in "Flying". Or perhaps a general "Rangefinder assisted flight" topic that includes all the info in the config section here.

en/sensor/rangefinders.md
```diff
@@ -68,31 +68,43 @@ PX4 also supports the Bebop rangefinder.
...
+   
+1. Enable *Range Aid* by setting [EKF2_RNG_AID=1](../advanced_config/parameter_reference.md#EKF2_RNG_AID). 
+   - This makes the vehicle use the rangefinder as the primary source of height estimation *when it is safe to use* (during low speed/low altitude operation), but will otherwise use the sensor specified in `EKF2_HGT_MODE`.
+   - It is primarily intended for takeoff and landing, where baro interference from rotor wash is excessive and can corrupt EKF state estimates. 
+   - Specifically, the rangefinder is enabled when:
+     - velocity < [EKF2_RNG_A_VMAX](../advanced_config/parameter_reference.md#EKF2_RNG_A_VMAX)
+     - distance to ground < [EKF2_RNG_A_HMAX](../advanced_config/parameter_reference.md#EKF2_RNG_A_HMAX)
+   - Other parameters affecting "Range aid" are prefixed with `EKF2_RNG_A_`.
+
+1. Enable rangefinder-supported *terrain following* by setting [MPC_ALT_MODE=1](../advanced_config/parameter_reference.md#MPC_ALT_MODE).
```
bkueng: Might be worth noting that this applies to Position & Altitude modes (i.e. not Mission).

en/sensor/rangefinders.md
bkueng: Yes, same question for me :)

en/sensor/rangefinders.md
hamishwillee: That's a pity because it means you can't use it as a primitive form of obstacle avoidance when flying missions, or for avoiding things during RTL.

en/sensor/rangefinders.md
hamishwillee: @bkueng Not clear to me either - who would know?

EKF is only used on copters/VTOL right? (ie I was thinking maybe the RNG_AID might be useful on platforms other than copter, and this could be the essential difference?)

What is clear is that the docs on EKF2_RNG_AID will need an update because they attempt and fail to answer this question (ie they indicate that MPC_ALT_MODE  should be used instead unless there are baro issues, but in fact the description of behaviour is the same).

>
I guess setting EKF2_RNG_AID=1

hamishwillee: I think you'd be mad to set EKF2_HGT_MODE to rangefinder unless you knew you would never go above the effective range, but if you did it would be the same as Range Aid.
 
From what I can tell fusion of the altitude just means "picking a particular source under certain circumstances" not "using some algorithm to work out a best guess by merging the values from different sources". As a result, if you set the primary source as rangefinder and you go out of range there is no fallback to baro - that means you'd just get random crap from EKF for the height.

en/flying/terrain_following_holding.md
```diff
@@ -0,0 +1,40 @@
+# Terrain Following/Holding
```
hamishwillee: @bkueng I agree with you re importance of this feature. So what I have done is added this as a stand alone doc here. I have not yet updated the rangefinder doc to cross link and doubtless this will change as we clarify Range Aid.

This is worth looking at now though. 

My gut feeling is that Range aid is very similar to the terrain hold but applies to more modes - specifically to landing and hold. So it would be worth documenting it in this section too. 
Similarly, range aid provides the same benefits of terrain following to missions, return and similar modes - albeit it does have a user-defined velocity restriction. My leaning therefore is to simply make this clear as an option alongside the other too and make it clear that it will need to be tuned for speed.

Thoughts?

en/sensor/rangefinders.md
bkueng: Right, I think you would get that if you use the range finder as primary height source. And you would use relative height in missions. But this is where it gets tricky, and I really don't know how it works.

en/flying/terrain_following_holding.md
bkueng: > My gut feeling is that Range aid is very similar to the terrain hold but applies to more modes - specifically to landing and hold.

That's also my understanding.

> Similarly, range aid with a large velocity limit will emulate terrain following - but allow them in missions, return and similar modes.

Right, but it was not designed for that. I'm not sure if that has other consequences.

en/flying/terrain_following_holding.md
```diff
@@ -0,0 +1,40 @@
...
+
+PX4 supports *Terrain Following* and *Terrain Hold* on **multicopters** that have a [distance sensor](../sensor/rangefinders.md).
+
+The features are available in manual altitude-control modes: [Position](../flight_modes/position_mc.md), [Altitude](../flight_modes/altitude_mc.md).
+They are not available in acrobatic modes or automatic modes (e.g. Return, Land, Mission, etc.).
```
bkueng: - `acrobatic modes` I would take that to be obvious, no?
- `automatic modes`: I would document (or link) EKF2_HGT_MODE/EKF2_RNG_AID here as well, to make clear we support terrain following in missions too (after we clarified how/if that all works).

en/flying/terrain_following_holding.md
hamishwillee: I'm sure it will. More generally I wonder if the EKF compensates the altitude for vehicle roll etc. At a certain point I assume the rangefinder might provide meaningless results.

en/flying/terrain_following_holding.md
```diff
@@ -0,0 +1,81 @@
...
+This allows a vehicle to avoid altitude changes due to barometer drift or excessive barometer interference from rotor wash.
+
+> **Note** Terrain hold is enabled only in *multicopters* and *VTOL vehicles in MC mode*, and can be used in [position](../flight_modes/position_mc.md) and [altitude modes](../flight_modes/altitude_mc.md).
+
+When moving horizontally (`speed >` [MPC_HOLD_MAX_XY](../advanced_config/parameter_reference.md#MPC_HOLD_MAX_XY), or above the altitude where the distance sensor is providing valid data, the vehicle will instead use the *primary source of altitude data* ([EKF2_HGT_MODE](../advanced_config/parameter_reference.md#EKF2_HGT_MODE)).
+Typically this is the barometer.
+
+Terrain holding is enabled by setting [MPC_ALT_MODE](../advanced_config/parameter_reference.md#MPC_ALT_MODE) to `2`.
+
+> **Note** *Terrain hold* is implemented similarly to *terrain following.
```
hamishwillee: @RomanBapst IS this true? Essentially I am saying Terrain hold is Terrain following but without the sideways velocity limitation?

hamishwillee: Except that you're not moving, so something like the wind pushes you down, the estimated height above ground changes, so your altitude setpoint is adjusted in turn. 
Which seems a bit wrong - the ground hasn't moved, so it "feels" like the altitude setpoint should be unchanged - you've been pushed away from it and should be pushed back to it by EKF.

en/flying/terrain_following_holding.md
```diff
@@ -0,0 +1,81 @@
...
+## Distance Sensor as Primary Source of Height {#distance_sensor_primary_altitude_source}
+
+PX4 allows you to make a distance sensor the *primary source of altitude data*.
+This may be useful when no barometer is available, or for applications when the vehicle is *guaranteed* to only fly over a near-flat surface (e.g. indoors).
+
+> **Tip** The default and preferred altitude sensor for most use cases is the barometer (when available).
+
+When using a distance sensor as the primary source of height, fliers should be aware:
+- Flying over obstacles can lead to the estimator rejecting data (due to internal data consistency checks).
+  This will result in TBD.
```
hamishwillee: @RomanBapst What happens if the estimator rejects the data?

en/flying/terrain_following_holding.md
```diff
@@ -0,0 +1,81 @@
...
+  > **Note** A scenario where this can occur is when the vehicle is ascending a slope at a near-constant height above ground;
+    The measured/estimated altitude does not change, but consistency checks show an error against the barometer measurements.
+- The local NED origin will move up and down with ground level.
```
hamishwillee: @RomanBapst So is there an example of why this might be a problem?

en/flying/terrain_following_holding.md
```diff
@@ -0,0 +1,81 @@
...
+- Rangefinder performance over uneven surfaces (e.g. trees) can be very poor, resulting in noisy and inconsistent data.
```
hamishwillee: Again, if we see noisy data presumably this means that we might have a rapidly varying estimate of position? What happens then ?

en/flying/terrain_following_holding.md
```diff
@@ -0,0 +1,81 @@
...
+*Range Aid* is enabled by setting [EKF2_RNG_AID=1](../advanced_config/parameter_reference.md#EKF2_RNG_AID) (when the primary source of altitude data ([EKF2_HGT_MODE](../advanced_config/parameter_reference.md#EKF2_HGT_MODE)) is *not* the rangefinder).
+
+Range aid is further configured using the `EKF2_RNG_A_` parameters:
+- [EKF2_RNG_A_VMAX](../advanced_config/parameter_reference.md#EKF2_RNG_A_VMAX): Maximum horizonatal speed, above which range aid is disabled.
+- [EKF2_RNG_A_HMAX](../advanced_config/parameter_reference.md#EKF2_RNG_A_HMAX): Maximum height, above which range aid is disabled.
+- [EKF2_RNG_A_IGATE](../advanced_config/parameter_reference.md#EKF2_RNG_A_IGATE): Range aid consistency checks "gate" (a measure of the error before range aid is disabled).
+
+
+   <!-- 
+If vehicle motion causes repeated switching between the primary height sensor and range finder, an offset in the local position origin can accumulate.
```
hamishwillee: @RomanBapst What does this mean? I *think* perhaps every time you go to range finder a local position origin gets set, and movement is then relative to that origin. But my guess would be that when you cycle to primary height sensor and back, doesn't the origin get reset? It is probable that I don't know what local position origin means :-)

In any case, the main question is "what problem does getting errors in the local position offset cause"?

en/flying/terrain_following_holding.md
```diff
@@ -0,0 +1,81 @@
+# Terrain Following/Holding & Range Aid
+
+PX4 supports [Terrain Following](#terrain_following) and [Terrain Hold](#terrain_hold) in [Position](../flight_modes/position_mc.md) and [Altitude modes](../flight_modes/altitude_mc.md), on *multicopters* and *VTOL vehicles in MC mode* that have a [distance sensor](../sensor/rangefinders.md).
+
+PX4 also supports [using a distance sensor as the primary source of altitude data](#distance_sensor_primary_altitude_source) in any mode, either all the time, or just when flying at low altitudes ([Range Aid](#range_aid)).
```
RomanBapst: @hamishwillee ...at low altitudes AND low horizontal velocities. Remember there is the parameter EKF2_RNGA_VMAX.

en/flying/terrain_following_holding.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Terrain Following {#terrain_following}
+
+When *terrain following* is enabled, a vehicle will automatically maintain a relatively constant height above *ground level* if traveling at low altitudes.
+This is useful for avoiding obstacles and for maintaining constant height (e.g. for aerial photography) when flying over varied terrain.
+
+*Terrain following* uses the output of the EKF estimator to provide the altitude estimate, and the estimated terrain altitude (calculated from distance sensor measurements using another estimator) to provide the altitude setpoint.
+As the distance to ground changes, the altitude setpoint adjusts to keep the height above ground constant.
+
+At higher altitudes (when the estimator reports that the distance sensor data is invalid) the vehicle switches to *altitude following*, and will typically fly at a near-constant height above mean sea level (AMSL) using the barometer for altitude data.
```
RomanBapst: @hamishwillee ...using its primary height sensor for estimating altitude. (This could be baro or GPS, see EKF2_HGT_MODE)

RomanBapst: Ah, you actually write this below, forget about it then.

en/flying/terrain_following_holding.md
```diff
@@ -0,0 +1,81 @@
...
+
+
+## Terrain Hold {#terrain_hold}
+
+Terrain holding uses a distance sensor to help a vehicle to better maintain a constant height above ground in [position](../flight_modes/position_mc.md) or [altitude](../flight_modes/altitude_mc.md) modes, when horizontally stationary at low altitude.
+This allows a vehicle to avoid altitude changes due to barometer drift or excessive barometer interference from rotor wash.
+
+> **Note** Terrain hold is enabled only in *multicopters* and *VTOL vehicles in MC mode*, and can be used in [position](../flight_modes/position_mc.md) and [altitude modes](../flight_modes/altitude_mc.md).
+
+When moving horizontally (`speed >` [MPC_HOLD_MAX_XY](../advanced_config/parameter_reference.md#MPC_HOLD_MAX_XY), or above the altitude where the distance sensor is providing valid data, the vehicle will instead use the *primary source of altitude data* ([EKF2_HGT_MODE](../advanced_config/parameter_reference.md#EKF2_HGT_MODE)).
```
RomanBapst: @hamishwillee I think you mean the right thing but it's not formulated correctly. I think it should be:
...or above the altitude where the distance sensor is providing valid data, the vehicle will switch into *altitude following*.
It's important to understand that his has nothing to do with the estimator, it's the position controller switching from terrain hold into altitude following.

en/flying/terrain_following_holding.md
RomanBapst: @hamishwillee Yes, this is true. Terrain hold is terrain following if the vehicle is not moving faster than the threshold. All your assumptions above are correct.

en/flying/terrain_following_holding.md
```diff
@@ -0,0 +1,81 @@
...
+Typically this is the barometer.
+
+Terrain holding is enabled by setting [MPC_ALT_MODE](../advanced_config/parameter_reference.md#MPC_ALT_MODE) to `2`.
+
+> **Note** *Terrain hold* is implemented similarly to *terrain following.
+  It uses the output of the EKF estimator to provide the altitude estimate, and the estimated terrain altitude (calculated from distance sensor measurements using another estimator) to provide the altitude setpoint.
```
RomanBapst: @hamishwillee Can we maybe write a "separate, single state terrain estimator"?

en/flying/terrain_following_holding.md
RomanBapst: @hamishwillee the estimator will reject the data for a certain amount of time until the estimated variance (uncertainty) of that state exceeds a threshold. After that the estimator will perform a state reset which should make data consistent again. Think about it like "ok, my data is inconsistent, let me add an offset to the estimated altitude such that the range finder data is consistent with my altitude again. "

RomanBapst: @hamishwillee In practice the user could notice a drift in altititude (e.g. drone not maintaining altitude well) during the time when range data is rejected.

RomanBapst: @hamishwillee We are talking about a duration of roughly 6 seconds.

en/flying/terrain_following_holding.md
```diff
@@ -0,0 +1,81 @@
...
+PX4 allows you to make a distance sensor the *primary source of altitude data*.
+This may be useful when no barometer is available, or for applications when the vehicle is *guaranteed* to only fly over a near-flat surface (e.g. indoors).
+
+> **Tip** The default and preferred altitude sensor for most use cases is the barometer (when available).
+
+When using a distance sensor as the primary source of height, fliers should be aware:
+- Flying over obstacles can lead to the estimator rejecting data (due to internal data consistency checks).
+  This will result in TBD.
+  > **Note** A scenario where this can occur is when the vehicle is ascending a slope at a near-constant height above ground;
+    The measured/estimated altitude does not change, but consistency checks show an error against the barometer measurements.
```
RomanBapst: @hamishwillee There will be an inconsistency between range data and accelerometer data (the barometer is not used at all, since the range finder is the primary source of height).
When the vehicle move up a hill slowly and maintaining altitude above the ground the accelerometer data will indicate a climb whereas the range finder data claims that the drone is not increasing altitude.

en/flying/terrain_following_holding.md
RomanBapst: @hamishwillee Poor altitude hold

en/flying/terrain_following_holding.md
hamishwillee: Thanks.

en/flying/terrain_following_holding.md
hamishwillee: Sure. I don't know why that is useful to know, but it is done :-)

en/flying/terrain_following_holding.md
hamishwillee: @RomanBapst I'll rewrite this, but I'm really not 100% certain how time vs variance vs reset works

You say "for a certain amount of time until the estimated variance (uncertainty) of that state exceeds a threshold" and "We are talking about a duration of roughly 6 seconds.". Which makes me think that the threshhold/uncertainty is defined in terms of time? 

So consider this scenario were vehicle moving slowly up hill and maintaining altitude above ground. Accelerometer data will indicates a climb whereas the range finder data claims that the drone is not increasing altitude.
I think you are saying that:
- the estimator will *immediately* reject the rangefinder (?) data
- at this point the altitude starts to drift - ie it is hard to maintain constant altitude.
- after 6 or so seconds the system resets and the estimated altitude matches. 
- If the drone was to start moving over a level field the data would still be inconsistent right because of the existing offset - so there would still be a reset after 6 seconds right?

hamishwillee: PS Made minor change to this section, but did not capture all of above.

RomanBapst: @hamishwillee The reset is triggered based on time, see [here](https://github.com/PX4/ecl/blob/master/EKF/control.cpp#L738)
So basically if the range finder data is rejected for 5 seconds then a reset will be performed.
Whether range data is rejected or not depends on innovation consistency checks. Those checks take into account the error between measurement and current state as well as the estimated variance of the state and the variance of the measurement itself.
If the filter is very confident about the estimated state (low variance) then a measurement will be rejected "quicker".

RomanBapst: @hamishwillee 

> the estimator will immediately reject the rangefinder (?) data as soon as they differ

As explained above it depends on innovation consistency checks, it's not just a fixed value.
So no, it will not immediately reject the data.

> at this point the altitude starts to drift as rangefinder data is rejected and altitude is maintained based on the accelerometer - ie it is hard to maintain constant altitude

yes

> There is some threshhold for trusting accelerometer data only - about 6 seconds or so

5 seconds, see [here](https://github.com/PX4/ecl/blob/master/EKF/control.cpp#L738)

> As I see it the only way that the rangefinder can come consistent again is if the vehicle then goes down the hill by the same amount, within the 6 seconds

Yes, that would work

> Anyway, if they aren't consistent after that time, the estimator resets and effectively adds an offset to the rangefinder estimated altitude to make it match what the acc. thinks it is.

No, the estimator resets the state (in this case height) to the current distance sensor data.


> The altitude drifts during the 6 seconds where they don't match because the accelerometer estimate is going to be increasingly unreliable

Predicting the state reliably  using only the accelerometer cannot be done for long with the sensors we are using.

PR #458, opened 2019-03-20
FlightReporting: Improve with auto log upload via QGC

hamishwillee:
QGC supports automatic upload of logs to flight review in 3.5. This update adds that information and simplifies the description appropriately.

### DIFF
--- a/en/getting_started/flight_reporting.md
+++ b/en/getting_started/flight_reporting.md
@@ -1,31 +1,34 @@
 # Flight Reporting
 
-PX4 logs detailed aircraft performance data once the system has been armed until it is disarmed. These flight logs can also be used to analyze performance issues.
+PX4 logs detailed aircraft state and sensor data, which can be used to analyze performance issues.
+This topic explains how you can download and analyse logs, and share them with the development team for review.
 
-> **Tip** Keeping flight logs is a legal requirement in some jurisdictions. 
+> **Tip** Keeping flight logs is a legal requirement in some jurisdictions.
 
-## Downloading Logs from the Flight Controller
+## Upload and Share Logs via Flight Review
 
-Logs can be downloaded using [QGroundControl](http://qgroundcontrol.com/) (v3.2 or later): **[Analyze View > Log Download](https://docs.qgroundcontrol.com/en/analyze_view/log_download.html)**.
+*Flight Review* (http://logs.px4.io) is a web based tool for PX4 flight log analysis.
+In addition to being an effective analysis tool, it the recommended mechanism for sharing logs with the development team (and other users).
 
-![Flight Log Download](../../assets/qgc/analyze/log_download.jpg)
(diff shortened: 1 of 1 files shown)


en/getting_started/flight_reporting.md
```diff
@@ -1,31 +1,34 @@
...
-## Downloading Logs from the Flight Controller
+## Upload and Share Logs via Flight Review
 
-Logs can be downloaded using [QGroundControl](http://qgroundcontrol.com/) (v3.2 or later): **[Analyze View > Log Download](https://docs.qgroundcontrol.com/en/analyze_view/log_download.html)**.
+*Flight Review* (http://logs.px4.io) is a web based tool for PX4 flight log analysis.
+In addition to being an effective analysis tool, it the recommended mechanism for sharing logs with the development team (and other users).
 
-![Flight Log Download](../../assets/qgc/analyze/log_download.jpg)
+You can configure *QGroundControl* to automatically log and upload flight logs to *Flight Review* by following the instructions in:
+[Settings View > MAVLink > MAVLink 2 Logging (PX4 only)](https://docs.qgroundcontrol.com/en/SettingsView/MAVLink.html#logging).
```
bkueng: We need to clarify this further (on both sides): The log streaming (`MAVLink 2 Logging (PX4 only)`) is something different from SD card logging and the QGC log download page.
The log streaming happens in real-time, and requires a constant high-rate MAVLink connection to the vehicle (a telemetry link won't work, a WiFi might).
Since that connection is typically not reliable, we avoid using it for general setups. So I would not recommend it too much, even though it's a nice feature.

hamishwillee: Thanks & curses, foiled again. That is most of the information I am still hoping to get out of @dogmaphobic in https://github.com/mavlink/qgc-user-guide/pull/201#issuecomment-474621966

So even though the approach for getting the log is different, the log itself is the same as would be downloaded normally? 
Also, I assume the data is streamed just to QGC, and not "all the way" to flight review? Ie there is no particular benefit to the streaming right now because there is no dynamic analysis of the log?

I'll leave this open for the discussion, but my plan is to close this and start again once I have the full picture.

bkueng: > So even though the approach for getting the log is different, the log itself is the same as would be downloaded normally?

The data is the same, with the following differences:
- The time of log start and end can be different (e.g. if you have log from boot enabled, your log on the SD card will start from boot). For log streaming, QGC decides when to start log streaming (and stop it).
- The amount of dropouts will differ (expected more for log streaming, since less reliable)

> Also, I assume the data is streamed just to QGC, and not "all the way" to flight review?

Correct

> Ie there is no particular benefit to the streaming right now because there is no dynamic analysis of the log?

Partially, you also don't have to get it from the SD card first.

hamishwillee:
Closing, replaced by https://github.com/PX4/px4_user_guide/pull/463 and https://github.com/mavlink/qgc-user-guide/pull/208

PR #459, opened 2019-03-20
esp8266: Setup instruction improvements

hamishwillee:
This fixes the ESP8266 setup instructions in light of discussion on connecting it in https://github.com/mavlink/qgroundcontrol/issues/7189#issuecomment-469452863

Specifically
- There is no need to install master as this previously started
- the instructions now mention the baud rate requirement and provide link to info on the serial port setup.
- The FC can't be updated until it can be connected, so this now states that you have to have a physical/USB connection to set up the port before you can do the other stuff.


hamishwillee:
@dogmaphobic Please review this when you get a chance.

hamishwillee:
Thanks Gus. In it goes then.

PR #470, opened 2019-04-17
pixhawk-2: fix ADC description

julianoes:
Apparently this was a type and there is only one 3v3 ADC.

This came up in http://discuss.px4.io/t/inquiry-about-the-pixhawk-adc-6-6v/10448.

We should probably check the description of the other Pixhawks as well.


hamishwillee:
PS @julianoes I am very happy for you to just merge this sort of one liner yourself - no one can be more canonical a source than Philip re this FC.

PR #472, opened 2019-05-01
Orbit mode docs

### LINKED ISSUE #397: Add Orbit Mode
Need to document the Orbit mode. 

Some key links
- [Orbit: Send telemetry for the UI #10989](https://github.com/PX4/Firmware/pull/10989)
- [Enable Orbit Flight Mode](https://github.com/PX4/Firmware/pull/10907)
- QGC: [Orbit supports display/changing rotation direction #7064](https://github.com/mavlink/qgroundcontrol/pull/7064)
- [MAV_CMD_DO_ORBIT](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_ORBIT)
- [ORBIT_EXECUTION_STATUS](https://mavlink.io/en/messages/common.html#ORBIT_EXECUTION_STATUS)

As I understand it, 
- this "mode" is initiated by the [MAV_CMD_DO_ORBIT](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_ORBIT). 
  - specify circle 3D centre point, radius, tangential velocity (speed and direction), and yaw (direction vehicle faces)
  - It says "Center point latitude (if no MAV_FRAME specified) / X coordinate according to MAV_FRAME."
    - In the second case it says X co-ordinate in MAV_FRAME (which is clear). But the first cases is "Center point latitude (if no MAV_FRAME specified)" 
      - but that doesn't tell us the frame if no MAV_FRAME is specified? 
      - Is that only a concern if mission is sent in COMMAND_LONG?
- While flying the mode vehicle should output [ORBIT_EXECUTION_STATUS](https://mavlink.io/en/messages/common.html#ORBIT_EXECUTION_STATUS)
  - publishes 3D centre point, radius and travel direction.

Questions:
- Release - when will this all go live? Ie when in master, and when expected in a release?
- Can this only be entered via the command? 
  - ie there is no way to invoke this from an RC switch?
- This is a true mode like Acro etc right? ie not just a mission command.
- Can it be run in a mission?
- What path does the vehicle fly to start the circle? Just straight line to closest edge at given altitude, or is there some pretty path? Does it matter?
- Are there any parameters that affect the behaviour? (ie can it be tuned)
- Does it work for VTOL in MC mode.
- What is the expected UI interaction? 
  - My understanding is that user will only be able to set centre point with default radius on QGC, and then will be able to drag the radius in and out.
   - Will they also be able to change speed and direction from UI?
   - This is another case of where I can imagine a user might want to have control from RC. On Solo for example you can set the ROI point and then orbit around the point, changing the radius with your joystick.

@MaEtUgR

hamishwillee:
Fixes #397 

This is first draft of Orbit mode docs. Some things are known "not great". Still need to verify that the mode will be present in v1.9.0

Rendered version [here](https://hamishwillee.gitbooks.io/ham_px4userguide/content/v/orbit_mode/en/flight_modes/orbit.html). Some other references [here](https://hamishwillee.gitbooks.io/ham_px4userguide/content/v/orbit_mode/en/getting_started/flight_modes.html#orbit_mc) and [here](https://hamishwillee.gitbooks.io/ham_px4userguide/content/v/orbit_mode/en/flight_modes/#orbit_mc).

@MaEtUgR can you please review?


hamishwillee:
@MaEtUgR One thing I'm concerned about is the usage image:

image

The diagram shows the right stick as angle control. I.e. setpoint reflects a fixed rate of rotation, rather than a delta in increased rate of rotation. Is that correct?
Would it be better just to remove the "Angle control" information. Is it otherwise OK?

Note, the instructions for setting the orbit via QGC are linked to in QGC docs. Decided not to duplicate them.

Is there anything else you think we need to tell people?

MaEtUgR:
> This is first draft of Orbit mode docs.

Thanks! Awsome!!

> Still need to verify that the mode will be present in v1.9.0

It is, did someone plan to disable it? I wouldn't know why.

> speed will ramp up in a particular direction and/or the rate at which the radius will change in a particular direction. Is that correct?

Yes exactly, the stick axis accelerates CW/CCW and it does so from the vehicle yaw perspective. The vehicle always points the nose to the center and hence right is CCW.

en/flight_modes/README.md
```diff
@@ -356,6 +356,19 @@ th {
...
 
+<tr id="orbit_mc">
+ <td><a href="../flight_modes/orbit.md">Orbit</a>
+ <p><a href="#key_difficulty"><img src="../../assets/site/difficulty_easy.png" title="Easy to fly" width="20px" /></a></p>
+ </td>
+ <td>-</td>
+ <td>-</td>
+ <td>-</td>
+ <td></td>
+ <td><p>GCS/RC mode for flying a circle, always facing the centre.</p>
```
MaEtUgR: No offense, I really like British English. I'm only suggesting to change it to match the code and in the code it's Ami because I thought that's more generally understandable to say it carefully.
```suggestion
 <td><p>GCS/RC mode for flying a circle, always facing the center.</p> 
```

en/flight_modes/README.md
```diff
@@ -356,6 +356,19 @@ th {
...
+<tr id="orbit_mc">
+ <td><a href="../flight_modes/orbit.md">Orbit</a>
+ <p><a href="#key_difficulty"><img src="../../assets/site/difficulty_easy.png" title="Easy to fly" width="20px" /></a></p>
+ </td>
+ <td>-</td>
+ <td>-</td>
+ <td>-</td>
+ <td></td>
+ <td><p>GCS/RC mode for flying a circle, always facing the centre.</p> 
+  <p>Mode must be started from GCS, specifying centre point and initial radius and altitude. RC control can then optionally be used to set the orbit altitude, radius, speed, and direction. Altitude control is the same as for Position Mode.</p>
```
MaEtUgR:
```suggestion
  <p>Mode must be started from GCS, specifying center point and initial radius and altitude. RC control can then optionally be used to set the orbit altitude, radius, speed, and direction. Altitude control is the same as for Position Mode.</p>
```

en/flight_modes/orbit.md
```diff
@@ -0,0 +1,46 @@
+# Orbit (MC)
+
+[<img src="../../assets/site/difficulty_easy.png" title="Easy to fly" width="30px" />](../getting_started/flight_modes.md#key_difficulty)&nbsp;[<img src="../../assets/site/remote_control.svg" title="Manual/Remote control required" width="30px" />](../getting_started/flight_modes.md#key_manual)&nbsp;[<img src="../../assets/site/position_fixed.svg" title="Position fix required (e.g. GPS)" width="30px" />](../getting_started/flight_modes.md#key_position_fixed)
+
+The *Orbit* flight mode allows you to fly a multicopter (or VTOL in multicopter mode) in a circle, yawing so that it always faces towards the centre.
+
+![Orbit Mode - MC](../../assets/flying/orbit.jpg)
+
+*QGroundControl* (or other compatible GCS or MAVLink API) is *required* to enable the mode, and to set the centre position, initial radius and altitude of the orbit.
```
MaEtUgR:
```suggestion
*QGroundControl* (or other compatible GCS or MAVLink API) is *required* to enable the mode, and to set the center position, initial radius and altitude of the orbit.
```

en/flight_modes/orbit.md
```diff
@@ -0,0 +1,46 @@
...
+- **Left stick:** up/down for altitude change as in [Position mode](../flight_modes/position_mc.md).
+- **Right stick:** right/left controls speed and direction (counter-clockwise/clockwise), up/down controls orbit radius (smaller/bigger).
+
+The diagram below shows the mode behaviour visually (for a [mode 2 transmitter](../getting_started/rc_transmitter_receiver.md#transmitters-for-aircraft)).
+
+![Orbit Mode - MC](../../images/flight_modes/orbit_MC.png)
+
+The mode can be stopped by switching to any other flight mode (using RC or QGC).
+
+<!--
```
MaEtUgR: Should I describe how it's implemented or an example how to start it via command here?

en/getting_started/flight_modes.md
```diff
@@ -35,7 +36,7 @@ Fixed Wing:
 * Autonomous: [Hold](#hold_fw), [Return](#return_fw), [Mission](#mission_fw), [Takeoff](#takeoff_fw), [Land](#land_fw)
 
 Multicopter: 
-* Manual-Easy: [Position](#position_mc), [Altitude](#altitude_mc), [Manual/Stabilized](#manual_stabilized_mc)
+* Manual-Easy: [Position](#position_mc), [Altitude](#altitude_mc), [Manual/Stabilized](#manual_stabilized_mc), [Orbit](#orbit_mc)
```
MaEtUgR: I think we should also mention that you can use Orbit without any remote? There's no failsafe needed when you loose RC because it's optional.

en/flight_modes/orbit.md
```diff
@@ -0,0 +1,46 @@
...
+
+*QGroundControl* (or other compatible GCS or MAVLink API) is *required* to enable the mode, and to set the centre position, initial radius and altitude of the orbit.
+Once enabled the vehicle will fly as fast as possible to the closest point on the commanded circle trajectory and do a slow (1m/s) clockwise orbit on the planned circle, facing the center.
+Instructions on how to start an orbit can be found here: [FlyView > Orbit Location](https://docs.qgroundcontrol.com/en/FlyView/FlyView.html#orbit) (*QGroundControl* guide).
+
+> **Note** RC control cannot be used to start the mode (if you switch to the mode via RC it will sit idle).
+
+Once orbit mode is active, the RC controller can be used to change the orbit altitude, radius, speed, and orbit direction:
+- **Left stick:** up/down for altitude change as in [Position mode](../flight_modes/position_mc.md).
+- **Right stick:** right/left controls speed and direction (counter-clockwise/clockwise), up/down controls orbit radius (smaller/bigger).
```
MaEtUgR: Like you wrote in the pr conversation comment https://github.com/PX4/px4_user_guide/pull/472#issuecomment-488187345 the stick ramps up and down the linear velocity (m/s not °/s) and is hence used to accelerate, decelerate, stop, start spinning in the other direction and the speed is kept for further semi-autonomous flight afterwards. Without it keeping the speed by itself it would not be possible to use it "hands-off".

hamishwillee:
Thanks @MaEtUgR . I'm too tired to do "final check and merge" but will do so on Monday.

>> Still need to verify that the mode will be present in v1.9.0

> It is, did someone plan to disable it? I wouldn't know why.

Thanks! No, just couldn't find the PR in the release notes list.

en/flight_modes/orbit.md
hamishwillee: No. For the manual flight modes we tend to explain in bullet form the detail of the way that the sticks control movement here. However this is a "guided" mode, and I don't think we need such a section.

In MAVLink docs we probably should have info on the Orbit Mode protocol as a microservice, but in no way urgent.

en/getting_started/flight_modes.md
hamishwillee: Absolutely. I'll update that now and remove the RC required icon from the top of the page. Thanks.

en/flight_modes/orbit.md
hamishwillee: @MaEtUgR That isn't quite what I was querying in the above comment, and now I am confused. Can you confirm:

Sticks can either angle control or angle rate control a behaviour - say the speed around the orbit in m/s. 
- If the speed is angle controlled, then if I have the stick at 30% (say) from centre, then the speed will be some fixed value - and as I move it back to the centre the value would reduce. In the centre the speed is zero. If move the stick down then I change the setpoint to a fixed speed in the other direction.
- If the speed is _angle rate_ controlled and I have the stick at 30% say, then this controls the rate of increase of speed. If I leave it there for a bit it will keep increasing. If I move it back to center the speed will stop increasing, but will stay the same. If I move the stick lower then I start applying a negative change to the angle rate.

My believe is that the radius and speed are angle rate controlled - ie if I move the sticks back to the centre or RC drops out the current values will be kept. If I want to change direction, I have to provide a reverse value. That is correct right?

So for right stick I might change
> right/left controls speed and direction (counter-clockwise/clockwise), up/down controls orbit radius (smaller/bigger).

to say:

> right/left controls rate of change of speed in a particular direction (counter-clockwise/clockwise), up/down controls rate of change of orbit radius (smaller/bigger).

Is that correct?

If so, then the diagram is wrong too and I think should be as per the purple writing. Ie the left stick is actually angle controlled
image

[This implies all my diagrams are wrong for position and altitude modes - these show the altitude as rate controlled when I THINK that actually Yaw is rate controlled on left stick, but altitude is angle controlled - right?]

MaEtUgR: Now I'm also confused.

> If the speed is angle rate controlled then the stick angle controls the rate of change of speed. If I leave it at some positive value for a bit it will keep increasing at the rate defined by the angle. If I move it back to center the speed will stop increasing, but will stay the same final speed. If I move the stick lower then I start applying a negative change to the angle rate and will slow it down.

That's correct.

Let me write it as clear as possible:

|                        | Centered/RC loss                    | Deflected Down/Left                     | Deflected Up/Right                   |
|------------------------|-------------------------------------|-----------------------------------------|--------------------------------------|
| Left stick down/up     | Altitude is locked                  | Vehicle has downwards speed             | Vehicle has upwards speed            |
| Left stick left/right  | -                                   | -                                       | -                                    |
| Right stick down/up    | Keep current radius                 | Radius gets bigger until maximum (100m) | Radius gets lower until minimum (1m) |
| Right stick left/right | Keep current m/s speed around orbit | Accelerate CW/Decelerate CCW *          | Decelerate CW/Accelerate CCW *       |

* When Decelerating to zero speed and staying on the stick it will switch direction of rotation and accelerate again. Maximum velocity is 10m/s and it's even more limited to keep the centripetal acceleration below 2m/s^2.

hamishwillee: @MaEtUgR Ha ha. I like the table - will steal it.
Still need to know whether rate or angle controlled because that affects the diagrams. Fortunately this is now mostly answered - see queries inline.

From the table I can say that 
- altitude (Left stick Down/UP) is angle controlled. The speed setpoint is mapped to the angle of the stick.
- radius (Right stick down up) looks like angle rate control -ish
  - does the rate of radius change change based on the stick angle or is it basically a fixed rate of change when stick is not centred?
 - speed of rotation (Right stick, left right) - this looks angle rate controlled, with acceleration based on stick position, but limited as you indicated?

Upshot, I do need to change the left stick labels. 
For right stick, depends on your answer above, but probably left/right at least is rate controlled.

hamishwillee:
@MaEtUgR Thanks for all your help on this - I've done some minor updates and merged. Realised I don't need clarification on the angle vs rate stuff. Mostly end users might care about whether stick controls position, speed or acceleration, so those are more useful terms to use.

MaEtUgR:
I have to thank you for the awsome documentation work! 👍

KnightHawk06:
Is https://mavlink.io/en/messages/common.html#MAV_CMD_DO_ORBIT able to implemented in Mission mode?

PR #479, opened 2019-05-10
Add doc for lanbao cm8jl65 distance sensor

hamishwillee:
Just saw this new sensor mentioned in slack. As driver is present in 1.9, adding so that it will go in the 1.9 cut of docs. 

@bkueng As I understand it key selling point of this is that it is small and light - so that is what I said. Anything else of interest to say?


jkflying:
I've worked with it a fair bit - main thing is small and light with 8m range.

hamishwillee:
Thanks @jkflying . I think I captured that in the docs. 

Next time you have a look at that sensor, can you add module docs in-source - you can see we've been trying to do that for all of them: https://dev.px4.io/master/en/middleware/modules_driver_distance_sensor.html

PR #482, opened 2019-05-16
FlightREview: Add general refs re vibration analysis

hamishwillee:
Some references to general vibration analysis from @bresch: https://github.com/PX4/px4_user_guide/issues/434#issuecomment-492543293

These have been added to the "Solutions" section of the flight review for want of a better place.

Beat, do you think we might reasonably pull out this whole section and add it as a stand along doc under Basic assembly? I like the idea because this doc would then show up clearly in search and can be linked from several places. Only downside is that there isn't all that much information. At least it would become obvious where to put more information in future.

I am leaning towards this.

### DIFF
--- a/en/assembly/vibration_isolation.md
+++ b/en/assembly/vibration_isolation.md
@@ -19,12 +19,12 @@ Further isolation may be needed in order to reduce vibration to the level that s
 ## Basic Vibration Fixes
 
 A few of simple steps that may reduce vibrations are:
-- Use a vibration-isolation method to mount the autopilot.
-  Many flight controllers come with *mounting foam* that you can use for this purpose, while others have inbuilt vibration-isolation mechanisms.
 - Make sure everything is firmly attached on the vehicle (landing gear, GPS mast, etc.).
 - Use balanced propellers.
 - Make sure to use high-quality components for the propellers, motors, ESC and airframe.
   Each of these components can make a big difference.
+- Use a vibration-isolation method to mount the autopilot.
+  Many flight controllers come with *mounting foam* that you can use for this purpose, while others have inbuilt vibration-isolation mechanisms.
 - As a *last* measure, adjust the software filters (see [here](../config_mc/racer_setup.md#filters)).
   It is better to reduce the source of vibrations, rather than filtering them out in software.
   
--- a/en/assembly/vibration_isolation.md
+++ b/en/assembly/vibration_isolation.md
@@ -5,7 +5,7 @@ This topic shows how to determine whether vibration levels are too high, and lis
 ## Overview
 
 Flight Control boards with in-built accelerometers or gyros are sensitive to vibrations.
-High vibration levels can cause sensor clipping and sensor failures, resulting in a range of problems, from reduced flight efficiency/performance (leading to shorter flights and increased vehicle wear-and-tear), through to fly-aways and crashes.
+High vibration levels can cause a range of problems, including reduced flight efficiency/performance, shorter flight times and increased vehicle wear-and-tear. In extreme cases vibration may lead to sensor clipping/failures, possibly resulting in estimation failures and fly-aways.
 
 Well-designed airframes damp/reduce the amplitude of specific structural resonances at the autopilot mounting location.
 Further isolation may be needed in order to reduce vibration to the level that sensitive components can handle (e.g. some flight controllers must be attached to the airframe using some form of anti-vibration foam/mount - while others are internally isolated).
--- a/en/assembly/vibration_isolation.md
+++ b/en/assembly/vibration_isolation.md
@@ -1,0 +1,36 @@
+# Vibration Isolation
+
+This topic shows how to determine whether vibration levels are too high, and lists some simple steps to improve vibration performance.
+
+## Overview
+
+Flight Control boards with in-built accelerometers or gyros are sensitive to vibrations.
+High vibration levels can cause sensor clipping and sensor failures, resulting in a range of problems, from reduced flight efficiency/performance (leading to shorter flights and increased vehicle wear-and-tear), through to fly-aways and crashes.
+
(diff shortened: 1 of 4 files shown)


bkueng:
> Beat, do you think we might reasonably pull out this whole section and add it as a stand along doc under Basic assembly?

Yes I would do that, given it's an important topic.

hamishwillee:
Thanks @bkueng . OK, updated as stand-alone topic. Please re-review.

en/log/flight_review.md
```diff
@@ -181,21 +181,13 @@ Very high (unsafe) vibration levels.
...
-Solutions and steps to reduce vibrations include:
-- Make sure everything is firmly attached on the vehicle (landing gear, GPS mast, etc.)
-- Use balanced propellers.
-- Make sure to use high-quality components for the propellers, motors, ESC and airframe.
-  Each of these components can make a big difference.
-- Use a vibration-isolation method to mount the autopilot.
-- As a *last* measure, adjust the software filters (see [here](../config_mc/racer_setup.md#filters)).
-  It is better to reduce the source of vibrations, rather than filtering them out in software.
+In this case the vehicle should be inspected.
+[Vibration Isolation](../assembly/vibration_isolation.md) explains some basic things you can check (and do) to improve vibration performance.
```
bkueng: Maybe like this?
```suggestion
[Vibration Isolation](../assembly/vibration_isolation.md) explains some basic things you can check (and do) to reduce vibration levels.
```

en/assembly/vibration_isolation.md
```diff
@@ -0,0 +1,36 @@
+# Vibration Isolation
+
+This topic shows how to determine whether vibration levels are too high, and lists some simple steps to improve vibration performance.
```
bkueng: `vibration performance` sounds a bit strange, Mathieu suggested `vibration characteristics` instead.

en/assembly/vibration_isolation.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Overview
+
+Flight Control boards with in-built accelerometers or gyros are sensitive to vibrations.
+High vibration levels can cause sensor clipping and sensor failures, resulting in a range of problems, from reduced flight efficiency/performance (leading to shorter flights and increased vehicle wear-and-tear), through to fly-aways and crashes.
```
bkueng: The way I understand this, the clipping and failures result in a range of problems. However event w/o clipping or failures you can have problems.
Do I understand this incorrect?

en/assembly/vibration_isolation.md
```diff
@@ -0,0 +1,36 @@
...
+
+Airframes are designed to damp/reduce the amplitude of specific structural resonances at the autopilot mounting location.
```
bkueng: I wish all of them were! :)

en/assembly/vibration_isolation.md
```diff
@@ -0,0 +1,36 @@
...
+Users should further isolate the flight controller from airframe (using foam or some other isolation mount) in order to reduce vibration to the level that the autopilot can handle.
```
bkueng: I would not generally suggest that. Some boards are isolated internally, and in some cases hard-mounting works too.

en/assembly/vibration_isolation.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Vibration Analysis
+
+[Log Analysis using Flight Review > Vibration](../log/flight_review.md#vibration) explains how to use logs to confirm whether vibration is a probable cause of flight problems.
+
+
+## Basic Vibration Fixes
+
+A few of simple steps that may reduce vibrations are:
+- Use a vibration-isolation method to mount the autopilot.
```
bkueng: I'd keep the original order: reducing vibrations at the source (e.g. props) is better than adding vibration-isolation for the autopilot.

en/assembly/vibration_isolation.md
hamishwillee: :-)

I changed this to "Well-designed airframes damp/reduce the amplitude of specific structural resonances at the autopilot mounting location."

en/assembly/vibration_isolation.md
hamishwillee: I've generalised this to:

> Further isolation may be needed in order to reduce vibration to the level that sensitive components can handle (e.g. some flight controllers must be attached to the airframe using some form of anti-vibration foam/mount - while others are internally isolated).

FYI, what I'm trying to capture here is Paul's comment:

> Damping as an objective is normally something done to the airframe itself to reduce the amplitude of specific structural resonances at the the autopilot mounting location to levels that the isolation mount + autopilot can handle.

en/assembly/vibration_isolation.md
hamishwillee: What this says is that it is the sensor and clipping failures (caused by high vibration) that result in the listed problems.

I think you are saying that some of these problems are not due to the sensor and clipping failures - so you can still have the problems.

I can reword, but would be useful if you can help me understand which problems are associated just with clippin?

bkueng: Exactly. Clipping is the extreme case that leads to (position) estimation failures and fly-aways.
This holds for the accel, I've never seen a gyro clipping, but likely similar bad things happen.

hamishwillee: Thanks @bkueng . So would you be happy with:

> High vibration levels can cause a range of problems, including reduced flight efficiency/performance, shorter flights and increased vehicle wear-and-tear. In extreme cases vibration may lead to sensor clipping/failures, possibly resulting in estimations and fly-aways.

bkueng: I'd say 'shorter flight time' and 'estimation failures', otherwise looks good.

hamishwillee: Thanks. Fixed and nearly merged. Those were pretty rubbish typos I hope I would have fixed before submitting.

PR #484, opened 2019-05-22
Safety: Clarify Data Link Loss applies in missions

hamishwillee:
As discussed in https://github.com/PX4/Firmware/issues/12015#issuecomment-493983484

I said "when flying a mission". I presume that we don't need to be very pedantic and say "when not landed in a mission"?


hamishwillee:
@bresch YOu were too slow. Decided this is accurate. If not, please let me know.

bresch:
@hamishwillee sorry, yes, this is all good! :)

PR #489, opened 2019-05-27
Trimble serial port baud rate

hamishwillee:
The old serial config param is gone - I think this has changed to SER_GPS1_BAUD

@bkueng Is this the same as all other serial port things? I.e. Shold I further state that you need to set the actual GPS port you have connected and the config?


bkueng:
Yes we can do that.
SET_GPS1_BAUD was just a typo, this parameter never existed.

hamishwillee:
Thanks @bkueng - happy with this?

PR #496, opened 2019-05-27
Fix up crazyflie defconfig docs

### LINKED ISSUE #486: Crazyflie 2.0 Where is CONFIG_DEV_LOWCONSOLE/nuttx-config?
[Crazyflie 2.0 · PX4 User Guide](https://docs.px4.io/master/en/complete_vehicles/crazyflie2.html#flashing-px4)

Has this line

> If QGroundControl does not connect with the vehicle, ensure that in nuttx-config for crazyflie # CONFIG_DEV_LOWCONSOLE is not set is replaced by CONFIG_DEV_LOWCONSOLE=y

But nuttx-config is not at https://github.com/PX4/Firmware/blob/master/platforms/nuttx/nuttx-configs/crazyflie/nsh/defconfig#L934 and search does not find CONFIG_DEV_LOWCONSOLE

What should this say/where is the link?

hamishwillee:
Fixes #486

### DIFF
--- a/en/complete_vehicles/crazyflie2.md
+++ b/en/complete_vehicles/crazyflie2.md
@@ -77,7 +77,7 @@ After setting up the PX4 development environment, follow these steps to install 
   ```
   make bitcraze_crazyflie_default menuconfig
   ```
-  or *qconfig*:
+  or *qconfig* (Check *Low-level console support* under *Serial Driver Support* in GUI):
   ```
   make bitcraze_crazyflie_default qconfig
   ```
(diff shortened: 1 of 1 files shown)


dagar:
For reference, this is what it looks like in `qconfig`.

Screenshot from 2019-05-27 20-01-23

hamishwillee:
Thanks @dagar . I decided not to include the qconfig gui reference but did state the name of the checkbox so it is easier to find.

PR #508, opened 2019-07-03
Add ADSB/Flarm docs

hamishwillee:
This adds ADSB/FLarm docs.

@julianoes There are some open questions, which I hope you can help me with. Specifically:
1. What hardware do we support for this purpose - ie:
   - is there a "generic" definition/protocol/type that people can use.
   - What specific types have we tested against, and where do we recommend people getting it from
2. What hardware setup is required. Ie what bus/port, and what parameters are needed? Do the boards come with appropriate connectors? 
3. How can this feature be simulated/tested?

Is there anything else that should be mentioned?

### DIFF
--- a/en/advanced_features/traffic_avoidance_adsb.md
+++ b/en/advanced_features/traffic_avoidance_adsb.md
@@ -1,22 +1,81 @@
 # Air Traffic Avoidance: ADS-B/FLARM
 
-PX4 uses [ADS-B](https://en.wikipedia.org/wiki/Automatic_dependent_surveillance_%E2%80%93_broadcast) and [FLARM](https://en.wikipedia.org/wiki/FLARM) transponders to support simple air traffic avoidance in [missions](../flight_modes/mission.md).
-If a potential collision is detected, PX4 can *warn*, immediately [land](../flight_modes/land.md), or [return](../flight_modes/return.md) - depending on the value of [NAV_TRAFF_AVOID](#NAV_TRAFF_AVOID).
+PX4 can use [ADS-B](https://en.wikipedia.org/wiki/Automatic_dependent_surveillance_%E2%80%93_broadcast) or [FLARM](https://en.wikipedia.org/wiki/FLARM) transponders to support simple air traffic avoidance in [missions](../flight_modes/mission.md).
+If a potential collision is detected, PX4 can *warn*, immediately [land](../flight_modes/land.md), or [return](../flight_modes/return.md) (depending on the value of [NAV_TRAFF_AVOID](#NAV_TRAFF_AVOID)).
 
 
-## Setup
+## Supported Hardware {#supported_hardware}
 
-### Hardware Setup
+PX4 traffic avoidance works with ADS-B or FLARM products that supply transponder data using the MAVLink [ADSB_VEHICLE](https://mavlink.io/en/messages/common.html#ADSB_VEHICLE) message.
+
+It has been tested with the following devices:
+- [PingRX ADS-B Receiver](https://uavionix.com/product/pingrx/) (uAvionix)
+- [FLARM](https://flarm.com/products/powerflarm/uav/)
 
 
+## Hardware Setup
 
-### Software Configuration (Parameters)
+Either device can can be connected to any free/unused serial port on the flight controller.
+Most commonly it they are connected to TELEM2 (if this is not being use for some other purpose).
+
+### PingRX
+
+The PingRX MAVLink port uses a JST ZHR-4 mating connector with pinout as shown below.
+
+Pin | Signal | Volt
+--- | --- | ---
+1 (red) | RX (IN)  | +5V tolerant
+2 (blk) | TX (OUT) | 
+3 (blk) | Power  | +4 to 6V
+4 (blk) | GND    | GND
+
+The PingRX comes with connector cable that can be attached directly to the TELEM2 port (DF13-6P) on an [mRo Pixhawk](../flight_controller/mro_pixhawk.md).
+For other ports or boards, you will need to obtain your own cable.
 
 
+## FLARM
+
+FLARM has an on-board DF-13 6 Pin connector that has an identical pinout to the [mRo Pixhawk](../flight_controller/mro_pixhawk.md).
+
+Pin | Signal | Volt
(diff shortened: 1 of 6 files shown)


julianoes:
Sorry, I know nothing about ADSB!

LorenzMeier:
Here is the ADSB Device:
https://uavionix.com/product/pingrx/

FLARM:
https://flarm.com/products/powerflarm/uav/

LorenzMeier:
Configuration: Set TELEM2 to 57600 baud and a low-bandwidth profile and connect either module to it.

hamishwillee:
A few more notes:
1. This should work with any avoidance system that implements [ADSB_VEHICLE](https://mavlink.io/en/messages/common.html#ADSB_VEHICLE)
2. PX4 also relays the MAVLink data from the reciever to the GCS.

en/advanced_features/traffic_avoidance_adsb.md
```diff
@@ -0,0 +1,97 @@
...
+### Port Configuration
+
+Flarm/PingRX are configured in the same way as any other [MAVLink Peripheral](../peripherals/mavlink_peripherals.md). 
+The only *specific* setup is that the port baud rate must be set to 57600 and the a low-bandwidth profile (`MAV_X_MODE`).
+
+Assuming you have connected the device to the TELEM2 port, [set the parameters](../advanced_config/parameters.md) as shown:
+
+- [MAV_1_CONFIG](../advanced_config/parameter_reference.md#MAV_1_CONFIG) = TELEM 2
+- [MAV_1_MODE](../advanced_config/parameter_reference.md#MAV_1_MODE) = Normal
+- [MAV_1_RATE](../advanced_config/parameter_reference.md#MAV_1_RATE) = 0 (default sending rate for port).
```
hamishwillee: @DanielePettenuzzo @bkueng Why do we need both MAV_1_RATE and SER_TEL2_BAUD - ie to set rate for both mavlink and the port? 
- Daniele indicated that we can set the rate to 0 to use default for the RATE - correct?
- What if we don't set this?

en/advanced_features/traffic_avoidance_adsb.md
```diff
@@ -0,0 +1,97 @@
...
+
+### Port Configuration
+
+Flarm/PingRX are configured in the same way as any other [MAVLink Peripheral](../peripherals/mavlink_peripherals.md). 
+The only *specific* setup is that the port baud rate must be set to 57600 and the a low-bandwidth profile (`MAV_X_MODE`).
+
+Assuming you have connected the device to the TELEM2 port, [set the parameters](../advanced_config/parameters.md) as shown:
+
+- [MAV_1_CONFIG](../advanced_config/parameter_reference.md#MAV_1_CONFIG) = TELEM 2
+- [MAV_1_MODE](../advanced_config/parameter_reference.md#MAV_1_MODE) = Normal
```
hamishwillee: Is Normal correct, or should it be *Minimal*

en/advanced_features/traffic_avoidance_adsb.md
bkueng: RATE=0 is good, it's the maximum data sending rate.

en/advanced_features/traffic_avoidance_adsb.md
hamishwillee: Normal, Onboard, Config work. Minimal does not. So this is OK.

en/advanced_features/traffic_avoidance_adsb.md
hamishwillee: @bkueng According to https://dev.px4.io/master/en/advanced/parameter_reference.html#MAV_1_RATE it is half the theoretical maximum bandwidth - is that equivalent to the maximum data sending rate,?

Also, the docs for MAV_1_RATE say "If this is set to 0, a value of /20 is used, which corresponds to half of the theoretical maximum bandwidth." - what is /20 supposed to mean? "20B/s" or "1/2"?
My guess is that it is a typo  for half, where the maximum data rate is based on the baud rate for the port. Is that correct?

hamishwillee:
@DanielePettenuzzo Can you please sanity check this? I'd like to merge.

en/advanced_features/traffic_avoidance_adsb.md
bkueng: A couple of things:
- bit != byte
- baudrate != bitrate

We have: baudrate / 10 = maximum datarate in bytes. Following baudrate/20 = half of the maximum. Which is close to the practical limit that you can push through (on an average setup).
So not a typo, but the explanation might be improved.

hamishwillee: Thanks @bkueng - Arrrrggg! OK, makes sense now - my brain was reading "Bytes/sec as bits/sec".

Is it exactly 10 or "around 10"? I'm *guessing* that this depends on the channel - ie parity bits etc?

In any case, the comment minimally needs "baudrate" added to make any sense at all. 

> Configure the maximum sending rate for the MAVLink streams in Bytes/sec. If the configured streams exceed the maximum rate, the sending rate of each stream is automatically decreased. If this is set to 0, a value of **baudrate**/20 is used, which corresponds to half of the theoretical maximum bandwidth.

But perhaps even better as:

> Configure the maximum sending rate for the MAVLink streams in Bytes/sec. If the configured streams exceed the maximum rate, the sending rate of each stream is automatically decreased. If this is set to 0 a value of half of the theoretical maximum bandwidth is used. This corresponds to baudrate/20 Bytes/s (baudrate/10=maximum data rate).

hamishwillee:
I think this is good now. Any further suggestions? @DanielePettenuzzo @bkueng

en/advanced_features/traffic_avoidance_adsb.md
bkueng: > Is it exactly 10 or "around 10"? I'm guessing that this depends on the channel - ie parity bits etc?

Exactly, but we generally use 8N1 config, for which it is true.

Both of your suggestions look good.

hamishwillee:
@bkueng Thanks for review and patience. I am not waiting on @DanielePettenuzzo because he is unreliable :-)

PR #510, opened 2019-07-05
Add safe landing feature page

hamishwillee:
Hi Martina,

This is minimal PX4-side introduction to Safe Landing. Can you please review?

What is basically supposed to explain is:
- what the safe landing feature does and does not do, 
- that you use COM_OBS_AVOID to enable this on PX4 (along with every other service that uses the trajectory interface). 
- It then points to PX4/avoidance for the rest of the information.

What is largely empty is the section on the safe landing interface. What we want to do here is explain how the integration works at high level, so that someone else could replace the companion side interface if they wanted. 
Some questions:
- How does this differ from [obstacle avoidance](https://docs.px4.io/en/computer_vision/obstacle_avoidance.html#mission_avoidance_interface). Ie what gets sent from PX4 when you're landing? Presumably not 3 points - last, current, next waypoint.
- How does the companion know you're landing rather than doing some other mission behaviour?
- Does it just send back the same format trajectory as it does for obstacle avoidance?
- How does it know when you're landed an you don't want more points?
- Can you run multiple planners at the same time? ie It isn't clear when scanning the avoidance repo what you do to run landing, collision prevention, and obstacle avoidance at the same time.
- You mentioned in the dev summit that you send null points when doing calculations that PX4 is expected to ignore (but use to confirm that companion is alive and not enter failsafe). How are these formatted?
- Perhaps worth adding message flow for Safe landing? https://github.com/PX4/avoidance#message-flows


en/computer_vision/safe_landing.md
```diff
@@ -0,0 +1,49 @@
...
+
+The *Safe Landing* feature ensures that vehicles only land on flat terrain.
+
+The feature can be enabled in both [Land mode](../flight_modes/land.md) and [Mission mode](../flight_modes/mission.md) on multicopter vehicles (and VTOL vehicles in MC mode) that have a companion computer running the appropriate vision software.
+
+If commanded to land, the vehicle first descends to a height where it can measure the surface (companion computer `loiter_height` parameter).
+If the landing area is not sufficiently flat, the vehicle moves outwards in a square-spiral pattern, periodically stopping to re-check the terrain for a landing spot that isn't too rough.
+
+> **Warning** The feature will not prevent a vehicle from landing on a road (even if a car is detected it will be "forgotten" once it moves past).
+  It will prevent landing on water, as the cameras used do not "see" the water surface at all.
```
hamishwillee: FYI, this came up at dev summit, and this was Julian Kent's response. Is it true? I mean it is true for our setup, but will it be true in future if someone used a different camera?

hamishwillee:
PS YOu can see this in a rendered gitbook here: https://hamishwillee.gitbooks.io/ham_px4userguide/content/v/safe_landing/en/computer_vision/safe_landing.html

mrivi:
> How does this differ from obstacle avoidance. Ie what gets sent from PX4 when you're landing? Presumably not 3 points - last, current, next waypoint.

The interface is exactly the same as for OA. Firmware sends type adapted current waypoint from FlightTaskAutoMapper, current waypoint as it is in mission item, next waypoint as it is in mission item

> How does the companion know you're landing rather than doing some other mission behaviour?

On the companion side we know the px4 mode we're in e.g. AUTO LAND, MISSION, AUTO RTL and the command associated with the waypoint to know if it's a land waypoint within the mission

> Does it just send back the same format trajectory as it does for obstacle avoidance?

yes

> How does it know when you're landed an you don't want more points?

Firmware detects that the vehicle has landed and disarms. Companion subscribes to the vehicle state and knows the vehicle disarmed 

> Can you run multiple planners at the same time? ie It isn't clear when scanning the avoidance repo what you do to run landing, collision prevention, and obstacle avoidance at the same time.

No, you can't at the moment. Collision Prevention can't overlap with the OA and SafeLanding because it only runs in Position Control mode while the other two run in Auto modes. 
It would be interesting to have the OA and Safe Landing running at the same time such that we are sure to never crash into things when the Safe Landing is doing the spiral. However this is not yet supported.

> You mentioned in the dev summit that you send null points when doing calculations that PX4 is expected to ignore (but use to confirm that companion is alive and not enter failsafe). How are these formatted?

We sent out an HEARTBEAT mavlink message from the companion to confirm healtiness. Corresponding mavros plugin `companion_process_status`. Commander checks the heartbeat if COM_OBS_AVOID is enabled.

> Perhaps worth adding message flow for Safe landing? https://github.com/PX4/avoidance#message-flows

Sure!

en/computer_vision/safe_landing.md
jkflying: This is true for stereo cameras. They won't generate points on water that is still due to there being no distinguishing features to do the triangulation with. And water that is moving enough to have nice features won't be flat enough for the safe landing planner.

jkflying: Radar or ultrasound might get returns on still water though. LIDAR shouldn't.

ndepal: Is it actually true that no features will prevent landing? Or will it cause a descent to find features, effectively resulting in a water landing?

mrivi: No features will prevent landing

hamishwillee: @mrivi Just FMI (further to @ndepal 's point), what I think will happen there is that the vehicle will descend to `loiter_height`. It won't detect any features with a stereo camera, so it will decide that it doesn't have enough information to land. Presumably at this point it will start doing its spiral?
Does it send any setpoints to the vehicle at this point - or just heartbeats?

hamishwillee: @jkflying Thanks. I have moved the clarification on water landing into a "clarifications section".

hamishwillee:
@mrivi Thanks! One clarification:

>> Can you run multiple planners at the same time? ie It isn't clear when scanning the avoidance repo what you do to run landing, collision prevention, and obstacle avoidance at the same time.

> No, you can't at the moment. Collision Prevention can't overlap with the OA and SafeLanding because it only runs in Position Control mode while the other two run in Auto modes.
It would be interesting to have the OA and Safe Landing running at the same time such that we are sure to never crash into things when the Safe Landing is doing the spiral. However this is not yet supported.

The case I was actually thinking of is that someone is flying a mission in OA. They then finish the mission and go to position mode and want to avoid collisions. Then they decide they want to land safely. From what you are saying above, they can only run one planner on the companion at a time? So you can't have one vehicle that supports all of these vision options?

If so that sounds like a more critical omission than not supporting OA and safe landing in the same operation.

hamishwillee:
@mrivi More questions

>> How does this differ from obstacle avoidance. Ie what gets sent from PX4 when you're landing? Presumably not 3 points - last, current, next waypoint.

> The interface is exactly the same as for OA. Firmware sends type adapted current waypoint from FlightTaskAutoMapper, current waypoint as it is in mission item, next waypoint as it is in mission item

1. What if the land item is the last item in the mission - what should PX4 put in the "next waypoint" (or more generally, what should go in next waypoint slot for the last waypoint?

2. What about when landing in land mode. In this case we aren't in a mission, so PX4 can't really specify last, current, or next waypoint. Does it send anything? Or does the companion just monitor the land mode and start sending the sentpoints back?

hamishwillee:
FYI ... on a separate note... As standalone docs I think what we've done is OK, but I might need to play with this structure. 

The problem is that when we did the object avoidance in missions I was thinking of the integration using the trajectory interface as "just for missions". Now we're looking at it more as a generic interface for PX4 integration. So the obvious thing is to pull this out a bit and make it feel more "reusable". Depends a bit on how landing on land mode is done. 

For now the minimal changes I will do are:
1. Generic MAVLink update to capture further expected usage:  https://mavlink.io/en/services/trajectory.html
   - Add note that autopilot-side implementations should use HEARTBEAT to monitor that external interface is alive.
   - Add note that the companion side can provide mode-specific behaviour by monitoring autopilot state - ie modes, armed state etc
2. Same think for the Obstacle avoidance docs: http://localhost:4000/en/computer_vision/obstacle_avoidance.html#mission_avoidance_interface

But I may need to pull out docs on the interfaces used in computer vision into a dev guide page (or a common parent "implementation" page in the user guide). Not sure yet.

en/computer_vision/obstacle_avoidance.md
```diff
@@ -56,11 +54,41 @@ The difference when avoidance is active are:
 If PX4 stops receiving setpoint updates then obstacle avoidance will be disabled, and the mission continues under normal PX4 [Mission mode](../flight_modes/mission.md) control. 
 
 
-###  Mission Mode Avoidance Interface {#mission_avoidance_interface}
+### Path Planning Interface {#path_planning_interface}
```
hamishwillee: FYI, this is a restructure, but not quite finished. Basically what the companion sends to PX4is always the same, so I put that first. However i THINK that PX4 probably sends something different in mission mode or land mode outside of missions, because the desired path is different in each case. So that is now second, and I'll expand that out IFF the assumption is correct (and if not will revert)

en/computer_vision/obstacle_avoidance.md
```diff
@@ -71,43 +99,26 @@ The fields are set as shown:
   - `acc_x[0]`, `acc_y[0]`, `acc_z[0]`: x-y-z NED acceleration setpoint
   - `pos_yaw[0]`: Current yaw angle
   - `vel_yaw[0]`: NaN
+  - `command[0]`: NaN
```
hamishwillee: @mrivi Is the command 0 for the current position in missions an NaN? Are the other two correct?

en/computer_vision/obstacle_avoidance.md
```diff
@@ -160,6 +171,12 @@ The array waypoints contains all NAN expect for index 0:
 The setpoints are tracked by the multicopter position controller.
 
 
+## Testing
```
hamishwillee: FYI moved this to end. I suspect this is no longer true - ie we use a different platform. Do we have any detailed instructions for what people can do to duplicate our hardware setup?

en/computer_vision/safe_landing.md
```diff
@@ -0,0 +1,59 @@
...
+The *Safe Landing* feature ensures that vehicles only land on flat terrain.
+
+The feature can be enabled in both [Land mode](../flight_modes/land.md) and [Mission mode](../flight_modes/mission.md) on multicopter vehicles (and VTOL vehicles in MC mode) that have a companion computer running the appropriate vision software.
+
+If commanded to land, the vehicle first descends to a height where it can measure the surface (companion computer `loiter_height` parameter).
+If the landing area is not sufficiently flat, the vehicle moves outwards in a square-spiral pattern, periodically stopping to re-check the terrain for a landing spot that isn't too rough.
+
+{% youtube %}https://youtu.be/9SuJYcT0Mgc{% endyoutube %}
+
+## Limitations/Capabilities
```
hamishwillee: FYI, note, these are now in own section so we can expand out more easily

en/computer_vision/safe_landing.md
mrivi: Yes, that's correct. In order to do the spiral we send setpoints to the FCU.

mrivi:
> The case I was actually thinking of is that someone is flying a mission in OA. They then finish the mission and go to position mode and want to avoid collisions. Then they decide they want to land safely. From what you are saying above, they can only run one planner on the companion at a time? So you can't have one vehicle that supports all of these vision options?

Yes, only one planner can run on the companion. You could have CollisionPrevention with distance sensors once https://github.com/PX4/Firmware/pull/12179 is in and SafeLanding on the companion side. 


>    What if the land item is the last item in the mission - what should PX4 put in the "next waypoint" (or more generally, what should go in next waypoint slot when there is no next waypoint (because you're on the last one).

If the next triplet is not valid, it gets set to the current triplet. This is a general to FlighTaskAuto it isn't specific to the OA interface

>   What about when landing in land mode. In this case we aren't in a mission, so PX4 can't really specify last, current, or next waypoint. Does it send anything? Or does the companion just monitor the land mode and start sending the sentpoints back?`

Auto Land goes trough FlightTaskAuto as well so the interface is filled the same as for a mission. Current triplet is where we want to land. Next triplet is set to not valid by navigator and FlightTaskAuto sets it equal to the current triplet.
The companion knows that the vehicle is in AutoLand by monitoring the navigation mode, while for a land waypoint in a mission we look for the type associated with a point.

> Is there any impact if the vehicle is not configured to disarm on landing? Presumably it will just sit there getting setpoints. If it decides to take off then it would ignore setpoints from the companion anyway.

Correct

en/computer_vision/obstacle_avoidance.md
mrivi: This changed. We are sending the type adapted current triplet in Point0. This is the information we get out of FlightTaskAutoMapper.
FlightTaskAuto gets the triplets from navigator and reprojects them in local coordinates. FlightTaskAutoMapper takes into consideration the type of waypoint e.g. Land, Loiter, Takeoff... and sets velocity components if needed. In the case of a Land waypoint, the z component of the position is set to NAN and the z component of the velocity is set. In this way, xy is position controlled and z is velocity controlled.

en/computer_vision/obstacle_avoidance.md
mrivi: We have a list of tested companion computers and cameras in the Avoidance repository readme

hamishwillee: Thanks. I'll just link then.

en/computer_vision/obstacle_avoidance.md
hamishwillee: I think I'm going to need a bit more information to understand this. Point0 is the current information about the vehicle right? 
Why is it not useful to send the z component of the position? I can see why you might say "I want you to velocity control the target point" but I would have thought that would be something in point1.

mrivi: No, it isn't anymore. It's the current waypoint in local coordinated changed by FLightTaskAutoMapper based on the type of the waypoint.

hamishwillee: Then what are points 1 and 2?

hamishwillee:
@mrivi Thanks for all your help! I have now simplified this quite a lot. Since obstacle avoidance and safe landing use the same interfaces I have made those pages a lot simpler - mostly about how to enable the features and how they differ from "just running a mission" or "just landing".

They now link through to a path planning page (peer level) that explains the integration layer. I haven't tried to do "all" the detail - just the PX4 interfaces, and an outline of where you go on PX4/avoidance for the planner side information. 

Anyway, there are some questions inline, but I think it is ready for your review. You're looking at a few things:
- is this enough for someone to use the feature
- is this enough for someone to be able to work out how to write their own vision system (with some digging possibly).
- is it all correct/factually true
- any glaring omissions.

en/computer_vision/path_planning_interface.md
```diff
@@ -0,0 +1,90 @@
...
+- Current vehicle information:
+  - `pos_x[0]`, `pos_y[0]`, `pos_z[0]`: x-y-z NED vehicle local position setpoint
+  - `vel_x[0]`, `vel_y[0]`, `vel_z[0]`: x-y-z NED velocity setpoint
+  - `acc_x[0]`, `acc_y[0]`, `acc_z[0]`: NaN
+  - `pos_yaw[0]`: Yaw angle setpoint
+  - `vel_yaw[0]`: Yaw speed setpoint
+  - `command[0]`: NaN.
+- All other indices/fields are set as NaN.
+
+PX4 will stream trajectory messages ... TBD.
```
hamishwillee: @mrivi The main remaining TBD is how the systems handle fail cases.
1. If I enable COM_OBS_AVOID and go into mission mode, what happens if I get no heartbeat from companion? What if I have heartbeat and setpoints messages, but then they drop out?
   - I assume that in either case there will be some wait period, and then a failsafe. I further assume that the failsafe for land would be to go to normal Land mode, and failsafe for obstacle avoidance would be to drop to normal missions. Is this the case? If so, is this immediate or is there a wait period?
   - What about if you're in land mode on PX4 but the companion is set up for obstacle avoidance. You'll have a live HEARTBEAT and be sending messages to the companion. The companion will know that it isn't able to support land mode and will not send messages back. What will PX4 do?

en/computer_vision/obstacle_avoidance.md
```diff
@@ -2,13 +2,19 @@
 
 *Obstacle Avoidance* enables a vehicle to navigate around obstacles when following a preplanned path.
 
-The feature requires a companion computer that is running computer vision software. 
+The feature requires a companion computer that is running computer vision software.
```
hamishwillee: Do you have a prefered video you'd like at top of this doc?

en/computer_vision/path_planning_interface.md
jkflying: If the planner stops sending HEARTBEAT messages while in an AUTO mode the vehicle will switch to LOITER. It will also switch to LOITER if the trajectory messages stop being sent.

I don't believe there is much more in way of edge case handling.

mrivi: If the HEARTBEAT stops coming in, a message will show up in QGC say "Avoidance system lost" or "Avoidance system timeout" dipending on the MAV_STATE. However no failsafe action is taken.

If the trajectory message is older than 0.5 seconds and the vehicle is in auto with COM_OBS_AVOID=1, the vehicle will switch into Loiter.

en/computer_vision/path_planning_interface.md
```diff
@@ -0,0 +1,99 @@
...
+  - `command[0]`: NaN.
+- All other indices/fields are set as NaN.
+
+
+
+## Companion Failure Handling
+
+PX4 safely handles the case where messages are not received from the offboard system:
+- If the `HEARTBEAT` is lost, PX4 will emit a status message (which is displayed in QGroundControl) stating either "Avoidance system lost" or "Avoidance system timeout" (depending on the vehicle state).
+- When external path planning is enabled (`COM_OBS_AVOID=1`) and the vehicle is in an autonomous mode (Hold, Return, Mission, Takeoff, Land) the vehicle will switch into [Hold mode](../flight_modes/hold.md) a trajectory message is not received for more than 0.5 seconds.
```
hamishwillee: OK, @mrivi Info on failure from https://github.com/PX4/px4_user_guide/pull/510/#discussion_r301443267 now added to this section.

1. This text slightly different than what you said - basically you talked about messages being old, but I have made it like "if message not received".
2. The assumption therefore is that the planner will always be able to calculate the next point within .5 second (so there is never a gap). Is this reasonable? Or does the planner send "old points" if it doesn't have new ones yet to ensure that it is always sending something?
3. So what if you enable avoidance but no planner is running? Does the same thing happen - just drop into hold immediately?
4. What if you have safe landing planner running, but you go into a mission. PX4 expects setpoints because it is in a mission but the safe landing planner isn't going to provide them until you get to a land mission item.  Won't this push you into hold mode?

en/computer_vision/path_planning_interface.md
```diff
@@ -0,0 +1,99 @@
...
+
+Path planning is activated in PX4 by [setting](../advanced_config/parameters.md) the [COM_OBS_AVOID](../advanced_config/parameter_reference.md#COM_OBS_AVOID) to 1.
+
+## Companion Computer Setup
+
+Companion-side hardware setup and hardware/software configuration is provided in the [PX4/avoidance](https://github.com/PX4/avoidance#obstacle-detection-and-avoidance) Github repo.
+
+The actual setup/configuration required depends on the planner being used.
+
+> **Warning** Only one planner can run on the companion computer at a time (at time of writing).
```
hamishwillee: Is this correct?ie you can only run one planner, and therefore get whatever features it offers.

Is there a plan to work out how to support multiple planners or some other approach to support all the features at the same time? It is a bit of a marketing fail to say we support obstacle avoidance or safe landing as opposed to saying we support "and".

en/computer_vision/path_planning_interface.md
jkflying: 2. On the planner side we use a lightweight algorithm to give just the next point on the expensively-computed path, this should always be available in 0.5s.
3. Only if you switch to an AUTO mode
4. The Safe Landing system gives "I don't care what you do" setpoints (NaN setpoints), which lets the internal FCU setpoints continue with the mission undisturbed.

en/computer_vision/path_planning_interface.md
jkflying: At the moment it is 'or', as the safe landing planner is brand new. We'll work on getting them running together, but it isn't top priority.

en/computer_vision/path_planning_interface.md
mrivi: > So what if you enable avoidance but no planner is running? Does the same thing happen - just drop into hold immediately?

If COM_OBS_AVOID is enabled from boot and no planner is running, the preflight checks will fails no matter in which mode the vehicle is. So you won't be able to fly until COM_OBS_AVOID is set to 0

en/computer_vision/obstacle_avoidance.md
```diff
@@ -18,146 +28,59 @@ The desired route comes from a [ROS](http://dev.px4.io/en/ros/) node running on
 This is passed into an obstacle avoidance module (another ROS node).
 The avoidance software sends the planned path to the flight stack as a stream of `SET_POSITION_TARGET_LOCAL_NED` messages.
 
-> **Note** The only required PX4-side setup is to put PX4 into *Offboard mode*.
-  While the `SET_POSITION_TARGET_LOCAL_NED` setpoints come from a ROS collision avoidance node, to PX4 could be from any MAVLink system.
+While the `SET_POSITION_TARGET_LOCAL_NED` setpoints come from a ROS collision avoidance node, to PX4 they could be from any MAVLink system.
```
mrivi: I am not sure I understand what this means. 
Also collision avoidance is confusing since we either have collision prevention or obstacle avoidance in the PX4 jargon.

en/computer_vision/obstacle_avoidance.md
```diff
@@ -18,146 +28,59 @@ The desired route comes from a [ROS](http://dev.px4.io/en/ros/) node running on
...
+- If a waypoint is *inside* an obstacle it may unreachable (and the mission will be stuck).
   - If the vehicle projection on the line previous-current waypoint passes the current waypoint, the acceptance radius is enlarged such that the current waypoint is set as reached
   - If the vehicle within the x-y acceptance radius, the altitude acceptance is modified such that the mission progresses (even if it is not in the altitude acceptance radius).
 - The original mission speed (as set in *QGroundControl*/PX4) is ignored.
   The speed will be determined by the avoidance software:
   - *local planner* mission speed is around 3 m/s.
   - *global planner* mission speed is around 1-1.5 m/s.
 
-If PX4 stops receiving setpoint updates then obstacle avoidance will be disabled, and the mission continues under normal PX4 [Mission mode](../flight_modes/mission.md) control. 
+If PX4 stops receiving setpoint updates then obstacle avoidance will be disabled, and the mission continues under normal PX4 [Mission mode](../flight_modes/mission.md) control.
```
mrivi: Now we have proper failsafe. The vehicle will go into Loiter if the setpoints are older than half a second

mrivi: I see that it's covered below. Maybe we can remove this sentence

en/computer_vision/obstacle_avoidance.md
hamishwillee: What this was supposed to say  is that PX4 does not care about what generates the offboard messages - it just happens to be the obstacle avoidance ROS software. This isn't particularly useful to tell people, so I have deleted it.

Yes, thanks, fixed the terminology!

en/computer_vision/obstacle_avoidance.md
hamishwillee: Thanks. I corrected but kept the sentence. It is a level of duplication, but I think users will appreciate knowing about this at this level rather than hidden in the other doc.

hamishwillee:
Thanks @jkflying and @mrivi . Your points now integrated. I think I capture everything, and I have no more questions. Can you please carefully review as per https://github.com/PX4/px4_user_guide/pull/510#issuecomment-509531272.

hamishwillee:
Thanks you two! Much appreciated.

PR #516, opened 2019-07-14
Add estimation video link from PX4 Dev conf.

hamishwillee:
Hi Paul, this just adds your video from the Dev conference for those who are interested. It is added twice - the link up the top might disappear over as it is no longer topical, but the link at the bottom would remain for people who prefer getting their overview from video.


hamishwillee:
I am self merging as I see no harm in this.

priseborough:
Fine with me.

PR #519, opened 2019-07-15
Highlight dual GPS setup

hamishwillee:
Makes this easier to find by putting it in the GPS configuration section. I cross link from the serial port config section to this doc.


LorenzMeier:
Hamish, please start a project to cover all the „commercial“ needs in documentation. Coming from the user perspective we need to document what you need to do for a certain use case.

PR #528, opened 2019-07-18
Safety: kill switch disarms after engaged for 5 seconds

MaEtUgR:
Here's the part of the code that does this: https://github.com/PX4/Firmware/blob/aaad71faab6bdb507bb9525f1c9d64edc33632bc/src/modules/commander/Commander.cpp#L1648-L1653

Related:
- Introduced in https://github.com/PX4/Firmware/pull/10175
- Discussed the documentation is off in https://github.com/PX4/Firmware/pull/12346#issuecomment-512592191 Thanks @hamishwillee !

### DIFF
--- a/en/config/safety.md
+++ b/en/config/safety.md
@@ -262,8 +262,8 @@ This section lists the available safety switches.
 
 ### Kill Switch {#kill_switch}
 
-A kill switch immediately stops all motor outputs (if flying the vehicle will start to fall).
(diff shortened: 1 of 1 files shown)


hamishwillee:
Thanks @MaEtUgR - appreciate that. This was clear, but I still tidied it a little to my taste :-0 : https://github.com/PX4/px4_user_guide/commit/ee7afcf20c637f819cf51254fc01d553844fc446

PR #532, opened 2019-07-19
MC position control - Add documentation for trajectory generators

bresch:
@hamishwillee I started a minimal documentation for the jerk-limited trajectory generator. Since the logic is a bit different from the previous one, I added two pages and moved the previous documentation (still relevant) into one of those.

@hamishwillee Thanks a lot for the review and the typo fixes and formatting! I'll try to update the PR in the next few days.

* [ ] Link to this from the MC position mode and mission mode docs (as post process)


en/config_mc/position_mode_smooth.md
```diff
@@ -0,0 +1,91 @@
+# Smooth and Sport Modes
+
```
hamishwillee: 1. "Modes" feels like an overloaded term, when we are in position mode or some other mode already.  Is there some other acceptable way to name these? My preference would be to refer to these as configurations - ie Smooth and Sport Configurations. Other options - trajectory, path, options, model, scheme, config. 
2. It would be good here to have a paragraph or sentence explaining what smooth and sports modes "deliver" - perhaps by contrasting to Jerk limited modes. What differentiates smooth from sport?

en/config_mc/position_mode_smooth.md
```diff
@@ -0,0 +1,91 @@
...
+These two parameters only have effect during the transition from **velocity-control** to **position-control**.
+The purpose of these two parameters are to minimize the jerk introduced from forward flight to hover (please see [MPC_ACC_HOR and MPC_DEC_HOR_SLOW](#mpc_acc_hor-and-mpc_dec_hor_slow)).
+
+The jerk-parameter controls the rate limit with which the acceleration limit can change to `MPC_ACC_HOR_MAX`.
+The actual jerk-value is a linear map from velocity speed to jerk where full speed maps to [MPC_JERK_MAX](../advanced_config/parameter_reference.md#MPC_JERK_MAX) and zero speed to [MPC_JERK_MIN](../advanced_config/parameter_reference.md#MPC_JERK_MIN).
+The smoothing can be turned off by setting `MPC_JERK_MAX` to a value smaller than `MPC_JERK_MIN`.
+
+## Mission Mode
+
+In [Mission mode](../flight_modes/mission.md) the vehicle always follows a straight line from the previous waypoint to the current target.
```
hamishwillee: I guess this will remain still technically "true" even when we add a velocity control curve at the waypoint.

en/config_mc/position_mode_smooth_vel.md
```diff
@@ -0,0 +1,65 @@
+# Jerk-limited Modes
+
```
hamishwillee: As above "Jerk-limited Configuration"?

en/config_mc/position_mode_smooth_vel.md
```diff
@@ -0,0 +1,65 @@
...
+This type of setpoints generator is active in:
```
hamishwillee: As above, would be good to have a paragraph explaining the broad aims of the config, contrasted against sports/smooth (I am assuming these are mutually exclusive).

en/config_mc/position_mode_smooth_vel.md
```diff
@@ -0,0 +1,65 @@
...
+- [Position mode](../flight_modes/position_mc.md) when [MPC_POS_MODE=3](../advanced_config/parameter_reference.md#MPC_POS_MODE) (_Smooth position control (Velocity)_).
+- Auto modes when [MPC_AUTO_MODE=1](../advanced_config/parameter_reference.md#MPC_AUTO_MODE) (_Jerk-limited trajectory_).
```
hamishwillee: You say auto modes - do you mean just "Mission mode" or will this also affect landing, takeoff, return modes?

hamishwillee:
PS Should we link to this from the position mode and mission mode docs? http://localhost:4000/en/flight_modes/position_mc.html

en/config_mc/position_mode_smooth.md
bresch: What about "Slew-rate type" (instead of smmoth and sport) and "Jerk-limited type" ? I had smooth and sport in the same file because the implementation is actually the same and "sport" just uses a few different parameters; but isn't really used ...

bresch:
@hamishwillee Thanks a lot for the review and the typo fixes and formatting! I'll try to update the PR in the next few days.

> PS Should we link to this from the position mode and mission mode docs? http://localhost:4000/en/flight_modes/position_mc.html

Yes, good idea!

en/config_mc/position_mode_smooth.md
hamishwillee: "Type" works well too - I'm not sure it is better than "config" but it is certainly no worse (and lots better than mode.

Will leave these fixes to you!

hamishwillee:
Great. I'll leave with you then. Let me know when you are ready for re-review.

FYI, I added a checkbox to remind me to do the crosslink. I won't do it just yet because there are some other changes in those files going in, and putting it off will save me some merging effort later.

hamishwillee:
Hi @bresch Any chance you could do the further updates so we can get this in?

bresch:
@hamishwillee I already made a few changes locally, I'll try to finish it today :)

hamishwillee:
@bresch I'm guessing you "nearly made it". Today?

bresch:
@hamishwillee I have something! :D

hamishwillee:
@bresch Thanks for that. I have gone through and made some more mods:
1. Added more complete introductions to the types. These do duplicate parent information, but you can't know how people will find these page - perhaps by search.
2. Renamed the docs - they are not just position mode as names inferred
3. Made it clear in docs intros that types are not just for smoothing stick input - they apply to both positions and missions.


I think this is good, bar two possible caveat
- If you were a random user, would you expect to find information about tuning the response to stick and mission changes under "Trajectory"? Is that what you might search for? Would you look at Trajectory Tuning and think "Hey, that's about choosing a smooth or jerky response to user input or mission changes"? Is there a more obvious option? E.g. "Setpoint Response Tuning (Trajectory Generator)" or "Input Response Tuning (Trajectory Generator)"
- What is the default mode, and does it vary by frame? We should probably say that.

bresch:
@hamishwillee Thanks for the review! We decided that "Setpoint tuning (Trajectory Generator)" was a good title

hamishwillee:
Hi @bresch Awesome. I have merged this :-)

There was one open question "What is the default mode, and does it vary by frame?"

bresch:
Hmm, I forgot to mention that. The default is the jerk-limited type for both manual and auto modes; it's the same for all the frames.

hamishwillee:
Great thanks. Added that note inhttps://github.com/PX4/px4_user_guide/commit/92502b2bc4a52167aa059bccad1f3d4470d87146 and https://github.com/PX4/px4_user_guide/commit/94867a0d20c5e965185d6904a88587a49a8ac2d5

PR #533, opened 2019-07-23
CUAV v5+ and CUAV5nano wiring clarifications

hamishwillee:


en/assembly/quick_start_cuav_v5_nano.md
```diff
@@ -14,8 +14,8 @@ We'll go through each of these in detail in the following sections.
 
 | Main interface | Function |
 | :--- | :--- |
-| POWER1 | Connect Power module;  power input & AD voltage  and current detection. |
-| POWER2 | Connect i2c smart battery. |
+| Power | Connect Power module;  power input & AD voltage  and current detection. |
```
hamishwillee: @cuhome - I changed this to `Power` - that is what is on the device right?

en/assembly/quick_start_cuav_v5_nano.md
```diff
@@ -14,8 +14,8 @@ We'll go through each of these in detail in the following sections.
...
+| PM2 | Connect i2c smart battery. |
```
hamishwillee: @cuhome - I changed this to `PM2` - that is the label on the device right?

Is there a particular smart battery you are recommending? Is there any advice for this power connection?

en/assembly/quick_start_cuav_v5_nano.md
```diff
@@ -14,8 +14,8 @@ We'll go through each of these in detail in the following sections.
...
 | TF CARD |Insert SD card for log storage|
```
hamishwillee: @cuhome - Do they need to insert the card on the nano, or is this pre-inserted in the factory as stated below (on nano)?

en/assembly/quick_start_cuav_v5_nano.md
```diff
@@ -49,12 +49,13 @@ Connect to the flight control GPS interface using a cable.
...
 The dedicated safety switch that comes with the V5+ is only required if you are not using the recommended Neo GPS (which has an inbuilt safety switch).
-If you are flying without the GPS you must attach the switch directly to the  GPS1  port in order to be able to arm the vehicle and fly (If you use the old 6-pin GPS, please read the definition of the bottom interface to change the line.).
+
+If you are flying without the GPS you must attach the switch directly to the GPS1 port in order to be able to arm the vehicle and fly (If you use the old 6-pin GPS, please read the definition of the bottom interface to change the line).
 
 
 ## Buzzer
 
-If you do not use the recommended GPS, there may be problems with the buzzer not working.
+If you do not use the recommended GPS the buzzer may not work.
```
hamishwillee: Is there any information on how/where you should attach an external buzzer if not using the recommended GPS?

en/assembly/quick_start_cuav_v5_nano.md
```diff
@@ -71,13 +72,13 @@ The figure below shows how you can access your remote receiver (please find the
 ## Spektrum Satellite Receivers
 
 The V5 nano has a dedicated DSM cable.
-The Spektrum satellite receiver should be connected to the flight control DSM/SBUS/RSSI interface.
+The Spektrum satellite receiver should be connected to the flight control DSM/SBUS/RSSI interface.
```
hamishwillee: @cuhome - Do you mean "the Spektrum" receiver (ie a particular one supplied) or "a Spektrum" receiver (i.e. if you are using a spektrum receiver, this is where it would be attached)?

en/assembly/quick_start_cuav_v5_nano.md
```diff
@@ -71,13 +72,13 @@ The figure below shows how you can access your remote receiver (please find the
...
 The V5 nano has a dedicated DSM cable.
-The Spektrum satellite receiver should be connected to the flight control DSM/SBUS/RSSI interface.
+The Spektrum satellite receiver should be connected to the flight control DSM/SBUS/RSSI interface. 
 
 
 ## Power
 
 The v5 nano kit includes HV\_PM, which supports 2~10S LiPo battery.
-Please connect the 6pin connector of the HW\_PM module to the flight control `Power1` interface.
+Please connect the 6pin connector of the HW\_PM module to the flight control `Power` interface.
```
hamishwillee: @cuhome 
- I changed this to `Power` - that is what is on the device right?
- This shows how to power the Flight controller right? Are we assuming that people will not use this with anything that needs to power the servo rail (e.g. fixed wing vehicles)? If not, we should explain how to do this. Does the power module provide a BEC or similar to do this?
- The image below shows the outputs, for which we do NOT describe the power. Any particular reason for this choice?
- Would be great if we could have a note about the use of PM2 here - does not need to be detailed, but we should note the additional power system, and link to more information?

en/assembly/quick_start_cuav_v5_nano.md
```diff
@@ -87,7 +88,7 @@ Please connect the 6pin connector of the HW\_PM module to the flight control `Po
 A telemetry system allows you to communicate with, monitor, and control a vehicle in flight from a ground station (for example, you can direct the UAV to a particular position, or upload a new mission).
 
 The communication channel is via Telemetry Radios.
-The vehicle-based radio should be connected to the **TELEM1/TELEM2** port (if connected to this port, no further configuration is required).
+The vehicle-based radio should be connected to the **TELEM1** or **TELEM2** port (if connected to these ports, no further configuration is required).
```
hamishwillee: @cuhome - FYI only I split this out because otherwise it reads like we have just one port for both telem1 and telem2

en/assembly/quick_start_cuav_v5_plus.md
```diff
@@ -66,11 +67,13 @@ The figure below shows how you can access your remote receiver (please find the
 
 ## Spektrum Satellite Receivers
 
-The V5+ has a dedicated DSM cable. The Spektrum satellite receiver should be connected to the flight control DSM/SBUS/RSSI interface.
+The V5+ has a dedicated DSM cable.
+The Spektrum satellite receiver should be connected to the flight control DSM/SBUS/RSSI interface.
```
hamishwillee: @cuhome - As above, need to confirm you mean "A spektrum ..." rather than "The spektrum"

en/assembly/quick_start_cuav_v5_plus.md
```diff
@@ -66,11 +67,13 @@ The figure below shows how you can access your remote receiver (please find the
...
 
 ## Power
 
-The V5+ kit includes HV\_PM, which supports 2~10s LiPo battery. Please connect the 6pin connector of the HW\_PM module to the flight control power1 interface.
+The V5+ kit includes the HV\_PM, which supports 2~10s LiPo battery.
```
hamishwillee: @cuhome - As above, would be good to outline how to power the servo rail. Also a note on how to use Power2.

en/assembly/quick_start_cuav_v5_nano.md
cuhome: No, can't be modified like this, power2 is the i2c interface, it is not an adc interface.

cuhome: It is power1 and power2 on the silkscreen, so the modification may not know which power interface is.

en/assembly/quick_start_cuav_v5_nano.md
cuhome: This modification is very good.

cuhome: They don't need to insert the card into the nano because the factory has already inserted the card into the nano.

en/assembly/quick_start_cuav_v5_nano.md
cuhome: This can be a bit difficult.

en/assembly/quick_start_cuav_v5_nano.md
cuhome: The HV_PM module supplies power to the flight controller, but it does not power the servos, and the v5+/v5 nano servo rails cannot provide power to the flight controller. Power2 powered, I may need a separate description.

en/assembly/quick_start_cuav_v5_nano.md
cuhome: ok

en/assembly/quick_start_cuav_v5_plus.md
cuhome: A spektrum

en/assembly/quick_start_cuav_v5_plus.md
cuhome: I will describe how to modify it to make it similar to power1 with adc detection.

cuhome: Need independent bec to power the servo rail

en/assembly/quick_start_cuav_v5_nano.md
hamishwillee: @cuhome I'm confused. This is for the nano doc, which does not have power 1 and power 2 as silkscreen, but instead `Power` and (I think) `PM2` on the side?

en/assembly/quick_start_cuav_v5_nano.md
hamishwillee: Yes. So what do you advise? 

It would be better to point them to guidance on how to do this, but if there is none, how about we just remove the text "You may need to add an external buzzer instead."?

en/assembly/quick_start_cuav_v5_nano.md
hamishwillee: I think you do.  These docs are an OK example of how you might approach this: http://docs.px4.io/master/en/assembly/quick_start_pixhawk4.html#power

en/assembly/quick_start_cuav_v5_nano.md
hamishwillee: @cuhome Not as far as I can see on the Nano. Are you sure? Point is, the label should match what they see on the device.

hamishwillee:
Merging as-is. More work needs to be done though.

PR #536, opened 2019-07-24
COM_RC_OVERRIDE enabled by default

hamishwillee:
This documents the changes from https://github.com/PX4/Firmware/pull/12495

@MaEtUgR Can you please review.

Note, the assumption here is that follow me, orbit and offboard are not affected by this setting. Note that offboard is changed in a different PR.


hamishwillee:
@MaEtUgR I have merged this, because I think it is correct, and because other changes to the page are making this hard to keep up to date. 

Would still appreciate a sanity check though.

PR #537, opened 2019-07-24
Collision prevention using FC-distance sensor

hamishwillee:
Explains how to enable the FC-attached distance sensor for collision prevention, as per https://github.com/PX4/Firmware/pull/12179

### DIFF
--- a/en/computer_vision/collision_prevention.md
+++ b/en/computer_vision/collision_prevention.md
@@ -2,7 +2,7 @@
 
 *Collision Prevention* may be used to automatically slow and stop a vehicle before it can crash into an obstacle.
 
-It can be enabled for multicopter vehicles in [Position mode](../flight_modes/position_mc.md), and at time of writing requires a companion computer.
+It can be enabled for multicopter vehicles in [Position mode](../flight_modes/position_mc.md), and can use sensor data from an offboard companion computer, a rangefinder attached to the flight controller, or both (fused).
 
 > **Warning** Collision prevention may not prevent a crash if your vehicle is moving too fast!
   This feature has only been tested (at time of writing) for a vehicle moving at 4 m/s.
@@ -12,10 +12,7 @@ It can be enabled for multicopter vehicles in [Position mode](../flight_modes/po
 
 *Collision Prevention* is enabled/configured on PX4 by setting the parameter for minimum allowed distance ([MPC_COL_PREV_D](../advanced_config/parameter_reference.md#MPC_COL_PREV_D)).
 
-The feature requires obstacle information from either an external system (sent using the [OBSTACLE_DISTANCE](https://mavlink.io/en/messages/common.html#OBSTACLE_DISTANCE) message) or a [distance sensor](../sensor/rangefinders.md) connected to the flight controller.
-
-> **Note** *Collision Prevention* currently only works with a companion computer! 
-  Very soon, we hope to also enable it for distance sensors attached to the flight controller.
+The feature requires obstacle information from either an external system (sent using the MAVLink [OBSTACLE_DISTANCE](https://mavlink.io/en/messages/common.html#OBSTACLE_DISTANCE) message) or a [distance sensor](../sensor/rangefinders.md) connected to the flight controller.
 
 The vehicle starts braking as soon as it detects an obstacle.
 The velocity setpoint towards the obstacle is reduced linearly such that it is set to zero at the point when the vehicle reaches the minimum allowed distance.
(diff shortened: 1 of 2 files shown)


en/computer_vision/collision_prevention.md
```diff
@@ -2,7 +2,7 @@
 
 *Collision Prevention* may be used to automatically slow and stop a vehicle before it can crash into an obstacle.
 
-It can be enabled for multicopter vehicles in [Position mode](../flight_modes/position_mc.md), and at time of writing requires a companion computer.
+It can be enabled for multicopter vehicles in [Position mode](../flight_modes/position_mc.md), and can use sensor data from an offboard companion computer, a rangefinder attached to the flight controller, or both (fused).
```
hamishwillee: @mrivi HOw does "fusing" work? ie this could just be taking lots of messages from different orientations and passing them into avoidance code, or it could be that this does some magic if multiple sensors cover the same orientation ... or both.

en/computer_vision/collision_prevention.md
```diff
@@ -30,16 +27,38 @@ The user is notified through *QGroundControl* while *Collision Prevention* is ac
...
+If you are using a companion computer to provide obstacle information see [companion setup](#companion).
+
+
+## PX4 Distance Sensor {#rangefinder}
+
+At time of writing PX4 allows you to use the [Lanbao PSK-CM8JL65-CC5](../sensor/cm8jl65_ir_distance_sensor.md) IR distance sensor for collision prevention "out of the box", with minimal additional configuration:
+- First [attach and configure the sensor](../sensor/cm8jl65_ir_distance_sensor.md), and enable collision prevention (as described above, using `MPC_COL_PREV_D`).
+- Set the sensor orientation using [SENS_CM8JL65_R_0](../advanced_config/parameter_reference.md#SENS_CM8JL65_R_0).
+
+<!-- ROTATION_FORWARD_FACING - Does it matter what angles? - ie is collision prevention active in 3 D? -->
```
hamishwillee: @mrivi 
1. What is the field of collision prevention? is it just a single field in front of just one sensor in horizontal 2 d space, or does it cover 3d space (in front in a horizontal and vertical plane?)
2. Given that, are there orientations that make more or less sense? Reason I'm asking is that forward facing sensor makes best sense for range traveling horizontally - I'm wondering if a backwards facing one is useful, or if one at 45 degrees just uses the forward component (implying you need twice the range for the same distance of avoidance).

en/computer_vision/collision_prevention.md
```diff
@@ -30,16 +27,38 @@ The user is notified through *QGroundControl* while *Collision Prevention* is ac
...
 
-## Companion Setup
+Other sensors may be enabled, but this requires modification of driver code to set the sensor orientation and field of view.
+- Attach and configure the distance sensor on a particular port (see [sensor-specific docs](../sensor/rangefinders.md)) and enable collision prevention using `MPC_COL_PREV_D`.
+- Modify the driver to set the orientation. 
+  This should be done by mimicking the `SENS_CM8JL65_R_0` parameter (though you might also hard-code the orientation in the sensor *module.yaml* file to something like `sf0x start -d ${SERIAL_DEV} -R 25` - where 25 is equivalent to `ROTATION_DOWNWARD_FACING`).
```
hamishwillee: @mrivi This is worded as it is, because we'd really like to be able to support all sensors "out of the box", so this encourages a good copy. 
If instructions are clear enough we might be able to encourage @mcsauder to backfill this.

en/computer_vision/collision_prevention.md
mrivi: It's 360 around the vehicle. So if you have a distance sensor backwards, it will refrain the vehicle from moving backwards if there is an obstacle in that direction. 
The prevention is 2D, so you will never fly over/under stuff

en/computer_vision/collision_prevention.md
mrivi: Yes, I agree. I didn't want to add two many useless parameters. However this PR https://github.com/PX4/Firmware/pull/12567/files seems to go already in the direction we need. 
I really would like to have the rotation paramter to be common for all the distance sensor and not specific to one.

mhkabir: There is also this rangefinder, which is specifically for collision prevention, and includes per-sensor rotation parameters already: https://github.com/PX4/Firmware/pull/12147

I'm going to be testing them this week.

hamishwillee: @mrivi I guess the advantage of individual params is that the driver doesn't have to know anything else about the system config - e.g. if there are multiple sensors.

@mhkabir Mappydot looks very cool but appears to lack [module](http://dev.px4.io/master/en/middleware/modules_driver_distance_sensor.html) or [end user](https://docs.px4.io/v1.9.0/en/sensor/rangefinders.html) docs? Do you think you could help (or help me) create that when you have this tested?

Also to use a slightly different naming convention for the rotation param SENS_MPDTn_ROT vs SENS_CM8JL65_R_n (where `n`, is sensor number). Nice to be consistent (personally I like the use of ROT.

en/computer_vision/collision_prevention.md
hamishwillee: Awesome! Added a note about that: https://github.com/PX4/px4_user_guide/pull/537/commits/e9e9bee3c171b122f13d1bee9e75c49e6c875f1e

hamishwillee:
Thanks for the review @mrivi - merging. @mhkabir As per note, would love to update this with the mappydot information, but worth waiting until it is tested. Would also be good to get docs for that ....

PR #539, opened 2019-07-26
Doc other px4flow params

hamishwillee:
Specifically this adds the EKF params and the additional PX4 flow ones.

### DIFF
--- a/en/sensor/px4flow.md
+++ b/en/sensor/px4flow.md
@@ -68,12 +68,26 @@ The recommended mounting orientation (meaning: zero rotation) is defined as Y on
 ![PX4Flow align with Pixhawk](../../assets/hardware/sensors/px4flow/px4flowalignwithpixhawk.jpg)
 
 On **PX4**, the orientation should be set using the parameter [SENS_FLOW_ROT](../advanced_config/parameter_reference.md#SENS_FLOW_ROT).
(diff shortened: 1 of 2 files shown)


en/sensor/px4flow.md
```diff
@@ -68,12 +68,26 @@ The recommended mounting orientation (meaning: zero rotation) is defined as Y on
...
+> **Warning** PX4Flow emits a significant amount of electromagnetic radiation, and should be placed as far away from other electronics (in particular GPS modules) as possible (see [Hardware/issues/8](https://github.com/PX4/Hardware/issues/8) for more information).
+  
+
+## PX4 Configuration {#px4_configuration}
+
+The PX4Flow parameters that you may need to configure are listed below.
+
+Parameter | Description
+--- | ---
+<span id="SENS_FLOW_MAXHGT"></span>[SENS_FLOW_MAXHGT](../advanced_config/parameter_reference.md#SENS_FLOW_MAXHGT) | Maximum height above ground when reliant on optical flow.
```
hamishwillee: @mhkabir FYI, I put these in PX4flow doc because they appear to be part of that software not "generic" to all sensors.

hamishwillee:
> We should probably document all the EKF2 sensor offset parameters somewhere - not just for flow. This looks great though!

Thanks. Re offsets, yes we should - normally I like to doc params near the sensors that use them.

PR #542, opened 2019-08-04
Clarify auto-disarm params

hamishwillee:
Docs to match (and dependent on) https://github.com/PX4/Firmware/pull/12608 + a few typos.


hamishwillee:
@RomanBapst @bresch @catch-twenty-two FYI as https://github.com/PX4/Firmware/pull/12608 has merged, I am merging the associated docs. This is not reviewed, but I am happy that it is accurate.

PR #545, opened 2019-08-05
Clarify which setpoints supported

### LINKED ISSUE #514: Not clear OFFBOARD mode
Here https://docs.px4.io/en/flight_modes/offboard.html it's written that in OFFBOARD mode the vehicle position, velocity, or thrust (SET_POSITION_TARGET_LOCAL_NED) can be controlled.

What does the OR stands for? What if I send a SET_POSITION_TARGET_LOCAL_NED with position AND velocity (AND thrust)? Or should I only pick one to send (position or velocity or thrust)?

Can we have more info on how the setpoint is created and if it is useful or useless to send both position and velocity.

hamishwillee:
Fixes #514 

@Stifael @MaEtUgR Would appreciate your review. Comments inline.

### DIFF
--- a/en/flight_modes/offboard.md
+++ b/en/flight_modes/offboard.md
@@ -36,7 +36,7 @@ The action is defined in the parameters [COM_OBL_ACT](#COM_OBL_ACT) and [COM_OBL
     * Velocity setpoint (only)
     * *Thrust* setpoint (only)
       > **Note** Acceleration setpoint values are mapped to create a normalized thrust setpoint (i.e. acceleration setpoints are no "properly" supported).
-    * Position setpoint **and** velocity setpoint (velocity used as feedforward)
+    * Position setpoint **and** velocity setpoint (the velocity setpoint is used as feedforward; it is added to the output of the position controller and the result is used as the input to the velocity controller).
   - PX4 supports the coordinate frames (`coordinate_frame` field): [MAV_FRAME_LOCAL_NED](https://mavlink.io/en/messages/common.html#MAV_FRAME_LOCAL_NED) and [MAV_FRAME_BODY_NED](https://mavlink.io/en/messages/common.html#MAV_FRAME_BODY_NED).
 * [SET_ATTITUDE_TARGET](https://mavlink.io/en/messages/common.html#SET_ATTITUDE_TARGET) - Control vehicle attitude/orientation.
 
--- a/en/flight_modes/offboard.md
+++ b/en/flight_modes/offboard.md
@@ -30,8 +30,13 @@ The action is defined in the parameters [COM_OBL_ACT](#COM_OBL_ACT) and [COM_OBL
 
 ### Copter/VTOL
 
-* [SET_POSITION_TARGET_LOCAL_NED](https://mavlink.io/en/messages/common.html#SET_POSITION_TARGET_LOCAL_NED) - Control vehicle position, velocity, or thrust. 
-  - Acceleration setpoints are combined to create a 'thrust' setpoint.
+* [SET_POSITION_TARGET_LOCAL_NED](https://mavlink.io/en/messages/common.html#SET_POSITION_TARGET_LOCAL_NED)
+  * The following input combinations are supported: <!-- https://github.com/PX4/Firmware/blob/master/src/lib/FlightTasks/tasks/Offboard/FlightTaskOffboard.cpp#L166-L170 -->
+    * Position setpoint (only)
+    * Velocity setpoint (only)
+    * *Thrust* setpoint (only)
+      > **Note** Acceleration setpoint values are mapped to create a normalized thrust setpoint (i.e. acceleration setpoints are no "properly" supported).
+    * Position setpoint **and** velocity setpoint (velocity used as feedforward)
   - PX4 supports the coordinate frames (`coordinate_frame` field): [MAV_FRAME_LOCAL_NED](https://mavlink.io/en/messages/common.html#MAV_FRAME_LOCAL_NED) and [MAV_FRAME_BODY_NED](https://mavlink.io/en/messages/common.html#MAV_FRAME_BODY_NED).
 * [SET_ATTITUDE_TARGET](https://mavlink.io/en/messages/common.html#SET_ATTITUDE_TARGET) - Control vehicle attitude/orientation.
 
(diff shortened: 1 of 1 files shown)


en/flight_modes/offboard.md
```diff
@@ -30,16 +30,21 @@ The action is defined in the parameters [COM_OBL_ACT](#COM_OBL_ACT) and [COM_OBL
...
+    * Velocity setpoint (only)
+    * *Thrust* setpoint (only)
+      > **Note** Acceleration setpoint values are mapped to create a normalized thrust setpoint (i.e. acceleration setpoints are no "properly" supported).
+    * Position setpoint **and** velocity setpoint (velocity used as feedforward)
   - PX4 supports the coordinate frames (`coordinate_frame` field): [MAV_FRAME_LOCAL_NED](https://mavlink.io/en/messages/common.html#MAV_FRAME_LOCAL_NED) and [MAV_FRAME_BODY_NED](https://mavlink.io/en/messages/common.html#MAV_FRAME_BODY_NED).
 * [SET_ATTITUDE_TARGET](https://mavlink.io/en/messages/common.html#SET_ATTITUDE_TARGET) - Control vehicle attitude/orientation.
 
 ### Fixed Wing
 
 * [SET_ATTITUDE_TARGET](https://mavlink.io/en/messages/common.html#SET_ATTITUDE_TARGET) - Control vehicle attitude/body rates.
```
hamishwillee: @Stifael @MaEtUgR For FW this says "Control vehicle attitude/body rates.". Looking at the code it just seems to obey thrust setpoints - not position or velocity. What is the correct thing to say here?

en/flight_modes/offboard.md
```diff
@@ -30,16 +30,21 @@ The action is defined in the parameters [COM_OBL_ACT](#COM_OBL_ACT) and [COM_OBL
...
 
-* [SET_POSITION_TARGET_LOCAL_NED](https://mavlink.io/en/messages/common.html#SET_POSITION_TARGET_LOCAL_NED) - Control vehicle position, velocity, or thrust. 
-  - Acceleration setpoints are combined to create a 'thrust' setpoint.
+* [SET_POSITION_TARGET_LOCAL_NED](https://mavlink.io/en/messages/common.html#SET_POSITION_TARGET_LOCAL_NED)
+  * The following input combinations are supported: <!-- https://github.com/PX4/Firmware/blob/master/src/lib/FlightTasks/tasks/Offboard/FlightTaskOffboard.cpp#L166-L170 -->
+    * Position setpoint (only)
+    * Velocity setpoint (only)
+    * *Thrust* setpoint (only)
+      > **Note** Acceleration setpoint values are mapped to create a normalized thrust setpoint (i.e. acceleration setpoints are no "properly" supported).
+    * Position setpoint **and** velocity setpoint (velocity used as feedforward)
```
hamishwillee: What does "velocity used as feedforward" mean?

Stifael: it means that the given feedforward-velocity setpoint `v_sp` is added to the output of the position P-controller:
`P(p_sp) + v_sp = v_sp`  where `P` is a function representing a position-controller (in the case of PX4, it is just a gain of the position setpoint and the position estimation)

en/flight_modes/offboard.md
Stifael: Don't know the current state of FW offboard support.

en/flight_modes/offboard.md
tuloski: Maybe call differently v_sp at left and at the right of "=" because they are two different things ;) . The one of the left is the setpoint given by the "user". The one of the right is the output of the controller which is the reference for the velocity controller.

en/flight_modes/offboard.md
hamishwillee: @RomanBapst Can you comment on this?

en/flight_modes/offboard.md
hamishwillee: Thanks for the clarification. I changed this to 

>    Position setpoint **and** velocity setpoint (the velocity setpoint is used as feedforward; it is added to the output of the position controller and the result is used as the input to the velocity controller).

I'm going to merge now.

tuloski:
At line 38
no "properly" supported --> noT "properly" supported

hamishwillee:
Thanks @tuloski - typo fixed inhttps://github.com/PX4/px4_user_guide/commit/b13886e6bab366d7de60a63039f3953eee54748e

en/flight_modes/offboard.md
RomanBapst: @mrivi You are a regular offboard user, do you know about this?

mrivi: @RomanBapst @hamishwillee - as far as I know position setpoint support for FW was brought in with https://github.com/PX4/Firmware/pull/12532 . However I have never tried offboard with FW myself

Jaeyoung-Lim: @RomanBapst I have tested position setpoints on offboard with FW on the PR @mrivi mentioned. Also with attitude/body rates from my experience works reasonably

hamishwillee: @Jaeyoung-Lim Thanks!
1. What do you mean attitude/body rates? Dumb question? Well the message refers to position, velocity and acceleration and I want to make sure that we mean the same thing - so how to they map?
2.  If you look at the copter docs, they are very precise about what we support: http://docs.px4.io/master/en/flight_modes/offboard.html#coptervtol - Can you provide similar for FW?

Or to put it another way, it sounds like we support messages that contain a position setpoint (only), and a velocity setpoint (only). However it is not clear for FW whether we support acceleration setpoints, and whether any combinations are supported.

Jaeyoung-Lim: @hamishwillee By attitude / bodyrates, I meant the support for SET_ATTITUDE_TARGET message.  I think I will be able to add information for FW in the doc.

hamishwillee: @Jaeyoung-Lim ! That makes a lot more sense - sorry. A PR would be great - do you think you can do this next week?

Jaeyoung-Lim: @hamishwillee Sure, just to be sure this is not about https://github.com/PX4/Firmware/pull/12532 but more clarity on the SET_ATTITUDE_TARGET message? I will add a PR for https://github.com/PX4/Firmware/pull/12532 on a separate PR

hamishwillee: @Jaeyoung-Lim To be clear, it is about making our docs on [offboard mode](http://docs.px4.io/master/en/flight_modes/offboard.html#fixed-wing) accurately document what commands are supported - and _how_ they are supported if they do not support all combinations options in the message definition :-) So it is about both. But it may be that we already say all we need to about SET_ATTITUDE_TARGET.

Right now we say that:
 
>  "[SET_ATTITUDE_TARGET](https://mavlink.io/en/messages/common.html#SET_ATTITUDE_TARGET) - Control vehicle attitude/body rates.

The assumption is that this supports all options/variants as per the linked definition. If that is the case then we don't need to say more about this message. If we don't do every possible permutation of that message we should state how we differ/what is not supported.

For https://github.com/PX4/Firmware/pull/12532 ... we currently say:

> Position setpoints are not supported on Fixed Wing offboard mode (SET_POSITION_TARGET_LOCAL_NED).

So basically when https://github.com/PX4/Firmware/pull/12532 goes in we will need to update the docs to remove that note and state exactly what set of combinations of position, velocity, accel setpoints we support. 

Hope that clarifies.

PR #548, opened 2019-08-06
Fix typo in accelerometer calib docs

hamishwillee:
@bkueng Typo is fairly self explanatory. Created PR for these questions:

1. What are symptoms of a bad accelerometer calibration? Things like ... fly away, not arming, not holding position (and in what way - circling?) etc etc ..
2. Can the accelerometer calibration drift or is it a "do once, trusted forever"? If it can, how often should it be run?
3. Does it ever have to change for other reasons. Ie if I change the payload weight or position?


bkueng:
> @bkueng Typo is fairly self explanatory.

Looks correct :)

> What are symptoms of a bad accelerometer calibration? Things like ... fly away, not arming, not holding position (and in what way - circling?) etc etc ..

It is mostly caught by preflight checks and arming denied messages, such as high accelerometer bias or consistency check failures.
I think if it manages to slip through the checks, it can lead to incorrect z velocity estimates, but the checks generally prevent that in the first place.

> Can the accelerometer calibration drift or is it a "do once, trusted forever"? If it can, how often should it be run?

There's a temperature-dependency, but you need temperature calibration to account for that optimally. If you don't have one, it might be useful to re-calibrate in winter (or generally when the average ambient temperature changes). Other than that you don't have to recalibrate.

> Does it ever have to change for other reasons. Ie if I change the payload weight or position?

No, except maybe if you change the orientation relative to the vehicle.

hamishwillee:
Thanks very much. I think it is worth letting people know when they might need to to this. Can you please re-review.

hamishwillee:
Merging prior to v1.10 release. This seems OK to me based on discussion in thread.

PR #550, opened 2019-08-06
px4flow now enabled via SENS_EN_PX4FLOW parameter

dagar:
Should we mention this more prominently near the top? 

 - https://github.com/PX4/Firmware/issues/12571


hamishwillee:
> Should we mention this more prominently near the top?

Yes @dagar - done in https://github.com/PX4/px4_user_guide/pull/551. I will merge but if you could give it a quick check that would be good.

PR #551, opened 2019-08-06
PX4flow: More info about enabler parameter

hamishwillee:


hamishwillee:
Self merging. But would be good if it could be sanity checked.

PR #555, opened 2019-08-13
Explain MPC_COL_PREV_ANG

hamishwillee:
This attempts to explain MPC_COL_PREV_ANG as introduced (and discussed in) https://github.com/PX4/Firmware/pull/12561

@jkflying and @baumanta you'll recognise the information and diagrams :-)

What has changed a bit here is that I believe that the angle MPC_COL_PREV_ANG is used for all movement, not just going backwards. So its tuning affects what happens all the time. 

The other thing that I've tried to capture here is that when you go forward the distance calculated is based on the minimum distance from the whole set of included sensors. So if you're travelling along a wall and including a sensor, that is the distance, that gets included, not the empty space in front of you.

Anyway, please check I got it right. This was hard to understand and convey.


en/computer_vision/collision_prevention.md
```diff
@@ -12,25 +12,26 @@ It can be enabled for multicopter vehicles in [Position mode](../flight_modes/po
...
+The feature requires obstacle information from an external system (sent using the MAVLink [OBSTACLE_DISTANCE](https://mavlink.io/en/messages/common.html#OBSTACLE_DISTANCE) message) and/or a [distance sensor](../sensor/rangefinders.md) connected to the flight controller.
 
 > **Note** Multiple sensors can be used to get information about, and prevent collisions with, objects *around* the vehicle. 
   If multiple sources supply data for the *same* orientation, the system uses the data that reports the smallest distance to an object.
 
-The vehicle starts braking as soon as it detects an obstacle.
-The velocity setpoint towards the obstacle is reduced linearly such that it is set to zero at the point when the vehicle reaches the minimum allowed distance.
+Collision prevention determines the desired speed using the minimum distance from *the set of all included* sensors in the commanded direction (where the "included" set of sensors is tuned using [MPC_COL_PREV_ANG](#angle_tuning)).
```
baumanta: This sounds not entirely correct to me. I would go for something like :

The data from all sensors is fused into a range scan, which is the input to the collision prevention algorithm. The parameter MPC_COL_PREV_ANG determines which part of the of the range scan is used for the calculation of the velocity limit. The considered data from the range scan is centered around the commanded direction and spreads by MPC_COL_PREV_ANG to both sides.

en/computer_vision/collision_prevention.md
```diff
@@ -40,6 +41,32 @@ Enabled collision prevention by [setting the following parameter](../advanced_co
 If you are using a distance sensor attached to your flight controller for collision prevention, it will need to be [attached and configured](#rangefinder) as described in the next section.
 If you are using a companion computer to provide obstacle information see [companion setup](#companion).
 
+### MPC_COL_PREV_ANG Angle Tuning {#angle_tuning}
+
+Collision prevention calculates the vehicle speed using the minimum distance from *the set of all included sensors* in the commanded direction, where the "included set" is tuned using [MPC_COL_PREV_ANG](../advanced_config/parameter_reference.md#MPC_COL_PREV_ANG).
```
baumanta: " Collision prevention calculates the allowed vehicle speed using a subset of the range scan provided by all sensors. The size of the this subset (centered around the commanded direction) can be tuned by the parameter MPC_COL_PREV_ANG."

en/computer_vision/collision_prevention.md
```diff
@@ -40,6 +41,32 @@ Enabled collision prevention by [setting the following parameter](../advanced_co
...
+
+The angle affects both what objects will be detected in the direction that the vehicle is moving, and also the angle at which it can withdraw from an obstacle after stopping.
```
baumanta: instead of "moving" I would put "is commanded to move". It depends on the command not the actual velocity of the vehicle

en/computer_vision/collision_prevention.md
```diff
@@ -40,6 +41,32 @@ Enabled collision prevention by [setting the following parameter](../advanced_co
...
+
+Using a larger angle includes more sensors and reduces the chance of clipping obstacles.
```
baumanta: "more sensor data" . we do not discard all data form a sensor, but just a part of the range scan

en/computer_vision/collision_prevention.md
```diff
@@ -40,6 +41,32 @@ Enabled collision prevention by [setting the following parameter](../advanced_co
...
+- The blue line shows the angle where received sensor data *can be used* (and what will be discarded).
+  - This is equal to double `MPC_COL_PREV_ANG` degrees centred on the commanded setpoint.
+  - On the left diagram the `MPC_COL_PREV_D` angle is 90 degrees, so all sensors on the right hand side of the vehicle are included.
+  - On the right the angle is 45 degrees, resulting in a smaller set of sensors being used (note the sensors close to the obstacle are omitted!)
+- The red arc shows the angle where sensor data *will be used*.
+  This is the subset of sensors that are both in range of an obstacle, and within the angle defined by the blue line.
+
+![MPC_COL_PREV_ANG image](../../assets/computer_vision/collision_prevention_angle.png)
+
+Because it has such a large angle the vehicle on the left includes sensors that are within the 4m minimum range; the distance used for calculation is 4m, and the commanded setpoint is ignored.
```
baumanta: just to clarify: We do not only use the miniumum distance to calculate the velocity but all sensor data in the red arch. It is a geometric calculation of the breaking distance considering all data inside MPC_COL_PREV_ANG. But of course the core statement is correct: Because in the left case the data contains closer bins, the velocity is more limited.

en/computer_vision/collision_prevention.md
hamishwillee: Thanks for those clarifications. I was assuming a simplistic algorithm, but of course you use data from across the considered (what I have called "included") data.

hamishwillee: @baumanta THanks! The term "range scan" is a bit domain specific - not sure if there is something more mainstream or if we should define it?

I assume this means a 360 degree map around the vehicle containing all sensor input (ie with indications of distance to obstacle, out of range, or no sensor here)

What I put in here was ....

> The data from all sensors is fused into a range scan (a 360 degree map of the sensor data/state from around the vehicle), which is then input to the collision prevention algorithm.
The collision prevention algorithm uses/considers only part of the range scan data in the calculation of the velocity limit - the data within two [MPC_COL_PREV_ANG](../advanced_config/parameter_reference.md#MPC_COL_PREV_ANG) degree arcs centered around the commanded direction.
For more information see [MPC_COL_PREV_ANG Angle Tuning](#angle_tuning).

en/computer_vision/collision_prevention.md
hamishwillee: Ah, so the large angle means that we use sensor data that is within the minimum distance, but that just means it is part of the calculation. It might have a small impact because of the geometry. In this case perhaps the vehicle would still move, just slower due to those setpoints?

Assuming that is correct, I have reworded. Is this OK?

> Because it has such a large angle, the vehicle on the left includes sensor data that is within the 4m minimum range.
This data is included in the velocity calculation, and may result in the commanded setpoint being ignored (vehicle doesn't move) or obeyed at a reduced velocity. 
The vehicle on the right uses an angle that causes the collision prevention algorithm to discard the sensor data that is closest to the obstacle.
The vehicle is therefore more likely to be able to move in the commanded direction.

baumanta: Exactly! It only stops completely if there is one sensor bin indicating 4m or closer. Otherwise it will just move the slower the closer the obstacles are.

hamishwillee: @baumanta Actually my text doesn't quite say what you think it does. 

Yes it says that on left the range data includes a bin within 4m min range. BUT it also says that "This may result in the commanded setpoint being ignored (vehicle doesn't move) *or obeyed at a reduced velocity.*"

So my take here was that geometry is important too. I.e. you have a whole scan of data, just because one bucket is too close doesn't mean you necessarily can't move - the geometry of all the buckets affects velocity.

But now it sounds like I misinterpretted you - any buckets are in the minimum cutoff then all movement is dead. Is that correct. 

Presumably it is still true that the velocity depends on all the buckets, it is just that part of that calculation is that if any buckets are within the minimum velocity must be zero?

baumanta: So what actually happens is the following: If any bucket in the considered data range is below the specified limit, all movement is dead. The reason for this is, that because the maximum angle for COL_PREV_ANG is 90deg, we consider at most the hemisphere of buckets centered at the command. That means any execution of the command (however slow) will bring you even closer to the bucket which is already below the limit. Therefore, if one bucket is too close -> no movement. If they just contain other distances, the executed speed depends on the geometry, actual bin values etc.

hamishwillee:
Thanks for that last clarifications @baumanta - that made it clear. I have done a further update to tidy and capture this and merged. I think this is all good now. Thanks also @jkflying .

PR #559, opened 2019-08-18
[WIP] PMW3901 page (optical flow)

DanielePettenuzzo:
In PX4 we are now supporting several optical flow boards based on the PMW3901 sensor:
- Bitcraze and Tindie (SPI)
- Hex (UAVCAN)
- Trone (UART) --> work in progress 

The goal of this page is to compare the 4 types of boards and then to be able to use the selected one with PX4. 

This is not finished. The page still has the PX4FLOW youtube video because I'm trying to find some PMW3901 videos that I took a few months ago. The diagrams are hand made by me and can be improved.

FYI @hamishwillee @jkflying @mhkabir


DanielePettenuzzo:
TODO before merging:
- find decent youtube video
- add some sensor specifications
- update voltages and sizes in the main table
- add orientation diagrams (I need to verify these with @jkflying because I don't remember)
- add troubleshooting if we can think of anything useful

DanielePettenuzzo:
Moreover I'm not sure if the small module from TIndie is actually designed by them. We need to check this to put the right name of the manufacturer

TSC21:
@DanielePettenuzzo in terms of cabling and adaptors: you have schematics for connecting on the Pixhawk 4. My question is, from these sensors, which ones do require some specific wiring and buy out adaptor plugs to connect to the Pixhawk 4 pins? Or do they bring already the required cabling? That should be something to be documented as that would help people understand if they need to buy extra plugs and cabling to connect it to the Flight Controller.

DanielePettenuzzo:
And we recently made changes to the pixracer board config to be able to get a SPI port on the WIFI port. We should also add a small section about this because it's a very small (in terms of size) setup.

DanielePettenuzzo:
@TSC21 except for the HEX uavcan module I think the other 3 don't come with the wire and the user will have to solder a cable for it. Yes I agree that we need to specify this

hamishwillee:
Also see https://github.com/PX4/Firmware/pull/12738 and https://github.com/PX4/Firmware/pull/12477#issuecomment-521831178

en/sensor/pmw3901.md
```diff
@@ -0,0 +1,76 @@
+# PMW3901 Flow Sensor
+
+PMW3901 is an [optical flow](../sensor/optical_flow.md) asic that computes the flow internally and provides a difference in pixels between 
+each frame over a SPI interface. It is basically a tracking sensor similar to what you would find in a computer mouse, but adapted to work between
```
hamishwillee: "an SPI" (FYI, the rule in English is that the "an" is used before word that _sounds_ like it has a vowel first. So "an esssPI").

hamishwillee: People do search on component part numbers, but they just as much (or more) look for products. So it is worth highlighting the relevant products as early as possible.

How about after the sentence ending in "SPI interface" have another one listing the manufacturer parts  - "Bitcraze" then a paragraph break and following sentence.

en/sensor/pmw3901.md
```diff
@@ -0,0 +1,76 @@
...
+
+* TODO
+* TODO
+* TODO
+
+## Pixhawk Setup
+
+### Bitcraze
+
+The Bitcraze optical flow module exposes directly the SPI interface from the PMW3901 module. The board also has a distance sensor that can be wired via I2C. This distance sensor is the VL53L0x ToF sensor for STMicroelectronics. The range of the sensor is very limited (2 meters) and will be reduced when flying in the sunlight. We therefore highly recommend to use another external distance sensor. This board was mainly tested with the [Lidar Lite V3](../sensor/lidar_lite.md) on larger vehicles and with the [Lanbao CM8JL65](../sensor/cm8jl65_ir_distance_sensor.md) on smaller vehicles.
```
hamishwillee: Not essential, but consider putting each sentence on a new line. Makes things slightly easier for translators.

hamishwillee: for or "from" STMicroelectronics?

en/sensor/pmw3901.md
```diff
@@ -0,0 +1,76 @@
...
+
+!!!REPLACE WITH RIGHT VIDEO
+{% youtube %}
+https://youtu.be/0Jpq6DU_HVg
+{% endyoutube %}
+
+## Boards using PMW3901
+
+The following table shows some of the boards available on the market with the pmw3901 optical flow sensor. The number interfaces, the number of sensors, the input voltage and the size are shown. The board name contains the link to where to buy the specific board.
+
```
hamishwillee: - Should we also list the range/height at which these can be used?
- Should we add a note that for cases with out a rangefinder one must be used as well?

en/sensor/pmw3901.md
hamishwillee: What is a reasonable range/good range for a distance sensor?

en/sensor/pmw3901.md
```diff
@@ -0,0 +1,76 @@
...
+![Tindie PH4 Pinout](../../assets/hardware/sensors/pmw3901/ph4-tindie-flow-pinout.png)
+
+### Hex
+
+The Hex optical flow module is a very small board containing the PMW3901 flow module, the VL53L1X distance sensor and an IMU to better synchronize the flow data with the gyro data. These three sensors are then sampled by a microcontroller onboard. The goal of the microcontroller is to then expose 2 UAVCAN messages containing all the necessary data.
+
+This board can be connected to the CAN port on any Pixhawk board. All tests were performed connecting it to the CAN1 port on a Pixhawk 4. Differently from the other sensors, this one needs UAVCAN running on PX4. To enable UAVCAN you will have to set the UAVCAN_ENABLE parameter to 2 (if you have have UAVCAN sensors but no motor controllers) or to 3 (if you have also your motor controllers on CAN).
+
```
hamishwillee: If I use an external distance sensor, is there anything special I need to do to make sure that PX4 properly fuses the external one rather than using the inbuilt one that comes with UAVCAN message?

hamishwillee:
This is a great start at getting the information and makes it very easy to compare the different options. 

The problem for me is that this doesn't quite fit into the existing docs structure. My thinking is that generally you want to be able to look at the to level [optical flow](http://docs.px4.io/master/en/sensor/optical_flow.html) page to get a good idea of the options and any common setup, then drill down to more detail. The detail pages then link back up to the top page for the common setup. The reason that works is that it is obvious how you extend when new things come along

So what I'd kind of like to do is pull this apart. The top part of the page would go under [cameras here](http://docs.px4.io/master/en/sensor/optical_flow.html#cameras) - this would include the comparision information. Each of the different flow products would get their own pages that contains wiring disagram and specifications. 

Does that make sense?

en/sensor/pmw3901.md
```diff
@@ -0,0 +1,76 @@
...
+As for the other optical flow boards, it is recommended to use an external distance sensor instead of the onboard one due to the limited range of the VL53L1X. 
+
+![Hex PH4 Pinout](../../assets/hardware/sensors/pmw3901/ph4-hex-optical-flow.png)
+TODO
+
+Regarding rotations, also this board will have a default orientation and if necessary this could be modified through a parameter.
+
+### Thone
+WORK IN PROGRESS
+
```
jkflying: I've also tested this board, the Alientek atk-PMW3901 https://www.aliexpress.com/i/32979605707.html
Had to reverse-engineer some pinouts, only trust the SPI pinouts and V+/GND

image

jkflying:
It would be good to mention the orientation of the sensor for the SENS_FLOW_ROT parameter. On the direct-access SPI devices this has the notch in the chip housing facing backwards when SENS_FLOW_ROT = 0.

I'm not sure if the Thone or other boards change this.

image

hamishwillee:
>It would be good to mention the orientation of the sensor for the SENS_FLOW_ROT parameter. On the direct-access SPI devices this has the notch in the chip housing facing backwards when SENS_FLOW_ROT = 0.

@jkflying Can you confirm that you saying zero rotation is when the board is aligned on the vehicle with that notch facing the back? ie

image

Is this the _recommended_ rotation? (for PX4FLOW the recommendation is 270)

jkflying:
> Can you confirm that you saying zero rotation is when the board is aligned on the vehicle with that notch facing the back?

Yes

> Is this the recommended rotation? (for PX4FLOW the recommendation is 270)

I don't think it has a 'recommended' orientation. Any should work fine, as long as SENS_FLOW_ROT is correct.

hamishwillee:
@DanielePettenuzzo Can you address the comments/questions please?

en/sensor/pmw3901.md
DanielePettenuzzo: not sure anymore. I need to check this one

DanielePettenuzzo:
@hamishwillee regarding the SENS_FLOW_ROT being default to 270 for the PX4 Flow this is because with this rotation the forward facing arrow that is drawn on the module will be aligned with the orientation of the autopilot.

DanielePettenuzzo:
@hamishwillee I addressed all your questions/comments and pushed a new commit. I still didn't change the structure since I was super busy lately. Would you have time to change the structure to what you think is best and then I can go in and add all the missing parts.

tstellanova:
@DanielePettenuzzo I'm testing with the Tindie SPI board (made by "Pesky Products" https://www.tindie.com/products/onehorse/pmw3901-optical-flow-sensor/) and getting ECL rejecting the XY flow (see https://github.com/PX4/ecl/issues/646).  It appears SPI communications is working and that the sensor is oriented correctly.  Any suggestions?

hamishwillee:
@DanielePettenuzzo Thanks very much. I don't think I can alter structure while it is on your repo - can you push to *this repo* and change what the PR is comparing? Then yes, I can play with this!

en/sensor/pmw3901.md
hamishwillee: @jkflying Further down your linked page they had this:
image

Does that help you with the missing pins?

jkflying: Those are the pins on the parent board this is mounted to, and reflect the labels of the schematic/microcontroller on the other side. There are a bunch of CS lines for example, I had to trace the circuit board to figure out which one is the real one on this board.

hamishwillee: Thanks. I've sorted by just showing the SPI pins in the connection diagram in other doc. Have included your doc too "for the moment".

PR #570, opened 2019-09-03
MC PID rate tuning - Add explanation and diagrams of the parallel/ideal/mixed forms

bresch:
fixes https://github.com/PX4/Devguide/issues/847

@hamishwillee I think it makes more sense if this documentation lives in the user guide rather than in the devguide. do you agree?


hamishwillee:
@bresch Can you please look at my comments above? https://github.com/PX4/px4_user_guide/pull/570#pullrequestreview-283333627

Is this impacted by this change? https://github.com/PX4/Firmware/pull/12986

hamishwillee:
@bresch Ping? Would be cool to get this into the v1.10 library

hamishwillee:
@bresch Can we move this along?

bresch:
> The name "Ideal" is not ideal, unless this is how it is commonly referred to in the literature. Is there a more formal name?

This is a common name, but "Standard" is also widely used, I added that

> For the ideal format image, can you update the labels to match the mixed format - ie use K, I, D, rather than P, I, D`. It makes it easier to work out what is going on.

Updated

> Are I and D the same in both cases? ie in one you say "they are unitless" and in the other "I and D terms represent the integral and derivative time constants". I'm not sure what that means for the user.

They have the same effect, but the scales are not the same. I reworded a bit, it's not important to add too many details, I added references

> The comparison between the two forms does not capture (to my mind) why the parallel case is flawed and the ideal case is better. More specifically, it says "More intuitive to tune than the parallel form" but not why? I think you explained this in the original PR - something to do with the P and I being linked in the parallel case while in the ideal case you can tune them independently.
    You might be better off just having a paragraph above explaining the differences rather than try to capture that in the bullet points

Yep, I added a small paragraph
    
> What is G(s), u, y

I added that

bresch:
I'll remove the "Ideal" completely, it's not clear in the literature which form is the ideal one. In some cases it's the Standard one and in some cases the Parallel.

en/config_mc/pid_tuning_guide_multicopter.md
```diff
@@ -49,14 +51,50 @@ Here are some general points to follow when tuning:
...
+  ![PID_Standard](../../images/mc_pid_tuning/PID_algorithm_Standard.png)
+
+References:
+- https://www.controleng.com/articles/not-all-pid-controllers-are-the-same/
+- https://en.wikipedia.org/wiki/PID_controller#Standard_versus_parallel_(ideal)_PID_form
+
+#### Rate PID Tuning
+
+The related parameters for the tuning of the PID rate controllers are:
+- Roll rate control ([MC_ROLLRATE_P](../advanced_config/parameter_reference.md#MC_ROLLRATE_P), [MC_ROLLRATE_I](../advanced_config/parameter_reference.md#MC_ROLLRATE_I), [MC_ROLLRATE_D](../advanced_config/parameter_reference.md#MC_ROLLRATE_D), [MC_ROLLRATE_K](../advanced_config/parameter_reference.md#MC_ROLLRATE_K))
```
hamishwillee: @bresch
1.  I believe from all this that you EITHER tune in the standard or in the parallel form - correct? 
    - So you would start by setting `MC_ROLLRATE_P` OR `MC_ROLLRATE_K` to 1 (but not both)? (That is implied but not stated).
    - If so, then IMO the P and D params should be grouped here to make it obvious.

en/config_mc/pid_tuning_guide_multicopter.md
```diff
@@ -49,14 +51,59 @@ Here are some general points to follow when tuning:
...
+##### Parallel Form {#parallel_form}
+
+The *parallel form* is the simplest form, and is (hence) commonly used in textbooks.
+In this case the output of the controller is simply the sum of the proportional, integral and derivative actions.
+
+![PID_Parallel](../../images/mc_pid_tuning/PID_algorithm_Parallel.png)
+
+##### Standard Form {#standard_form}
+
+The *standard form* is mathematically equivalent to the *parallel form* but facilitates manual tuning as the K gain represents the global "aggressiveness" of the controller.
```
hamishwillee: Why this is a good think isn't clear to me - i.e. why should user try this one or the other one?

In your original post you said something like "facilitate the tuning of a new airframe as it decouples properly the integral and derivative time constants. A user should be able to take the gains of a similar platform (size/inertia) and simply adjust this gain to have it flying properly.
Until now, with the pure parallel form, if the thrust to weight ratio is different from the reference config, the tuning has to be done almost from scratch."

Is that true? Does this make tuning easier in other ways? To me the aggressiveness etc doesn't mean all that much.

en/config_mc/pid_tuning_guide_multicopter.md
bresch: > I believe from all this that you EITHER tune in the standard or in the parallel form - correct? 

Correct

> Is it correct that can MC_ROLLRATE_K can have a max value of 5, but MC_ROLLRATE_P can only have max of 0.5 (i.e. just making sure that this is reasonable and not a typo).

Thanks for rising that. I'll change the `MC_(ROLL|PITCH|YAW)RATE_P` maximum value to 1

en/config_mc/pid_tuning_guide_multicopter.md
bresch: Ok, I'll reincorporate that explanation then. Just for you to know, I see no advantage in the parallel form. Most likely, this became the default because it came from a student project.

hamishwillee: Cool. The main thing is to give people direction - which will be easier? When should they use one over the other? Why did we choose to give them two choices?

hamishwillee:
Thanks @bresch , your work is done :-). I think pretty clear now.

PR #571, opened 2019-09-05
Add Return mode docs for mission-based RTL

hamishwillee:
This is a serious revision of Return mode docs to include info on the RTL types introduced in https://github.com/PX4/Firmware/pull/8749

Fixes https://github.com/PX4/px4_user_guide/issues/565
Fixes https://github.com/PX4/px4_user_guide/issues/137

Overview:
1. Code for different vehicle types is the same, it is just the config that is changed. So this pulls the "Cone" documentation into its own section on Return Altitude.
2. The vehicle-specific docs now just explain "default behaviour".
3. This mixes home location and safe point terminology a bit. I'm hoping to do a revision in the near future to add the new rally point support. It isn't "broken" as such now, but it will make more sense when that goes in.

### DIFF
--- a/en/flight_modes/return.md
+++ b/en/flight_modes/return.md
@@ -2,33 +2,107 @@
 
 [<img src="../../assets/site/position_fixed.svg" title="Position fix required (e.g. GPS)" width="30px" />](../getting_started/flight_modes.md#key_position_fixed)
 
-The *Return* flight mode causes the vehicle to ascend to a safe height, then return to its home position where it may then either wait (hover or circle) or land.
+The *Return* flight mode is used to *fly a vehicle to safety*  on an unobstructed path to a safe destination, where it may either wait (hover or circle) or land.
 
-> **Note** This mode is also known as *Return to Launch* (RTL) and *Return to Home* (RTH) 
+PX4 provides several mechanisms for choosing a safe return path, destination and landing, including using safe points, mission paths, and mission landing sequences.
 
-<span></span>
+The following sections explain how to configure the [return type](#return_types), [return altitude](#return_altitude) and [landing/arrival behaviour](#arrival).
+At the end there are sections explaining the *default* (preconfigured) behaviour for each [vehicle type](#default_configuration).
+
 > **Note** 
->  * This mode requires GPS.
+> * This mode requires GPS.
 > * This mode is automatic - no user intervention is *required* to control the vehicle.
 > * RC control switches can be used to change flight modes on any vehicle.
     The effect of RC stick movement depends on the vehicle type.
 
-The specific behaviour for each vehicle type is described below.
 
-## Multi-Copter (MC)
+## Return Types (RTL_TYPE) {#return_types}
 
-A copter/multi-rotor will first ascend to the return altitude (*by default*, [RTL_RETURN_ALT](#RTL_RETURN_ALT)) and then fly to the home position in a straight line and constant altitude (if already above the return altitude it will return at its current altitude).
+PX4 allows you to specify one of three different approaches for finding an unobstructed path to a safe destination and/or landing:
+- [Safety point return](#safety_point_return) ([RTL_TYPE=0](#RTL_TYPE)): Ascend to safe altitude and return to a safe point (home location)
+- [Mission landing return](#mission_landing_return) (`RTL_TYPE=1`): Ascend to a safe altitude, fly direct to start of mission landing (if defined) and land.
+- [Mission path return](#mission_path_return) (`RTL_TYPE=2`): Use mission path and continue to mission landing (if defined). If no mission landing defined, reverse mission to home.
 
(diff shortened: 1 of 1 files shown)


en/flight_modes/return.md
```diff
@@ -2,33 +2,107 @@
...
+
+This return type uses a *mission landing pattern* (if defined) to set the return destination and landing behaviour.
+
+> **Note** A mission landing pattern consists of a [MAV_CMD_DO_LAND_START](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_LAND_START), one or more position waypoints, and a [MAV_CMD_NAV_LAND](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_LAND).
 
-A copter/multi-rotor will first ascend to the return altitude (*by default*, [RTL_RETURN_ALT](#RTL_RETURN_ALT)) and then fly to the home position in a straight line and constant altitude (if already above the return altitude it will return at its current altitude).
+If a mission landing has been defined, the vehicle will:
+- Ascend to a safe [return altitude](#return_altitude) above any expected obstacles.
+- Fly a direct path to the start of a mission landing sequence
+- Descend/ascend to the specified waypoint altitude
```
hamishwillee: Would be good if this could be confirmed (ie that we fly to safe altitude), then to start of landing sequence, descend to its altitude, then continue.

en/flight_modes/return.md
```diff
@@ -2,33 +2,107 @@
...
+### Mission Path Return Type (RTL_TYPE=2) {#mission_path_return}
+
+This return type uses the mission (if defined) to provide a safe return *path*, and the mission landing pattern (if defined) to provide landing behaviour.
+If there is a mission but no mission landing pattern, the mission is flown *in reverse*.
+
+> **Note** The behaviour is fairly complex because it depends on the flight mode, and whether a mission and mission landing are defined.
+
+Mission *with* landing pattern:
+- **Mission mode:** Mission is continued in "fast-forward mode" (jumps, delay and any other non-position commands ignored, loiter and other position waypoints converted to simple waypoints) and then lands.
+- **Auto mode other than mission mode:**
```
hamishwillee: In non-mission auto modes you join the mission while in manual modes you just fly to the landing waypoint. FWIW that seems a bit inconsistent in that if I fly around in manual and put myself into hold mode, then my return path will be different than if I just do RTL when flying.

en/flight_modes/return.md
```diff
@@ -2,33 +2,107 @@
...
+  - Ascend to a safe [return altitude](#return_altitude) above any expected obstacles.
+  - Fly directly to landing sequence position and descend to waypoint altitude
+  - Land using mission landing pattern
+
+Mission *without* landing pattern defined:
+- **Mission mode:** 
+  - Mission flown "fast-backward" (in reverse) starting from the previous waypoint 
+    - Jumps, delay and any other non-position commands ignored, loiter and other position waypoints converted to simple waypoints.
+    - VTOL vehicles transition to FW mode (if needed) before flying the mission in reverse.
+  - On reaching waypoint 1, the vehicle ascends to the [return altitude](#return_altitude) and flies to the home position (where it [lands or waits](#arrival)).
```
hamishwillee: Can this be confirmed. Ie if I get to waypoint 1, it then ascends to RTL altitude and flies home - ie it effectively treats the first waypoint as the beginning of a normal RTL.

What if waypoint=1 is a takeoff waypoint? Will it reverse that and land?

hamishwillee:
No comments?

en/flight_modes/return.md
RomanBapst: @hamishwillee Yes, that sounds correct.

en/flight_modes/return.md
RomanBapst: @hamishwillee Yes, I agree that this can be a bit confusing. I think the main point was to make sure that if you are in manual mode but far away from the mission, which might still be on the vehicle, you don't actually try to go join the mission in order to go to the landing spot. We'd need to get fancier and actually start calculating distances to do this properly.

en/flight_modes/return.md
RomanBapst: @hamishwillee I think you are right, but let me double check the takeoff waypoint.

RomanBapst:
@hamishwillee I will double check the case when it flies the mission in reverse, I never actually tried that.
Otherwise it looks really good! Thanks a lot!

en/flight_modes/return.md
hamishwillee: Thanks for the clarification. I guess at this point just documenting what happens is a good start.

en/flight_modes/return.md
hamishwillee: I tried this and takeoff waypoint just gets treated like any other waypoint. It heads to it, and when it gets acceptance it then goes back to home. This therefore is strictly correct.

hamishwillee:
Hi @RomanBapst 
As per note, I checked the return myself and it appears to behave as documented. So I'm going to merge - no need for you to check that.
Thanks very much for the review and your help on this.
Cheers
H

PR #572, opened 2019-09-11
Make it easier to work out why parameters are missing

hamishwillee:
We still get quite a lot of questions about why parameters might be missing, and most of the time this is due to using FMUv2 hardware.

This fix adds generic docs on why a parameter might be missing/not visible and how you can solve that problem.

It also adds info about the FMUv2 limitation in a number of places (to make it more obvious) and removes the FMUv2 boards from "recommended lists".


hamishwillee:
@bkueng Decided to self-merge. I think this is all good, or at least "not bad"

PR #578, opened 2019-09-23
add a documentation page for DShot

bkueng:
Docs for https://github.com/PX4/Firmware/pull/12854

I'm not sure if we should add something uder `Basic Configuration` as well. For example, I think reversing the motor direction is helpful to know about.

### DIFF
--- a/en/peripherals/dshot.md
+++ b/en/peripherals/dshot.md
@@ -35,7 +35,7 @@ If using a Pixhawk that has ports labeled AUX and MAIN, set [SYS_USE_IO=0](../ad
 
 Enable DShot with the [DSHOT_CONFIG](../advanced_config/parameter_reference.md#DSHOT_CONFIG) parameter (if the parameter does not exist, the board does not support DShot).
 
-DShot comes with different speed options: *DShot150*, *DShot300*, *DShot600* and *DShot1200*, where the number indicates the speed in bits per second.
+DShot comes with different speed options: *DShot150*, *DShot300*, *DShot600* and *DShot1200*, where the number indicates the speed in kilo-bits/second.
 You should set the parameter to the highest speed supported by your ESC (according to its datasheet) and then reboot the vehicle.
 
 Then connect the battery and arm the vehicle.
--- a/en/peripherals/dshot.md
+++ b/en/peripherals/dshot.md
@@ -28,9 +28,9 @@ If the motors do not spin at lowest throttle, increase [DSHOT_MIN](../advanced_c
 ## Sending Commands
 
 Commands can be sent to the ESC via the [MAVLink shell](https://dev.px4.io/master/en/debug/system_console.html#mavlink_shell).
-Type `dshot help` for a full list.
+See [here](https://dev.px4.io/master/en/middleware/modules_driver.html#dshot) for a full reference of the supported commands.
 
-The most important commands are:
+The most important ones are:
 - Make the first motor beep (helps with identifying motors):
   ```
   dshot beep1 -m 1
--- a/en/peripherals/dshot.md
+++ b/en/peripherals/dshot.md
@@ -1,10 +1,10 @@
 # DShot ESCs
 
 DShot is an alternative ESC protocol that has several advantages over PWM or OneShot:
-- Reduced latency
-- Increased robustness via a checksum
-- No need for ESC calibration as mechanism uses digital encoding
-- Telemetry feedback is available/supported on some ESCs
+- Reduced latency.
+- Increased robustness via a checksum.
+- No need for ESC calibration as the protocol uses digital encoding.
+- Telemetry feedback is available/supported on some ESCs.
 - Can reverse motor spin directions via commands when needed (rather than physically moving wires/re-soldering).
 - Other useful commands are supported.
 
@@ -15,8 +15,18 @@ This topic shows how to connect and configure DShot ESCs.
 
 DShot ESCs are connected and wired the same way as [PWM ESCs](pwm_escs_and_servo.md), and you can switch between these protocols just by changing software parameters (ESCs automatically detect the selected protocol on startup).
 
-On Pixhawk flight controllers that have both MAIN and AUX ports, DShot can only be used on the MAIN ports if the AUX port is disabled (see configuration below).
-This means that if you use DShot ESCs you're restricted to [airframes](../airframes/airframe_reference.md) that don't *require* AUX ports for essential controls (you can use airframes where an AUX port is used for an optional output).
+If using a Pixhawk flight controller that only has a MAIN port, connect the pins according to the [airframe reference](../airframes/airframe_reference.md) for your vehicle.
+
+If using a Pixhawk that has ports labeled AUX and MAIN, set [SYS_USE_IO=0](../advanced_config/parameter_reference.md#SYS_USE_IO) and connect your ESCs to the AUX-labeled outputs *as though they were labeled MAIN*.
+
+> **Note** A Pixhawk flight controller that has both FMU and IO will label these ports as AUX and MAIN respectively.
+> DShot can only be used on the FMU ports (labeled AUX), which is a problem because ESC/motor outputs are typically assigned to the MAIN port in the [airframe reference](../airframes/airframe_reference.md).
+>
+> To use DShot you therefore normally set `SYS_USE_IO=0` (which makes the ports labeled AUX behave *as though* they were the ports labeled MAIN), and connect your ESCs to the corresponding AUX-labeled outputs.
+> Any outputs that would normally be assigned to AUX ports in the [airframe reference](../airframes/airframe_reference.md) are no longer available.
+
+<span></span>
+> **Note** Developers might alternatively modify the [airframe AUX mixer](http://dev.px4.io/master/en/airframes/adding_a_new_frame.html#mixer-file) so that the multirotor outputs are on the AUX port rather than MAIN.
 
 
 ## Configuration {#configuration}
@@ -32,12 +42,6 @@ Then connect the battery and arm the vehicle.
 The ESCs should initialize and the motors turn in the correct directions.
 - If the motors do not spin in the correct direction (for the [selected airframe](../airframes/airframe_reference.md)), reverse them by sending an [ESC Command](#commands).
 - Adjust [DSHOT_MIN](../advanced_config/parameter_reference.md#DSHOT_MIN) so that the motors spin at lowest throttle (but the vehicle does not take off).
-
-On boards that have both MAIN and AUX ports, set [SYS_USE_IO=0](../advanced_config/parameter_reference.md#SYS_USE_IO).
-
-> **Note** DShot is only supported on FMU pins, which map to the AUX ports on boards that have both MAIN and AUX.
-  Setting `SYS_USE_IO=0` disables the IO board, and remaps the FMU pins to the MAIN port.
-  This is needed because most [airframes](../airframes/airframe_reference.md) specify that ESCs/motors are connected to MAIN pins. 
 
 
 ## ESC Commands {#commands}
@@ -80,7 +84,7 @@ Some ESCs are capable of sending telemetry back to the flight controller, includ
 - accumulated current consumption
 - RPM values
 
-These Dshot ESCs will have an additional telemetry wire.
+These DShot ESCs will have an additional telemetry wire.
 
 To enable this feature (on ESCs that support it):
 1. Join all the telemetry wires from all the ESCs together, and then connect them to one of the RX pins on an unused flight controller serial port.
@@ -90,6 +94,5 @@ After a reboot you can check if telemetry is working (make sure the battery is c
 ```
 dshot esc_info -m 1
 ```
-
 
 > **Tip** You may have to configure [MOT_POLE_COUNT](../advanced_config/parameter_reference.md#MOT_POLE_COUNT) to get the correct RPM values.
--- a/en/peripherals/dshot.md
+++ b/en/peripherals/dshot.md
@@ -1,0 +1,78 @@
+# DShot
+
+DShot is an alternative ESC protocol to traditional PWM or OneShot.
+It has several advantages compared to those:
+- Reduced latency
+- Checksum (increased robustness)
+- Digital encoding (no need for ESC calibrations)
+- Possibility to get feedback (telemetry)
+- Sending commands (such as reversing the motor spin direction)
+
+ESC's are connected and wired the same way as for [PWM Outputs](pwm_escs_and_servo.md).
+It is possible to switch between both protocols by changing software parameters only.
+ESC's automatically detect the protocol on startup.
+
+## Configuration
+DShot comes with different speed options: *DShot150*, *DShot300*, *DShot600* and *DShot1200*.
+It defines the speed in number of bits per second.
+Select the highest speed supported by the ESC, which can be found in the datasheet.
+
+> **Warning** As usual: change ESC configuration parameters without props mounted.
+
+Enable DShot with the [DSHOT_CONFIG](../advanced_config/parameter_reference.md#DSHOT_CONFIG) parameter by selecting the appropriate speed and reboot the vehicle.
+
+> **Note** If the parameter does not exist, the board does not support DShot.
+
+<span></span>
+
+> **Tip** On boards with an IO, DShot is only supported on the AUX pins.
+> It is best to disable the IO via [SYS_USE_IO](../advanced_config/parameter_reference.md#SYS_USE_IO) and then connect the motors to the AUX pins instead of MAIN.
+
+If connected via battery, the ESC's should initialize and the motors turn when armed.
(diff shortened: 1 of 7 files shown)


en/peripherals/dshot.md
```diff
@@ -0,0 +1,78 @@
...
+
+<span></span>
+
+> **Tip** On boards with an IO, DShot is only supported on the AUX pins.
+> It is best to disable the IO via [SYS_USE_IO](../advanced_config/parameter_reference.md#SYS_USE_IO) and then connect the motors to the AUX pins instead of MAIN.
+
+If connected via battery, the ESC's should initialize and the motors turn when armed.
+In case the motors do not spin at lowest throttle, increase [DSHOT_MIN](../advanced_config/parameter_reference.md#DSHOT_MIN) until they spin.
+
+## Sending commands
```
hamishwillee: Link to module docs for dshot when they exist? 
I'd be tempted to push these examples into dshot module docs along with a link to this doc. Then have a very much reduced section here.

en/peripherals/dshot.md
```diff
@@ -0,0 +1,79 @@
...
+You should set the parameter to the highest speed supported by your ESC (according to its datasheet) and then reboot the vehicle.
+
+Then connect the battery and arm the vehicle.
+The ESCs should initialize and the motors turn.
+If the motors do not spin at lowest throttle, increase [DSHOT_MIN](../advanced_config/parameter_reference.md#DSHOT_MIN) until they spin.
+
+> **Tip** On boards with an IO, DShot is only supported on the AUX pins.
+> It is best to disable the IO via [SYS_USE_IO](../advanced_config/parameter_reference.md#SYS_USE_IO) and then connect the motors to the AUX pins instead of MAIN.
+
+## Sending Commands
```
hamishwillee: Link to module docs for dshot when they exist? Those docs would ideally link back here.

en/peripherals/dshot.md
```diff
@@ -0,0 +1,79 @@
...
+
+Some ESCs are capable of sending telemetry back to the flight controller, including:
+- temperature
+- voltage
+- current
+- accumulated current consumption
+- RPM values
+
+To enable this feature (on ESCs that support it):
+1. Connect all wires from the ESCs together to one of the RX pins of an unused serial port of the flight controller.
```
hamishwillee: I subedited this section, and now it is clear that I don't know what it means :-). 

Docs indicate the esc has 2 wires Power VBAT and GND. I read this as taking all these wires for all the ESCS (so on a quadcopter, 8 wires) and joining them to just one RX pin on one serial port. Is this correct?

en/peripherals/dshot.md
```diff
@@ -0,0 +1,79 @@
...
+
+DShot comes with different speed options: *DShot150*, *DShot300*, *DShot600* and *DShot1200*, where the number indicates the speed in bits per second.
+You should set the parameter to the highest speed supported by your ESC (according to its datasheet) and then reboot the vehicle.
+
+Then connect the battery and arm the vehicle.
+The ESCs should initialize and the motors turn.
+If the motors do not spin at lowest throttle, increase [DSHOT_MIN](../advanced_config/parameter_reference.md#DSHOT_MIN) until they spin.
+
+> **Tip** On boards with an IO, DShot is only supported on the AUX pins.
+> It is best to disable the IO via [SYS_USE_IO](../advanced_config/parameter_reference.md#SYS_USE_IO) and then connect the motors to the AUX pins instead of MAIN.
```
hamishwillee: On boards with IO Dshot is only supported on the AUX ports (FMU). This tips suggests that we disable the IO board and connect to MAIN (Same physical ports on FMU), but it isn't obvious why.

What I think (?) you are saying is that DSHOT only works on FMU pins, so if you want to map outputs to motors as per the airframe reference you're going to have to disable the IO so that those pins are MAIN. Correct?

Further implication being that anything mapped to AUX ports will have to be unavailable if you use DSHOT ESC.

bkueng: > What I think (?) you are saying is that DSHOT only works on FMU pins, so if you want to map outputs to motors as per the airframe reference you're going to have to disable the IO so that those pins are MAIN. Correct?

Correct.

> Further implication being that anything mapped to AUX ports will have to be unavailable if you use DSHOT ESC.

Right.

en/peripherals/dshot.md
bkueng: Yes I considered that, adding it now.

bkueng:
Thanks for the updates and review.

> We don't have anything else on motors in basic configuration - you're basically suggesting we have a section on configuration to make sure that motors are spinning in the right direction?

Yes, as testing the motor assignment and spin direction is one of the things you always have to do when setting up a vehicle. What you normally have to do is swapping 2 of the 3 motor wires, which means resoldering, if the motors are not attached via bullet connectors. Using dshot commands simplifies this.

en/peripherals/dshot.md
```diff
@@ -0,0 +1,79 @@
...
+
+DShot ESCs are connected and wired the same way as [PWM ESCs](pwm_escs_and_servo.md), and you can switch between these protocols just by changing software parameters (ESCs automatically detect the selected protocol on startup).
+
+## Configuration
+
+> **Warning** Remove propellers before changing ESC configuration parameters!
+
+Enable DShot with the [DSHOT_CONFIG](../advanced_config/parameter_reference.md#DSHOT_CONFIG) parameter (if the parameter does not exist, the board does not support DShot).
+
+DShot comes with different speed options: *DShot150*, *DShot300*, *DShot600* and *DShot1200*, where the number indicates the speed in bits per second.
```
hamishwillee: Is it bits per second? You other docs indicate kb/s

en/peripherals/dshot.md
hamishwillee: @bkueng OK, I think I fixed this by adding a line:

> Join all the telemetry wires from all the ESCs together, and then connect them to one of the RX pins on an unused flight controller serial port.

I clarifying the specific line that gets connected. Please check/correct?

en/peripherals/dshot.md
hamishwillee: OK, so to address this I've put wiring into its own section and mentioned the issue there, and more clearly stated that you have to make the change in the configuration section. Please check.

en/peripherals/dshot.md
```diff
@@ -0,0 +1,95 @@
+# DShot ESCs
+
+DShot is an alternative ESC protocol that has several advantages over PWM or OneShot:
+- Reduced latency
```
hamishwillee: FYI only, I turned these around so the advantages are stated then the feature that grants it (since we're listing advantages).

hamishwillee:
> Yes, as testing the motor assignment and spin direction is one of the things you always have to do when setting up a vehicle. What you normally have to do is swapping 2 of the 3 motor wires, which means resoldering, if the motors are not attached via bullet connectors. Using dshot commands simplifies this.

Then I think it makes sense to do so. Perhaps as a separate PR (up to you). Do you want to take first stab at this.

I've made changes to address raised issues for your review.

en/peripherals/dshot.md
bkueng: Perfect.

en/peripherals/dshot.md
bkueng: That's good.

en/peripherals/dshot.md
```diff
@@ -2,14 +2,24 @@
 
 DShot is an alternative ESC protocol that has several advantages over PWM or OneShot:
 - Reduced latency
-- Checksum (increased robustness)
-- Digital encoding (no need for ESC calibrations)
-- Possibility to get feedback (telemetry)
-- Sending commands (such as reversing the motor spin direction)
+- Increased robustness via a checksum
+- No need for ESC calibration as mechanism uses digital encoding
```
bkueng: Maybe: No need for ESC calibration as the protocol uses digital encoding

en/peripherals/dshot.md
```diff
@@ -19,13 +29,18 @@ DShot comes with different speed options: *DShot150*, *DShot300*, *DShot600* and
...
+The ESCs should initialize and the motors turn in the correct directions.
+- If the motors do not spin in the correct direction (for the [selected airframe](../airframes/airframe_reference.md)), reverse them by sending an [ESC Command](#commands).
+- Adjust [DSHOT_MIN](../advanced_config/parameter_reference.md#DSHOT_MIN) so that the motors spin at lowest throttle (but the vehicle does not take off).
+
+On boards that have both MAIN and AUX ports, set [SYS_USE_IO=0](../advanced_config/parameter_reference.md#SYS_USE_IO).
 
-> **Tip** On boards with an IO, DShot is only supported on the AUX pins.
-> It is best to disable the IO via [SYS_USE_IO](../advanced_config/parameter_reference.md#SYS_USE_IO) and then connect the motors to the AUX pins instead of MAIN.
+> **Note** DShot is only supported on FMU pins, which map to the AUX ports on boards that have both MAIN and AUX.
+  Setting `SYS_USE_IO=0` disables the IO board, and remaps the FMU pins to the MAIN port.
```
bkueng: > and remaps the FMU pins to the MAIN port

There is no physical remapping: the FMU pins still go to AUX (according to the board label). What is different is that the mixer for MAIN is loaded on the pins for AUX. That is why you have to replug your motors from MAIN to AUX.

en/peripherals/dshot.md
```diff
@@ -2,14 +2,24 @@
...
+
+This topic shows how to connect and configure DShot ESCs.
+
+
+## Wiring/Connections {#wiring}
 
 DShot ESCs are connected and wired the same way as [PWM ESCs](pwm_escs_and_servo.md), and you can switch between these protocols just by changing software parameters (ESCs automatically detect the selected protocol on startup).
 
-## Configuration
+On Pixhawk flight controllers that have both MAIN and AUX ports, DShot can only be used on the MAIN ports if the AUX port is disabled (see configuration below).
```
bkueng: > DShot can only be used on the MAIN ports if the AUX port is disabled (see configuration below).

Technically this is not correct: the SYS_USE_IO=0 method is only one way of doing it. You could also change the mixer for AUX to add multi-rotor outputs for example, and then still use the MAIN pins for whatever you want.

> DShot can only be used on the MAIN ports

Same as below: it is only used on the AUX ports (if you are refering to the pin labels of the board).

en/peripherals/dshot.md
hamishwillee: @bkueng Bugger. Thought I understood this from http://dev.px4.io/master/en/hardware/reference_design.html#mainio-function-breakdown

So what I previously understood was that .... if you had a board with only an FMU this only had MAIN ports and this would be labelled MAIN. If you had a board with FMU (AUX) and IO (MAIN) and you set `SYS_USE_IO=0` then the IO board is disabled and the ports labelled MAIN magically get remapped to whatever was on AUX. 

But what I think you are now saying is that (please confirm):
- FMU is always connected to AUX physical port. AUX is the "name" of the FMU port.
- IO is always connected to MAIN physical port. Main is the "name" of the IO port.
- On FMU-only system the FMU is connected to the AUX port (still) but it will load the MAIN mixer and will be *screen printed with MAIN*. The screen printing is what tells the user where they connect their wires.
- On FMU and AUX board with  `SYS_USE_IO=1`  we have FMU port AUX loaded with AUX mixer (screen printed AUX), IO port MAIN loaded with MAIN mixer (screen printed MAIN)
- On FMU and AUX board after setting `SYS_USE_IO=0`  we have FMU port AUX loaded with **MAIN** mixer (screen printed **AUX**), IO port MAIN disabled (screen printed MAIN)

So basically a user has to connect to the ports screen printed AUX as though they were labelled MAIN. Correct?

I wonder if we could fix this by naming port and mixer separately. Is there another generic way to refer to FMU AUX port that conveys this is a bank or PWM outputs pins? Ie FMU PWM Port?

en/peripherals/dshot.md
hamishwillee: Thanks @bkueng , I think I get the distinction now - waiting on your confirmation in https://github.com/PX4/px4_user_guide/pull/578#discussion_r328026199

>> DShot can only be used on the MAIN ports if the AUX port is disabled (see configuration below).

So changing my terminology (avoiding using the term mixer below)...is it reasonable?

A Pixhawk board that has both FMU and IO will label them as AUX and MAIN respectively. Normally you would connect your ESCs to the ports labelled MAIN, but DShot can only be used on the FMU ports (labeled `AUX`). In order to use these boards for controlling motors, you will typically set `SYS_USE_IO=0` (which makes the ports labeled `AUX` behave as though they were the ports labeled `MAIN`) and connect your ESCs to the corresponding AUX-labeled outputs. 

> Technically this is not correct: the SYS_USE_IO=0 method is only one way of doing it. You could also change the mixer for AUX to add multi-rotor outputs for example, and then still use the MAIN pins for whatever you want.

OK, I get that, but probably a bit hard for most.

en/peripherals/dshot.md
bkueng: > On FMU-only system the FMU is connected to the AUX port (still) but it will load the MAIN mixer and will be screen printed with MAIN. The screen printing is what tells the user where they connect their wires.

You can call it that way if you want to. But FMU-only boards simply don't have any notion of AUX.

The rest is all correct.

> I wonder if we could fix this by naming port and mixer separately. Is there another generic way to refer to FMU AUX port that conveys this is a bank or PWM outputs pins? Ie FMU PWM Port?

Maybe just FMU Outputs/Port? If you add PWM in the name, it won't be correct for DShot.

en/peripherals/dshot.md
bkueng: >  So changing my terminology (avoiding using the term mixer below)...is it reasonable?

Yes, looks good.

> OK, I get that, but probably a bit hard for most.

Right, we don't need to explain it, I just want to ensure we know about it.

en/peripherals/dshot.md
hamishwillee: @bkueng ^^^

bkueng: Hmm I thought I replied here as well.
Anyway, the speed would be `150'000' b/s, which is shortened in the name as `DShot150`. So kb/s would be more appropriate here indeed.

hamishwillee: Thanks @bkueng . Who knows - sometimes github does odd shit :-)

I changed this to kilobits/second. I know that contracts to Kb, but for some reason I always look at that and think - typo, they mean KB (kilobytes).

hamishwillee:
Hi @bkueng 

OK, updated the sections on using the escs on AUX port again. This is only in the wiring section and I think is quite thorough. I didn't put in the config section, because they would already have done this work by the time they get to the config section.

I also created https://github.com/PX4/Devguide/pull/871 to clarify what happens - but in this case it refers to the mixer being loaded.

Any idea when Dshot stuff can merged?

bkueng:
Looking good now.

> Any idea when Dshot stuff can merged?

It's in a good state now, we need some more flight testing and enabling Pixhawk 2. I'm not sure if people want to add it to v1.10 though.

hamishwillee:
Great. Feel free to merge it yourself when ready! Thanks for your patience on the clarifications.

hamishwillee:
@bkueng Do you think it is worth adding a "Purchase" section that states "Any Dshot ESCs should work. We have tested on ... (and then add these ones https://github.com/PX4/Firmware/pull/12854#issuecomment-536187998). 

If so, is it worth mentioning specifically  BLHeli32 ESC ?

bkueng:
I would only add such a list if we see people reporting issues on some ESC's. Otherwise I expect it to work everywhere. We should test on Kiss ESC's as well quickly to be sure, I'll check if I can find some.
The other important thing is that it's likely that we need 1kHz rate loop for some ESC's. But that is the default now, it's only different if someone would change `IMU_GYRO_RATEMAX`.

bkueng:
> We should test on Kiss ESC's as well quickly to be sure, I'll check if I can find some.

Found some, thanks to @MaEtUgR, they work as well.

hamishwillee:
Thanks. Let's leave this alone then, as you have suggested.

smpotter:
Digging up bones...

After remapping from main to auxiliary output for Dshot ESC. Did you lose rc commands to the motors? ie. No yaw, pitch or roll??? I have reached my witz end.

Pix2.1 xrotor 60 32bit Orange Cube.

I can arm without fail and I do have throttle. I also made a two channel gimbal as well as a bait release.

ARTF channels 1-4 and gimbal on 8-9 Release 10

smpotter:
Ooops ps I see the sticks movement on the calibration screen in Mission Planner. I have attempted to move output mapping to no avail. Is this bumping heads with Dshot ESC motors output?

hamishwillee:
@smpotter Generally the right place to ask this is on the PX4 discussion boards or slack. Bigger pool of people who might know the answer. You should note what version of PX4 you are using.

If you then find that the documentation is incorrect or incomplete we'd welcome an issue.

PR #581, opened 2019-09-25
Fix Parameter Reference

Finwood:
The _Voltage divider_ section referred wrongly to the parameter BAT_A_PER_V, which belongs to the current divider's configuration.


hamishwillee:
Thanks very much @Finwood ! Merged.

PR #582, opened 2019-09-26
basic configuration: add motors section

bkueng:
PR against #578, since it adds links to DShot.

Checking motor direction and assignment will become easier once we have the motor test command and UI in QGC - which is soon hopefully.

### DIFF
--- a/en/config/motors.md
+++ b/en/config/motors.md
@@ -1,4 +1,4 @@
-# Motor Checks
+# Motor/Servo Checks
 
 After the airframe is setup and configured you should validate the motor assignment, spin direction and servo response:
 
--- a/en/config/motors.md
+++ b/en/config/motors.md
@@ -2,9 +2,8 @@
 
 After the airframe is setup and configured you should validate the motor assignment, spin direction and servo response:
 
-1. Remove all propellers, switch to Manual flight mode and arm the vehicle.
-1. Increase the throttle a bit, so the vehicle does not automatically disarm.
-   At the same time check that the motors respond to throttle changes.
+1. Remove all propellers, switch to [Stabilised mode](../flight_modes/manual_stabilized_mc.md) (Multicopter) or [Manual mode](../flight_modes/manual_fw.md) (Fixed Wing) and arm the vehicle.
+1. Increase the throttle a little (so the vehicle does not automatically disarm) and check that the motors respond to throttle changes.
 1. Check that all motors turn to the right direction according to the configured [airframe](../airframes/airframe_reference.md).
    If any spin the wrong direction, they need to be reversed.
    There are several options to achieve that:
--- a/en/config/motors.md
+++ b/en/config/motors.md
@@ -1,0 +1,12 @@
+# Motor Checks
+
+After the airframe is configured and the vehicle is setup, it is necessary to validate the motor assignment, spin direction and servo response.
+
+1. First, remove all propellers, switch to Manual flight mode and arm the vehicle.
+1. Increase the throttle a bit, so the vehicle does not automatically disarm. At the same time check that the motors are responding to throttle changes.
+1. Then check that all motors turn to the right direction according to the configured [airframe](../airframes/airframe_reference.md). If one or more spin the wrong direction, they need to be reversed. There are several options to achieve that:
+   1. Swap 2 of the 3 motor cables (it does not matter which ones). If they are not connected via bullet-connectors, re-soldering is required. Therefore the next option is preferable.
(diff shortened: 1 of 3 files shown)


en/config/motors.md
```diff
@@ -0,0 +1,12 @@
+# Motor Checks
+
+After the airframe is configured and the vehicle is setup, it is necessary to validate the motor assignment, spin direction and servo response.
+
+1. First, remove all propellers, switch to Manual flight mode and arm the vehicle.
+1. Increase the throttle a bit, so the vehicle does not automatically disarm. At the same time check that the motors are responding to throttle changes.
+1. Then check that all motors turn to the right direction according to the configured [airframe](../airframes/airframe_reference.md). If one or more spin the wrong direction, they need to be reversed. There are several options to achieve that:
+   1. Swap 2 of the 3 motor cables (it does not matter which ones). If they are not connected via bullet-connectors, re-soldering is required. Therefore the next option is preferable.
+   1. Reverse the direction via [DShot commands](../peripherals/dshot.md#commands) if the ESCs support DShot. DShot should be preferred over PWM in general, as it has further advantages.
```
ndepal: It might be worth mentioning that even when not using DShot it's possible to reverse the motor direction without soldering. E.g. BLHeli ESCs can be configured though the 1 wire interface and their direction can be changed, even without DShot. You can either use a USB linker or a flight controller that supports FC passthrough (like betaflight). The latter is convenient because you can configure all 4 ESCs at the same time (so you can set the same beep volume etc for all of them).

bkueng: Yes I'm aware of the passthrough, but I wasn't sure if it's worth mentioning since PX4 does not support it. That would be convenient as well though.
I didn't know about the USB linker. Is it something easy to find and buy?

ndepal: > Is it something easy to find and buy?

Actually, not really. You will find them on hobbyking etc, but it's almost impossible to find out if it will work with your ESC (depends on the MCU), so it's a bit of trial and error. The demand for them seems to be quite low, any guide online will only suggest FC passthrough.

hamishwillee: IMO not worth mentioning as the process of getting passthrough working likely to be similar effort to re-soldering and just makes this very complicated. Marking as resolved on that basis.

en/config/motors.md
```diff
@@ -0,0 +1,14 @@
+# Motor Checks
```
hamishwillee: Is this/should this be "Motor/Servo Checks"?

en/config/motors.md
```diff
@@ -0,0 +1,14 @@
...
+
+After the airframe is setup and configured you should validate the motor assignment, spin direction and servo response:
+
+1. Remove all propellers, switch to Manual flight mode and arm the vehicle.
+1. Increase the throttle a bit, so the vehicle does not automatically disarm.
+   At the same time check that the motors respond to throttle changes.
+1. Check that all motors turn to the right direction according to the configured [airframe](../airframes/airframe_reference.md).
+   If any spin the wrong direction, they need to be reversed.
+   There are several options to achieve that:
+   - If using ESCs that support [DShot](../peripherals/dshot.md) you can reverse the direction via [DShot commands](../peripherals/dshot.md#commands).
```
hamishwillee: Note, I switched this over to be "easiest way first".  I haven't gone into detail on the other factors about using Dshot. IMO there are several reasons you might not want to use them too. I did state that there are other reasons you might prefer Dshot.

en/config/motors.md
```diff
@@ -0,0 +1,14 @@
+# Motor Checks
+
+After the airframe is setup and configured you should validate the motor assignment, spin direction and servo response:
+
+1. Remove all propellers, switch to Manual flight mode and arm the vehicle.
```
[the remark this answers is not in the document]
bkueng: It is specifically the Manual mode, since you then have direct throttle (and roll/pitch) control. It applies to both FW or MC (where Manual==Stabilized on MC).

en/config/motors.md
bkueng: Yes that's fine.

en/config/motors.md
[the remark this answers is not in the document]
bkueng: Would you still link to MC and FW Manual modes?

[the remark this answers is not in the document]
hamishwillee: Yes. I will fix and merge now.

hamishwillee:
OK, merged. Thanks for doing this @bkueng ! @ndepal if needed, the conversation can continue on https://github.com/PX4/px4_user_guide/pull/578

en/config/motors.md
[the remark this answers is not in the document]
bkueng: Thanks

PR #583, opened 2019-09-26
Collision prevention: include guidance

jkflying:
This reworks the collision prevention, and includes the new guidance feature.

Still needs some diagrams, but this is a good start.


en/computer_vision/collision_prevention.md
```diff
@@ -2,78 +2,92 @@
...
-The diagrams below are used to illustrate how the setting works, where:
-- The minimum distance (separation) `MPC_COL_PREV_D = 4m`, and the sensor range is 12m.
-- The pink line shows where the user is trying to move the vehicle (this is the same for both diagrams).
-- The blue line shows the angle where received sensor data *can be used* (and what will be discarded).
-  - This is equal to double `MPC_COL_PREV_ANG` degrees centred on the commanded setpoint.
-  - On the left diagram the `MPC_COL_PREV_D` angle is 90 degrees, so all sensor data from the right hand side of the blue line is included.
-  - On the right the angle is 45 degrees, resulting in a smaller set of sensor data being used (note the sensor data close to the obstacle is excluded!)
-- The red arc shows the angle where sensor data *will be used*.
-  This is the subset of sensors that are both in range of an obstacle, and within the angle defined by the blue line.
+There are two main sources of delay which should be accounted for: *sensor delay*, and vehicle *velocity setpoint tracking delay*.
```
hamishwillee: How do you calculate the sensor delay/What if you have both a sensor and a setpoint tracking delay. 

I would guess that this is all calculated using the logs as indicated, but that isn't clear.

en/computer_vision/collision_prevention.md
```diff
@@ -2,78 +2,92 @@
...
 
-It can be enabled for multicopter vehicles in [Position mode](../flight_modes/position_mc.md), and can use sensor data from an offboard companion computer, a rangefinder attached to the flight controller, or both (fused).
+It can be enabled for multicopter vehicles in [Position mode](../flight_modes/position_mc.md), and can use sensor data from an offboard companion computer, offboard rangefinders over MAVLink, a rangefinder attached to the flight controller, or any combination of the above.
 
-> **Warning** Collision prevention may not prevent a crash if your vehicle is moving too fast!
-  This feature has only been tested (at time of writing) for a vehicle moving at 4 m/s.
+Collision prevention may restrict vehicle maximum speed if the sensor range isn't large enough!
+It also prevents motion in directions where no sensor data is available (i.e. if you have no rear-sensor data, you will not be able to fly backwards).
 
+> **Tip** If high flight speeds are critical, consider disabling collision prevention when not needed.
```
hamishwillee: FYI Note, I split the specific tip from the associated cause. Not 100% sure this is better, but I think so.

en/computer_vision/collision_prevention.md
jkflying: Characterizing sensor delay is a huge pain. We did it by overlaying a known-low-latency sensor onto our unknown sensor, and then applying movements. I'm not sure if there is a better way.

jkflying:
Thanks for the cleanup! I think we can actually merge this as is. I don't have time in the next week or so to add diagrams, and I'd prefer to get this out.

en/computer_vision/collision_prevention.md
hamishwillee: @jkflying Thanks, but I have no idea what that means. I would have thought that since there is only one parameter the tracking delay must incorporate the sensor delay?

Anyway, I will merge this, but am assuming you will come back and clarify :-)

hamishwillee:
Thanks @jkflying - completely agree. Merging!

en/computer_vision/collision_prevention.md
jkflying: Yes, the single parameter incorporates both delays, you just add them together.

hamishwillee: Shouldn't the measured tracking error incorporate the sensor delay by default? (ie you get the measured velocity delay in part due to the sensor delay?). Then you'd be adding a delay to something that is already "factored in".

@jkflying I could be being stupid here  ...

PR #589, opened 2019-10-18
Provide THR_MDL_FAC Example

Finwood:
[As discussed on the PX4 forums](https://discuss.px4.io/t/thrust-curve-compensation-through-thr-mdl-fac/13207/), the documentation was so far lacking
an example for the thrust curve linearization through THR_MDL_FAC.
This commit adds a mathematical approach as an example, hoping that it
might help understand the concept behind the modelling parameter.

The approach references [this gist](https://gist.github.com/Finwood/19fe4504fab043d35b5f71bc990e5855) – might this be too technical for user docs?


en/config_mc/pid_tuning_guide_multicopter.md
```diff
@@ -152,18 +152,33 @@ The following parameters can also be adjusted. These determine the maximum rotat
...
+  > **Note** The mapping between PWM and static thrust depends highly on the battery voltage.
+
+  [THR_MDL_FAC_Calculation]: https://gist.github.com/Finwood/19fe4504fab043d35b5f71bc990e5855
+
+  If you don't have access to a thrust stand, you can also tune the modelling factor empirically.
+  Start off with 0.3 and increase it by 0.1 at a time.
+  If it is too high, you will start to notice oscillations at lower throttle values,
+  if it is too low you'll notice oscillations at higher throttle values.
   
   > **Note** The rate controller must be re-tuned if you change this parameter.
```
hamishwillee: This note probably should go up to top near line 161 in new version.

en/config_mc/pid_tuning_guide_multicopter.md
```diff
@@ -152,18 +152,33 @@ The following parameters can also be adjusted. These determine the maximum rotat
...
+
+  If you have the possibility to _measure_ thrust (e.g. with a [thrust stand](https://www.rcbenchmark.com/pages/series-1580-thrust-stand-dynamometer)),
+  you can measure the relation between the PWM control signal and the motor's actual thrust, and fit a function to the data.
+  [This Notebook][THR_MDL_FAC_Calculation] shows how the thrust model factor `THR_MDL_FAC` may be calculated from previously measured thrust data.
+
+  [![Thrust Curve Compensation](../../images/mc_pid_tuning/thrust-curve-compensation.svg)][THR_MDL_FAC_Calculation]
+
+  > **Note** The mapping between PWM and static thrust depends highly on the battery voltage.
+
+  [THR_MDL_FAC_Calculation]: https://gist.github.com/Finwood/19fe4504fab043d35b5f71bc990e5855
```
hamishwillee: 1. This is not appearing in the gitbook generated page. Is it supposed to be part of the note? Generally prefer not to have links to things outside of control of the library - i.e. that you might delete yourself accidentally. Perhaps put the file in the gitbook tree (I can do s a post process if needed). 
1. What I think you've done here is provided an anchor that is linked to from "This Notebook" above - correct? This works in github, but may not work in gitbook. I will test after this is technical reviewed.

hamishwillee:
@MaEtUgR Can you please give this a technical review and confirm you're OK with the the changes?

en/config_mc/pid_tuning_guide_multicopter.md
Finwood: 1. I could render the file as static html and add it to the gitbook tree if that is the preferred method. Where would I place it?
1. Yes, this is a markdown reference-style link. I've linked from both "This Notebook" and the image itself and didn't want to repeat myself, that's why I did not include inline links.

en/config_mc/pid_tuning_guide_multicopter.md
Finwood: Right, would be better. Fixed in ae67a27 :+1:

en/config_mc/pid_tuning_guide_multicopter.md
hamishwillee: @Finwood 
There isn't a "normal" case for Notebooks. Perhaps though it could go in **px4_user_guide/assets/config/mc/ThrustCurve.ipynb** - then we could still link to the file in the same way as we do for the gist but it would be in the tree.

2.The markdown link works fine. Thanks!

hamishwillee:
Thanks @Finwood - this looks great. Really appreciate you taking the time to provide a real example. 

I merged this and did a minor subedit. There are two open issues still:
1. Commend from @jkflying above 
   > As an aside - the variation across different battery voltages is interesting - maybe we should add a battery voltage correction factor?

2. Location of the notebook, - suggestion here: https://github.com/PX4/px4_user_guide/pull/589#discussion_r337251581

Finwood:
@hamishwillee I've moved the notebook to the specified location and updated the link.

hamishwillee:
@Finwood I think you might have to do a new PR as the old one is merged?

I note also that you converted to HTML. I'm OK with that, but I was thinking perhaps we keep in the notepad format and let github do the rendering?

hamishwillee:
@Finwood I fixed the link up now. Thanks very much again for this!!!

PR #590, opened 2019-10-20
Return - add warning about mission upload

hamishwillee:
As discussed in https://github.com/PX4/Firmware/pull/12470#issuecomment-543834071

### DIFF
--- a/en/flight_modes/return.md
+++ b/en/flight_modes/return.md
@@ -48,7 +48,7 @@ This return type uses a *mission landing pattern* (if defined) to set the return
 > **Note** A mission landing pattern consists of a [MAV_CMD_DO_LAND_START](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_LAND_START), one or more position waypoints, and a [MAV_CMD_NAV_LAND](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_LAND).
 
 <span></span>
-> **Warning** When this type is set PX4 will reject upload of a mission without a valid landing pattern.
+> **Warning** When this type is set PX4 will reject any mission without a valid landing pattern.
 
 If a mission landing has been defined, the vehicle will:
 - Ascend to a safe [return altitude](#return_altitude) above any expected obstacles.
--- a/en/flight_modes/return.md
+++ b/en/flight_modes/return.md
@@ -20,7 +20,7 @@ At the end there are sections explaining the *default* (preconfigured) behaviour
 
 PX4 allows you to specify one of three different approaches for finding an unobstructed path to a safe destination and/or landing:
 - [Safety point return](#safety_point_return) ([RTL_TYPE=0](#RTL_TYPE)): Ascend to safe altitude and return to a safe point (home location)
-- [Mission landing return](#mission_landing_return) (`RTL_TYPE=1`): Ascend to a safe altitude, fly direct to start of mission landing (if defined) and land.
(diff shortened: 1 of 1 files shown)


en/flight_modes/return.md
```diff
@@ -44,8 +44,12 @@ In this return type the vehicle:
 
 This return type uses a *mission landing pattern* (if defined) to set the return destination and landing behaviour.
 
+<span id="mission_landing_pattern"></span>
 > **Note** A mission landing pattern consists of a [MAV_CMD_DO_LAND_START](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_LAND_START), one or more position waypoints, and a [MAV_CMD_NAV_LAND](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_LAND).
 
+<span></span>
+> **Warning** When this type is set PX4 will reject upload of a mission without a valid landing pattern.
```
hamishwillee: @Antiheavy Does this do the job? I don't think at this point we need to have detail on what does the checking (toehr than PX4). The "Warning" means this is pretty obvious, and it should be one of the first things a user sees if they select this mode.

Antiheavy: well....  technically this should say: 
" **Warning** When this type is set PX4 will reject any mission missing a valid landing pattern. "

...not to open a can of worms, but PX4 absolutely DOES allow the upload of the mission.  PX4 allows upload of ANY mission as far as I can tell.  Only after the mission has been uploaded and previous mission deleted will PX4 run the MissionFeasibilityChecker.  This is a fundamental flaw in PX4 (I wish I had the skills to fix it).  The reason  this is a problem is that PX4 will delete a good mission and replace with an unknown/unverified mission which could potentially be rejected by PX4, leaving the vehicle in limbo.  In the context of RTL_TYPE=1 this means the vehicle no longer has a safe RTL landing!!!  This PX4 issue summarizes the situation: https://github.com/PX4/Firmware/issues/12473

hamishwillee: @Antiheavy Thanks, I was aware of that PR and the distinction. I'm not sure I want to open that can of worms. Perhaps can improve by testing my assumptions:
1. I presume the check is done only after upload? So you could upload a mission without a landing pattern with RTL_TYPE=0 and then change to RTL_TYPE=1 and in this case you will just fall back to using home landing (but not lose the mission)?
2. After mission rejection, is the mission just cleared/dumped?

Antiheavy: >     1. I presume the check is done only after upload? So you could upload a mission without a landing pattern with RTL_TYPE=0 and then change to RTL_TYPE=1 and in this case you will just fall back to using home landing (but not lose the mission)?

Hmmm... I'm not exactly sure what would happen in this scenario.



> After mission rejection, is the mission just cleared/dumped?

no, the mission stays onboard the vehicle.  However, it will be be executed.

hamishwillee: > Hmmm... I'm not exactly sure what would happen in this scenario.

Well the original docs said the thing would fall back to home position if no mission landing specified, so unless that code was pulled out, that is what I would think would happen. Can you find out please?

> no, the mission stays onboard the vehicle. However, it will be be executed.

@Antiheavy Do you mean "not executed"? Are there circumstances where it comes live again or is dead until (say) there has been a new successful upload?

hamishwillee:
Thanks for your help @Antiheavy . I have done minor tidy following our discussion. I'm going to merge as this is "better". This ensures the PX4v1.10 branch I will take shortly is more accurate.

PR #595, opened 2019-10-28
Battery failsafe behaviour simplified

hamishwillee:
Options changed in dialog. Came from https://github.com/PX4/Firmware/pull/13295

This is not expected in PX4 v 1.10 so hold off merging for now.


julianoes:
No, I don't think this should go into v1.10 already.

hamishwillee:
@julianoes I think this change didn't go ahead, so closing.

PR #598, opened 2019-11-01
Update motors page, add link to QGC UI

bkueng:
From https://github.com/mavlink/qgc-user-guide/pull/226.

This should go in after v1.10 (or after we branched off for that).


bkueng:
Sounds good.

PR #600, opened 2019-11-05
on fmu-v5 only the first 4 pins can be used for dshot

bkueng:


hamishwillee:
Thanks for adding. Worth also making the explicit comment that when Dshot is used you cant use the other pins for anything else?

bkueng:
They can still be used for camera triggering, for example.

hamishwillee:
Yes, but they can't be used for servos right?

bkueng:
So you're talking about the general case, not Pixhawk 4, pins 5-8? In that case, yes, if the servos don't support DShot (I'm not aware of any), then they cannot be used.
Essentially it's all DShot or all PWM for the FMU pins.

PR #603, opened 2019-11-08
Add Geofence information

hamishwillee:
This adds info/links to create Geofence plans. It also clarifies that the safety geofence failsafe is separate.


hamishwillee:
FYI, did not seek review of this, but have tested the behaviour.

PR #614, opened 2019-11-19
add section for data loss

baumanta:
Adding a paragraph to describe collision prevention behavior in case of range data loss


hamishwillee:
Thanks @baumanta - awesome. I did a minor subedit just for style and to link to associated mode docs.

PR #617, opened 2019-11-21
Draft connector selection

hamishwillee:
Hi @davids5 ,

OK, taken your advice and tried to put together a guide for users to work out what connectors are what.
It is DRAFT - I'm checking that this is what you had in mind at high level?

This PR puts it in the getting started section, but might be better as part of peripherals, or even part of the debugging doc in dev guide. On one hand it is useful generally for people to have a view on cables and this can be extended to batteries etc. 

Can you please confirm the part numbers - they are a guess. I am still seeking better images for the ports etc.


hamishwillee:
FYI, updated. More like expected structure. Still some "?" marks for things I don't know. If you have good images of 6 and 10 pin SUR connectors in place, would be helpful. Can you help find missing info?

davids5:
@DuganY can you please check the part numbers.

DuganY:
@hamishwillee the best way to get pictures and data sheets of connectors is to go to Digikey site and search on their site for the part number, for example: https://www.digikey.com/products/en?keywords=SAM8796-ND will give a picture for the Arm 10-pin.  I will upload a spreadsheet with part numbers and links to Digikey that shows what connectors are used on Pixhawk Debug adapter and what range of connectors can be used on Target boards.  I say range of connectors because Targets can use either vertical or side entry versions of the JST connectors and still be compatible.

[Connectors.xlsx](https://github.com/PX4/px4_user_guide/files/3882979/Connectors.xlsx)

hamishwillee:
@DuganY Thanks for the information in the spreadsheet.

> the best way to get pictures and data sheets of connectors is to go to Digikey site and search on their site for the part number, for example: https://www.digikey.com/products/en?keywords=SAM8796-ND will give a picture for the Arm 10-pin. 

Again thanks. I did actually do this for most cases, and it is definitely my fallback option. The problem is that in-situ the connectors often do not look that much like the what is in digikey. So it is actually more useful to show a picture of the connector taken from a real device.

image

What I'm also generally hoping for is some pictures of the boards with covers taken off so I can see the hidden ports - e.g. Pixhawk 4/4 mini. Do either of you guys also know where you can get the tool for opening the Pixhawk 4 screws?

DuganY:
I’m not certain, but it may be a Torx, see https://www.amazon.com/Kingsdun-Screwdriver-Screwdrivers-Stainless-Repairing/dp/B00MUJU33S/ref=sr_1_4?crid=19T43PC52GMST <https://www.amazon.com/Kingsdun-Screwdriver-Screwdrivers-Stainless-Repairing/dp/B00MUJU33S/ref=sr_1_4?crid=19T43PC52GMST&keywords=screwdriver+torx&qid=1574657079&sprefix=screw%2Caps%2C248&sr=8-4> &keywords=screwdriver+torx&qid=1574657079&sprefix=screw%2Caps%2C248&sr=8-4

davids5:
@hamishwillee This will help you https://www.newark.com/platinum-tools/19101/precision-driver-33-piece-set/dp/26Y8526?CMP=KNC-GUSA-GEN-KWL-DSA-PLATINUM-TOOLS&mckv=sde3AgXyg_dc|pcrid|384524617183|plid||kword||match|b|slid||product||pgrid|17246645061|ptaid|dsa-96407080821|&gclid=CjwKCAiAlO7uBRANEiwA_vXQ-_634pipWjEYgVHjwCA1Ij4ilTQVkdSRqA7EschF8D0vsIwWd5syJBoChTkQAvD_BwE

PR #630, opened 2019-12-19
Battery calibration - multiple batteries.

hamishwillee:
Hi @ItsTimmy,
This is battery calibration update. I have made 2 assumptions:
1. This is in master, not 1.10
2. QGC has not been updated to allow calibration of additional battery via the battery config UI (only via parameters)
3. There are still no battery failsafe settings. 

OK to merge?


ItsTimmy:
This is all good. But perhaps it might be a good idea to add information about migration from the old `BAT_` parameters to the new `BAT1_` parameters?

https://github.com/PX4/Firmware/blob/master/src/lib/battery/battery_params_deprecated.c

Whenever either the old parameter OR new parameter is changed, the value is copied into the other parameter, so they are always kept in sync in both directions. If the 2 parameters are different at boot, then the value of the old `BAT_` parameter is copied into the new `BAT1_` parameter.

hamishwillee:
Thanks @ItsTimmy ! 
I wasn't completely sure of the need for the parameter sync info as the need for it will disappear once people have updated. But I added it anyway, and also made changes to make it clear that QGC currently only sets battery 1. 

Merged!

PR #631, opened 2019-12-20
Update arming gesture info

hamishwillee:
Hi @MaEtUgR,
Finally got to this. Initially just the "advanced" page. When that is good will also update in other places.
Some questions inline.

Todo:
- [x] Update [Flying 101 > Arming](http://docs.px4.io/master/en/flying/basic_flying.html#arm)

### DIFF
--- a/en/advanced_config/prearm_arm_disarm.md
+++ b/en/advanced_config/prearm_arm_disarm.md
@@ -34,7 +34,9 @@ The button should be held down for ([nominally](#COM_RC_ARM_HYST)) one second to
 
 A two-position switch can also be used for arming/disarming, where the respective arm/disarm commands are sent on switch *transitions*.
 
-> **Tip** Use of a two-position switch is not recommended because it is easy to get the switch into an inconsistent state (i.e. if arming is rejected, the switch will stay armed even though the vehicle is not armed).
+> **Tip** Two-position arming switches are primarily used in/recommended for racing drones.
+
+is not recommended because it is easy to get the switch into an inconsistent state (i.e. if arming is rejected, the switch will stay armed even though the vehicle is not armed).
 
 The switch or button is assigned (and enabled) using [RC_MAP_ARM_SW](#RC_MAP_ARM_SW), and the switch "type" is configured using [COM_ARM_SWISBTN](#COM_ARM_SWISBTN).
 
--- a/en/advanced_config/prearm_arm_disarm.md
+++ b/en/advanced_config/prearm_arm_disarm.md
@@ -47,8 +47,6 @@ A two-position switch can also be used for arming/disarming, where the respectiv
 
 > **Tip** Two-position arming switches are primarily used in/recommended for racing drones.
 
-is not recommended because it is easy to get the switch into an inconsistent state (i.e. if arming is rejected, the switch will stay armed even though the vehicle is not armed).
-
 The switch or button is assigned (and enabled) using [RC_MAP_ARM_SW](#RC_MAP_ARM_SW), and the switch "type" is configured using [COM_ARM_SWISBTN](#COM_ARM_SWISBTN).
 
 
--- a/en/advanced_config/prearm_arm_disarm.md
+++ b/en/advanced_config/prearm_arm_disarm.md
@@ -30,7 +30,7 @@ RC controllers will have different gestures [based on their mode](../getting_sta
   - *Disarm:* Left stick to the bottom left.
 - **Mode 1**: 
   - *Arm:* Left-stick to right, right-stick to bottom.
-  - *Disarm:* - Left-stick to left, right-stick to the bottom.
+  - *Disarm:* Left-stick to left, right-stick to the bottom.
 
 The required hold time can be configured using [COM_RC_ARM_HYST](#COM_RC_ARM_HYST).
 
--- a/en/advanced_config/prearm_arm_disarm.md
+++ b/en/advanced_config/prearm_arm_disarm.md
@@ -4,13 +4,42 @@ Vehicles may have moving parts, some of which are potentially dangerous when pow
 
 To reduce the chance of accidents, PX4 has explicit state(s) for powering the vehicle components:
 - **Disarmed:** There is no power to motors or actuators.
-- **Pre-armed:** Actuators and other non-dangerous electronics are powered.
-  - In this state you can move ailerons, flaps etc, but motors/propellers are locked.
-- **Armed:** Vehicle is fully powered, including motors/propellers.
+- **Pre-armed:** Motors/propellers are locked but actuators for non-dangerous electronics are powered (e.g. ailerons, flaps etc.).
+- **Armed:** Vehicle is fully powered. Motors/propellers may be turning (dangerous!)
 
-A [safety switch](../getting_started/px4_basic_concepts.md#safety_switch) (if available) is used as a precondition for arming the vehicle, and may also be used to enable the pre-armed state. When permitted, arming is enabled using an arming sequence, switch or MAVLink command. 
+A [safety switch](../getting_started/px4_basic_concepts.md#safety_switch) (if available) is used as a precondition for arming the vehicle, and may also be used to enable the pre-armed state.
 
-PX4 allows you to configure how prearming, arming and disarming work using parameters (which can be edited in *QGroundControl* via the [parameter editor](../advanced_config/parameters.md)), as described in the following sections.
+When permitted, arming is enabled using an [arming gesture](#arm_disarm_gestures), [arming switch](#arm_disarm_switch) or MAVLink command. 
+
+PX4 allows you to configure how pre-arming, arming and disarming work using parameters (which can be edited in *QGroundControl* via the [parameter editor](../advanced_config/parameters.md)), as described in the following sections.
+
+> **Note** Arming/disarming parameters can be found in [Parameter Reference > Commander](../advanced_config/parameter_reference.md#commander) (search for `COM_ARM_` and `COM_DISARM_*`).
+
+## Arming Gesture {#arm_disarm_gestures}
+
+By default the vehicle is armed and disarmed by moving the RC throttle/yaw stick into a particular position and holding for for 1 second:
+- **Arming:** stick to *bottom right*.
+- **Disarming:** stick to *bottom left*. 
+
+The hold time can be configured using [COM_RC_ARM_HYST](#COM_RC_ARM_HYST).
+
+Parameter | Description
+--- | ---
+<span id="COM_RC_ARM_HYST"></span>[COM_RC_ARM_HYST](../advanced_config/parameter_reference.md#COM_RC_ARM_HYST) | Time RC stick must be held in arm/disarm position before arming/disarming occurs (default: 1 second).
+
+## Arming Switch {#arm_disarm_switch}
+
+An *Arm switch* can be configured to trigger arm/disarm, and will then be used instead of a [gesture](#arm_disarm_gestures) (which is disabled).
+
+The switch is assigned (and enabled) using [RC_MAP_ARM_SW](#RC_MAP_ARM_SW).
+The behaviour of the switch is then further configured using [COM_ARM_SWISBTN](#COM_ARM_SWISBTN).
+
+Parameter | Description
+--- | ---
+<span id="RC_MAP_ARM_SW"></span>[RC_MAP_ARM_SW](../advanced_config/parameter_reference.md#RC_MAP_ARM_SW) | Arm switch channel (default: 0 - unassigned). If defined the specified RC channel is used instead of stick.
+<span id="COM_ARM_SWISBTN"></span>[COM_ARM_SWISBTN](../advanced_config/parameter_reference.md#COM_ARM_SWISBTN) | Arm switch is only a button. 0: Arm switch is a switch that stays on when armed, 1: Arm switch is a button that only triggers arming and disarming.
+
+> **Note** The switch can also be set as part of *QGroundControl* [Flight Mode](../config/flight_mode.md) configuration.
 
 
 ## Auto-Disarming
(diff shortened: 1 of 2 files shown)


en/advanced_config/prearm_arm_disarm.md
```diff
@@ -4,13 +4,42 @@ Vehicles may have moving parts, some of which are potentially dangerous when pow
...
+
+An *Arm switch* can be configured to trigger arm/disarm, and will then be used instead of a [gesture](#arm_disarm_gestures) (which is disabled).
+
+The switch is assigned (and enabled) using [RC_MAP_ARM_SW](#RC_MAP_ARM_SW).
+The behaviour of the switch is then further configured using [COM_ARM_SWISBTN](#COM_ARM_SWISBTN).
+
+Parameter | Description
+--- | ---
+<span id="RC_MAP_ARM_SW"></span>[RC_MAP_ARM_SW](../advanced_config/parameter_reference.md#RC_MAP_ARM_SW) | Arm switch channel (default: 0 - unassigned). If defined the specified RC channel is used instead of stick.
+<span id="COM_ARM_SWISBTN"></span>[COM_ARM_SWISBTN](../advanced_config/parameter_reference.md#COM_ARM_SWISBTN) | Arm switch is only a button. 0: Arm switch is a switch that stays on when armed, 1: Arm switch is a button that only triggers arming and disarming.
```
hamishwillee: @MaEtUgR Don't understand at all what COM_ARM_SWISBTN is supposed to do - it what the difference between switch and button you are trying to imply is.

Is it that if set as a switch (0=default) the button state is locked? - so essentially if you move switch to armed state the arming command is "held down". Presumably this means that
-  if you can't arm you'll get a stream of messages "can't arm" back from the vehicle?
- in this case to disarm you move back to the disarm state.

It could alternatively be that when used as a switch you move to the position and hold for at least 1 second, wait for arming or disarming, then move back to the off position when that completes.

While for value of 1 perhaps it is a transitory switch. So command sent once. If you set to arm and it doesn't work, to try again you have to toggle back to disarm (command ignored) and then try again.

Can you confirm which/what?

MaEtUgR: > the button state is locked?
No, it's handled differently you have to retry again either way just on the switch it's the transition that triggers a command on the button it's holding it for the same amount of time like the stick gesture that triggers a command once.

Summary:
- If the arm switch RC channel **is a switch** set `COM_ARM_SWISBTN` 0 and
the **transition** of the switch position from disarmed to armed commands arm,
the transition from armed to disarm commands disarm.
If the command is rejected the switch is out of sync with the arm state and does not retry.
- If the arm switch RC channel **is a button or momentary switc**h set `COM_ARM_SWISBTN` 1
**holding the button down** while disarmed does the same as holding the stick arm gesture,
holding it down when armed does the same as holding the stick disarm gesture.
If you press the button indefinitely it only does the first thing no retry, no arm, disarm, arm disarm.

I consider the switch implementation is not that useful for a position-controlled flying product but rather for hobby pilots flying more manual modes e.g. FPV racing because in acro the disarm command usually succeeds and it's kind of your kill/disarm failsafe. While the button implementation is meant for hover platforms like bigger drones with a remote that has a button e.g. like the H520/ST16 has.

Does that help? 😇

hamishwillee:
Hi @MaEtUgR ,

Sorry for the delay. Thanks for your explanation - what really helped was the use of the term "momentary switch", which helped me understand what you mean by a button.

I have updated this section for your review. I also simplified the "Basic Concepts" section  because that went into too much detail for a beginner (who just needs to know that arming/disarming are  about safety).

en/advanced_config/prearm_arm_disarm.md
```diff
@@ -4,13 +4,47 @@ Vehicles may have moving parts, some of which are potentially dangerous when pow
...
+<span id="COM_RC_ARM_HYST"></span>[COM_RC_ARM_HYST](../advanced_config/parameter_reference.md#COM_RC_ARM_HYST) | Time that RC stick must be held in arm/disarm position before arming/disarming occurs (default: 1 second).
+
+## Arming Button/Switch {#arm_disarm_switch}
+
+An *arming button* or "momentary switch" can be configured to trigger arm/disarm *instead* of [gesture-based arming](#arm_disarm_gestures).
+The button should be held down for ([nominally](#COM_RC_ARM_HYST)) one second to arm (when disarmed) or disarm (when armed).
+
+A two position-switch can also be used for arming/disarming, where the respective arm/disarm commands are sent on switch *transitions*.
+
+> **Tip** Use of a two-position switch is not recommended because it is easy to get the switch into an inconsistent state (i.e. if arming is rejected, the switch will say armed even though the vehicle is not armed).
```
MaEtUgR:
```suggestion
> **Tip** Use of a two-position switch is not recommended because it is easy to get the switch into an inconsistent state (i.e. if arming is rejected, the switch will stay armed even though the vehicle is not armed).
```
Maybe it's worth noting this feature is intended for more of a race-style completely manual flying use case? An arm switch is the industry standard for racing drones.

en/advanced_config/prearm_arm_disarm.md
```diff
@@ -4,13 +4,47 @@ Vehicles may have moving parts, some of which are potentially dangerous when pow
...
+
+The switch or button is assigned (and enabled) using [RC_MAP_ARM_SW](#RC_MAP_ARM_SW), and the switch "type" is configured using [COM_ARM_SWISBTN](#COM_ARM_SWISBTN).
+
+
+Parameter | Description
+--- | ---
+<span id="RC_MAP_ARM_SW"></span>[RC_MAP_ARM_SW](../advanced_config/parameter_reference.md#RC_MAP_ARM_SW) | Arm switch channel (default: 0 - unassigned). If defined the specified RC channel is used instead of stick.
```
MaEtUgR: It should maybe be stressed that the default stick arming gesture is not being applied anymore when the switch/button method is configured. **Note:** That's RC only and does not apply for joysticks that send MAVLink commands through QGC.

en/getting_started/px4_basic_concepts.md
```diff
@@ -129,24 +129,20 @@ PX4 uses SD memory cards for storing [flight logs](../getting_started/flight_rep
...
 Vehicles may have moving parts, some of which are potentially dangerous when powered (in particular motors and propellers)!
 
-To reduce the chance of accidents, PX4 has explicit state(s) for powering the vehicle components:
-- **Disarmed:** There is no power to motors or actuators.
-- **Pre-armed:** Actuators and other non-dangerous electronics are powered.
-  - In this state you can move ailerons, flaps etc, but motors/propellers are locked.
-- **Armed:** Vehicle is fully powered, including motors/propellers.
+To reduce the chance of accidents:
+- PX4 vehicles are *disarmed* (unpowered) when not in use, and must be explicitly *armed* before taking off.
+- Some vehicles additionally require a [safety switch](../getting_started/px4_basic_concepts.md#safety_switch) be disengaged before arming can be attempted.
```
MaEtUgR:
```suggestion
- Some vehicles additionally require a [safety switch](../getting_started/px4_basic_concepts.md#safety_switch) be disengaged before arming can succeed.
```
Not sure here, just should say you can attempt but it will not arm since it's part of the checks.

en/advanced_config/prearm_arm_disarm.md
```diff
@@ -4,13 +4,47 @@ Vehicles may have moving parts, some of which are potentially dangerous when pow
...
+When permitted, arming is enabled using an [arming gesture](#arm_disarm_gestures), [arming switch](#arm_disarm_switch) or MAVLink command. 
+
+PX4 allows you to configure how pre-arming, arming and disarming work using parameters (which can be edited in *QGroundControl* via the [parameter editor](../advanced_config/parameters.md)), as described in the following sections.
+
+> **Note** Arming/disarming parameters can be found in [Parameter Reference > Commander](../advanced_config/parameter_reference.md#commander) (search for `COM_ARM_` and `COM_DISARM_*`).
+
+## Arming Gesture {#arm_disarm_gestures}
+
+By default the vehicle is armed and disarmed by moving the RC throttle/yaw stick into a particular position and holding for 1 second:
+- **Arming:** stick to *bottom right*.
```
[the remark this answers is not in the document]
hamishwillee: Ah, I'd assume that the mapping was based on the throttle stick, irrespective of mode. But what I think you're saying is that it is based on the value of throttle and yaw irrespective of sticks.

So *Arming* : Mode 2  (left) throttle/yaw stick to bottom right = throttle min (bottom)+ yaw max (right) = Mode 1 (left) yaw stick max (to right) and (right) throttle stick min (to bottom)

en/advanced_config/prearm_arm_disarm.md
```diff
@@ -4,13 +4,47 @@ Vehicles may have moving parts, some of which are potentially dangerous when pow
...
+Parameter | Description
+--- | ---
+<span id="COM_RC_ARM_HYST"></span>[COM_RC_ARM_HYST](../advanced_config/parameter_reference.md#COM_RC_ARM_HYST) | Time that RC stick must be held in arm/disarm position before arming/disarming occurs (default: 1 second).
+
+## Arming Button/Switch {#arm_disarm_switch}
+
+An *arming button* or "momentary switch" can be configured to trigger arm/disarm *instead* of [gesture-based arming](#arm_disarm_gestures).
+The button should be held down for ([nominally](#COM_RC_ARM_HYST)) one second to arm (when disarmed) or disarm (when armed).
+
+A two position-switch can also be used for arming/disarming, where the respective arm/disarm commands are sent on switch *transitions*.
```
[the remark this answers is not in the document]
hamishwillee: yes!

en/advanced_config/prearm_arm_disarm.md
hamishwillee: Thanks for that - I was wondering why we supported this. I've just changed this to say "Two-position arming switches are primarily used in/recommended for racing drones.". If you're in that business then you know that is the option to use. Otherwise not.

en/advanced_config/prearm_arm_disarm.md
```diff
@@ -4,13 +4,47 @@ Vehicles may have moving parts, some of which are potentially dangerous when pow
 
 To reduce the chance of accidents, PX4 has explicit state(s) for powering the vehicle components:
 - **Disarmed:** There is no power to motors or actuators.
```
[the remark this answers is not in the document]
hamishwillee: Not entirely sure about this, because we're talking about states, not what you see in QGC. I've added a note though "Ground stations may display *disarmed* for both disarmed and pre-armed states.
  While not technically correct for pre-armed vehicles, it is "safe"."

en/advanced_config/prearm_arm_disarm.md
hamishwillee: I had covered this, but have now put much stronger note, and added note about the joystick ...

en/getting_started/px4_basic_concepts.md
hamishwillee: Yes, this is a fine point, but a good one. I will just accept your suggestion.

en/advanced_config/prearm_arm_disarm.md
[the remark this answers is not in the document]
hamishwillee: OK, this is updated.

en/advanced_config/prearm_arm_disarm.md
```diff
@@ -4,13 +4,60 @@ Vehicles may have moving parts, some of which are potentially dangerous when pow
...
+## Arming Button/Switch {#arm_disarm_switch}
+
+An *arming button* or "momentary switch" can be configured to trigger arm/disarm *instead* of [gesture-based arming](#arm_disarm_gestures) (setting an arming switch disables arming gestures).
+The button should be held down for ([nominally](#COM_RC_ARM_HYST)) one second to arm (when disarmed) or disarm (when armed).
+
+A two-position switch can also be used for arming/disarming, where the respective arm/disarm commands are sent on switch *transitions*.
+
+> **Tip** Two-position arming switches are primarily used in/recommended for racing drones.
+
+is not recommended because it is easy to get the switch into an inconsistent state (i.e. if arming is rejected, the switch will stay armed even though the vehicle is not armed).
```
MaEtUgR:
```suggestion
The use of a two-position switch if not explicitly required is not recommended because it is easy to get the switch into an inconsistent state (i.e. if arming is rejected, the switch will stay armed even though the vehicle is not armed).
```

en/advanced_config/prearm_arm_disarm.md
```diff
@@ -4,13 +4,60 @@ Vehicles may have moving parts, some of which are potentially dangerous when pow
...
-PX4 allows you to configure how prearming, arming and disarming work using parameters (which can be edited in *QGroundControl* via the [parameter editor](../advanced_config/parameters.md)), as described in the following sections.
+When permitted, arming is enabled using an [arming gesture](#arm_disarm_gestures), [arming switch](#arm_disarm_switch) or MAVLink command. 
+
+PX4 allows you to configure how pre-arming, arming and disarming work using parameters (which can be edited in *QGroundControl* via the [parameter editor](../advanced_config/parameters.md)), as described in the following sections.
+
+> **Note** Arming/disarming parameters can be found in [Parameter Reference > Commander](../advanced_config/parameter_reference.md#commander) (search for `COM_ARM_` and `COM_DISARM_*`).
+
+## Arming Gesture {#arm_disarm_gestures}
+
+By default, the vehicle is armed and disarmed by setting RC throttle/yaw to particular values and holding for 1 second
```
MaEtUgR:
```suggestion
By default, the vehicle is armed and disarmed by moving RC throttle/yaw sticks to particular extremes and holding them for 1 second.
```

en/advanced_config/prearm_arm_disarm.md
hamishwillee: Whoops. That bit is supposed to be deleted.

hamishwillee:
Thanks @MaEtUgR !!!

PR #632, opened 2019-12-20
Update pre_flight_checks.md

NicolasM0:
Fix a typo (form copy/paste error) COM_ARM_EKF_POS --> COM_ARM_EKF_YAW in PREFLIGHT FAIL: EKF YAW ERROR section

remove COM_ARM_EKF_YAW section because it is redundant with the corrected section above and the first sentence is confusing (not a threshold in radians but an innovation ratio)

### DIFF
--- a/en/flying/pre_flight_checks.md
+++ b/en/flying/pre_flight_checks.md
@@ -29,6 +29,8 @@ The following errors (with associated checks and parameters) are reported by the
 * This error is produced when the yaw angle estimated using gyro data and the yaw angle from the magnetometer or external vision system are inconsistent.
 * Check the IMU data for large yaw rate offsets and check the magnetometer alignment and calibration.
 * The check is controlled by the [COM_ARM_EKF_YAW](../advanced_config/parameter_reference.md#COM_ARM_EKF_YAW) parameter
+* The default value of 0.5 allows the differences between the navigation yaw angle and magnetic yaw angle (magnetometer or external vision) to be no more than 50% of the maximum tolerated by the EKF and provides some margin for error increase when flight commences.
+* It can fail if the yaw gyro has a large offset or if the vehicle is moved or rotated in the presence of a bad magnetic interference or magnetometer calibration.
 
 `PREFLIGHT FAIL: EKF HIGH IMU ACCEL BIAS`:
 * This error is produced when the IMU accelerometer bias estimated by the EKF is excessive. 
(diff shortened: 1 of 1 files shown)


en/flying/pre_flight_checks.md
```diff
@@ -28,7 +28,7 @@ The following errors (with associated checks and parameters) are reported by the
 `PREFLIGHT FAIL: EKF YAW ERROR`:
 * This error is produced when the yaw angle estimated using gyro data and the yaw angle from the magnetometer or external vision system are inconsistent.
 * Check the IMU data for large yaw rate offsets and check the magnetometer alignment and calibration.
-* The check is controlled by the [COM_ARM_EKF_POS](../advanced_config/parameter_reference.md#COM_ARM_EKF_POS) parameter
+* The check is controlled by the [COM_ARM_EKF_YAW](../advanced_config/parameter_reference.md#COM_ARM_EKF_YAW) parameter
```
hamishwillee: @NicolasM0 Thanks for this. I'm a little concerned that some of the information from the removed section might be useful for people to know - e.g. the stuff around the default value setting. Thoughts?

@MaEtUgR Can you confirm this is OK, or suggest who might also review?

NicolasM0: I didn't add information from the removed section to be homogeneous with the other sections. But if you think it can be useful, I will add a point with: 
"The default value of 0.5 allows the differences between the navigation yaw angle and magnetic yaw angle (magnetometer or external vision) to be no more than 50% of the maximum tolerated by the EKF and provides some margin for error increase when flight commences. It can fail if the yaw gyro has a large offset or if the vehicle is moved or rotated in the presence of a bad magnetic interference or magnetometer calibration."

hamishwillee: Thanks! I'm hoping @MaEtUgR can comment; I'm no expert in this so just attempting to facilitate by bringing in someone who actually knows what they are talking about to to review :-)

MaEtUgR: I honestly don't know what the parameter exactly does, would need to check in the code. @bresch was likely already looking into it recently. I'll ask him.

hamishwillee: Thanks, looking forward to @bresch commenting!

bresch: Good catch @NicolasM0 , it's absolutely correct.
Can you add the point with your description of the test ratio?

NicolasM0:
@bresch done

hamishwillee:
Thanks @NicolasM0 ! (and @bresch  for review). Merged.

PR #636, opened 2020-01-08
added term "...valid mission landing..."

Antiheavy:
based on good suggestion in this comment:  https://github.com/PX4/Firmware/pull/13873#issuecomment-571825054

Added term "...valid mission landing..." to description for RTL_TYPE=1 in several places.

@hamishwillee can you check.  also, how does this tie into the Parameter descriptions?

### DIFF
--- a/en/flight_modes/return.md
+++ b/en/flight_modes/return.md
@@ -46,8 +46,8 @@ In this return type the vehicle:
 In this return type the vehicle:
 - Ascends to a safe [return altitude](#return_altitude) (above any expected obstacles).
 - Flies via direct path to a rally point or the start of a [mission landing pattern](#mission_landing_pattern) (whichever is closest).
-  If no mission landing or rally points are defined the vehicle instead returns home via a direct path.
-- If the destination is a mission landing pattern it will follow the pattern to land.
+  If no valid mission landing or rally points are defined the vehicle instead returns home via a direct path.
+- If the destination is a valid mission landing pattern it will follow the pattern to land.
 - If the destination is a rally point or home it will [land or wait](#arrival) at descent altitude (depending on landing parameters).
 
 <span id="mission_landing_pattern"></span>
@@ -158,7 +158,7 @@ Multicopters use a [home location return](#home_return) by default (and the foll
 ### Fixed Wing (FW)
 
 Fixed-wing aircraft use a [mission landing return type](#mission_landing_return) by default: 
-- If a mission landing is defined, fly direct to the mission landing start point and then land.
+- If a valid mission landing is defined, fly direct to the mission landing start point and then land.
 - Otherwise fly directly to the home position and circle above it at radius [NAV_LOITER_RAD](#NAV_LOITER_RAD).
 
 If not following a mission landing, and [RTL_LAND_DELAY](#RTL_LAND_DELAY) is set to -1, the vehicle will land as described in the topic: [Landing (Fixed Wing)](../flying/fixed_wing_landing.md).
@@ -168,7 +168,7 @@ RC stick movement is ignored.
 ### VTOL
 
 VTOL aircraft use a [mission landing return type](#mission_landing_return) by default: 
-- If a mission landing is defined, fly direct to the mission landing start point and then land.
+- If a valid mission landing is defined, fly direct to the mission landing start point and then land.
 - Otherwise fly directly to the home position, transition to multicopter mode, and land as a multicopter.
   > **Note** If not in a mission landing, a VTOL in FW mode will *always* transition back to MC just before landing (ignoring [NAV_FORCE_VT](../advanced_config/parameter_reference.md#NAV_FORCE_VT)).
 


sfuhrer:
Let's wait with that till we have https://github.com/PX4/Firmware/pull/13873 in. I basically think the "valid" makes sense, but it could also confuse the users, or I could be missing something.

hamishwillee:
@sfuhrer I was thinking of valid to differentiate the case where there is a mission start landing defined, and mission landing becomes the selected return destination, but for some reason it fails (i.e. it is "invalid") and some fallback is engaged.

Can you confirm that if mission landing is selected as a destination then it is valid. If that is the case then there is no need to really specify "valid" here.

sfuhrer:
> @sfuhrer I was thinking of valid to differentiate the case where there is a mission start landing defined, and mission landing becomes the selected return destination, but for some reason it fails (i.e. it is "invalid") and some fallback is engaged.
> 
> Can you confirm that if mission landing is selected as a destination then it is valid. If that is the case then there is no need to really specify "valid" here.

Okay I've dug a bit deeper in the code.

Currently all that is checked is if there is either a LAND or a DO_LAND_START during a "return to mission landing", so if RTL_TYPE =mission landing. For this type of RTL_TYPE and if a fixed-wing vehicle, the mission validity check checks if there is a DO_LAND_START marker in the mission, and otherwise rejects the mission. For fixed-wing and RTL_TYPE=mission landing, it is thus impossible to have a "invalid" landing. 
For VTOL it is not required to have either a DO_LAND_START marker or a simple LAND, so the landing can be "invalid" (in the meaning that there is NO landing) and thus fallback to RTL to HOME is possible.
For MC the same as for VTOL.

Summary: The is no valid or invalid landing, there either is or there is not (for MC and VTOL), and there always is a valid on for FW :)

The only corner case would be if the user changes RTL_TYPE during flight from 0 to 1, but doesn't upload a proper FW landing. Then the vehicle will still try to land at the LAND wp, without the proper approach, loiter etc.

Antiheavy:
> For fixed-wing and RTL_TYPE=mission landing, it is thus impossible to have a "invalid" landing.

ahhhhh  I wish this were the case!  Unfortunately it is trivially easy to have an "invalid" landing in PX4.  Actually, you can have an invalid mission for any reason.  All you have to do is upload it during flight!  Our users do this all the time.  Once uploaded, the mission is on the vehicle and the old "valid" mission is gone.  Mission feasibility checker will reject executing the new mission so it is true that it is impossible for PX4 to fly an invalid mission.  As you can see, this leads to a big problem with RTL_TYPE=mission landing and the vehicle will not actually come land at the mission landing (which is why the fallback to return to home was added to RTL_TYPE=mission landing in the first place).

Here is the fundamental problem: https://github.com/PX4/Firmware/issues/12473

hamishwillee:
@Antiheavy It's even messier now, because with  RTL_TYPE=1 we now select the closest destination of rally points or valid mission landing. So f you have no mission landing it should select an rally point. If you have a valid mission landing that is close it should be selected, and I would expect the destination to be re-evaluated if the mission changed (e.g. to select rally point or whatever).

Anyway .... @sfuhrer 

> For VTOL it is not required to have either a DO_LAND_START marker or a simple LAND, so the landing can be "invalid" (in the meaning that there is NO landing) and thus fallback to RTL to HOME is possible. For MC the same as for VTOL.

I think you are saying that a mission landing is defined as being a valid Fixed Wing landing with DO_LAND_START marker etc. So a MC or VTOL that just had a LAND command would not "count" as having a mission landing? That means that rally point would always be selected if defined in this case?

sfuhrer:
> > For fixed-wing and RTL_TYPE=mission landing, it is thus impossible to have a "invalid" landing.
> 
> ahhhhh I wish this were the case! Unfortunately it is trivially easy to have an "invalid" landing in PX4. Actually, you can have an invalid mission for any reason. All you have to do is upload it during flight! Our users do this all the time. Once uploaded, the mission is on the vehicle and the old "valid" mission is gone. Mission feasibility checker will reject executing the new mission so it is true that it is impossible for PX4 to fly an invalid mission. As you can see, this leads to a big problem with RTL_TYPE=mission landing and the vehicle will not actually come land at the mission landing (which is why the fallback to return to home was added to RTL_TYPE=mission landing in the first place).
> 
> Here is the fundamental problem which leads to the requirement to alert the user that the drone is not actually going to come back and land using the mission landing pattern like they want it to: [PX4/Firmware#12473](https://github.com/PX4/Firmware/issues/12473)

But this means not that the landing is invalid, it means that there is **no** landing (because there is no mission) in case it got rejected when in-air, right? This sounds like it should be fixed asap, agree..

sfuhrer:
> @Antiheavy It's even messier now, because with RTL_TYPE=1 we now select the closest destination of rally points or valid mission landing. So f you have no mission landing it should select an rally point. If you have a valid mission landing that is close it should be selected, and I would expect the destination to be re-evaluated if the mission got updated (e.g. to select rally point or the mission landing if closer and still valid).
> ie. there is no longer the assumption that if you select RTL_TYPE=1 you really really want to use the mission landing.
> 
> Anyway .... @sfuhrer
> 
> > For VTOL it is not required to have either a DO_LAND_START marker or a simple LAND, so the landing can be "invalid" (in the meaning that there is NO landing) and thus fallback to RTL to HOME is possible. For MC the same as for VTOL.
> 
> I think you are saying that a mission landing is defined as being a valid Fixed Wing landing with DO_LAND_START marker etc. So a MC or VTOL that just had a LAND command would not "count" as having a mission landing? That means that rally point would always be selected if defined in this case?

Yes, for the fixed-wing mission to not be rejected it needs to contain a DO_LAND_START marker. 
For VTOL and MC, this is though not required, neither is is required to have a mission with a LAND. That's all concerning the mission feasibility. The other part is the RTL logic, which (obviously) only can execute a mission landing if there is any landing defined (either with or without a DO_LAND_START). As MC/VTOL mission do not have to contain any landing, it is possible to have the fallback to RTL to HOME.

Antiheavy:
> Yes, for the fixed-wing mission to not be rejected it needs to contain a DO_LAND_START marker.
> For VTOL and MC, this is though not required, neither is is required to have a mission with a LAND.

@sfuhrer is this actually true for the specific case of RTL_TYPE=1 (use mission landing)?  It seems to me that if RTL_TYPE=1 is selected by the user or system designer, the intention is to require a mission landing "pattern" for RTL purposes.  This is often the best choice for fixed wings, but can apply to VTOL and even MC too.  

In the general case, a DO_LAND_START is not required for a fixed wing mission (neither is a LAND for that matter).  When there is a LAND, the fixed wing auto land feature works fine without a DO_LAND_START, it just needs a preceding WP to define the approach.  The DO_LAND_START becomes important and useful for things like RTL_TYPE=1 to mark the location we'd like to use as the landing pattern portion of the mission.  This is why the landing pattern requirement (DO_LAND_START, minimum of 1 WP, and a LAND) was added to mission feasibility checker in the specific case of RTL_TYPE=1.  By selecting RTL_TYPE=1 the user/designer is choosing to require those things in the mission for the purpose of RTL.

If the user/designer does not want to require those things (for example because their use cases allow them to execute RTL at other places like Home, Rally Points, etc) then there are better RTL_TYPE options for them to choose.



> But this means not that the landing is invalid, it means that there is **no** landing (because there is no mission) in case it got rejected when in-air, right?

well, from the PX4 Navigator perspective this is partially true.  Technically the mission is there onboard, but PX4 refuses to fly it.  This PR attempts to clarify the documentation around the idea that the onboard mission also needs to be "valid" (or "feasible" in PX4 speak) in order to use the mission landing for RTL.



> This sounds like it should be fixed asap, agree..

Yes!  I've been advocating for it to be "fixed" for a long time.  I don't have the skills to fix it myself so I can't complain too loudly (but I have offered to 'pay' people).  There are several approaches to solving it, but they aren't easy so it hasn't been done (yet).

hamishwillee:
@sfuhrer

> @sfuhrer is this actually true for the specific case of RTL_TYPE=1 (use mission landing)? It seems to me that if RTL_TYPE=1 is selected by the user or system designer, the intention is to require a mission landing "pattern" for RTL purposes. This is often the best choice for fixed wings, but can apply to VTOL and even MC too.

This is what I was trying to convey in our earlier RTL discussions. The RTL_TYPE=1  ONLY exists because FW needs them and needs to always land using a landing pattern. The fallback to home should in theory never be needed on FW. Choosing a rally point is also always a mistake with this type. Yes I know that people should "just not define rally points" or "they defined rally points so they must have known they wanted to use them", but in fact if they use this RTL type we're saying "no, we really really really want to be able to land using the mission".

The type makes no sense for VTOL or MC, unless we're forcing VTOL to land using FW semantics (we're not).

Rant over :-)

en/flight_modes/return.md
```diff
@@ -168,7 +168,7 @@ RC stick movement is ignored.
 ### VTOL
 
 VTOL aircraft use a [mission landing return type](#mission_landing_return) by default: 
-- If a mission landing is defined, fly direct to the mission landing start point and then land.
+- If a valid mission landing is defined, fly direct to the mission landing start point and then land.
```
hamishwillee: Is this true? VTOL uses mission landing by default? Why?

hamishwillee:
> The other part is the RTL logic, which (obviously) only can execute a mission landing if there is any landing defined (either with or without a DO_LAND_START). As MC/VTOL mission do not have to contain any landing, it is possible to have the fallback to RTL to HOME.

Does not quite answer the question. It is clearly stated that there must be an FW mission landing pattern defined (DO_LAND_START, minimum of 1 WP, and a LAND) for mission landing to be selected as the destination. Correct? 

For MC/VTOL and RTL_TYPE=1 what would cause the mission landing to be selected? Is it the full landing pattern as for FW, or some other landing pattern definition?

I'm pretty sure that this PR is "ok". Ie the addition of valid is pointless but does not harm. Because a mission landing pattern as defined above is valid, and anything else is a mission plan without a landing pattern.

Antiheavy:
>  The RTL_TYPE=1 ONLY exists because FW needs them and needs to always land using a landing pattern.

The FW need is why we added it to begin with, but there is no reason a VTOL or MC can't use this if they really wanted to.



> It is clearly stated that there must be an FW mission landing pattern defined (DO_LAND_START, minimum of 1 WP, and a LAND) for mission landing to be selected as the destination. Correct?

To my above point, the existing requirement isn't necessarily to be a"FW landing pattern".  It's generalize to just a "landing pattern".   There was some recent work for landing patterns for VTOL's that include a DO_LAND_START, a WP to start a fixed wing glide slope, and they add a FW->MC transition command prior to vertical landing at the LAND waypoint.  This is completely within the existing functionality of RTL_TYPE=1.  Theoretically someone flying pure MC could still define a mission landing pattern that had them fly a set route to avoid flying over or into certain areas during transit back to a LAND waypoint - again all within the current boundaries of RTL_TYPE=1.   If MissionFeasibiltyChecker rejects the mission  for any reason (or the mission is deleted by the user) during flight then there will no longer be a valid mission landing pattern to follow for RTL_TYPE=1.   Hence the need to still have the basic fallback to Return to Home.



> Choosing a rally point is also always a mistake with this type. Yes I know that people should "just not define rally points" or "they defined rally points so they must have known they wanted to use them", but in fact if they use this RTL type we're saying "no, we really really really want to be able to land using the mission".

@hamishwillee you and I are in agreement here, but that is maybe another topic.

hamishwillee:
@Antiheavy 

> The FW need is why we added it to begin with, but there is no reason a VTOL or MC can't use this if they really wanted to.
> It's generalized to just a "landing pattern". 

Sure. but I need that to be clearly stated by whoever owns that - hence the question above "For MC/VTOL and RTL_TYPE=1 what would cause the mission landing to be selected? Is it the full landing pattern as for FW, or some other landing pattern definition?"

Yes, we probably should off the broader discussion :-)

en/flight_modes/return.md
sfuhrer: Yes, because when you plan the VTOL mission appropriately (which is with a DO_LAND_START and a loiter to altitude before transition) then is he much safer/nicer landing compared to RTL_HOME.

sfuhrer:
> Does not quite answer the question. It is clearly stated that there must be an FW mission landing pattern defined (DO_LAND_START, minimum of 1 WP, and a LAND) for mission landing to be selected as the destination. Correct?

Where is that stated? AFAIK it only requires a DO_LAND_START marker.

sfuhrer:
> For MC/VTOL and RTL_TYPE=1 what would cause the mission landing to be selected? Is it the full landing pattern as for FW, or some other landing pattern definition?

DO_LAND_START or LAND

sfuhrer:
> This is what I was trying to convey in our earlier RTL discussions. The RTL_TYPE=1 ONLY exists because FW needs them and needs to always land using a landing pattern. The fallback to home should in theory never be needed on FW. Choosing a rally point is also always a mistake with this type. Yes I know that people should "just not define rally points" or "they defined rally points so they must have known they wanted to use them", but in fact if they use this RTL type we're saying "no, we really really really want to be able to land using the mission".

>The type makes no sense for VTOL or MC, unless we're forcing VTOL to land using FW semantics (we're not).

This type makes a **lot** of sense for us using VTOL, even if we don't enforce atm having a landing pattern. Because if we make the landing pattern (what we internally always do), then (I'm repeating myself) this RTL is much better then HOME. For VTOL it is though also totally legitimate to have Rally points, that's why this RTL type makes for VTOL the most sense. 
If it is important to you both to not enable rally points at all in this mode for FW, then we should make yet another separate RTL_TYPE to cover that . Or in the mission feasibility check: do not allow missions with a rally point being updated when RTL_TYPE=1 and fixed-wing). What do you think?

hamishwillee:
I have decided to close this. I don't think the term adds enough value when compared to the confusion of the actual implementation.

PR #640, opened 2020-01-20
Traffic Collision Avoidance, Treat UAVS differently from manned aviation

LowOrbitIonCannon:
UPDATING Docs to [#13159](https://github.com/PX4/Firmware/pull/13159)

@hamishwillee do you know of something else that needs to be changed?

This pull will need updates too:[px4_user_guide/pull/527](https://github.com/PX4/px4_user_guide/pull/527)

### DIFF
--- a/en/advanced_features/traffic_avoidance_adsb.md
+++ b/en/advanced_features/traffic_avoidance_adsb.md
@@ -84,9 +84,11 @@ Parameter | Description
 
 PX4 listens for valid transponder reports during missions.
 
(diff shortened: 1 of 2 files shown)


hamishwillee:
Thanks very much. I have subedited. Would be good if you could also tidy the params as I have indicated in https://github.com/PX4/Firmware/pull/13159/files#r372709230

hamishwillee:
Also fixed up https://github.com/PX4/px4_user_guide/pull/527 and merged. Thanks for noting that!

PS Sorry for delay on this. Was on holiday.

PR #642, opened 2020-01-28
missions: clarify how to set yaw

hamishwillee:
This comes up quite a lot. Thought it useful to add a link. 

Note, behaviour differs slightly from MAVLink spec, which indicates that if yaw not set then the vehicle heading should be unchanged.


LorenzMeier:
@hamishwillee Lets augment the MAVlink spec to say that NaN means uncontrolled which means the vehicle stays in the current heading mode. I would add a heading mode command that allows us to ditch the parameter eventually which does the same as the param.

PR #649, opened 2020-02-05
Add HKPilot32 SWD link

hamishwillee:
Adds link from HKPilot32 debug port info to 3DR Pixhawk 1 page - they are pretty much the same under the hood.

@davids5 Do you happen to know whether Pixfalcon has an SWD or Serial console interface? If so, do you have any imagery? No worries if not, I will just omit it for now.

### DIFF
--- a/en/flight_controller/HKPilot32.md
+++ b/en/flight_controller/HKPilot32.md
@@ -63,6 +63,11 @@ To [build PX4](https://dev.px4.io/master/en/setup/building_px4.html) for this ta
 make px4_fmu-v2_default
 ```
 
+## Debug Ports
+
+See [3DR Pixhawk 1 > Debug Ports](../flight_controller/pixhawk.md#debug-ports).
+
+
 ## Pinouts and Schematics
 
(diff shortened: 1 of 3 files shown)


davids5:
@hamishwillee if it is the same as the old 3dr PH mini, then no, neither, you have to solder wires to the board test pads for SWD and IC to a console.

hamishwillee:
@davids5 Thanks. FMI, what does it mean to say " and IC to a console."?

I have added debug port sections to the Pixfalcon and 3DR mini pages stating the port is not present and adding your text above. While it would be nice to add images showing the ports/wiring I figure anyone who wants to do this will be willing to do some digging. Making it clear that this won't be easy is enough to help most users.

davids5:
> @davids5 Thanks. FMI, what does it mean to say " and IC to a console."

Soldering wires on to the STM32F4 (IC) TX and RX to get a console.

hamishwillee:
David - thank you very much for clarification. Merged.

PR #665, opened 2020-02-18
Pixracer and pixhawk3 pro debug info

hamishwillee:
This move info about debug ports from DevGuide to user guide.

Specifically the Pixracer and Pixhawk3Pro apparently use the Pixhawk Debug Port standard - according to http://dev.px4.io/master/en/debug/system_console.html#pixracer--pixhawk-v3-wiring-the-console

So this adds that information and then links to the system console docs in devguide for wiring [which is in almost complete state here: https://github.com/PX4/Devguide/pull/953

@davids5 If you can confirm these use the debug port as per standard, that would be great!


hamishwillee:
FMI, when this merges, update system console doc in devguide to link here rather than duplicate

PR #671, opened 2020-02-26
UAVCAN_ESCs: SD Card dependency

hamishwillee:
Follows on from https://github.com/PX4/Firmware/issues/14206

- [x] update https://docs.px4.io/master/en/power_module/

This is dependent on answer here: https://github.com/PX4/Devguide/pull/962/files#r384538073
- ie "what are the end user symptoms" of having no SD card. The current text is just a placeholder.

### DIFF
--- a/en/peripherals/uavcan_escs.md
+++ b/en/peripherals/uavcan_escs.md
@@ -140,6 +140,10 @@ There is some guidance here: [Quick start guide for Myxa v0.1](https://forum.zub
 
 For VESC the preferred tool for motor enumeration is the [VESC tool](https://vesc-project.com/vesc_tool).
 
+### Troubleshooting
+
+PX4 requires an SD card for UAVCAN node allocation and firmware upgrade (both of which happen during boot).
+If motors simply don't start, check that there is a (working) SD card present and reboot.
 
 ## Further Information
 


hamishwillee:
Hi @bkueng 

As I understand it, UAVCan will fail if there is no SD card (at boot).  I added this troubleshooting note to (just the ESC docs) to note this in case. I have no idea what the symptoms or how to confirm, but I figure this is useful for end-users because at least they can try it. 

If you think this is useful, please approve.

----

At the moment UAVCAN docs are split between [devguide](http://dev.px4.io/master/en/uavcan/) and user guide ESC and UAVCAN PM docs. This is an evolution because originally UAVCAN was pretty theoretical, and now it has some reality.

Would appreciate your advice on:
1. Is it time to move all UAVCAN docs from devguide into user guide?
2. If not, what would trigger for that be?

bkueng:
I'd say we can move 'UAVCAN Enumeration' and FAQ/troubleshooting to the user-guide, but keep the bootloader and FW updates in the dev-guide as this is part of development and not what users need to do.

To '"what are the end user symptoms" of having no SD card', I don't know either, but is worth documenting. 
@cmic0 might?

LorenzMeier:
@hamishwillee Now that they are merged, could you update?

hamishwillee:
Yes. This is on my list but requires a little thought. I plan to move most of the UAVCAN setup into the main guide and just keep "development topics" in the developer section. Probably I will start again on the actual work but have been keeping this PR alive as a reminder.

LorenzMeier (approved):

hamishwillee:
Closing. Everything done in another PR.

PR #673, opened 2020-03-04
Documentation of the new rpm sensor

kaklik:
User documentation of  RPM sensor added in following pull-request: https://github.com/PX4/Firmware/pull/14018


hamishwillee:
Two of the images have large file sizes - can you please reduce the size so that there is less download cost for people with poor internet (e.g. across great wall in china). For example, for tfrpm01_electronics.jpg you can cut out all the empty whitespace around the image and resize width to about 500px and it shrinks to 79K without losing any information in rendered version.

en/sensor/rpm.md
```diff
@@ -0,0 +1,40 @@
+# ThunderFly TFRPM01 tachometer sensor
```
hamishwillee: This document will have to be added to en/SUMMARY.md or it will not render (just follow pattern of other docs).

en/getting_started/sensor_selection.md
```diff
@@ -35,6 +35,15 @@ They are so important because the autopilot does not have other means to detect
...
 
+## Tachometer
+
+[Revolution-counter sensors](https://en.wikipedia.org/wiki/Tachometer#In_automobiles,_trucks,_tractors_and_aircraft) are *highly recommended* for rotor-wing frames.
+
+They are so important because the autopilot does not have other means to detect stall or another rotor failure. For rotor-wing flight it is the rotation of blades that guarantees lift not airspeed or ground speed.
+
+![Digital RPM sensor](../../images/tachometer_sensor.jpg)
+
+For more information and recommended hardware see: [Tachometer Sensors](../sensor/rpm.md).
```
hamishwillee: If you look at the docs we typically have a parent page explaining what the sensor does, then "instances" below. Can you create a parent page under sensors for tachometers and then link to that (rather than going straight to your implementation). 
Note, otherwise what happens is that these is no obvious way for people to extend and add other implementations.
The parent page can have most of the same info you have here and any **generic** config for tachometers - though it doesn't look like there is any of that (I assume `PCF8583_POOL` etc are specific to your tachometer?)

en/sensor/rpm.md
```diff
@@ -0,0 +1,40 @@
...
+
+The [TFRPM01](https://github.com/ThunderFly-aerospace/TFRPM01) is a small, and low system demanding revolution-counter.
+It must be connected to a I²C bus.
+
+![TFRPM01A](../..//assets/hardware/sensors/tfrpm/tfrpm01_electronics.jpg)
+
```
hamishwillee: I think at this point it is also worth adding a note that it is generic counter that must be connected to a specific sensor instance - then maybe link down to examples

en/sensor/rpm.md
kaklik: Document listing added.

hamishwillee:
Thanks @kaklik . I have merged this so I can update it further (see https://github.com/PX4/px4_user_guide/pull/676)

I still have some questions:
1. You say "The sensor/probe hardware needs an pulse signal." - is there any definition of this signal? ie. voltage, max/min frequency etc? E.g. if someone wants to use a probe how can they check it is compatible?
2. What diagnostics does the LED provide? ie. just "on/off" or something else? Good to add that information.
3. Is there a purchase link we can add to "5100 Miniature Flange Mounting Proximity Sensor"

PR #674, opened 2020-03-05
Add Freefly RTK GPS

raulrff:
Adding newly-released RTK GPS by Freefly Systems.


hamishwillee:
Hi @raulrff 

Thanks - I did some minor layout fixes but this was good.

PR #680, opened 2020-03-11
Add information to path planning re: Bezier curves

jkflying:
This is a first pass. I'm not really happy with the formatting, but I leave that to the expert :-p

Right now there is nothing available to send the Bezier curves, so there it isn't much more to add yet.

### DIFF
--- a/en/computer_vision/path_planning_interface.md
+++ b/en/computer_vision/path_planning_interface.md
@@ -1,13 +1,15 @@
 # Path Planning Offboard Interface
 
 PX4 uses a number of MAVLink interfaces for integrating path planning services from a companion computer (including obstacle avoidance in missions, [safe landing](../computer_vision/safe_landing.md), and future services):
-- The [MAVLink Path Planning Protocol](https://mavlink.io/en/services/trajectory.html) message [TRAJECTORY_REPRESENTATION_WAYPOINTS](https://mavlink.io/en/messages/common.html#TRAJECTORY_REPRESENTATION_WAYPOINTS) is used to sent the current and next waypoint, and receive a stream of setpoints for the new path.
-- Alternatively, via mavlink the [TRAJECTORY_REPRESENTATION_BEZIER](https://mavlink.io/en/messages/common.html#TRAJECTORY_REPRESENTATION_BEZIER) can be used, parameterized by time, to specify the moving position setpoint of the vehicle over a given amount of time.
+- There are two [MAVLink Path Planning Protocol](https://mavlink.io/en/services/trajectory.html) interfaces:
+  > **Warning** Route planning software should not mix these interfaces while executing a task (PX4 will use the last received message of either type).
+
+  - The [Waypoint Trajectory Interface](#waypoint_interface) uses [TRAJECTORY_REPRESENTATION_WAYPOINTS](https://mavlink.io/en/messages/common.html#TRAJECTORY_REPRESENTATION_WAYPOINTS) to sent the current and next waypoint, and to receive a stream of setpoints for the new path.
+  - The [Bezier Trajectory Interface](#bezier_interface) uses [TRAJECTORY_REPRESENTATION_BEZIER](https://mavlink.io/en/messages/common.html#TRAJECTORY_REPRESENTATION_BEZIER) to specify the moving position setpoint of the vehicle over a given amount of time.
 - The [HEARTBEAT/Connection Protocol](https://mavlink.io/en/services/heartbeat.html) is used for "proof of life" detection.
 - [LOCAL_POSITION_NED](https://mavlink.io/en/messages/common.html#LOCAL_POSITION_NED) and [ALTITUDE](https://mavlink.io/en/messages/common.html#ALTITUDE) send the vehicle local position and altitude, respectively.
 
 > **Tip** The message flows from PX4 UORB topics, through MAVLink, to ROS and back again are all documented in: [PX4/avoidance > Message Flows](https://github.com/PX4/avoidance#message-flows).
-
 
 All services that use this interface send and receive messages of the same type/format.
 Developers can therefore use this interface to create their own new companion-side path planning services, or tweak the existing planner software.
@@ -110,14 +112,18 @@ PX4 safely handles the case where messages are not received from the offboard sy
 
 ## Bezier Trajectory Interface {#bezier_interface)
 
-If the obstacle avoidance interface is enabled, it can also receive mavlink messages in the [TRAJECTORY_REPRESENTATION_BEZIER](https://mavlink.io/en/messages/common.html#TRAJECTORY_REPRESENTATION_BEZIER) format. This is parsed as follows:
+If the obstacle avoidance interface is enabled, it can also receive MAVLink messages in the [TRAJECTORY_REPRESENTATION_BEZIER](https://mavlink.io/en/messages/common.html#TRAJECTORY_REPRESENTATION_BEZIER) format.
+This is parsed as follows:
 
-- The number of Bezier control points determines the degree of the Bezier curve. For example, 3 points makes a quadratic bezier curve with constant acceleration.
-- The Bezier curve must be the same degree in x,y,z and yaw, with all Bezier control points finite
-- The `delta` array should have the value corresponding with the last Bezier control point indicate the duration that the waypoint takes to execute the curve to that point, from beginning to end. Other values in the `delta` array are ignored.
-- The timestamp of the mavlink message should be the time that the curve starts, and communication delay and clock mismatch will be compensated for on the flight controller via the timesync mechanism.
+- The number of bezier control points determines the degree of the bezier curve.
+  For example, 3 points makes a quadratic bezier curve with constant acceleration.
+- The bezier curve must be the same degree in x, y, z, and yaw, with all bezier control points finite
+- The `delta` array should have the value corresponding with the last bezier control point indicate the duration that the waypoint takes to execute the curve to that point, from beginning to end.
+  Other values in the `delta` array are ignored.
+- The timestamp of the MAVLink message should be the time that the curve starts, and communication delay and clock mismatch will be compensated for on the flight controller via the timesync mechanism.
 - The control points should all be specified in local coordinates.
-- Bezier curves do not only expire after 0.5s like the waypoint interface, but also after the execution time of the bezier curve has been reached. Be careful not to specify such short times for the bezier curve that you cannot send the next one in time.
+- Bezier curves do not only expire after 0.5s like the waypoint interface, but also after the execution time of the bezier curve has been reached.
+  Be careful not to specify such short times for the bezier curve that you cannot send the next one in time.
 
 ## Supported Hardware
 
@@ -125,6 +131,6 @@ Tested companion computers and cameras are listed in [PX4/avoidance](https://git
 
 
 <!-- ## Further Information -->
-<!-- @mrivi is expert! -->
+<!-- @mrivi and @jkflying are the experts! -->
 <!-- Issue with discussion : https://github.com/PX4/Devguide/issues/530 -->
-<!-- PR for mavlink docs: https://github.com/mavlink/mavlink-devguide/pull/133 -->
+<!-- PR for MAVLink docs: https://github.com/mavlink/mavlink-devguide/pull/133 -->
--- a/en/computer_vision/path_planning_interface.md
+++ b/en/computer_vision/path_planning_interface.md
@@ -1,4 +1,4 @@
-# Path Planning Offboard Interface
+# Path Planning Interface
 
 PX4 uses a number of MAVLink interfaces for integrating path planning services from a companion computer (including obstacle avoidance in missions, [safe landing](../computer_vision/safe_landing.md), and future services):
 - There are two [MAVLink Path Planning Protocol](https://mavlink.io/en/services/trajectory.html) interfaces:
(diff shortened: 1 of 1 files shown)


jkflying:
Whoops, the Bezier isn't available in Offboard (I think). That would need to be clarified as well. It is only available in COM_OBS_AVOID=1 mode.

en/computer_vision/path_planning_interface.md
```diff
@@ -107,6 +108,17 @@ PX4 safely handles the case where messages are not received from the offboard sy
...
+
+- The number of Bezier control points determines the degree of the Bezier curve. For example, 3 points makes a quadratic bezier curve with constant acceleration.
+- The Bezier curve must be the same degree in x,y,z and yaw, with all Bezier control points finite
+- The `delta` array should have the value corresponding with the last Bezier control point indicate the duration that the waypoint takes to execute the curve to that point, from beginning to end. Other values in the `delta` array are ignored.
+- The timestamp of the mavlink message should be the time that the curve starts, and communication delay and clock mismatch will be compensated for on the flight controller via the timesync mechanism.
+- The control points should all be specified in local coordinates.
+- Bezier curves do not only expire after 0.5s like the waypoint interface, but also after the execution time of the bezier curve has been reached. Be careful not to specify such short times for the bezier curve that you cannot send the next one in time.
```
hamishwillee: 1. What happens if the curve expires? Is this the same as message not being received for 0.5s? 
2. Are the fault handling rules for losing the companion-side otherwise the same [if so, I might move the section "Companion Failure Handing" down to a common area.

en/computer_vision/path_planning_interface.md
```diff
@@ -107,6 +108,17 @@ PX4 safely handles the case where messages are not received from the offboard sy
...
 
+## Bezier Trajectory Interface {#bezier_interface)
+
+If the obstacle avoidance interface is enabled, it can also receive mavlink messages in the [TRAJECTORY_REPRESENTATION_BEZIER](https://mavlink.io/en/messages/common.html#TRAJECTORY_REPRESENTATION_BEZIER) format. This is parsed as follows:
+
+- The number of Bezier control points determines the degree of the Bezier curve. For example, 3 points makes a quadratic bezier curve with constant acceleration.
+- The Bezier curve must be the same degree in x,y,z and yaw, with all Bezier control points finite
+- The `delta` array should have the value corresponding with the last Bezier control point indicate the duration that the waypoint takes to execute the curve to that point, from beginning to end. Other values in the `delta` array are ignored.
+- The timestamp of the mavlink message should be the time that the curve starts, and communication delay and clock mismatch will be compensated for on the flight controller via the timesync mechanism.
+- The control points should all be specified in local coordinates.
```
hamishwillee: This means FRD relative to vehicle center-of-gravity and as shown here right?: https://docs.px4.io/master/en/getting_started/px4_basic_concepts.html#heading-and-directions ?

en/computer_vision/path_planning_interface.md
```diff
@@ -107,12 +110,27 @@ PX4 safely handles the case where messages are not received from the offboard sy
...
+
+- The number of bezier control points determines the degree of the bezier curve.
+  For example, 3 points makes a quadratic bezier curve with constant acceleration.
+- The bezier curve must be the same degree in x, y, z, and yaw, with all bezier control points finite
+- The `delta` array should have the value corresponding with the last bezier control point indicate the duration that the waypoint takes to execute the curve to that point, from beginning to end.
+  Other values in the `delta` array are ignored.
+- The timestamp of the MAVLink message should be the time that the curve starts, and communication delay and clock mismatch will be compensated for on the flight controller via the timesync mechanism.
+- The control points should all be specified in local coordinates.
+- Bezier curves do not only expire after 0.5s like the waypoint interface, but also after the execution time of the bezier curve has been reached.
+  Be careful not to specify such short times for the bezier curve that you cannot send the next one in time.
```
hamishwillee: 1. What happens if the curve expires? Is this the same as message not being received for 0.5s? 
2. Are the fault handling rules for losing the companion-side otherwise the same [if so, I might move the section "Companion Failure Handing" down to a common area.

hamishwillee:
@jkflying I've done a minor restructure and added a few questions, but bigger picture I still don't understand this.

1. For waypoint trajectory PX4 sends desired path and planner sends back setpoints. What gets sent out by PX4 in order to get bezier setpoints?
2. [TRAJECTORY_REPRESENTATION_BEZIER.pos_x](https://mavlink.io/en/messages/common.html#TRAJECTORY_REPRESENTATION_BEZIER) is an array of 5 values with description "X-coordinate of starting bezier point, set to NaN if not being used". I'm guessing this an array of X starting points for 5 points? So the description should actually be "X-coordinates of bezier points (index 0 is the start point), set to NaN if not being used"
3.  I think we need a diagram because from this description I don't understand the message flow, or what the "start of the curve" means in relations to the timestamp. I envisage PX4 might send a couple of waypoints to the companion. The companion works out a nice curve to that/several points, and gets back a curve that does the first .5s of that. Otherwise much like the waypoint approach. 
I also don't get what "start of the curve" means for first point and how the delta thingy works. It might be that the curve defines a whole path and each message says we're just going to do 0.5seconds travel of that.
Anyway, this is clearly obvious to you, but I hope you can see that this is not enough to understand what gets sent/received.

en/computer_vision/path_planning_interface.md
```diff
@@ -1,12 +1,18 @@
...
+  - The [Bezier Trajectory Interface](#bezier_interface) uses [TRAJECTORY_REPRESENTATION_BEZIER](https://mavlink.io/en/messages/common.html#TRAJECTORY_REPRESENTATION_BEZIER) to specify the moving position setpoint of the vehicle over a given amount of time.
 - The [HEARTBEAT/Connection Protocol](https://mavlink.io/en/services/heartbeat.html) is used for "proof of life" detection.
 - [LOCAL_POSITION_NED](https://mavlink.io/en/messages/common.html#LOCAL_POSITION_NED) and [ALTITUDE](https://mavlink.io/en/messages/common.html#ALTITUDE) send the vehicle local position and altitude, respectively.
 
-> **Tip** The message flows from PX4 UORB topics, through MAVLink, to ROS and back again are all documented in: [PX4/avoidance > Message Flows](https://github.com/PX4/avoidance#message-flows).
+Path planning is enabled on PX4 in automatic modes (landing, takeoff, hold, mission, return) if [COM_OBS_AVOID=1](../advanced_config/parameter_reference.md#COM_OBS_AVOID).
```
hamishwillee: > Whoops, the Bezier isn't available in Offboard (I think). That would need to be clarified as well. It is only available in COM_OBS_AVOID=1 mode.

@jkflying I added this paragraph to cover the case - ie that it is enabled in this set of modes when COM_OBS_AVOD is set to one, and that the planner must supply setpoints even if it cannot do proper planning for the mode. 

OK?

en/computer_vision/path_planning_interface.md
jkflying: No, it is NED. This is the same frame as the waypoint trajectory message, BTW.

en/computer_vision/path_planning_interface.md
jkflying: 1. Yes

2. Yes. Go for it :+1:

jkflying:
>     1. For waypoint trajectory PX4 sends desired path and planner sends back setpoints. What gets sent out by PX4 in order to get bezier setpoints?

PX4 still sends back the waypoint trajectories. These are just the next N waypoints in the mission, so nothing changed that side. The difference is that you can now specify from the companion the exact vehicle motion, not just what the next waypoint should be.

>     2. [TRAJECTORY_REPRESENTATION_BEZIER.pos_x](https://mavlink.io/en/messages/common.html#TRAJECTORY_REPRESENTATION_BEZIER) is an array of 5 values with description "X-coordinate of starting bezier point, set to NaN if not being used". I'm guessing this an array of X starting points for 5 points? So the description should actually be "X-coordinates of bezier points (index 0 is the start point), set to NaN if not being used"

Even better, it should be for X-coordinates of Bezier _control_ points. Bezier curves are funny in that the curves don't actually go through the points you specify, the points are just used to calculate parameters of the curve that is followed. So it's important to specify that they are control points, not just points.

>     3. I think we need a diagram because from this description I don't understand the message flow, or what the "start of the curve" means in relations to the timestamp. I envisage PX4 might send a couple of waypoints to the companion. The companion works out a nice curve to that/several points, and gets back a curve that does the first .5s of that. Otherwise much like the waypoint approach.
>        I also don't get what "start of the curve" means for first point and how the delta thingy works. It might be that the curve defines a whole path and each message says we're just going to do 0.5seconds travel of that.
>        Anyway, this is clearly obvious to you, but I hope you can see that this is not enough to understand what gets sent/received.

So this is more about understanding how Bezier curves work, which isn't immediately obvious from how they are parameterized. You specify a bunch of control points, and then a 'fraction traveled', often referred to as **_t_**, which shows you where on the curve you are. The only guarantees of the Bezier curve in terms of position, is that you start at the first point when **_t_**=0, and end at the last point when **_t_**=1. So what is happening with the _delta_ here is that we use it to say how long (in time) it takes to get from the first point to the last point, which in **_t_** terms is from **_t_**=0 to **_t_**=1. So what happens on the flight controller is it checks the current time, compares that to the timestamp the message was sent and the duration of the curve in specified seconds in `delta` to get **_t_**. Say the message was sent 0.1 seconds ago, and the duration is 0.3s, then **_t_** would be 0.33333 because it is 1/3 through the duration. It can then use that **_t_**, plus the control points, to get the exact position (and velocity, and acceleration) that the vehicle should have right now to follow the curve.

Clear as mud, I know. Please let me know if / what isn't making sense, so I can explain it better somehow :see_no_evil:

en/computer_vision/path_planning_interface.md
jkflying: LGTM.

en/computer_vision/path_planning_interface.md
```diff
@@ -107,12 +113,27 @@ PX4 safely handles the case where messages are not received from the offboard sy
...
+  Be careful not to specify such short times for the bezier curve that you cannot send the next one in time.
+
 ## Supported Hardware
 
 Tested companion computers and cameras are listed in [PX4/avoidance](https://github.com/PX4/avoidance#run-on-hardware).
 
 
 <!-- ## Further Information -->
-<!-- @mrivi is expert! -->
+<!-- @mrivi and @jkflying are the experts! -->
```
jkflying: :scream:

hamishwillee:
>> 1. For waypoint trajectory PX4 sends desired path and planner sends back setpoints. What gets sent out by PX4 in order to get bezier setpoints?

> PX4 still sends back the waypoint trajectories. These are just the next N waypoints in the mission, so nothing changed that side. The difference is that you can now specify from the companion the exact vehicle motion, not just overriding what the next waypoint should be.

Just what I needed to know.

I'm going to assume that PX4 doesn't know what trajectory messages it will get back so it always sends out the same waypoints. 

The reason being that right now this means that it is not the next "N" waypoints, but the current number of points it says in https://docs.px4.io/master/en/computer_vision/path_planning_interface.html#px4-waypoint-interface

hamishwillee:
FYI, updated definitions for the interface in mavlink and removed WIP. Please check them: https://github.com/mavlink/mavlink/pull/1321

hamishwillee:
> So this is more about understanding how Bezier curves work, which isn't immediately obvious from how they are parameterized ...
> ... Clear as mud, I know. Please let me know if / what isn't making sense, so I can explain it better somehow

Actually pretty good, though hard to document for externals without saying all this. 

What this tells me (that wasn't obvious previously) is that we're only supplying a curve for the next planned time fragment (around 0.5s), and we're supplying it from the point at which we sent the message [i.e. we're not supplying 2 minutes of curve, and telling them to just use some fragment of it].
It also tells me that the planner is sending these at a high enough rate that the next point can be calculated while we're still on the old curve - essentially that we're sending a time-window of the same curve most of the time, modified as we get obstacles.

hamishwillee:
OK. Updated. I think it is probably good enough to publish (if you could please sanity check). 

As an aside, we know that a planner that does not support a particular mode should mirror back the setpoints. What about a planner that does not support missions in mission mode? It is receiving waypoints not setpoints, so mirroring the incoming trajectory will send back the setpoint as the waypoint itself. Is that expected?

en/computer_vision/path_planning_interface.md
```diff
@@ -1,12 +1,18 @@
 # Path Planning Offboard Interface
```
jkflying: So I'm not sure if the Bezier curve works under the offboard interface. It might, but I'd be pleasantly surprised.

en/computer_vision/path_planning_interface.md
```diff
@@ -26,14 +32,21 @@ The actual setup/configuration required depends on the planner being used.
   This means that offboard features that use different planners cannot be enabled on the same vehicle. a vehicle at the same time (e.g. a vehicle can support obstacle avoidance and collision prevent, but not also safe landing - or visa versa).
 
 
-## Waypoint Trajectory Interface {#waypoint_interface)
+## Trajectory Interface {#waypoint_interface)
+
+PX4 sends information about the *desired path* to the companion computer (when `COM_OBS_AVOID=1`, in modes for which the path planning interface has been integrated), and receives back a stream of setpoints for the *planned path* from the path planning software.
```
jkflying:
```suggestion
PX4 sends information about the *desired path* to the companion computer (when `COM_OBS_AVOID=1`, in _auto_ modes), and receives back a stream of setpoints for the *planned path* from the path planning software.
```

en/computer_vision/path_planning_interface.md
hamishwillee: OK, now I understand. This title was supposed to indicate that Path Planning is an interface driven by a companion computer, not that it is an interface that is active in offboard mode. I will just rename it Path Planning Interface to reduce this confusion.

Thanks

hamishwillee:
Thanks @jkflying - Doubtless we could iterate forever, but I think this is more than good enough for people to understand what they need to do.

PR #628, opened 2019-12-17
Add S500 V2 Buildlog

Junkim3DR:
Full build log of S500 V2 frame with Pixhawk4 v5 autopilot.

PR #681, opened 2020-03-11
Buildlog_Images

Junkim3DR:
S500 v2 and X500 buildlog images.

PR #639, opened 2020-01-13
Add X500 Buildlog

Junkim3DR:
Full build log of X500 frame with Pixhawk4 v5 autopilot.

PR #638, opened 2020-01-13
Add S500 V2 Buildlog

Junkim3DR:
Full build log of S500 V2 frame with Pixhawk4 v5 autopilot.

PR #682, opened 2020-03-15
Airframe build guides for Holybro X500 & S500

mrpollo:
Adds two new airframe build guides

* Holybro X500
* Holybro S500

Closes #628 #638 #639 #681 

Co-authored by @Junkim3DR 🎉


hamishwillee:
This looks more than "good enough". I'll do a merge and request any fixes later.

FWIW it is fine if the images are numbered but it is better if they also have descriptive names. Makes them easier to find when you're doing a search of the file tree.

hamishwillee:
Hi @mrpollo and @Junkim3DR 

Firstly, this was in a reviewable state - great! 

@Junkim3DR THIS is the level I need to review a PR.
It is important to have one PR with both documentation and images so you can preview what it will look like. If the images are not in place you can see the steps, but not the images. It is also good to give the images descriptive names - I imagine it was a complete pain for Ramon to integrate them. 

The image size is good. Some advice for next time
1. Cut back "empty space" around the images as much as possible. Do this before scaling
2. Scale/resize resolution to reduce file size as much as is reasonable (you have done this). This has to be done sensibly though - the airframe was unreadable as S500 so I have regenerated that.
3. Consider if it is better to have landscape kind orientation for images - ie if you rotate can people see more on one page? I rotated the tools image for example, and this provides a quicker view for users.

I subedited and merged some changes in https://github.com/PX4/px4_user_guide/pull/684

Mostly this appears good (though obviously I haven't tried it). What did in https://github.com/PX4/px4_user_guide/pull/684 was subedit the S500 doc and not the X500 docs so you can see my preferred approach. This is to use BULLETS for steps, and not to label figures. 

If you use bullets you get the benefit that images align with sections. The figures comments apply to are obvious. It is more compact and readable.
You can also see that in some cases you might be better off adding more nesting. I haven't done that, but I can see this might be beneficial.

PR #687, opened 2020-03-21
Modify the description of CUAV v5 + and CUAV nano

cuhome:
Regarding the known issues section; the problem described is a difference from other pixhawk boards, not a bug; it will cause user misunderstanding.


hamishwillee:
I am closing. As per slack this is a safety and support issue. Parts can be modified, but that should be in a separate PR.

PR #690, opened 2020-03-23
RTK GPS: explicitly mention that F9P is supported

bkueng:
I've seen questions around that in slack.

### DIFF
--- a/en/gps_compass/rtk_gps.md
+++ b/en/gps_compass/rtk_gps.md
@@ -11,7 +11,7 @@ You will need:
 
 ## Supported RTK Devices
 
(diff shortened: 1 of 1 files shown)


LorenzMeier:
Thanks! These seemingly small things are important!

hamishwillee:
Yes. Have we tested any F9P devices? IF so, would be good to get them added to the list too.

bkueng:
- https://www.sparkfun.com/products/15136
- https://store-drotek.com/911-1010-sirius-rtk-gnss-rover-f9p.html#/158-sensor-no_magnetometer
- and the one from FreeFly

PR #701, opened 2020-03-30
Ulanding to Ainstein US-D1

hamishwillee:
I am sure, based on link redirects that Aerotenna became Ainstein. I'm _fairly_ sure based on other info links that ulanding became the US-D1.

Can anyone confirm or tell me who could confirm? 

Note, if this is approved, 
* [x] update SUMMARY.md and other linked docs.


bkueng:
I can't confirm, but looks like you are right.
@davidaroyer can you confirm?

davidaroyer:
US-D1 is the second generation of the uLanding radar altimeter, but the driver is essentially the same.

hamishwillee:
Thanks @davidaroyer/@bkueng . As it all looks very similar, I've essentially just search/replaced and link updated to reflect the new version. The information is pretty thin on the page so provided the driver works, this is still accurate.

Merging.

PR #702, opened 2020-03-31
VTOL Quad - replace VT_TRANS_THR

hamishwillee:
Hi @RomanBapst  (&/or @sanderux )

Link checker shows that `VT_TRANS_THR` disappeared at some point. Possibly replaced by `VT_F_TRANS_THR`

I've marked up the bits that are affected in the tuning guide. Can you please advise how this should "properly" change. 
There is a degree of urgency because we'll be cutting the PX4 v1.11 release soon, and it would be nice if the docs were accurate-ish at this point!

### DIFF
--- a/en/config_vtol/vtol_quad_configuration.md
+++ b/en/config_vtol/vtol_quad_configuration.md
@@ -39,12 +39,17 @@ Getting your transition tuning right is important for obtaining a safe entry int
 
 #### Transition Throttle 
 
-Parameter: [VT_TRANS_THR](../advanced_config/parameter_reference.md#VT_TRANS_THR) <!-- deleted - replaced by VT_F_TRANS_THR ? -->
+Parameter: [VT_F_TRANS_THR](../advanced_config/parameter_reference.md#VT_F_TRANS_THR)
 
-Transition throttle defines the maximum throttle to use during the transition.
+Front transition throttle defines the target throttle for the pusher/puller motor during the front transition.
 Don’t set this too low otherwise you will never reach the transition airspeed.
-If you set it too high it will just use more power than you may want.
-For your first transition you are better off higher than lower here. 
+If your vehicle is equiped with an airspeed sensor then you can increase this parameter to make the front transition complete quicker.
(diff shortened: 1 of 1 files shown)


xdwgood:
By the way, can we delete this in tailsitter?

This function also seems to have no effect（_thrust_transition always 0.0）：
https://github.com/PX4/Firmware/blob/master/src/modules/vtol_att_control/tailsitter.cpp#L270

RomanBapst:
@hamishwillee

> Possibly replaced by VT_F_TRANS_THR

Yes, you are absolutely right. We now have `VT_F_TRANS_THR` and `VT_B_TRANS_THR` for front-and backtransition respectively.

hamishwillee:
Thanks @RomanBapst - can you make the changes directly in this PR by editing the page? I'd prefer that :-) 

Otherwise I can do it and you review, but that's me guessing how this works, while since you "know" should be a few minutes change.

RomanBapst:
@hamishwillee Hang on, I wasn't done yet I need to add some more info. I will collect it here and can also do the changes :-)

RomanBapst:
Backtransition throttle VT_B_TRANS_THR:
- **only used for standard vtol** (you can set negative values and then you can have reverse thrust if your ESC supports that)
- tailsitters and tiltrotors use the throttle from the mc position controller

Front transition throttle VT_F_TRANS_THR:
- **only used for standard VTOL**, pusher throttle will be ramped to that value

RomanBapst:
@hamishwillee Aight, please have a look

hamishwillee:
Looks great. Thanks very much! Merging

hamishwillee:
PS Re https://github.com/PX4/px4_user_guide/pull/702#issuecomment-606450163

That is pretty interesting but not relevant to THIS doc which only applies to standard vtol (quad). You have comments about this in the param itself, so the info is captured.
There might be more work to be done in https://docs.px4.io/master/en/config_vtol/vtol_back_transition_tuning.html - a quick scan would be appreciated to confirm yay/nay :-)

PR #705, opened 2020-03-31
Trajectory interface changes

hamishwillee:
OK, this is first pass at fixing, but it still needs work. Probably from you. All I have done is deleted the slew rate page, and tried to tidy up mentions of the MPC_AUTO_MODE.

So previously we had to modes of trajectory path, and these could be set differently in position mode and in missions, using different params.
-  Now that MPC_AUTO_MODE is deleted does that mean that jerk limited is ALWAYS used in mission mode?

Then in position mode we still have [MPC_POS_MODE](https://docs.px4.io/master/en/advanced_config/parameter_reference.html#MPC_POS_MODE) with a number of options - Default, sport, jerk limited.

How do we describe the first two options? Ie these presumably define trajectories too - so how do they relate/compare?

I have left the top level doc that refers to the trajectories because previously we had 2, and this explained the difference. Will we have more in future? I.e. is it worth saying that, or should we merge the relevant info into the jerk limited trajectory doc?

PS please ignore all those merge banch 'master ... commits. Unless you now how to get rid of them of course. They always seem to appear, even though I am synced to head revision of docs before I create my branch :-(


hamishwillee:
@MaEtUgR Still working on this. 

You answered part of my question in slack "Just one sidenote to avoid confusion: There's only one trajectory generator/mode for auto/mission flying. There's still multiple options for manual flying MPC_POS_MODE."

So would be good to make clear what the other POS modes are?

en/config_mc/mc_trajectory_tuning.md
```diff
@@ -12,37 +15,39 @@ The algorithm that generates those setpoints is called a "trajectory generator".
...
 
-> **Tip** The demanded setpoint can change very quickly (e.g. if a user moves stick from zero to maximum value as a "step").
-  Vehicle flight characteristics are better if the corresponding desired setpoint changes as a "ramp".
+> **Tip** Poorly tuned [P/PID Gains](../config_mc/pid_tuning_guide_multicopter.md) can lead to instability.
+  Poorly tuned *setpoint values* cannot result in instability, but may result in either very jerky or very unresponsive reactions to setpoint changes.
 
-The setpoint-value tuning parameters can be split into two groups: tuning parameters for position mode and tuning parameters for mission mode.
-Some parameters will have an effect on both modes.
 
+<!--
```
hamishwillee, line 30: FYI, I commented this out because I don't think the definitions are used here or in the remaining document?

en/config_mc/mc_trajectory_tuning.md
```diff
@@ -12,37 +15,39 @@ The algorithm that generates those setpoints is called a "trajectory generator".
...
-
-- [Jerk-limited](../config_mc/mc_jerk_limited_type_trajectory.md) (`MPC_POS_MODE=3`, `MPC_AUTO_MODE=1`) - Used when smooth motion is required (e.g.: filming, mapping, cargo).
+- [Jerk-limited](../config_mc/mc_jerk_limited_type_trajectory.md) 
+  - Used when smooth motion is required (e.g.: filming, mapping, cargo).
   - Generates symmetric smooth S-curves where the jerk and acceleration limits are always guaranteed.
-
-> **Tip** The jerk-limited (smooth) type is used by default for all frames in both manual and auto modes; this may not be suitable for vehicles/use-cases that require a faster response - e.g. racer quads.
+  - May not be suitable for vehicles/use-cases that require a faster response - e.g. racer quads.
+  - Set in position mode using `MPC_POS_MODE=3`
+- ?
```
hamishwillee, line 48: What other modes are there?

MaEtUgR:
> My assumption is that in mission mode the jerk-limited mode is always used - correct?

Yes

> What are the other POS modes? When should we use them? how are they tuned?

The available modes are listed here, I just saw I need to improve the descriptions: https://github.com/PX4/Firmware/blob/2e2b9a97d3002d9f54dd7b3ae8a245e051c19397/src/modules/mc_pos_control/mc_pos_control_params.c#L707-L725
In a nutshell
 * @value 0 simple jerky direct velocity input, useful for velocity control tuning, development
 * @value 1 Smooth but outmoded, certain people still rely on it. We'll fade it out when we're sure the replacement covers all cases.
 * ~@value 2 simple jerky faster, **gone since this week**~
 * @value 3 Smooth position control (Velocity), with jerk limited trajectory, **default to be used**

en/config_mc/mc_trajectory_tuning.md
```diff
@@ -12,37 +15,45 @@ The algorithm that generates those setpoints is called a "trajectory generator".
...
 
-For the remainder of this topic the term **position-control** represents the case where both loops are active while **velocity-control** refers to the case when only the velocity control loop is in use.
+[Mission mode](../flight_modes/mission.md) used the [Jerk-limited](../config_mc/mc_jerk_limited_type_trajectory.md) trajectory all the time.
 
-## Implementations
+[Position mode](../flight_modes/position_mc.md) supports all the [trajectory types](#trajectory_implementation) listed below.
+It uses the [Jerk-limited](../config_mc/mc_jerk_limited_type_trajectory.md) trajectory by default; the other types can be set using [MPC_POS_MODE](../advanced_config/parameter_reference.md#MPC_POS_MODE).
 
-Two different implementations are available for each mode and can be selected using the parameters [MPC_POS_MODE](../advanced_config/parameter_reference.md#MPC_POS_MODE) and [MPC_AUTO_MODE](../advanced_config/parameter_reference.md#MPC_AUTO_MODE).
+No other modes support trajectory tuning.
```
hamishwillee, line 37: Is this correct? I was surprised that altitude hold does not support the trajectory.

hamishwillee:
Hi @MaEtUgR Thanks very much. 
Now I understand that slew rate tuning is not gone this makes more sense :-)

Only one confirmation left: https://github.com/PX4/px4_user_guide/pull/705#discussion_r402045119

Otherwise this should be good to publish. Please check it over and approve if you OK.

en/config_mc/mc_trajectory_tuning.md
bresch: The jerk-limited trajectory is also used for the vertical axis in altitude mode :)

en/config_mc/mc_slew_rate_type_trajectory.md
```diff
@@ -1,23 +1,28 @@
...
+This topic explains how the trajectory type can be tuned.
 
 
 ## Position Mode
 
 In [Position mode](../flight_modes/position_mc.md) the stick inputs are mapped to either **position-control** or **velocity-control**.
- 
+
+> **Note** The position controller ([diagram here](https://dev.px4.io/master/en/flight_stack/controller_diagrams.html#multicopter-position-controller)) consists of an outer **P** position-control loop and an inner **PID** velocity-control loop.
+Depending on the control (flight) mode either both loops are active or just the velocity control loop.
```
MaEtUgR, line 20:
```suggestion
Depending on the mode and situation either both loops are active or just the velocity control loop.
```

en/config_mc/mc_trajectory_tuning.md
```diff
@@ -1,6 +1,9 @@
 # Multicopter Setpoint Tuning (Trajectory Generator)
 
-This document provides an overview of the multicopter tuning parameters that change the *user experience*: how fast the vehicle reacts to stick movements or direction changes in missions, the maximum allowed velocity, etc. (i.e. the parameters that affect the value of a desired setpoint rather than those that affect how well the vehicle *tracks* the setpoint).
+This document provides an overview of the multicopter tuning parameters that change the *user experience*: how fast the vehicle reacts to stick movements or direction changes in missions, the maximum allowed velocity, etc.
+
+In other worlds, this topic explains how to tune the parameters that affect the value of a *desired setpoint* rather than those that affect how well the vehicle *tracks* the setpoint).
```
MaEtUgR, line 5:
```suggestion
In other words, this topic explains how to tune the parameters that affect the value of a *desired setpoint* rather than those that affect how well the vehicle *tracks* the setpoint).
```

en/config_mc/mc_trajectory_tuning.md
MaEtUgR: Altitude mode uses the type selected by `MPC_POS_MODE` just for the vertical component only.

hamishwillee: @MaEtUgR 

So just to be very clear, in both position mode and altitude mode the sticks control
- speed over ground in both directions
- rate or rotation above ground plane (yaw)
- speed of ascent/descent.

I thin you're are saying that in position mode, the jerk limiting is applied to all of these inputs. So a quick move on speed over ground stick will translate as a ramped demanded setpoint.

But in altitude mode this will only happen for the altitude stick. The other controls will default to the "position mode type 0 direct control type algorithm"?

Why is it done this way? i.e. why do we only choose to limit the jerkyness of the ascent/descent?

MaEtUgR: I think there's a remaining misconception:
- In Position mode you run the position controller on all 3 axes and it's smoothened using the `MPC_POS_MODE` you choose.
- In Altitude mode the position controller is only ran against the altitude (world frame local z-axis). The altitude is controlled and smoothed exactly the same way as in position mode with the method specified by `MPC_POS_MODE`. The horizontal stick input is mapped to the acceleration or vehicle attitude. That's why the vehicle needs constant manual input to not drift away because there's no horizontal position or velocity control.

Yaw is the same for both, stick results in yaw rotational velocity and letting go of the stick locks the absolute yaw orientation.

hamishwillee: Thanks very much. I've clarified and merged.

PR #718, opened 2020-04-07
Update AUX port description in airframe_reference

bys1123:
1. If "** **" has only one space, it will become "****" after translation.
2. Durandal、Pixhack、Cube also has AUX port. And Pixhawk4 has different label.

### DIFF
--- a/en/airframes/airframe_reference.md
+++ b/en/airframes/airframe_reference.md
@@ -1,7 +1,7 @@
 # Airframes Reference
-> **Note** **This list is auto-generated from the source code**.
-> 
-> The **AUX** channels are only available on Pixhawk Boards (labeled with **AUX OUT**).
+> **Note**
+> **This list is auto-generated from the source code**.
+> The **AUX** channels are only available on Pixhawk compatible Boards (has labeled with **AUX OUT**/**FMU PWM OUT**).
 > 
 
 This page lists all supported airframes and types including


en/airframes/airframe_reference.md
```diff
@@ -1,7 +1,7 @@
 # Airframes Reference
-> **Note** **This list is auto-generated from the source code**.
-> 
-> The **AUX** channels are only available on Pixhawk Boards (labeled with **AUX OUT**).
+> **Note**
+> **This list is auto-generated from the source code**.
```
hamishwillee, line 3: As it says here, the page is auto generated, so you can't modify it in the guide as it will be over-written. This specific text is in https://github.com/PX4/Firmware/blob/master/Tools/px4airframes/markdownout.py#L8

Are you saying this will help because the translators know that they shouldn't touch the Note part, so this splits it into lines?

en/airframes/airframe_reference.md
```diff
@@ -1,7 +1,7 @@
...
+> The **AUX** channels are only available on Pixhawk compatible Boards (has labeled with **AUX OUT**/**FMU PWM OUT**).
```
hamishwillee, line 4: The point here is to help people understand how they connect pins, and that the pins might not be present on some boards. Adding "compatible" doesn't help here because plenty of boards aren't compatible and do have AUX ports, and plenty are compatible and don't. 

I propose the following:(?)

> **AUX** channels may not be present on some flight controllers. If present, PWM AUX channels are commonly labelled **AUX OUT**.

I have explicitly said "PWM AUX" because there is nothing stopping us in future mapping the AUX channels via a mixer to UAVCAN or some other bus.

bys1123: Sounds awesome.

en/airframes/airframe_reference.md
bys1123: 1. Sorry I didn't find the file when I created this PR, maybe add this file link to comment {-- --} would help?
2. No. Just Crowdin can't handle “\*\*a\*\* \*\*b\*\*”, it has no space between in crowdin, looks like "<0>a</0><0>b</0>", then "\*\*a\*\*\*\*b\*\*", after build, it will be "**ab****"

bys1123: image

bys1123: 2. Sorry, Crowdin handle it well, translator deleted that space.

hamishwillee:
Cool. Fixed by https://github.com/PX4/Firmware/pull/14616 - closing!

PR #726, opened 2020-04-22
Basic concepts: Add MAIN/AUX output info

hamishwillee:
Was not covered. Have tried to capture the right level.
Deliberately did not capture everything in https://discuss.px4.io/t/beginner-question-regarding-aux-main-ports/16328/4 - too much for this level.


julianoes:
Great, some additions:
- The MAIN ports are connected to the separate IO board while the AUX ports are connected directly to the "main" FMU board.
- For racers or other airframes with very low control latency requirements the AUX ports can be used for better performance. Also, the AUX ports support Oneshot as far as I know, and not just PWM.

hamishwillee:
@julianoes Thanks for that feedback. I would point out that was exactly the level of detail I was trying to avoid :-).

To make this still work I pushed the images up the top and made the "important bits to me" first. Please check it.

Should not block merging this, but we do need to explain how Oneshot works (Dshot is covered). Who is right person to ask.

hamishwillee:
@julianoes I think this is OK and I'm going to merge. Obviously if you see an improvement that could be made then let me know!

PR #727, opened 2020-04-23
T265 setup

jkflying:
This adds a brief description on how to set up the T265. Devguide PR: https://github.com/PX4/Devguide/pull/1022

Don't mind the whitespace changes (I put it in a separate commit). There were a bunch of trailing spaces which my editor stripped, so I thought I might as well add that too.

### DIFF
--- a/en/peripherals/t265_vio.md
+++ b/en/peripherals/t265_vio.md
@@ -1,23 +1,31 @@
-# T265 Setup guide
+# T265 VIO Camera Setup Guide
 
-The T265 is a VIO (or V-SLAM) camera produced by Intel. It provides odometry information which can be used to augment or replace other positioning systems on PX4.
+The [Intel Realsense Tracking Camera T265](https://www.intelrealsense.com/tracking-camera-t265/) is a [VIO](https://dev.px4.io/master/en/advanced/computer_vision.html) (or V-SLAM) camera produced by Intel.
+It provides odometry information which can be used to augment or replace other positioning systems on PX4.
 
-## Instructions
+## Setup Instructions
+
+These instructions explain how to set up VIO using ROS (the supported method):
 
 - On the companion computer, install and configure [MAVROS](https://dev.px4.io/master/en/ros/mavros_installation.html) and verify it is connected successfully to the flight controller
 - Get the [VIO bridge ROS node from Auterion](https://github.com/Auterion/VIO_bridge)
-- Configure the VIO bridge launch file to reflect your camera orientation. Default is with the lenses downwards facing.
+- Configure the VIO bridge launch file to reflect your camera orientation.
+  Default is with the lenses downwards facing.
 - Follow the instructions at [External position estimation](https://dev.px4.io/master/en/ros/external_position_estimation.html#ekf2-tuningconfiguration) for tuning the EKF2.
 
-## T265 Debugging
+## Troubleshooting
 
-First make sure MAVROS is able to connect successfully to the flight controller. If you're 100% sure, here are some things to try:
+First make sure MAVROS is able to connect successfully to the flight controller.
+
+If you're certain it is connecting properly here are some things to try:
 
 - Problem: I get drift / flyaways when the drone flies, but not when I carry it around with the props off.
- - Try soft-mounting the T265. It is very sensitive to high frequency vibrations.
+  - Try soft-mounting the T265.
+    It is very sensitive to high frequency vibrations.
 
 - Problem: I get toilet-bowling when the VIO is enabled
- - Make sure the orientation of the camera matches the transform in the launch file. Use the mavlink inspector in QGC to verify that the velocities in the ODOMETRY message coming from MAVROS are aligned to the  FRD coordinate system.
+  - Make sure the orientation of the camera matches the transform in the launch file.
+    Use the *QGroundControl* [MAVLink Inspector](https://docs.qgroundcontrol.com/en/analyze_view/mavlink_inspector.html) to verify that the velocities in the `ODOMETRY` message coming from MAVROS are aligned to the FRD coordinate system.
 
 - Problem: I want to use vision position to do loop closing, and also want to run GPS.
- - This is really difficult, because when they disagree it will confuse the EKF. From testing it is more reliable to just use vision velocity. (If you figure out a way to make it reliable, let us know)
+  - This is really difficult, because when they disagree it will confuse the EKF. From testing it is more reliable to just use vision velocity. (If you figure out a way to make it reliable, let us know)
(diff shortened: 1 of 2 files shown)


jkflying:
@hamishwillee Can we also get the T265 added to the Peripherals section in the sidebar? I wasn't sure how to do that.

en/peripherals/t265_vio.md
```diff
@@ -0,0 +1,31 @@
...
+
+- On the companion computer, install and configure [MAVROS](https://dev.px4.io/master/en/ros/mavros_installation.html) and verify it is connected successfully to the flight controller
+- Get the [VIO bridge ROS node from Auterion](https://github.com/Auterion/VIO_bridge)
+- Configure the VIO bridge launch file to reflect your camera orientation.
+  Default is with the lenses downwards facing.
+- Follow the instructions at [External position estimation](https://dev.px4.io/master/en/ros/external_position_estimation.html#ekf2-tuningconfiguration) for tuning the EKF2.
+
+## Troubleshooting
+
+First make sure MAVROS is able to connect successfully to the flight controller.
```
hamishwillee, line 18: How can you tell?

en/peripherals/t265_vio.md
```diff
@@ -0,0 +1,31 @@
...
+The [Intel Realsense Tracking Camera T265](https://www.intelrealsense.com/tracking-camera-t265/) is a [VIO](https://dev.px4.io/master/en/advanced/computer_vision.html) (or V-SLAM) camera produced by Intel.
+It provides odometry information which can be used to augment or replace other positioning systems on PX4.
+
+## Setup Instructions
+
+These instructions explain how to set up VIO using ROS (the supported method):
+
+- On the companion computer, install and configure [MAVROS](https://dev.px4.io/master/en/ros/mavros_installation.html) and verify it is connected successfully to the flight controller
+- Get the [VIO bridge ROS node from Auterion](https://github.com/Auterion/VIO_bridge)
+- Configure the VIO bridge launch file to reflect your camera orientation.
```
hamishwillee, line 12: How? That bridge has no information at all on usage in its readme

en/peripherals/t265_vio.md
```diff
@@ -0,0 +1,31 @@
...
+
+The [Intel Realsense Tracking Camera T265](https://www.intelrealsense.com/tracking-camera-t265/) is a [VIO](https://dev.px4.io/master/en/advanced/computer_vision.html) (or V-SLAM) camera produced by Intel.
+It provides odometry information which can be used to augment or replace other positioning systems on PX4.
+
+## Setup Instructions
+
+These instructions explain how to set up VIO using ROS (the supported method):
+
+- On the companion computer, install and configure [MAVROS](https://dev.px4.io/master/en/ros/mavros_installation.html) and verify it is connected successfully to the flight controller
+- Get the [VIO bridge ROS node from Auterion](https://github.com/Auterion/VIO_bridge)
```
hamishwillee, line 11: How? Just install it on the same computer? Doe sit need a particular location?

hamishwillee:
Hi @jkflying 

This has no context - ie no one will find this, or know whether they should use it. Mostly because VIO is a developer feature right now. 

Proposal - drag the [computer vision page}(https://dev.px4.io/master/en/advanced/computer_vision.html) across and split out mocap and vio separately. Structure then looks like this:
- [Advanced Features](http://docs.px4.io/master/en/advanced_features/) 
  - Computer Vision
    - VIO
    - MoCap
    - [Obstacle Avoidance](http://docs.px4.io/master/en/computer_vision/obstacle_avoidance.html)
    - [Collision Prevention](http://docs.px4.io/master/en/computer_vision/collision_prevention.html)
    - Safe Landing

I'm tempted to move path planning into Developer section :-)

Is this camera ONLY for VIO or is it also for the other vision thingys we have - e.g. obstacle avoidance? 
The thing I'm trying to work out is where it should live in Peripherals - is isn't a [Camera](https://docs.px4.io/master/en/peripherals/camera.html) in the sense we mean. How would you describe this type of camera - I think perhaps it might live under [Companion computer peripherals](https://docs.px4.io/master/en/peripherals/companion_computer_peripherals.html#cameras)

en/peripherals/t265_vio.md
jkflying: The reason I put this here is because for avoidance a lot of people have problems because they didn't follow the MAVROS setup guide.

en/peripherals/t265_vio.md
jkflying: I'll add something there. However the default is "lenses pointing downwards". People shouldn't need to do any configuration if they mount the T265 this way, which from our testing was the most robust.

en/peripherals/t265_vio.md
jkflying: It's a ROS node, so it needs to be configured the same way as other ROS nodes. This (probably) means "clone into your catkin workspace source folder and build". I'll add this to the README on the repo.

jkflying:
> This has no context - ie no one will find this, or know whether they should use it. Mostly because VIO is a developer feature right now.
> 
> Proposal - drag the [computer vision page}(https://dev.px4.io/master/en/advanced/computer_vision.html) across and split out mocap and vio separately. Structure then looks like this:
> ...

That looks like an improvement. How about a small overview page as well?

> I'm tempted to move path planning into Developer section :-)

Considering you need to compile things to make it work, that is probably more correct :+1: 

> Is this camera ONLY for VIO or is it also for the other vision thingys we have - e.g. obstacle avoidance?

This camera _only_ does VIO. Which is a pity, because theoretically it should be able to give us back a (sparse) point cloud of the features in the world that it tracks, which would be nice to have for some minimal-effort obstacle avoidance. But it doesn't, so...

> The thing I'm trying to work out is where it should live in Peripherals - is isn't a [Camera](https://docs.px4.io/master/en/peripherals/camera.html) in the sense we mean. How would you describe this type of camera - I think perhaps it might live under [Companion computer peripherals](https://docs.px4.io/master/en/peripherals/companion_computer_peripherals.html#cameras)

It's a camera in the same sense that the PX4Flow is a camera :wink: It has image sensors, but it processes them onboard, and doesn't give back images. The difference is that this needs to connect to a companion, instead of speaking MAVLink directly to the flight controller.

hamishwillee:
I can't edit this on your private branch, and something needs to go in. So I will merge and do fixes as a post process. 

@jkflying When we you get back we NEED to revisit this topic. If I can't tell how you'd go about setting this up and what it is that we actually support, then i don't see how someone else might be able to do it without a lot of work. The Obstacle avoidance stuff is close to the right level. 

We need to similarly make it clear:
- what features are provided
- What the recommended setup is.
- What cameras are supported
- What other options are available for developers who don't want to use our recommended setup.

PR #729, opened 2020-04-28
VIO/Computer Vision Updates

hamishwillee:
#727 added the T265 VIO camera, but did not link it into the docs.  The problem with the way this was done is that it has no context - in the user guide VIO is meaningless, so this camera doc just "dangles pointlessly". 

What this does is create a structure for computer vision docs in the user guide by pulling across the page from the dev guide. It splits out VIO and MoCap from that page. The setup stuff from the T265 page them moves into VIO page, so  that we can in future use different cameras if we like.

NOte, this is still pretty poor in terms of usability. 
- I think a user would find it really hard to set up - by no means turnkey
- ROS is only one solution, and there is no way as is for people to work out other solutions.
- It would not be hard to fix this by pulling appart https://dev.px4.io/master/en/ros/external_position_estimation.html, but it needs someone with expertise who is not on holiday to help - @jkflying - when you're back.

@baumanta I have some questions that you might be able to help with. If you can comment would be great. If not, I may merge and then do post process with Julien in two weeks.


en/computer_vision/visual_inertial_odometry.md
```diff
@@ -0,0 +1,56 @@
...
+## Supported Sensors 
+
+- [T265 Intel Realsense Tracking Camera](../peripherals/camera_t265_vio.md)
+
+
+## Setup Instructions
+
+You will need to attach your camera to the companion computer and set up ROS.
+Generally the camera should be mounted facing down. 
+If using the [T265 ](../peripherals/camera_t265_vio.md) it should also be soft-mounted.
```
hamishwillee, line 21: @baumanta 
1. What does soft mounting mean, and how do you do it with this camera?
2. How is this connected? Is it just a USB port?
3. How does ROS know what camera it is talking to? I.e. it seems magic to me that you can just attach this camera and somehow ROS know that it is what should be used.

en/computer_vision/visual_inertial_odometry.md
```diff
@@ -0,0 +1,56 @@
...
+
+
+> **Note** ROS is not required to for VIO, but is the supported method. 
+
+These instructions explain how to set up VIO using ROS:
+- On the companion computer, install and configure [MAVROS](https://dev.px4.io/master/en/ros/mavros_installation.html) and verify it is connected successfully to the flight controller.
```
hamishwillee, line 27: @baumanta That link does not explain how to verify if MAVROS is successfully connected. How do you do that?

en/computer_vision/visual_inertial_odometry.md
```diff
@@ -0,0 +1,56 @@
...
+- Get the Auterion [VIO bridge ROS node](https://github.com/Auterion/VIO_bridge) (this is installed as a normal ROS node).
+- Configure the [VIO bridge launch file] to reflect your camera orientation.
```
hamishwillee, line 29: @baumanta Which launch file? There are 3. Which part is used to configure the orientation?

hamishwillee: Then how do you start VIO?
Do you need to configure anything with PX4 other than that?

en/computer_vision/visual_inertial_odometry.md
baumanta: 1. Kamil used some kind of "suckers". Like little springy things to attach the camera to frame in order to add some damping. The camera does not like vibrations.
2. yes as far as I remember a USB C. But I don't have a camera handy right now.
3. So from intel there is a package with the ROS wrapper for their cameras. You will need that and you need to strat the camera from ROS using the correct launch file

en/computer_vision/visual_inertial_odometry.md
baumanta: So after hearing that I would go to the mavlink console on QGC and check for the HEARTBEAT. Mavros/Offboard-component has a unique SYSTEM_ID and you could check if you get its heartbeat. Not sure if that's what the author meant, though.

en/computer_vision/visual_inertial_odometry.md
baumanta: the launch files say in their names what they do. The one containing sitl is for use with the PX4 sitl toolchain (it starts the vio bridge, mavros and sitl). The one bridge_mavros.launch starts the bridge and mavros which is most likely what you would need on a drone. Execpt if some other comonent is already starting mavros, then you would need the third, most basic one.

en/computer_vision/visual_inertial_odometry.md
hamishwillee: @baumanta Thank you. The SYSTEM_ID isn't a good thing for instructions because it changes based on setup. Can you confirm what ROS/our setup use for the MAV_TYPE or component ID, as these would be much easier to look at?

For example, if the [MAV_COMP_ID_VISUAL_INERTIAL_ODOMETRY](https://mavlink.io/en/messages/common.html#MAV_COMP_ID_VISUAL_INERTIAL_ODOMETRY) is used, that would be eaiser to check.

en/computer_vision/visual_inertial_odometry.md
hamishwillee: Thanks, that was helpful

en/computer_vision/visual_inertial_odometry.md
hamishwillee: Thanks!

en/computer_vision/visual_inertial_odometry.md
```diff
@@ -0,0 +1,138 @@
...
+    ```
+	cd ~/catkin_ws/src
+	git clone https://github.com/Auterion/VIO.git
+	```
+  - Build the package:
+    ```
+	cd ~/catkin_ws/src
+	catkin build px4_realsense_bridge
+	```
+- Configure the *VIO bridge launch file* (typically [bridge_mavros.launch](https://github.com/Auterion/VIO/blob/master/launch/bridge_mavros.launch) or [bridge.launch](https://github.com/Auterion/VIO/blob/master/launch/bridge.launch) ) to reflect your camera orientation.
```
hamishwillee, line 54: @baumanta I looked at the launch files, and it still isn't clear how you would specify a different orientation than the default (lenses down). Can you give me an example - e.g. what would I paste in to the file if the camera was pointed 45 degrees facing forward?

en/computer_vision/visual_inertial_odometry.md
```diff
@@ -0,0 +1,138 @@
...
+
+Most commonly you'll use `bridge_mavros.launch` as shown:
+```
+cd ~/catkin_ws/src
+roslaunch px4_realsense_bridge bridge_mavros.launch
+```
+
+
+
+### PX4 Tuning {#ekf2_tuning}
```
hamishwillee, line 75: FYI, I added this section as a tuning highlight.

hamishwillee:
Hi @baumanta 
Thanks for your help -- this is much better for it. I had an epiphany - this is the user guide so this doc is all about the "supported setup". So I've pulled in the EKF tuning stuff which a user needs, and pointed to the developer doc for people who want to understand the mavlink interface and/or LPE.

Can you sanity check the rendered page please: https://github.com/PX4/px4_user_guide/blob/a95926f32bbdb3839542aa1e218a0ad9e177f3ce/en/computer_vision/visual_inertial_odometry.md

There are just a couple of inline questions about how to check things are alive, and also how to update the orientation of the camera if it is not the default. 

Cheers
H

en/computer_vision/visual_inertial_odometry.md
hamishwillee: PS, there are instructions in a different doc, but I think you could only use these when you have got avoidance running: https://dev.px4.io/master/en/ros/external_position_estimation.html#check-external-estimate

So again, would really appreciate your advice.

baumanta: yes I would agree there to move the "check" to the end of the instructions. After the nodes are launched. It is confusing where it is because you are right at that point nothing is running yet. Mavlink Inspector is a good advice there you could basically check for any topis unique to VIO: The VIO heard beat (197=MAV_COMP_ID_VISUAL_INERTIAL_ODOMETRY), or the VIO data topic itsself (not sure how it is called but there is  mavlink message with the odometry data going to PX4

en/computer_vision/visual_inertial_odometry.md
baumanta: it is that line:
<node pkg="tf" type="static_transform_publisher" name="tf_baseLink_cameraPose"
        args="0 0 0 0 1.5708 0 base_link camera_pose_frame 1000"/>

This is a static transform that links the camera ROS frame `camera_pose_frame` to the mavros drone frame `base_link`. 

the first three numbers specify translation x,y,z (center of fcu to camera):
so if the camera is 10 cm in front of the pixhawk and 4 up, the first three numbers would be : [0.1, 0, 0.04,...]
And the next three numbers specify rotation (yaw, pitch, roll). 
So [... 0, 1.5708, 0] means pitch down by 90deg (facing the ground). facing straight forward would be [... 0 0 0].

en/computer_vision/visual_inertial_odometry.md
```diff
@@ -0,0 +1,177 @@
...
+A rough estimate of the delay can be obtained from logs by checking the offset between IMU rates and the EV rates:
+
+![ekf2_ev_delay log](../../assets/ekf2/ekf2_ev_delay_tuning.png)
+
+> **Note** A plot of external data vs. onboard estimate (as above) can be generated using [FlightPlot](../log/flight_log_analysis.md#flightplot) or similar flight analysis tools.
+
+The value can further be tuned by varying the parameter to find the value that yields the lowest EKF innovations during dynamic maneuvers.
+
+
+## Check/Verify VIO Estimate {#verify_estimate}
```
hamishwillee, line 116: @baumanta I copied this section from the [devguide here](https://dev.px4.io/master/en/ros/external_position_estimation.html). I believe that it is good advice for testing that the setup is good. My only concern is that original doc mixed mocap and VIO, so I worry this might not be true.

Can you also confirm that VIO streams ODOMETRY messages or VISION_POSITION_ESTIMATE or both? I think likely ODOMETRY  as these include velocities. 

Otherwise I think this is good enough. OK by you?

hamishwillee:
@baumanta I pushed this live, but would still appreciate answer to https://github.com/PX4/px4_user_guide/pull/729#discussion_r418398076 if you have time.

en/computer_vision/visual_inertial_odometry.md
baumanta: Actually I have no idea... I never used VIO so I'm not sure what messages are used, sorry

hamishwillee:
Thanks. It will be one of them :-) Ramon says he plans on doing some testing on this.

PR #730, opened 2020-05-05
Current propellers directions description in px4_vision_kit.md

bys1123:
reference:
image
image


hamishwillee:
Thanks Stone - much appreciated. I actually didn't know this, and it wasn't picked up in review.

bys1123:
@hamishwillee No problem, as it saids Early Adopter Edition, it's already very good, too much details in it. We have lots of time to fix the bugs next.

m1baldwin:
Thanks guys! Agreed with @bys1123 ,  documentation is already very good for something so early on.

PR #733, opened 2020-05-11
Optimize Compatibility Description for Holybro Pixhawk4

bys1123:
PX4 team is continuing support this board, thanks for everyone's work. 
So I think this description would be better, welcome to edit this commit to describe this in another better way.


hamishwillee:
Thanks @bys1123 ! Merged.

PR #738, opened 2020-05-18
Move Intel Aero to Complete Vehicles in SUMMARY.md

bys1123:


hamishwillee:
This is actually a weird one. Yes it is a complete vehicle, but it is also a) discontinued  b) a reference platform, c) treated as an autopilot from a PX4 codebase point of view. 

I'm happy enough to move if no one particularly cares, but this arguable. @bkueng ?

bkueng:
Yes it is a combination of all. But as it's discontinued I would not move it.
@bys1123 do you know of people still using it?

bys1123:
@bkueng people compare the specifications with new stuffs in the market.

hamishwillee:
I wouldn't have made the change, but I don't feel strongly enough to object. Thanks @bys1123

PR #747, opened 2020-06-03
Takeoff VTOL clarifications

hamishwillee:
Attempts to capture the information in https://github.com/PX4/Firmware/issues/12660

@sfuhrer I assume transition heading also not supported if the command is used as a command (not as a mission item)?


sfuhrer:
> Attempts to capture the information in [PX4/Firmware#12660](https://github.com/PX4/Firmware/issues/12660)
> 
> @sfuhrer I assume transition heading also not supported if the command is used as a command (not as a mission item)?

I honestly don't even know how to use these commands outside of a mission, but given that you don't have any "next" waypoint I would assume it just transitions in the direction it currently is. 
If you quickly show me how then I'm happy to test it in SITL.

hamishwillee:
@sfuhrer This does not appear to be exposed as a command in PX4 so my question was irrelevant. Thanks :-). Merged.

PR #748, opened 2020-06-03
Lidar lite - PWM is output of sensor.

hamishwillee:
This fixes the label on the data connection between Pixhawk AUX5 and the Lidarlite PWM output (to say that it is an output and not an input).

@bkueng Almost all the docs we have refer to using Pixhawk PWM ports as outputs, but this is a case where we specify it is an input. Some general questions, mostly FMI:
1. Would this be hard-coded to use only this port, and if so, does this mean you couldn't use lidarlite on a board that did not have an IO board?
1. How does this work with mixers? i.e. how does PX4 know that this is an input? How is confict resolved?
1. Do we need a topic "PWM ports as inputs" :-) ?

### DIFF
--- a/en/sensor/lidar_lite.md
+++ b/en/sensor/lidar_lite.md
@@ -28,7 +28,11 @@ The *Lidar-Lite v3* can be used with either PWM or I2C.
 PWM is recommended if using an older model.
 The rangefinder must be separately powered via some ESC/BEC (whether connected via PWM or I2C).
 
-> **Note** The I2C interface of non-blue-label Lidar-Lite (v1) devices has stability limitations and all silver-label generations of Lidar-Lite sensors are therefore excluded from the I2C interface. The use of the PWM interface (as detailed below) is advised for these sensors. The blue label (v2) devices can exhibit a constant offset if powered on with less than 5V under some conditions. This is currently (Q4/2015) under investigation by the manufacturer and potentially can be resolved by adhering to specific operational conditions. The recommended robust setup is a v1 device interfaced via PWM.
+> **Note** The I2C interface of non-blue-label Lidar-Lite (v1) devices has stability limitations and all silver-label generations of Lidar-Lite sensors are therefore excluded from the I2C interface.
+  The use of the PWM interface (as detailed below) is advised for these sensors.
+  The blue label (v2) devices can exhibit a constant offset if powered on with less than 5V under some conditions.
+  This is currently (Q4/2015) under investigation by the manufacturer and potentially can be resolved by adhering to specific operational conditions.
+  The recommended robust setup is a v1 device interfaced via PWM.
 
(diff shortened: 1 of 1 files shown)


bkueng:
> Would this be hard-coded to use only this port, and if so, does this mean you couldn't use lidarlite on a board that did not have an IO board?

Yes, hard-coded. On non-IO boards it would be the MAIN 5 pin.

> How does this work with mixers? i.e. how does PX4 know that this is an input? How is confict resolved?

The mixer will only use outputs 1 to 4. This is handled in the startup script.

> Do we need a topic "PWM ports as inputs" :-) ?

This is really a special case (and rarely used), and the deeper down you want to go, the more complicated it gets :) For example it is pin 5, because output pins are grouped and cannot be configured completely independent within a group. This is also board-specific, on pixhawks outputs 1 to 4 is a group (and thus it's pin 5).
This is also why it's not very flexible in terms of configurability.

en/sensor/lidar_lite.md
```diff
@@ -60,14 +68,16 @@ The wiring for LidarLite v2 is shown below. Lidar-Lite v3 is wired similarly, ex
 
 ### I2C Interface Wiring
 
-TBD
+The I2C wiring is the same for any other distance sensor.
```
hamishwillee, line 71: @bkueng Added this. Should be true.

en/sensor/lidar_lite.md
```diff
@@ -38,18 +42,22 @@ Lidar-Lite v2 and v3 are the same, except that the order of pins in the connecto
...
 2   | RESET | AUX 6 (bottom) | Reset line of the sensor
-3   | PWM   | AUX 5 (bottom) | PWM input of the Lidar Lite. **Needs a 470 Ohm pull-down (to GND), Do not use a 1 K0hm resistor.**
+3   | PWM   | AUX 5 (bottom) | PWM output of the Lidar Lite. **Needs a 470 Ohm pull-down (to GND), Do not use a 1 K0hm resistor.**
 4   | SCL   | - | Not connected
 5   | SDA   | - | Not connected
-6   | GND   | AUX 6 (top)    | Ground
+6   | GND   | AUX 6 (top) | Ground
 
-The wiring for LidarLite v2 is shown below. Lidar-Lite v3 is wired similarly, except that the pin-numbering on the connector is reversed.
+> **Note** On a flight controller that has no AUX port the equivalent MAIN pins are used (e.g. the PWM output on the lidar instead maps to MAIN 5).
```
hamishwillee, line 56: @bkueng This captures your note on what happens on boards with no IO. I'm going to merge, but please sanity check this when you get a chance.

bkueng: Yup, should be good.

hamishwillee: Thanks!

PR #751, opened 2020-06-10
flight review: add docs for high-rate (FIFO) IMU data logging

bkueng:
From https://github.com/PX4/Firmware/pull/15062.


en/log/flight_review.md
```diff
@@ -180,6 +180,30 @@ Very high (unsafe) vibration levels.
...
 
 
+### Raw hHgh-rate IMU Data Plots {#fifo_logging}
+
+For an in-depth analysis there is an option to log the raw IMU data at full rate (several kHz, depending on the IMU).
+This allows inspection of much higher frequencies than with normal logging, which can help when selecting vibration mounts or configuring low-pass and notch filters appropriately.
+
+To use it, some parameters need to be changed:
+- Set [IMU_GYRO_RATEMAX](../advanced_config/parameter_reference.md#IMU_GYRO_RATEMAX) to 400.
+  This is required to reduce the amount of logged data
```
hamishwillee, line 190: Reduce - I would have guessed increase? Do you mean that you want to reduce this so you can increase other things?

en/log/flight_review.md
```diff
@@ -180,6 +180,30 @@ Very high (unsafe) vibration levels.
...
+- Use a good SD card, as the IMU data requires a high logging bandwidth
+  > **Tip** See [Logging > SD Cards](https://dev.px4.io/master/en/log/logging.html#sd-cards) for a comparison of popular SD card.
+- Enable either the gyro or accel high-rate FIFO profile in [SDLOG_PROFILE](../advanced_config/parameter_reference.md#SDLOG_PROFILE) and disable the rest of the entries.
+  If you are using a really good SD card (Flight Review will show dropouts if the logging rate gets too high), you can:
```
hamishwillee, line 194: Should this bit "(Flight Review will show dropouts if the logging rate gets too high)," move up into my note on line 192 - i.e. the tip is "Here are some good SD cards, but you know they are not good enough when you start seeing dropouts?

en/log/flight_review.md
```diff
@@ -180,6 +180,30 @@ Very high (unsafe) vibration levels.
 ![Exceedingly high vibration in raw accel. plot](../../assets/flight_log_analysis/flight_review/vibrations_exceedingly_high_accel.png)
 
 
+### Raw hHgh-rate IMU Data Plots {#fifo_logging}
+
+For an in-depth analysis there is an option to log the raw IMU data at full rate (several kHz, depending on the IMU).
+This allows inspection of much higher frequencies than with normal logging, which can help when selecting vibration mounts or configuring low-pass and notch filters appropriately.
```
hamishwillee, line 186: I reworded this a little bit. Hope it is still correct.

en/log/flight_review.md
bkueng: It's a bit complicated: reducing the rate makes sure that more samples are packed into 1 fifo message (which is a fixed-size array), and thus reducing the amount of messages, but still containing the same amount of information.
Essentially wasting less unused buffer space.

en/log/flight_review.md
bkueng: Yes we can move it up, it applies in general, but particularly if you want to max it out.

hamishwillee: OK, I moved it up. This is the first place we mention using a good SD card, so to me this is the right place to mention how you know that it is "not good enough).

en/log/flight_review.md
hamishwillee: Ah, so I think you saying that [IMU_GYRO_RATEMAX](http://docs.px4.io/master/en/advanced_config/parameter_reference.html#IMU_GYRO_RATEMAX) affects the **internal rate** of publishing sensor data messages.

So the messages are always published in a fixed size buffer full of whatever data has come in.
- If `IMU_GYRO_RATEMAX` is high you'll get lots of messages that won't be filled
- - If `IMU_GYRO_RATEMAX` is lower you get fewer messages that are filled.
- If you get this too low for a particular sensor you might end up losing some data? Is this ever going to be a concern if we set to 400?

But why do we need to do this other than "reducing the wasted buffer space"? Presumably the log only stores the actual data that was sent. Or does "not doing this" somehow cause other problems with what data is stored?

If so, would it be correct to change:

> This is required to reduce the amount of logged data

TO

> This ensures that the raw sensor data is is more efficiently packed when sent from the sensor to the rest of the system (resulting in ...).

bkueng: Yes that is correct.

> If you get this too low for a particular sensor you might end up losing some data? Is this ever going to be a concern if we set to 400?

The array size is chosen that this is not possible. But you're right, it's something that we need to be careful about.

> Presumably the log only stores the actual data that was sent.

Unfortunately this is not true, uORB does not have a zero-trim feature like MAVLink.

> This ensures that the raw sensor data is is more efficiently packed when sent from the sensor to the rest of the system (resulting in ...).

Yes we can write that.

hamishwillee: Thanks @bkueng - so the whole array in the message that gets sent logged, even the "empty" bit at the end. Now your original comment makes sense to me.
I have updated and added a hidden comment on this. Merging now, but would appreciate you checking the final words.

bkueng: Thanks, looks good.

PR #755, opened 2020-06-24
FrSky telemetry: Single wire update.

hamishwillee:
This attempts to capture the single wire information in https://github.com/PX4/Firmware/pull/9855.

@bkueng I'm not convinced that I fully understand this/captured all the relevant information. Feel free to edit directly if I missed something.

@bys1123 FYI - and thanks for your help.


bkueng:
I'm waiting with this for David. He told me he will look at F7 support this week, so we can combine the information (basically for F7 you don't need an inverter anymore).

hamishwillee:
>  (basically for F7 you don't need an inverter anymore).

In this case, would the following note be true, and would it also apply for things like Durandal that uses H7. ....

> **Note**: For Pixhawk FMUv5 and later (or more generally, autopilots with STM32F7 or later) PX4 can read inverted S.Bus signals directly (i.e. no special inverter cable is required). 
  On these flight controllers the connection is then the same as for an uninverted signal; simply attach one of the UART's TX pins to the S.Bus port inverted or uninverted pin (and then configure PX4).

bkueng:
Yes, except `S.Bus` -> `SPort`

hamishwillee:
Thanks. OK, will wait on that update. Ping me!

hamishwillee:
@bkueng Did the required changes go in to allow this to be updated/merged?

bkueng:
Not yet, still waiting for David (he has not forgotten about it though)

bkueng:
Changes are coming: https://github.com/PX4/Firmware/pull/15492.

Do you still have unclarity?

hamishwillee:
@bkueng I've split this into sections for before/after FMUv5. Because this should be "simpler" going forward, I moved the ready made cables section down in the text. Can you please re-review?

FYI Craft and theory have (temporarily I hope) closed their doors.

> Do I have clarity? 

Well, not perfect clarity, but probably enough for this. 

If you have time answers to the following would interest me. 

1. Essentially it isn't clear why having an uninverted output always works   (any FMU version) just by connecting the TX, while if we have an inverted output, then for previous FMU versions we need to split the signal into uninverted TX and RX. My assumption is that the reason you can just use an uninverted signal in any FMU version is that if an uninverted signal is provided this is always going to be using a single-wire protocol. So you don't need both RX and TX wires - is that correct?

2. The  diagram for "before FMUv5" shows the cable has a TTL to RS232 thingy AND an SPC cable. Why do we need both/what do they each do? (ie is it that the TTL thing does the inversion, and the SPC does something else?).

en/peripherals/frsky_telemetry.md
```diff
@@ -1,221 +1,240 @@
...
+
+## Hardware Setup
+
+FrSky telemetry requires:
+
+* An [FrSky-compatible RC transmitter](#transmitters) like the FrSky Taranis X9D Plus.
+* An [FrSky telemetry-capable receiver](#receivers) like the XSR and X8R.
+* A cable to connect the FrSky receiver Smart Port (SPort) to a flight controller UART.
+
+First [connect the reciever for RC channels](../getting_started/rc_transmitter_receiver.md#connecting-receivers), e.g. connect the S.Bus ports on the reciever and the flight controller.
```
bkueng, line 18:
```suggestion
First [connect the receiver for RC channels](../getting_started/rc_transmitter_receiver.md#connecting-receivers), e.g. connect the S.Bus ports on the reciever and the flight controller.
```

en/peripherals/frsky_telemetry.md
```diff
@@ -1,221 +1,240 @@
...
+  
+If using an SPort receiver with a pin for *uninverted output* you can simply attach one of the UART's TX pins.
+
+Then [configure PX4](#configure).
+
+### Pixhawk FMUv5/STM32F7 and later
+
+For Pixhawk FMUv5 and later PX4 can read either inverted (or uninverted) S.Port signals directly - no special cable is required.
+
+> **Note** More generally this is true on autopilots with STM32F7 or later (e.g. [Durandal](http://docs.px4.io/master/en/flight_controller/durandal.html) has a STM32H7 and can read inverter or uninverted S.Port signals directly).
```
bkueng, line 45:
```suggestion
> **Note** More generally this is true on autopilots with STM32F7 or later (e.g. [Durandal](http://docs.px4.io/master/en/flight_controller/durandal.html) has a STM32H7 and can read inverted or uninverted S.Port signals directly).
```

hamishwillee:
Thanks very much! Merging.

bys1123:
> The diagram for "before FMUv5" shows the cable has a TTL to RS232 thingy AND an SPC cable. Why do we need both/what do they each do? (ie is it that the TTL thing does the inversion, and the SPC does something else?).

FrSky FUL-1 is for upgrade the firmware, one side connect to receiver, on side connect to back of transmitter, then use transmitter upgrade the firmware on receiver. not related to flight controller.

SPC wire convert S.port(inverted half-duplex single wire uart signal which F4 chip not supported) to uninverted TX RX signal to connect old flight controller.

PR #757, opened 2020-06-28
Fix mRo Pixhawk hardware version in SUMMARY

bys1123:
Referenced the content: https://docs.px4.io/master/en/flight_controller/mro_pixhawk.html
> The main difference is that it(mRo Pixhawk) is based on the Pixhawk-project FMUv3 open hardware design.


hamishwillee:
Thanks!

PR #758, opened 2020-07-01
Pixhawk Mini and HKPilot32 now sold by Holybro

hamishwillee:
This moves them out of "experimental/discontinued" and into Pixhawk supported. It does basic changes to update branding and fixes the store links.


jinger26:
Reminder to update also on px4.io @myself

PR #760, opened 2020-07-02
Return mode: Note that return altitude cone not used for FW

hamishwillee:
As per https://github.com/PX4/Firmware/pull/15216

@RomanBapst  Can you please review?


hamishwillee:
Roman says this is fine. Merged.

PR #762, opened 2020-07-03
Add holybro pix32 v5 information

vincentpoont2:
add assets, flight controller, and quick start. 
Update SUMMARY.md and autopilot_manufacturer_supported.md


hamishwillee:
Hi @vincentpoont2 
This was really good thanks. I have merged and done some tidy up in https://github.com/PX4/px4_user_guide/pull/767

Mostly just to reduced image sizes to make this a faster load for end users (and deleted unused images). Typically if you cut off all the whitespace around images, then convert to JPG they can be shrunk a lot). There is also no point in having a very wide image - 5K pixels. 1200 is more than enough for most cases.

One thing you might to is pull the power module information into its own page like this one: http://localhost:4000/en/power_module/holybro_pm07_pixhawk4_power_module.html

PR #764, opened 2020-07-06
Add GPS fix type 4, GNSS + Dead Reckoning

myself248:
As per https://discuss.px4.io/t/gps-fix-shows-as-3-or-4-but-only-3-5-6-are-valid/17496/2


LorenzMeier:
Thanks!

PR #772, opened 2020-07-15
Orbit: Clarify yaw can now have other options

hamishwillee:
Update docs in line with https://github.com/PX4/Firmware/pull/15319

This clarifies that center-facing is a default, and links to the other options.


hamishwillee:
Thanks @MaEtUgR ! Merged.

PR #781, opened 2020-08-05
camera: Add info about pin 7/8 usage

hamishwillee:
This attempts to capture info from https://github.com/PX4/Firmware/pull/15482

Questions inline.

### DIFF
--- a/en/peripherals/camera.md
+++ b/en/peripherals/camera.md
@@ -33,7 +33,10 @@ Mode | Description
 
 ## Trigger Hardware Configuration {#hardware_setup}
 
-You can choose which pins to use for triggering using the [TRIG_PINS](../advanced_config/parameter_reference.md#TRIG_PINS) parameter.
+On Pixhawk FMUv5x boards use the board-specific camera capture pin (PI0) for triggering
+(this is automatically enabled when `TRIG_INTERFACE` is 0/GPIO or 1:Seagull).
+
+For other FMUv versions you choose the pins to use for GPIO or Seagull-based triggering using the [TRIG_PINS](../advanced_config/parameter_reference.md#TRIG_PINS) parameter.
 The default is 56, which means that trigger is enabled on *FMU* pins 5 and 6.
 
 > **Note** On a Pixhawk flight controller that has both FMU and I/O boards these FMU pins map to `AUX5` and `AUX6` (e.g. Pixhawk 4, CUAV v5+). 
@@ -41,9 +44,9 @@ The default is 56, which means that trigger is enabled on *FMU* pins 5 and 6.
   At time of writing triggering only works on FMU pins - you can't trigger a camera using pins on the I/O board.
 
 <span></span>
-> **Warning** With `TRIG_PINS` set to its **default** value of 56, you can use the AUX pins 1, 2, 3 and 4 as actuator outputs (for servos/ESCs).
-  Due to the way the hardware timers are handled (1234 and 56 are 2 different groups handled by 2 timers), this is the ONLY combination which allows the simultaneous usage of camera trigger and FMU actuator outputs.
-  **DO NOT CHANGE THE DEFAULT VALUE OF `TRIG_PINS` IF YOU NEED ACTUATOR OUTPUTS.**
+> **Warning** With `TRIG_PINS=56` (default) you can use the AUX pins 1 to 4 as actuator outputs (for servos/ESCs).
+  With `TRIG_PINS=78`, you can use the AUX pins 1-6 as actuator outputs.
+  Any other combination of pins can be selected, but this will disable use of the other FMU pins as outputs.
 
 ## Trigger Interface Backends {#trigger_backend}
 
--- a/en/peripherals/camera.md
+++ b/en/peripherals/camera.md
@@ -33,10 +33,9 @@ Mode | Description
 
 ## Trigger Hardware Configuration {#hardware_setup}
 
-On Pixhawk FMUv5x boards use the board-specific camera capture pin (PI0) for capturing.
-(this is automatically enabled when `TRIG_INTERFACE` is 0/GPIO or 1:Seagull).
+On Pixhawk FMUv5x boards use the board-specific camera capture pin (PI0) to signal PX4 that an image has been captured (this is not available on earlier Pixhawk FMU versions).
 
-For other FMUv versions you choose the pins to use for GPIO, PWM or Seagull-based triggering using the [TRIG_PINS](../advanced_config/parameter_reference.md#TRIG_PINS) parameter.
+The pins used to trigger image capture for GPIO, PWM or Seagull-based triggering (i.e. when not using a MAVLink camera) are set using the [TRIG_PINS](../advanced_config/parameter_reference.md#TRIG_PINS) parameter.
 The default is 56, which means that trigger is enabled on *FMU* pins 5 and 6.
 
 > **Note** On a Pixhawk flight controller that has both FMU and I/O boards these FMU pins map to `AUX5` and `AUX6` (e.g. Pixhawk 4, CUAV v5+). 
@@ -55,9 +54,9 @@ The camera trigger driver supports several backends - each for a specific applic
 Number | Description
 --- | ---
 1 | enables the GPIO interface. The AUX outputs are pulsed high or low (depending on the `TRIG_POLARITY` parameter) every [TRIG_INTERVAL](../advanced_config/parameter_reference.md#TRIG_INTERVAL) duration. This can be used to trigger most standard machine vision cameras directly. Note that on PX4FMU series hardware (Pixhawk, Pixracer, etc.), the signal level on the AUX pins is 3.3v.
-2  | Enables the Seagull MAP2 interface. This allows the use of the [Seagull MAP2](http://www.seagulluav.com/product/seagull-map2/) to interface to a multitude of supported cameras. Pin 1 of the MAP2 should be connected to the lower AUX pin of `TRIG_PINS` (therefore, pin 1 to AUX 5 and pin 2 to AUX 6 by default). In this mode, PX4 also supports automatic power control and keep-alive functionalities of Sony Multiport cameras like the QX-1.
+2  | Enables the Seagull MAP2 interface. This allows the use of the [Seagull MAP2](http://www.seagulluav.com/product/seagull-map2/) to interface to a multitude of supported cameras. Pin/Channel 1 (camera trigger) and Pin/Channel 2 (mode selector) of the MAP2 should be connected to the lower and higher AUX pins of `TRIG_PINS`, respectively (therefore, channel/pin 1 to AUX 5 and channel/pin 2 to AUX 6 by default). Using Seagull MAP2, PX4 also supports automatic power control and keep-alive functionalities of Sony Multiport cameras like the QX-1.
 3 | Enables the MAVLink interface. In this mode, no actual hardware output is used. Only the `CAMERA_TRIGGER` MAVLink message is sent by the autopilot (by default, if the MAVLink application is in `onboard` mode. Otherwise, a custom stream will need to be enabled).
-4 | Enables the generic PWM interface. This allows the use of [infrared triggers](https://hobbyking.com/en_us/universal-remote-control-infrared-shutter-ir-rc-1g.html) or servos to trigger your camera.
+4 | Enables the generic PWM interface. This allows the use of [infrared triggers](https://hobbyking.com/en_us/universal-remote-control-infrared-shutter-ir-rc-1g.html) or servos to trigger your camera. The trigger signal is duplicated on both pins specified using `TRIG_PINS`.
 
 ## Other Parameters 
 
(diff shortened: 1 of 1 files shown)


en/peripherals/camera.md
```diff
@@ -33,17 +33,20 @@ Mode | Description
 
 ## Trigger Hardware Configuration {#hardware_setup}
 
-You can choose which pins to use for triggering using the [TRIG_PINS](../advanced_config/parameter_reference.md#TRIG_PINS) parameter.
+On Pixhawk FMUv5x boards use the board-specific camera capture pin (PI0) for triggering
+(this is automatically enabled when `TRIG_INTERFACE` is 0/GPIO or 1:Seagull).
```
hamishwillee, line 37: @bkueng is this line correct? Or is this also enabled if PWM interface is selected, so the line should be "(this is automatically enabled when `TRIG_INTERFACE` not set to MAVLink"?

en/peripherals/camera.md
```diff
@@ -33,17 +33,20 @@ Mode | Description
...
+
+For other FMUv versions you choose the pins to use for GPIO or Seagull-based triggering using the [TRIG_PINS](../advanced_config/parameter_reference.md#TRIG_PINS) parameter.
```
hamishwillee, line 39: Same question really - it isn't clear from the docs how the PWM option works - what pin is used for the output, and how you tell PW4 what PWM value must be used to trigger.

hamishwillee:
@bkueng In addition, a lot of the rest of the doc appears to be either wrong or confused - and I can't seem to test it. For example, I used Gazebo with a camera option and tried to set GPIO. As you can see TRIG_PINS does not seem to be offered following reboot, so I can't check if the [image for QGC is up to date](http://docs.px4.io/master/en/peripherals/camera.html#trigger_setup_qgc)

image

Upshot - how do I test this (is the QGC image still correct at that link). 
is there anything else you see that is clearly wrong in this doc.

en/peripherals/camera.md
bkueng: `TRIG_PINS` defines the output(s): e.g. 56 you will have pwm signal on pins 5+6 (duplicated). Seagull requires 2 signals, so you have to use 2 pins (lower pin number is the channel 2 (mode), higher pin number is the channel 1 (trigger)).
```suggestion
For other FMUv versions you choose the pins to use for GPIO, PWM or Seagull-based triggering using the [TRIG_PINS](../advanced_config/parameter_reference.md#TRIG_PINS) parameter.
```
Is that clear?

en/peripherals/camera.md
```diff
@@ -33,17 +33,19 @@ Mode | Description
 
 ## Trigger Hardware Configuration {#hardware_setup}
 
-You can choose which pins to use for triggering using the [TRIG_PINS](../advanced_config/parameter_reference.md#TRIG_PINS) parameter.
+On Pixhawk FMUv5x boards use the board-specific camera capture pin (PI0) to signal PX4 that an image has been captured (this is not available on earlier Pixhawk FMU versions).
```
hamishwillee, line 36: @bkueng Updated this to make it clear what capture is, and (my assumption) that camera capture indication is only supported on FMUv5. I know this was planned to also be indicated in MAVLink but I don't think Daniele ever finished the implemetation - correct?

en/peripherals/camera.md
```diff
@@ -52,9 +54,9 @@ The camera trigger driver supports several backends - each for a specific applic
...
-2  | Enables the Seagull MAP2 interface. This allows the use of the [Seagull MAP2](http://www.seagulluav.com/product/seagull-map2/) to interface to a multitude of supported cameras. Pin 1 of the MAP2 should be connected to the lower AUX pin of `TRIG_PINS` (therefore, pin 1 to AUX 5 and pin 2 to AUX 6 by default). In this mode, PX4 also supports automatic power control and keep-alive functionalities of Sony Multiport cameras like the QX-1.
+2  | Enables the Seagull MAP2 interface. This allows the use of the [Seagull MAP2](http://www.seagulluav.com/product/seagull-map2/) to interface to a multitude of supported cameras. Pin/Channel 1 (camera trigger) and Pin/Channel 2 (mode selector) of the MAP2 should be connected to the lower and higher AUX pins of `TRIG_PINS`, respectively (therefore, channel/pin 1 to AUX 5 and channel/pin 2 to AUX 6 by default). Using Seagull MAP2, PX4 also supports automatic power control and keep-alive functionalities of Sony Multiport cameras like the QX-1.
```
hamishwillee, line 57: FYI @bkueng I added info to clarify the seagull triggering here. It was a little confusing at first because the Seagull docs don't label the pins and refer to them as "channels". I'm hoping that by referring to them as both pins and channels might help.

hamishwillee:
@bkueng Thanks, ready for another review. Next time you do anything with a hardwired camera can you regenerate the UI diagram for me as per https://github.com/PX4/px4_user_guide/pull/781#issuecomment-668966435 ?

en/peripherals/camera.md
bkueng: Other boards use the FMU PWM pin 6 (hardcoded).
From the code, the `CAMERA_TRIGGER` mavlink message is sent out as well, so should be all there.

en/peripherals/camera.md
bkueng: Yes, found them confusing as well.

en/peripherals/camera.md
hamishwillee: @bkueng OK, I updated the capture info for FMU PWM pin 6. 

> the CAMERA_TRIGGER mavlink message is sent out as well, so should be all there.

Yes, but we're talking about camera *capture*, not trigger. That message should be sent by the camera and processed by PX4 (not "sent"). I have a feeling that we don't yet process that signal.

hamishwillee:
Otherwise good to go?

en/peripherals/camera.md
bkueng: See https://github.com/PX4/Firmware/blob/master/src/drivers/camera_capture/camera_capture.cpp#L159. I'm not sure how this is supposed to work, but that's how it is currently.
There's also a `CAMERA_CAPTURE` message. Quite confusing.

hamishwillee: This was discussed in https://github.com/PX4/Firmware/pull/10295 but eventually merged in https://github.com/PX4/Firmware/pull/11395 (for some reason I though it got abandoned). Anyway the theory here is that you publish the trigger message **after** you get the capture (from a hotshoe say), and this gives you a better timestamp for geotagging.

But I really don't get this. I mean I understand how it will work for a camera  attached to the FC, but if you have a mavlink camera then the CAMERA_TRIGGER is what sets off triggering, so this becomes recursive. 

Perhaps @DanielePettenuzzo can explain, but I think the mavlink case is an omission/enhancement that should not block this right?

bkueng: > But I really don't get this. I mean I understand how it will work for a camera attached to the FC, but if you have a mavlink camera then the CAMERA_TRIGGER is what sets off triggering, so this becomes recursive.

Exactly, my main point of confusion as well.

> OK, I updated the capture info for FMU PWM pin 6.

I don't see this?

hamishwillee: >> But I really don't get this. I mean I understand how it will work for a camera attached to the FC, but if you have a mavlink camera then the CAMERA_TRIGGER is what sets off triggering, so this becomes recursive.

> Exactly, my main point of confusion as well.

Arrgg, I got this wrong, and am sorry to have confused you/us. [CAMERA_TRIGGER](https://mavlink.io/en/messages/common.html#CAMERA_TRIGGER) is always a notification. It is emitted by either the camera driver or the camera capture driver. It will also be emitted by a MAVLink camera. 

If you use a camera capture there will be multiples of this flying around, with the same sequence number and different timestamp. So essentially you'll have some component that is smart enough to resolve this, as described in https://github.com/PX4/Firmware/pull/10295#issuecomment-425932056

en/peripherals/camera.md
```diff
@@ -33,17 +33,24 @@ Mode | Description
...
 
 <span></span>
-> **Warning** With `TRIG_PINS` set to its **default** value of 56, you can use the AUX pins 1, 2, 3 and 4 as actuator outputs (for servos/ESCs).
-  Due to the way the hardware timers are handled (1234 and 56 are 2 different groups handled by 2 timers), this is the ONLY combination which allows the simultaneous usage of camera trigger and FMU actuator outputs.
-  **DO NOT CHANGE THE DEFAULT VALUE OF `TRIG_PINS` IF YOU NEED ACTUATOR OUTPUTS.**
+> **Warning** With `TRIG_PINS=56` (default) you can use the AUX pins 1 to 4 as actuator outputs (for servos/ESCs).
+  With `TRIG_PINS=78`, you can use the AUX pins 1-6 as actuator outputs.
+  Any other combination of pins can be selected, but this will disable use of the other FMU pins as outputs.
+  
+Some cameras can provide image capture signalling to a flight controller camera-capture pin (this enables more accurate image timestamp than the time that the trigger command is sent).
```
hamishwillee, line 48: @bkueng FYI, I moved the camera capture information here.

en/peripherals/camera.md
```diff
@@ -33,17 +33,24 @@ Mode | Description
...
+- Pixhawk FMUv5x boards use the board-specific camera capture pin (PI0).
+- Other board use FMU PWM pin 6 (hardcoded) for camera capture.
+
+> **Note** The camera driver and camera capture driver emit the MAVLink [CAMERA_TRIGGER](https://mavlink.io/en/messages/common.html#CAMERA_TRIGGER) message (with timestamp and image sequence number) when image capture is triggered and signaled, respectively.
```
hamishwillee, line 52: @bkueng FYI here I added info abut the camera trigger. Seem correct to you? Looking at the code it looks like the camera driver just updates the uorb topic and that ultimately causes the message to be omitted.

It seems to me that a MAVLink camera should emit CAMERA_TRIGGER to.

Anyway, this make sense?

bkueng: Yes that helps, thanks.
The fmu will emit CAMERA_TRIGGER in any case though: https://github.com/PX4/Firmware/blob/master/src/drivers/camera_trigger/camera_trigger.cpp#L330.
Probably worth clarifying in the mavlink spec as well.

en/peripherals/camera.md
```diff
@@ -33,17 +33,24 @@ Mode | Description
...
+  PX4 determines which instance to use based on the (later) timestamp.
```
bkueng, line 53: Where is this coming from?

hamishwillee: @bkueng This is coming from https://github.com/PX4/Firmware/pull/10295#issuecomment-425932056 . I.e. the intent is clearly that PX4 will determine which value should be used for timestamping, or that another system should be able to work this out from the log.

However I don't truly know if this intent is implemented in code. I would love to be canonical, but I'm not digging in code for this. If you don't either, might this be better "softened" to ...(?)

"A geotagging system can determine which instance to use based on the (later) timestamp (e.g. from the log)."

en/peripherals/camera.md
hamishwillee: Hmmm. On further thought and reading I think the behaviour here will remain undefined - the commands that PX4 uses are not deprcated - do_trigger_control etc - but they are very much "old style". The new camera API uses new ones. 

The good news here is that I think it doesn't matter. A MAVLink camera should emit CAMERA_IMAGE_CAPTURED on every capture, with a timestamp. A geotagging system should simply use that by preference if both messages are present. Make sense?

bkueng: I really lack the background on all these messages to give you a good answer, but you're probably right. Daniele should have the overview, but he's out this week.

en/peripherals/camera.md
bkueng: Ah ok, clear now. I would change it then to (or similar):
```
**Note** PX4 emits the MAVLink [CAMERA_TRIGGER](https://mavlink.io/en/messages/common.html#CAMERA_TRIGGER) message (with timestamp and image sequence number).
```

hamishwillee: Works for me. Changed.

en/peripherals/camera.md
hamishwillee: OK. Gus pretty much said that the behaviour for the old commands on a mavlink camera is undefined. That said, he did indicate that if the camera did obey the command then it would emit `CAMERA_IMAGE_CAPTURED`. I think we need to get those old commands deprecated and support for the new commands added to PX4.

hamishwillee:
@bkueng I have merged this because I'm pretty sure "it is not wrong", though there is obviously more that could be done with camera docs.

That said, https://github.com/PX4/Firmware/issues/13980 asked for writing diagrams/docs for camera capture. Do we have any examples? Ie of a real camera with trigger input/output and how it is wired up. I am imagining that a camera that supports triggering via GPIO might have two inputs and one output and you just connect 3 wires. But real example with images removes all ambiguity.

PR #792, opened 2020-08-25
Camera capture - add info about CAM_CAP_FBACK

hamishwillee:
Follows on from https://github.com/PX4/Firmware/issues/15526#issuecomment-679008693

@bkueng This adds info about parameter for enabling camera feedback. There is another question outstanding somewhere about signal requirements/how this is wired (ie we know what pin to use, but can we just wire the hotshoe up) - I have no idea how this works, so can't comment.
We would include that in the "camera capture" section, but perhaps merge this first.

### DIFF
--- a/en/peripherals/camera.md
+++ b/en/peripherals/camera.md
@@ -33,6 +33,8 @@ Mode | Description
 
 ## Trigger Hardware Configuration {#hardware_setup}
 
+### Camera Triggering
+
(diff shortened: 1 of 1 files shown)


bkueng (approved):
Anything that goes beyond that (wiring, etc.) you should ask @DanielePettenuzzo, as I've never done this either.

hamishwillee:
Merging as is for now then.

PR #794, opened 2020-08-30
Holybro h-rtk f9p update

vincentpoont2:
Add holybro h-rtk f9p info and update h-m8p info


hamishwillee:
@vincentpoont2 Looks great thanks. 

FYI only, you will note that in paragraphs I put sentences on different lines. This is not strictly necessary (or I would mandate it via the toolchain). The reason I do this is that in the past some translators have told me it makes translation management a little easier. As I am not a translator I don't know for sure whether that is true, but I now do it by habit :-)

PR #800, opened 2020-09-09
Update battery config with LiIon info

hamishwillee:
This adds info about LiIon batteries in a table with existing info for LiPo batteries. Originates from discussion in Slack: https://px4.slack.com/archives/C629RE9H7/p1599236420010800

Please check. In particular the warnings and caveats have not changed from what they were for LiPo, and it may be that the LiIon behaviour is sufficiently different to require new comments.

### DIFF
--- a/en/config/battery.md
+++ b/en/config/battery.md
@@ -23,6 +23,10 @@ The approach you use will depend on whether the vehicle's power module can measu
 > **Note** The instructions below refer to battery 1 calibration parameters: `BAT1_*`. 
   Other batteries use the `BATx_*` parameters, where `x` is the battery number.
   All battery calibration parameters [are listed here](../advanced_config/parameter_reference.md#battery-calibration).
+  
+<span></span>
+> **Tip** In addition to PX4 configuration discussed here, you should ensure that the ESC's low voltage cutoff is either disabled or set below the expected minimum voltage.
+  This ensures that the battery failsafe behaviour is managed by PX4 (i.e. that the ESCs will not cut out while the battery still has charge according to the "empty-battery" setting that you have chosen).
 
 
 ## Basic Battery Settings (default) {#basic_settings}
--- a/en/config/battery.md
+++ b/en/config/battery.md
@@ -47,20 +47,25 @@ The sections below explain what values to set for each field.
 
 ### Number of Cells (in Series)
 
-This sets the number of cells connected in series in the battery. Typically this will be written on the battery as a number followed by "S" (e.g "3S", "5s").
+This sets the number of cells connected in series in the battery.
+Typically this will be written on the battery as a number followed by "S" (e.g "3S", "5S").
 
 > **Note** The voltage across a single galvanic battery cell is dependent on the chemical properties of the battery type.
-  The most common drone battery type (Lithium-Polymer - LiPo) has a nominal cell voltage of 3.7V.
+  Lithium-Polymer (LiPo) batteries are the most common type, and have a nominal cell voltage of 3.7V, while Lithium-Ion batteries have a nominal cell voltage of 3.6V.
   In order to achieve higher voltages (which will more efficiently power a vehicle), multiple cells are connected in *series*.
   The battery voltage at the terminals is then a multiple of the cell voltage.
 
-If the number of cells is not supplied you can calculate it by dividing the battery voltage by the nominal voltage for a single cell. The table below shows the voltage-to-cell relationship for LiPo batteries:
-* 1S - 3.7V
-* 2S - 7.4V
-* 3S - 11.1V
-* 4S - 14.8V
-* 5S - 18.5V
-* 6S - 22.2V
+If the number of cells is not supplied you can calculate it by dividing the battery voltage by the nominal voltage for a single cell.
+The table below shows the voltage-to-cell relationship for these batteries:
+
+Cells | LiPo (V) | LiIon (V)
+--- | --- | ---
+1S | 3.7 | 3.6
+2S | 7.4 | 7.2
+3S | 11.1 | 10.8
+4S | 14.8 | 14.4
+5S |18.5 | 18
+6S | 22.2 | 21.6
 
 > **Note** This setting corresponds to [parameters](../advanced_config/parameters.md): [BAT1_N_CELLS](../advanced_config/parameter_reference.md#BAT1_N_CELLS) and [BAT2_N_CELLS](../advanced_config/parameter_reference.md#BAT2_N_CELLS)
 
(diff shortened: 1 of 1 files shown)


ryanjAA (changes requested):
Cool this is getting into here. We use li-ion on pretty much everything.

A few things to change:

Lipo and Li-ion have the same nominal voltage (3.7v). Only older (you really won't find them anymore) cell chemistries have a 3.6v nominal (explanation is really how the test is done which is from 4.2v to 3.0v at 0.5c discharge. On Li-cobalt chems, the half point between full to empty is 3.6v but chemistries with a lower resistance [manganese and others] have a mid-way voltage of 3.7v. Lower resistance will also have higher discharge capabilities). It would be an extreme edge case at this point for 3.6v. I'd therefor keep the voltage table the same for both types.

Full voltage per cell should be the same for both. I think something that is important here is that li-ion cells can not have anywhere close to as high as a C rating as a lipo (since they have higher resistance) so while loaded down li-ions will see a decreased voltage under load (same as a lipo voltage sag just more pronounced), if operating the battery correctly (or within its spec) then it should be pretty close to a lipo in terms of voltage drop since you need to carry a larger battery to ensure your vehicle doesn't go outside the C rating. There are of course a few times this may not happen but generally it's a close. Probably slightly lower (0.5v lower) under load on a fully charged li-ion if you were pushing it but IIRC QGC has Li-po set at 4.05v which would be the same for li-ion.

The rule of thumb part:

3.7/3 (lipo/li-ion) Increase from 2.7v to 3v for li-ion. - It's worth noting that those numbers (both 3.7 and 3 are far from empty (hence why you have them in conservative) in both cases since the 3.7v is nominal voltage of the lipo (so it is definitely not its empty voltage and there should be 25-35% left) which means for an apple to apples comparison we should list the li-ion at the same voltage for that amount left which is 3.325v at 1c (25% left). That is very much li-ion type and C rating dependent.

Real minimum load while flying:
I would assume 1c for a li-ion as the data really is all over the place based on that variable so the assumption in place is important (as c rate increases, total capacity available decreases). At 1c, 2.7v for li-ion you have about 1.7-2% left on the battery and at 3.5v for a lipo (under load) you have 5%. 2.95v is 5% for a li-ion.

Damage battery:

There is a steep decline at this point but honestly, you don't empty the battery until 2.5v on a li-ion so that is true empty as well as the don't go under voltage. For lipo it is 3v instead of 3.5v.

Low Voltage cutoff:

It's really important to alter the ESC low voltage cutoff. Depending on the manufacturer of the ESC you may have to just disable it since large voltage sags will trigger it and that will depend on the battery type and health. Personally I would rather damage a battery an dispose of it than a vehicle... It is better anyway to let px4 deal with voltage management rather than a separate "dumb" system (ESC) that gets to supersede everything else and isn't always easily changeable.

Here is a graph.

lginr18650mj1tracking

hamishwillee:
Hi @ryanjAA,
Thanks for that. I've made your suggested changes - can you check I got it right.
Note that although some values are the same I decided to leave the columns separate for the battery types to make this easy to extend if we add other types later.

1. YOu said that the "S" ratings are lower for LiIon. What is the maximum you might reasonably see?

2. It might be good to have a section that captures your notes above "A note on battery types" but I'm not sure what would be in it (open to suggestions). Some of the takeaways appear to be (and sorry if I screwed this up):
   - LiIon 
      - higher resistance and hence have lower discharge capabilities
      - higher resistance and hence lower C ratings and greater drop in voltage under same load.
      - Steeper decline at lower voltage, so presumbly a little hard to get true empty separated from damage. 

   We could just have a link to a good article on the next if you know of one (or ignore this altogether).
  
3. I didn't yet add Low voltage cuttoff information - are you suggesting this for all batteries, or just LiIon because they experience deeper sag under load, and appear to have a sharper dropoff when nearing empty? My understanding is that this is outside of PX4 right?

4. Why do you choose to use LiIon batteries?

ryanjAA:
Thinking about this. One thing I’ve often wondered is the origins of the 4.05v used as the full voltage. Where did that number come from? I get that under a load we will see some voltage sag but I wonder if there is any data for that number and the effect it would have in terms of what constitutes a full battery.

hamishwillee:
@ryanjAA We nearly had comment cross over :-0. I suspect it comes from some battery that was on hand when this was written, mapped to the comment "The value should be set slightly lower that the nominal maximum cell voltage for the battery, but not so low that the estimated capacity is still 100% after a few minutes of flight.".

Does the real full battery value differ much/enough to make a difference. If so, perhaps we should suggest measuring it and then dropping by "x" to get a better value?

ryanjAA:
Hahaha. We definitely almost crossed! 

Thanks for the information re the origins of the initial info. I think we need to look into that a little. It may be correct but it is worth looking into. 

As for LVC on the esc. I recommend we say something about its importance of disabling or setting it correctly. Experience users will know but otherwise it is easy to overlook. We had a test plane make an emergency landing just after takeoff as the esc was set to stock and the battery had plenty of charge left but much too low for the esc setting. It was in mission mode and it couldn’t climb at all. It was just pulsing the throttle. We have the log somewhere. The other side of it is that some of our longer flight durations (3.5 hr+) had to have esc settings altered to get to it to actually use what’s in the battery so that’s equally important. Again, I’d really rather hurt a battery by accident (water enters the cells of a lipo essentially which irreversible) versus almost crashing from limited esc power that you can’t override in the air. 

I’ll start answering everything your note. I think what’s also important is that this varies greatly by li-ion cell type since their c rating range are so much smaller compared to lipos where they are magnitudes higher. The effect this has on voltage is less for all lipos in some regards with respect to voltage drop or out another way, that differing types will respond similarly whereas li-ions are all very different given the c rating. I’ll write something that fits though. You can’t group lions as much together except for their full and empty voltage. That’s the same. 

Ok let me get something started.

hamishwillee:
Thanks @ryanjAA 

I added a tip for the ESC low voltage cutoff which captures the essential point

> **Tip** In addition to PX4 configuration discussed here, you should ensure that the ESC's low voltage cutoff is either disabled or set below the expected minimum voltage.
  This ensures that the battery failsafe behaviour is managed by PX4 (i.e. that the ESCs will not cut out while the battery still has charge - according to the "empty-battery" setting that you have chosen).

ryanjAA:
I’m still working on this but it actually all comes down to energy density which is generally in battery watt hours per kg (wh/kg). That normalizes it across any battery and wh is nominal voltage x capacity. So for instance, a typical lipo battery has about 160 wh/kg. Higher energy density ones can reach 185 but li-ions are much higher (ours are 278 wh/kg). Higher energy density means the longer you can fly given the weight is the same but are holding more battery onboard. That said you have manage the battery specs. Mainly the c rating which is the maximum amount the battery can discharge safely. 1 c equals 1x the battery capacity in terms of current. So 10ah battery with a 1c discharge max rating could only discharge 10 amps.

ryanjAA:
Which is why people often like lipo that have 50c ratings. They can pull huge amps safely out of them

ryanjAA:
Which is why people often like lipo that have 50c ratings. They can pull huge amps safely out of them

ryanjAA:
https://github.com/PX4/px4_user_guide/pull/822 has added info as discussed.

hamishwillee:
Thanks - I'll look at that one then. Closing this as superseded by #822

PR #807, opened 2020-09-14
Camera Capture: Explain how capture pin wired

hamishwillee:
This explains how camera capture is wired - specifically that Pixhawk expects a rising signal of the right voltage (3.3V) on the camera capture pin and also reference GND and +3.3V on the servo rail [that might not be correct as a "requirement" but it is what the seagull sync2 does - please confirm)>

Open questions:
- Is 3.3V on the pin correct? Or does it depend on the FC. Please confirm
- The power/gnd 
  - are they "required"? 
  - What about for FMUv5x, which has a control pin/ is not part of PWM interface and hence has no GND/Power pins). Does that mean we'd have to find some appropriate other free pins to use for GND/+3.3.V?


hamishwillee:
@bkueng @DanielePettenuzzo 

I am merging this, because it usefully makes clear that you can't just connect any old wire - you need to match the voltages. 

Can you check this - specifically that the nominal voltage on the pin is 3.3V and it is the rising edge that is detected.

PR #820, opened 2020-10-01
Add note about thermal vs original calib. params

### LINKED ISSUE #753: Doc Bug: Sensor Thermal Compensation · PX4 User Guide
LOCATION:
In the section 'Offboard Calibration Procedure', there is no. 12 step as below:
"12. Power the board and perform a normal accelerometer sensor calibration using QGroundControl. It is important that this step is performed when board is within the calibration temperature range. The board must be repowered after this step before flying as the sudden offset changes can upset the navigation estimator and some parameters are not loaded by the algorithms that use them until the next startup."

PROBLEM:
But, after the accelerometer calibration is done, the parameter values for accelerometer calibration are not saved and maintain default values. So the step is meaningless actually. This problem can be reproduced in PX4 v1.10.2, v1.9.0, v1.7.2 and v1.6.5.

CONCLUSION:
Therefore, it needs to be checked whether the accelerometer calibration step is correct or not. And then the document can be clarified more.

Bug Page: [Sensor Thermal Compensation · PX4 User Guide](https://docs.px4.io/master/en/advanced_config/sensor_thermal_calibration.html)

hamishwillee:
@MaEtUgR I think this fixes #753 .

Can you also confirm we still don't support thermal calibration of the magnetometer?

### DIFF
--- a/en/advanced_config/sensor_thermal_calibration.md
+++ b/en/advanced_config/sensor_thermal_calibration.md
@@ -38,7 +38,7 @@ PX4 supports two calibration procedures:
 The offboard approach is more complex and slower, but requires less knowledge of the test setup and is easier to validate.
 
 > **Note** After thermal calibration the thermal calibration parameters (`TC_*`) are used for *all* calibration/compensation of the respective sensors.
-  The normal calibration parameters (e.g. `SYS_CAL_*`) are no longer used and are reset to their default values.
+  The normal `SYS_CAL_*` calibration parameters are no longer used (and in some cases may be reset).
 
 ### Onboard Calibration Procedure {#onboard_calibration}
 
--- a/en/advanced_config/sensor_thermal_calibration.md
+++ b/en/advanced_config/sensor_thermal_calibration.md
@@ -2,14 +2,19 @@
 
 PX4 contains functionality to calibrate and compensate rate gyro, accelerometer and barometric pressure sensors for the effect of changing sensor temperature on sensor bias.
 
-This topic details the [test environment](#test_setup) and [calibration procedures](#calibration_procedures). At the end there is a description of the [implementation](#implementation).
+This topic details the [test environment](#test_setup) and [calibration procedures](#calibration_procedures).
+At the end there is a description of the [implementation](#implementation).
 
-> **Note** At time of writing (June2019/PX4 v1.9) thermal calibration of the magnetometer is not yet supported.
-
+> **Note** After thermal calibration the thermal calibration parameters (`TC_*`) are used for *all* calibration/compensation of the respective sensors.
+  Any subsequent standard calibration will therefore update `TC_*` parameters and not the "normal"  `SYS_CAL_*` calibration parameters (and in some cases these parameters may be reset).
+  
+<span></span>
+> **Note** At time of writing (PX4 v1.11) thermal calibration of the magnetometer is not yet supported.
 
 ## Test Setup/Best Practice {#test_setup}
 
-The [calibration procedures](#calibration_procedures) described in the following sections are ideally run in an *environment chamber* (a temperature and humidity controlled environment) as the board is heated from the lowest to the highest operating/calibration temperature. Before starting the calibration, the board is first *cold soaked* (cooled to the minimum temperature and allowed to reach equilibrium).
+The [calibration procedures](#calibration_procedures) described in the following sections are ideally run in an *environment chamber* (a temperature and humidity controlled environment) as the board is heated from the lowest to the highest operating/calibration temperature.
+Before starting the calibration, the board is first *cold soaked* (cooled to the minimum temperature and allowed to reach equilibrium).
 
 For the cold soak you can use a regular home freezer to achieve -20C, and commercial freezers can achieve of the order of -40C. 
 The board should be placed in a ziplock/anti-static bag containing a silica packet, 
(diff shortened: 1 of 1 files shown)


en/advanced_config/sensor_thermal_calibration.md
```diff
@@ -4,7 +4,7 @@ PX4 contains functionality to calibrate and compensate rate gyro, accelerometer
 
 This topic details the [test environment](#test_setup) and [calibration procedures](#calibration_procedures). At the end there is a description of the [implementation](#implementation).
 
-> **Note** At time of writing (June2019/PX4 v1.9) thermal calibration of the magnetometer is not yet supported.
+> **Note** At time of writing (June2019/PX4 v1.9) thermal calibration of the magnetometer is not yet supported.
```
MaEtUgR, line 7:
```suggestion
> **Note** At time of writing (June2019/PX4 v1.9) thermal calibration of the magnetometer is not yet supported.
```

MaEtUgR (approved):
@hamishwillee Thanks for your efforts! I think @baumanta is more the expert in thermal calibration since she did quite some testing not too long ago. AFAIK you're right and we don't support mag temperature calibration. I would have guessed that's because it's not as useful to do it as with gyro and accel but I might be wrong.

hamishwillee:
@baumanta Hi Tanja, Matthias says you are the most recent person to touch this code and the most likely to be able to confirm this is correct. Can you please sanity check?

en/advanced_config/sensor_thermal_calibration.md
```diff
@@ -35,8 +35,10 @@ PX4 supports two calibration procedures:
 * [onboard](#onboard_calibration) - calibration is run on the board itself. This method requires knowledge of the amount of temperature rise that is achievable with the test setup.
 * [offboard](#offboard_calibration) - compensation parameters are calculated on a development computer based on log information collected during the calibration procedure. This method allows users to visually check the quality of the data and curve-fit.
 
-The offboard approach is more complex and slower, but requires less knowledge of the test setup and is easier to validate. 
+The offboard approach is more complex and slower, but requires less knowledge of the test setup and is easier to validate.
 
+> **Note** After thermal calibration the thermal calibration parameters (`TC_*`) are used for *all* calibration/compensation of the respective sensors.
+  The normal calibration parameters (e.g. `SYS_CAL_*`) are no longer used and are reset to their default values.
```
baumanta, line 42: It is correct that the calibration is a one time thing and the SYS_CAL_* parms specify the boundaries used in this process. So those params will not be used anymore after the caliibration has completed. But if I remember correctly only the boolean values are set back to 0. The SYS_CAL_TDEL, SYS_CAL_TMAX, SYS_CAL_TMIN params stay the way they have been set for the calibration.

baumanta (approved):

baumanta:
Also one thing which confused me a lot and might be worth clarifying:
- once you perform the thermal calibration you get thermal params out of it. Meaning e.g. for accel 0 it sets all the following

- if you then perform a standard calibration, the calibration modifies the TC_A0_X)_* and TC_A0_SCL-* params instead of writing the result to the CAL_ACC*_*OFF and CAL_ACC*_*SCALE params.

en/advanced_config/sensor_thermal_calibration.md
hamishwillee: Thanks. I modified this line to "The normal `SYS_CAL_*` calibration parameters are no longer used (and in some cases may be reset)."

I didn't want to be precise without actually testing.

hamishwillee:
@baumanta Great thanks. Re https://github.com/PX4/px4_user_guide/pull/820#issuecomment-706998504

> if you then perform a standard calibration, the calibration modifies the TC_A0_X0_* and TC_A0_SCL* params instead of writing the result to the CAL_ACC*OFF and CAL_ACC*SCALE params.

So what we say in the doc now is that 
- after doing a thermal calibration the "normal calibration parameters" are not used - which is correct. 
- the thermal calibration includes the impact of normal calibration

The assumption I made from that is that after thermal calibration there is no need or point in doing a normal calibration, and that if you do one, it will do noting. You are saying that if you do one, then if thermal calibration has been done/"is active" then it will have impact - just that impact will be to the thermal params.

So:
1. Is that correct?
2. Why/when would you do a normal calibration on gyro after having done a thermal one at some point (ditto for other sensors?)

My thinkin on this is that the note below is too late - it should move up to the intro section along with clarifications of points 1, 2 above. Reasonable?

> Note After thermal calibration the thermal calibration parameters (TC_*) are used for all calibration/compensation of the respective sensors. The normal SYS_CAL_* calibration parameters are no longer used (and in some cases may be reset).

3. Further, your image looks like a nice summary of what is changed. Do you think it worth having a section at the end of the doc that incorporates this "more or less as-is", perhaps with the title "Calibration Parameter Overview" ?

baumanta:
@hamishwillee No this is not correct. You still need to do the normal calibration after the thermal one. But when I tested it, the results of the standard calibration were merged into the thermal parameters instead of being stored in the usual ones. Which I found very confusing. But @dagar said that this has changed in the meantime, so maybe he can clarify what is the current situation?

hamishwillee:
@baumanta (and @MaEtUgR) I have modified this to match what was true when you last tested -  `TC_*` used for all parameters after thermal calibration. The note also makes it clear that you may still perform a standard calibration. It looks like this.

> **Note** After thermal calibration the thermal calibration parameters (`TC_*`) are used for *all* calibration/compensation of the respective sensors.
  Any subsequent standard calibration will therefore update `TC_*` parameters and not the "normal"  `SYS_CAL_*` calibration parameters (and in some cases these parameters may be reset).

I'm merging, as while it might not reflect the latest and greatest, it is closer than what we had.

@dagar Happy to amend if you can provide additional information.

PR #822, opened 2020-10-03
Updated Battery info

ryanjAA:
Expanded on battery info per https://github.com/PX4/px4_user_guide/pull/800


hamishwillee:
@ryanjAA Fantastic summary thank you. The place you had put it was a little odd, so I subedited it and moved it to the end but created a "Tip" and the beginning explaining the difference.

I am still not certain it is the right place. I was thinking perhaps that we might put this in its own page under "Basic Concepts", linked from here: http://docs.px4.io/master/en/getting_started/px4_basic_concepts.html#batterypower

Then that first tip would instead link to the complete discussion. My reasoning is that by the time you got this far, you should already have some view on batteries - I think where it is now is fine, but still too easy to miss.

Do you have an opinion?

PR #830, opened 2020-10-14
u-blox f9p heading docs

bkueng:
From https://github.com/PX4/Firmware/pull/15954


en/gps_compass/u-blox_f9p_heading.md
```diff
@@ -0,0 +1,22 @@
...
+  - Secondary GPS = Moving Base
+- Set [GPS_UBX_MODE](../advanced_config/parameter_reference.md#GPS_UBX_MODE) to Heading (1)
+- Reboot and wait until both devices have GPS reception.
+  `gps status` should then show the Main GPS going into RTK mode, which means the heading angle is available.
+
+> **Note** RTK with a fix base station is still possible.
+  In that case the secondary GPS will show the RTK state w.r.t. the base station.
+
+For heading estimation the two antennas need to be on the same level.
+The direction that they are facing does not matter as it can be configured with the [GPS_YAW_OFFSET](../advanced_config/parameter_reference.md#GPS_YAW_OFFSET) parameter.
```
hamishwillee, line 22: To me this implies that each GPS has a direction. I think what you mean is that the line between the two GPS forms a heading relative to the front of the vehicle - right?

en/gps_compass/u-blox_f9p_heading.md
```diff
@@ -0,0 +1,22 @@
...
+  - Main GPS = Rover
+  - Secondary GPS = Moving Base
+- Set [GPS_UBX_MODE](../advanced_config/parameter_reference.md#GPS_UBX_MODE) to Heading (1)
+- Reboot and wait until both devices have GPS reception.
+  `gps status` should then show the Main GPS going into RTK mode, which means the heading angle is available.
+
+> **Note** RTK with a fix base station is still possible.
+  In that case the secondary GPS will show the RTK state w.r.t. the base station.
+
+For heading estimation the two antennas need to be on the same level.
```
hamishwillee, line 21: Might be worth having a diagram showing the mounting and the angle you might have to set.

hamishwillee:
Looks pretty good. I think this basically says that you can mount two GPS F9Ps on a vehicle and use them for heading instead/as well as a compass. I was confused at first because of the concept of mobile base, and that RTK always comes up in the context of RTK positioning.

Anyway, I have "spelled that all out" - see if it is more obvious to you without being patronising.

Two things you _might_ do
- Link to an F9P that you know works (ie in first sentence do an "e.g.")
- Diagram showing the wiring setup to Pixhawk 4. 
- Diagram showing the yaw setting in practice.

If you don't think necessary, feel free to merge :-)

en/gps_compass/u-blox_f9p_heading.md
bkueng: Yes exactly. This was taken from https://github.com/PX4/px4_user_guide/blame/master/en/gps_compass/rtk_gps_trimble_mb_two.md#L51.

en/gps_compass/u-blox_f9p_heading.md
bkueng: Indeed. I don't think I can come up with a nice one though.

bkueng:
Right, the naming can be a bit confusing. It comes from the fact that it uses the same technology: RTK, but with a moving base station instead of a fixed one.
All F9P's will work, as long as UART2 is exposed.
A yaw diagram would be nice.

en/gps_compass/u-blox_f9p_heading.md
hamishwillee: OK. Me either.

hamishwillee:
Hi @bkueng 

I have given it a further minor subedit (please check). I have two more questions before merging.
- Is there is a minimal separation between the GPS modules?
- When you say that for heading antennas need to be on same level/horizonal plane for heading, do you mean that they need to be mounted on the same horizontal plane, or do you mean that the GPS can only _calculate_ the heading based on the  horizontal vector between them. Or in other words, if they are 1m apart on front and back of a plane, then as you dive the horizontal component of the separation shrinks, which I presume reduces the "effective" separation and perhaps then the accuracy?

Probably we should mount them on the wings :-)

bkueng:
Thanks, looks good.

> Is there is a minimal separation between the GPS modules?

I have not seen any recommendations by u-blox.

> Or in other words, if they are 1m apart on front and back of a plane, then as you dive the horizontal component of the separation shrinks, which I presume reduces the "effective" separation and perhaps then the accuracy?

Yes I expect that as well.

> Probably we should mount them on the wings :-)

Right or a big vehicle :) I believe it's about 50cm in our setup (front and back).

hamishwillee:
@bkueng Thanks very much. I merged following a minor update to mention the specific section in the Application note where the positioning is mentioned, and to note that the minimum separation is not known, but that 50cm works.

Didn't mention the other assumptions above, since they can't be confirmed. I think we've said what we can for now.

PR #831, opened 2020-10-14
uavcan_escs: Add VESC setup information

RicardoM17:
Signed-off-by: Ricardo Marques <marques.ricardo17@gmail.com>

### DIFF
--- a/en/peripherals/uavcan_escs.md
+++ b/en/peripherals/uavcan_escs.md
@@ -149,7 +149,7 @@ Parameter | Option
 --- | ---
 App to use | `No App`
 VESC ID | `1,2,...`
-CAN Status Message Mode | `CAN_STATUS_1_2_3_4_5`
+Can Status Message Mode | `CAN_STATUS_1_2_3_4_5`
 CAN Baud Rate | `CAN_BAUD_500K`
 CAN Mode | `UAVCAN`
 UAVCAN ESC Index | `0,1,...`
--- a/en/peripherals/uavcan_escs.md
+++ b/en/peripherals/uavcan_escs.md
@@ -141,8 +141,21 @@ There is some guidance here: [Quick start guide for Myxa v0.1](https://forum.zub
 
 ### VESC ESC Setup
 
-For VESC the preferred tool for motor enumeration is the [VESC tool](https://vesc-project.com/vesc_tool).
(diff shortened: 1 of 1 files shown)


RicardoM17:
@hamishwillee FYI

hamishwillee:
Thanks @RicardoM17 ! 

1. Does it matter the order that you do things? i.e This says you do normal setup in the VESC tool and also set up the app configuration - but I assume that you're saying first do the app configuration so that the VESC setup works?
2. Is it obvious where the app configuration is? I was looking for a screenshot on google play and there isn't an obvious app config place. It might be worth a screenshot.

RicardoM17:
Hi @hamishwillee both the motor configuration and the app configuration are done in the VESC Tool. For the motor configuration however, no pixhawk is needed for example. In order to run with px4 both need to be done but the order is not that important.

The place for the app configuration in the VESC Tool is quite obvious yes. If needed I can add a screenshot at a later date with a more detailed explanation? Not sure if it’s needed however.

P.s. I saw your subedit but I’ll leave it up to you. Unfortunately the VESC tool is not consistent and it has both Can and CAN. So I left it as it is in the VESC Tool. But either way it’s not a big deal and we can opt to be consistent on our side.

hamishwillee:
@RicardoM17 Thanks - I've reverted the parameter name and merged. If this is pretty much obvious once you're using the app then you've almost certainly hit the right level of detail - i.e. explain what they need to change and the essential params.

RicardoM17:
@hamishwillee No problem @hamishwillee , I'm happy to contribute :) 

I might do a follow up PR with a screenshot of the VESC Tool just to make it complete. It's never enough. And I'll also add a link to purchase VESC from https://trampaboards.com/vesc--c-1434.html , which is the official store for VESC. There are other alternatives but we've had quite a lot of issues with them and lost a ton of time in the process. I don't think those alternatives are appropriate for UAVs. What do you think?

hamishwillee:
>  I don't think those alternatives are appropriate for UAVs. What do you think?

I think that if there is a single great alternative that does the job and is "preferred" we use that. Perhaps just a note like "Other VESC configuration apps exist but these have not be validated/tested. These include (non-exhaustively): 

Yes I know you have tested them, the point is to recommend something good and to let people know the other choices if they want to look further. Don't list any you know are really crap for drones.

Cheers!

RicardoM17:
@hamishwillee my meaning with "what do you think" was mostly on adding the link to this "official" version, which we know works.

FYI I was talking about differente VESC manufacturers, not necessarily new configs (though those exist too).

That being said it does sound like a good idea so expect a PR from me soon with that info ;) 

Have a good weekend.

hamishwillee:
Thanks for the clarification. Yes :-)

PR #852, opened 2020-10-28
Improve sensor orientation info

hamishwillee:
This PR modifies the sensor/FCorientation sections to make them more accessible to new users. 
- Images of vehicles show roll/pitch/yaw and relationship to orientation config is more clear.
- Clarification that VTOL tailsitter is an exception in that its orientation is relative to takeoff position, not forward moving.

What this needs is a better image that shows the tailsitter - e.g. "like the image below but showing it landed orientation". Anyone got any drawing skills?

image

@RomanBapst Does this change make sense to you?


davidcgregory:
@hamishwillee I think it would be helpful to label the diagrams with a consistent set of nomenclature.  Sometimes rotations are referred to as roll, pitch, and yaw.  Sometimes Omega, Theta, phi.   And in the flight review gui, gyro rates are labeled "X,Y, and Z", actuator controls are labeled "Roll, pitch, yaw", and RC inputs are "Y/Roll, X/Pitch, Yaw".    !!!!!!

hamishwillee:
@davidcgregory If you want to follow up in each place making things consistent then pleas do. I agree it would be nice, but not as nice as the millions of other things on my list :-)

hamishwillee:
Note, merged. Better than it was and not worth waiting on @RomanBapst to comment.

PR #854, opened 2020-10-30
Update pid_tuning_guide_fixedwing.md

orkunasa:
Should be pitch angle because of under the pitch tuning title


RicardoM17 (approved):

RicardoM17:
Thanks for the fix contribution @orkunasa . Merged!

PR #897, opened 2020-11-19
fixed_wing_takeoff: replace explicit heading anchors

MaEtUgR:
image

Is this the right way to fix it? I copied from https://github.com/PX4/PX4-user_guide/pull/821


hamishwillee:
Thanks @MaEtUgR - yes, that's right. The other way is to go through all the links that point here and switch them to point to the auto-generated header anchor.

PR #905, opened 2020-11-25
Update contribution pages

hamishwillee:
Also covers translation and the main readme.

FYI @RicardoM17 :-)


hamishwillee:
Note, previously this was for all the libraries (e.g. mavlink too). However they use different infrastructure and IMO it is too early to migrate them. So for now, this is just tidied to be "PX4 User Guide" only.

RicardoM17:
Looks good @hamishwillee 

I'll try to read it today and see if I spot any improvements or errors :smile:

hamishwillee:
Thanks very much!

PR #909, opened 2020-11-25
Switch all adding_a_new_frame links from airframes/ to dev_airframes/ folder

MaEtUgR:
I got a report the second link "airfram_configurations" on this page is broken (404):
http://docs.px4.io/master/en/advanced/parameters_and_configurations.html

I found that the file is now under the `dev_airframes/` folder and replaced all links to that particular file with search and replace. 

@hamishwillee could you please double check since I could not test each and every link since I don't have the rendered docs and I'm not sure if it's the right way to change the link directly also in other languages.


hamishwillee:
Thanks very much @MaEtUgR ! I've replaced this with #910 to get this in quick. The reason is that you only should update the docs in the /en tree. The other docs are round-tripped through a translation website - so will automatically update. If we update them manually then it can be a bit of a pain to resolve conflicts

PR #938, opened 2020-12-02
Fix warning message for Raspbery Pi

elpiel:
@hamishwillee, I've moved the changes here.


hamishwillee:
Thanks very much! The link format wasn't quite right so I patched it. Basically to link to almost anything you have to go down one level to root then up to the file via its folder. You had just gone down one level, but not up again.

elpiel:
Oh yeah, np, I see there is still some work to be done ;)

Issue #941, opened 2020-12-02
Doc Bug: Acro (MC) · PX4 v1.8.2-1.11 User Guide


whataberk:
In the expo-describing equation: y=r(x^3+x(1−f))(1−g)/(1−g|x|) it does not appear to match the description or the code. I believe that it should be y=r(**f***x^3+x(1−f))(1−g)/(1−g|x|) based on the code from [here](https://github.com/PX4/PX4-Autopilot/blob/f13bbacd5277123d6af2cb5ed21587c220031353/src/lib/mathlib/math/Functions.hpp#L70), copied below:
```cpp
template<typename T>
const T expo(const T &value, const T &e)
{
	T x = constrain(value, (T) - 1, (T) 1);
	T ec = constrain(e, (T) 0, (T) 1);
	return (1 - ec) * x + ec * x * x * x;
}
```

This appears to affect all extant versions of the guide (1.8-1.11).

Bug Page: [Acro (MC) · PX4 v1.8.2 User Guide](https://docs.px4.io/v1.8.2/en/flight_modes/acro_mc.html)

RicardoM17:
Hi @whataberk thanks for raising this. I see that this was merged to master - V1.10 in #942 .

@hamishwillee I closed this. Is that ok or are you planning to back port to even older versions? At this point it doesn't make much sense to support such old firmware/documentation IMO.

hamishwillee:
@RicardoM17 Thanks, I merged the fix and backported to the last couple of releases. Just missed that this was linked :-)

PR #942, opened 2020-12-02
acro_mc.md: Add expo term

whataberk:
Add the expo factor to the cubic term in the exponential describing function. This needs to be merged to 1.9-1.11 too - I can create separate pull requests if necessary? 

#941

### DIFF
--- a/en/flight_modes/acro_mc.md
+++ b/en/flight_modes/acro_mc.md
@@ -28,7 +28,7 @@ This roll and pitch input stick response can be tuned using the [MC_ACRO_EXPO](#
 
 ![Acro mode - expo - pure linear input curve](../../images/flight_modes/acro_mc_input_curve_expo_linear.png) ![Acro mode - expo - pure cubic input curve](../../images/flight_modes/acro_mc_input_curve_expo_cubic.png)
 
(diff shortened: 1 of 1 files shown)


hamishwillee:
@whataberk Thanks! FYI, we generally start by fixing the master version of the docs first. The fix may then be back ported to the last version of the docs

@bresch Does this make sense to you? (does not to me, but then I am probably missing something)

whataberk:
@hamishwillee Would you like me to make a new branch and pull request off of `master`? I'd be happy to. 

The issue I added (#942) and mentioned in the PR noted was that the docs version of the function a) does not match the code and b) does not match the description of the algorithm. That is, if `f` goes to zero in the docs function then you still have a cubic component in the function.

hamishwillee:
@whataberk Let's get @bresch review and I can push it up to master. No point adding work if he has some other opinion. Just a note for next time :-)

bresch (approved):
That's correct, thanks

hamishwillee:
Thanks @whataberk and @bresch - merged, and I fixed up master, v1.11 and v1.10 in line with this.

PR #948, opened 2020-12-03
Update information about simulated batteries

potaito:
Information about minimum battery percentage is outdated. The lines referenced in mavlink_receiver.cpp no longer exist :sweat_smile: and devs also no longer have to modify this one line to change the minimum simulated battery percentage :+1: 

Relevant upstream changes that removed the battery sim from simulator_mavlink into its own module: https://github.com/PX4/PX4-Autopilot/commit/55808ed2f9c47bb85edc1dab36212faf1c09b4db#

Not sure if other information relevant to the simulation needs to be updated too. I only had a quick look and did not find anything.

### DIFF
--- a/en/simulation/failsafes.md
+++ b/en/simulation/failsafes.md
@@ -43,12 +43,12 @@ All parameters in SITL including this one get reset when you do `make clean`.
 The simulated battery is implemented to never run out of energy, and by default only depletes to 50% of its capacity and hence reported voltage.
 This enables testing of battery indication in GCS UIs without triggering low battery reactions that might interrupt other testing.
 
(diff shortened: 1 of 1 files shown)


hamishwillee:
Thanks @potaito - that's a good addition.

hamishwillee:
PS Do you happen to know if there are many other similar options? I was thinking this might be useful to add in https://docs.px4.io/master/en/simulation/jmavsim.html#usage-configuration-options and https://docs.px4.io/master/en/simulation/gazebo.html#usage-configuration-options
(or even the top level doc, but I think easier to find if duplicated by platform)

potaito:
@hamishwillee I wouldn't know right now, but I'll keep an eye open :)

hamishwillee:
FYI only added some of this info in https://github.com/PX4/PX4-user_guide/pull/981

PR #953, opened 2020-12-07
Update external_position_estimation.md

netkimjh:
- Section Reference Frames and ROS
FRD and FLU explanation is not correct for the figure. So I correct the explanation.
Also, switch the columns of ROS and PX4 in the table to keep the consistency of the side.


netkimjh:
I will resubmit the commit

PR #961, opened 2020-12-09
Position mode: Add note about Position Loss

hamishwillee:
Add note about what happens if you lose position estimate in position mode.

It is an edge case, but helpful because this is probably first place you'll look.

If this is seen as OK I'll add to the Position Mode FW, Land, RTL etc which depend on position.

### DIFF
--- a/en/flight_modes/position_mc.md
+++ b/en/flight_modes/position_mc.md
@@ -60,4 +60,5 @@ Parameter | Description
 ### Position Loss/Safety
 
 Position mode is dependent on having an acceptable position estimate.
-If the estimage falls below acceptable levels, for example due to GPS loss, this may trigger a [Position (GPS) Loss Failsafe](../config/safety.md#position-gps-loss-failsafe). Depending on configuration, whether you have a remote control, and whether there is an adequate altitude estimate, PX4 may switch to altitude mode, manual mode, land mode or terminate.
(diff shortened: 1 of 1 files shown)


AlexisTM:
I am uncertain if this specific case is in the configuration. The "going from position to altitude mode" when having no position (GPS/PX4Flow/Mocap) is mandatory as there is no data to control over.

hamishwillee:
> I am uncertain if this specific case is in the configuration. The "going from position to altitude mode" when having no position (GPS/PX4Flow/Mocap) is mandatory as there is no data to control over.

@AlexisTM  This particular PR doesn't make the statement about exactly what happens; what it does is highlight that you might switch out of the mode  if you lose position with links to more information. That is kind of the right level for THIS document.

Upshot, I'm going to merge this. 

But I still want to make sure that the linked documents are correct. W.r.t. to your comment according to the docs the failure action is controlled by COM_POSCTL_NAVL, based on whether RC control is assumed to be available and whether and altitude information is available:
- 0: Remote control available. Switch to Altitude mode if a height estimate is available, otherwise Stabilized mode.
- 1: Remote control not available. Switch to Land mode if a height estimate is available, otherwise enter flight termination.

Can you please explain what you mean above - are you saying that the docs on COM_POSCTL_NAVL are wrong or incomplete?

PR #966, opened 2020-12-11
uORB: make link to deprecated_msgs to a fixed commit

MaEtUgR:
The old link was pointing to the wrong line by now. Since the line number can switch pretty often and I think it's better to have the link to outdated code than potentially and in older versions to nowhere, I suggest linking to one specific commit.


bkueng (approved):

hamishwillee:
@MaEtUgR Yes. Thanks. 

FWIW 
An alternative I sometimes use is to do the link to master with no line, and create the context by adding the actual thing I want to show. So it might look like:

 - Removed or renamed messages must be added to the `deprecated_msgs` list in [msg/CMakeLists.txt](https://github.com/PX4/PX4-Autopilot/blob/master/msg/CMakeLists.txt) and the **.msg** file needs to be deleted.
   ```
   set(deprecated_msgs
	 ekf2_innovations.msg	# 2019-11-22, Updated estimator interface and logging; replaced by 'estimator_innovations'.
         <insert .msg here>
	 )
   ```

Completely up to you.

PR #970, opened 2020-12-16
Update flight_tasks.md

yutaide:
Added explanation for creating and testing new flight task with simple examples.


hamishwillee:
Thanks @yutaide - absolutely AWESOME to be getting some docs on this.

@Stifael @MaEtUgR Can you please review/suggest improvements?

My concerns are mostly about the fact that this is a tutorial and not a guide. The different is that a guide explains how things work, and can therefore be extended to any use case. By contrast a tutorial shows how you got a particular thing to work - and while it is helpful, it can mean that it is hard for someone else to extend your work to some other case.

So my thinking is:
1. Separate out what you chose to do vs. what you must do . i.e. 
   - this test flight task gets used in position control, but a task might be used in any mode. So it is important to explain what this shows, and what is different if you use other modes.
   - This uses ROS and simulation for testing. Is that something you HAVE to do? 
   - Do you HAVE to use the virtual joystick?
2. In the examples you have "copy this text" etc. That may well date badly if source code changes. What we could do here is have a template of some kind and instead of reproducing the text, link to it. @Stifael Is a template flight task like the one uses here "possible/practical".
3. I'd prefer to merge the overview and details. Lot of duplication there. 
4. It is a long time since I watched that video, but I recall there were concerns about using flight tasks that need to be captured. For example, if you switch tasks, do you need to do anything to ensure that the transitions is smooth - ie does the estimator smoothly switch to the new task/setpoint? In a guide, these are the most important things.
5. In a guide good to outline the classes you have to derive from up the top - ie explain the architecture first.

In summary, can we generalise this with a template flight task, architectural and "gotcha" overview?

yutaide:
@hamishwillee 
You're right! It's more a tutorial than a guide.
Please rework and improve this document as you like, to be more useful.

hamishwillee:
Thanks @yutaide - I'll try look over next couple of weeks. Will need your help to review changes I make. @MaEtUgR @Stifael yours too!

en/concept/flight_tasks.md
```diff
@@ -2,8 +2,141 @@
...
+     return ret;
+   }
+
+   bool FlightTaskMyTask::update()
+   {
+     PX4_INFO("FlightTaskMyTask update was called!"); // check updated.
+     return true;
+   }
+   ```
+5. Add the new task to the list of tasks to be build in [/src/lib/flight_tasks/tasks/CMakeLists.txt](https://github.com/PX4/PX4-Autopilot/blob/master/src/lib/flight_tasks/tasks/CMakeLists.txt):
```
hamishwillee, line 88: @yutaide @MaEtUgR This does not appear to be true for master - I THINK you don't need to do this because CMAKE just searches the current tree for file names. Correct? Or do you explicitly have to add them still?

hamishwillee:
@yutaide @MaEtUgR I have done first shot at making this more of a guide and less of a tutorial. Mostly this means explaining what we have to do more "generically". i.e. essentially you create a flight task by creating a normal module, deriving a class from `FlightTask`, and implementing `update()` and `activate()`. You also have to make sure it gets called "somehow" - this changes docs from "you do this" to "you normally activate with a parameter, this is how you do this in order to call the example in position mode".

For testing the main point is to highlight that you do need to test. I have simplified that to the bare minimum - set the parameter, takeoff, set the position mode, and clarified that this is for the current example, not the general case.

@MaEtUgR Can you please start thinking about this. We might for example be more methodical and talk about when you implement the OTHER virtual methods (or state that we don't). We might state what else FlightTasks does for you "by default". Also the problems with current framework. 

Can you review and help suggest next steps to updating?

en/concept/flight_tasks.md
yutaide: @hamishwillee 
Thank you for your rework!
I'm not sure if this step is necessary or not since I'm not familiar with CMAKE.
@MaEtUgR  What do you think about this?

en/concept/flight_tasks.md
```diff
@@ -2,8 +2,141 @@
...
+       case 2: //Add option for new task: MyTask
+          error =  _flight_tasks.switchTask(FlightTaskIndex::MyTask);
+          break;
+     ...
+     ```
+  
+
+##  Test New Flight Task
+
+To test the flight stack you need to run the vehicle with the task enabled.
```
yutaide, line 140: Is this `flight stack` or flight task ?
If this does not matter, please just ignore.

en/concept/flight_tasks.md
```diff
@@ -2,8 +2,141 @@
...
+For the example above, this means setting the parameter `MPC_POS_MODE=2`, taking off, and switching the vehicle to [Position mode](../flight_modes/position_mc.md).
+This would normally be done on the simulator.
```
yutaide, line 142: If someone follow this docs and test a flight with real vehicle, the vehicle may be start falling after switching MyTask mode because no setpoint is set.
Maybe some caution should be written...?

yutaide (commented):
@hamishwillee 
Thank you for your rework.
I've commented some.

en/concept/flight_tasks.md
hamishwillee: I've modified this to make the point that it won't fly. Good catch :-)

hamishwillee:
Thanks @yutaide ! @MaEtUgR can you please review and comment on https://github.com/PX4/PX4-user_guide/pull/970/files#r549215174 ?

en/concept/flight_tasks.md
MaEtUgR: Thank you so much @hamishwillee !
You still have to add the name of your task but it's here on master:
https://github.com/PX4/PX4-Autopilot/blob/master/src/modules/flight_mode_manager/CMakeLists.txt#L40

MaEtUgR:
> essentially you create a flight task by creating a normal module

It's not a separate module in the PX4/NuttX sense. All tasks (which probably get called modes in the future live in the flight mode manager module https://github.com/PX4/PX4-Autopilot/tree/master/src/modules/flight_mode_manager and a new task is just a new class inside the framework that produces setpoints.

@hamishwillee Thanks for all your work on this and sorry for my recurring absence from the discussion. I think we need at least a minimal documentation including example about the framework that is up to date and mentions the key concept. The more "more methodical" approach including all the possibilities probably makes sense to be built along with improvements to how it works at the moment. The intention is that it's really flexible and easy to use but some of the goals were not achieved yet. To find the remaining rough edges it very likely helps to try to explain in simple terms. I probably use everything to often to quickly see where other developers will struggle most. TL;DR Let's add documentation such that people don't get stranded and improve it along with framework improvements.

en/concept/flight_tasks.md
```diff
@@ -2,6 +2,158 @@
 
 *Flight Tasks* are used within [Flight Modes](../concept/flight_modes.md) to provide specific movement behaviours: e.g. follow me, or flight smoothing.
 
-At time of writing the best description of how flight modes work (in PX4 v1.9) is provided in the following video: **PX4 Flight Task Architecture Overview** - Dennis Mannhart, Matthias Grob (*PX4 Developer Summit 2019*).
+## Video
+
+The following vidoes provide an overview of flight tasks in PX4.
```
MaEtUgR, line 7:
```suggestion
The following videos provide an overview of flight tasks in PX4.
```

en/concept/flight_tasks.md
```diff
@@ -2,6 +2,158 @@
...
+The first covers the state of tasks in PX4 v1.9.
```
MaEtUgR, line 8:
```suggestion
The first covers the state of the flight task framework in PX4 v1.9.
```

en/concept/flight_tasks.md
```diff
@@ -2,6 +2,158 @@
...
+#### Overview of multicopter control from sensors to motors (PX4 Developer Summit Virtual 2020)
+
+@[youtube](https://youtu.be/orvng_11ngQ?t=560) <!-- datestamp:video:youtube:20200720:Overview of multicopter control from sensors to motors — PX4 Developer Summit Virtual 2020 From 9min20sec - Section on flight tasks-->
+
+The relevent section of this video is an update of flight tasks in PX4 v11.1 at (9min 20sec).
+The [slides can be found here (PDF)](https://static.sched.com/hosted_files/px4developersummitvirtual2020/1b/PX4%20Developer%20Summit%202020%20-%20Overview%20of%20multicopter%20control%20from%20sensors%20to%20motors.pdf) - Slides 9 and 12 are relevant.
+
+## Creating a Flight Task
+
+Flight tasks are [PX4 modules/applications](../modules/hello_sky.md) that contain a task definition class derived from [FlightTask](https://github.com/PX4/PX4-Autopilot/blob/master/src/lib/flight_tasks/tasks/FlightTask/FlightTask.hpp).
```
MaEtUgR, line 26:
```suggestion
A flight task is a class in the flight task framework derived from the base class [FlightTask](https://github.com/PX4/PX4-Autopilot/blob/master/src/modules/flight_mode_manager/tasks/FlightTask/FlightTask.hpp). Its goal is to generate setpoints for the controller from arbitrary input data, where each task implements the desired vehicle behavior for a specific mode.
```

en/concept/flight_tasks.md
```diff
@@ -2,6 +2,158 @@
...
+Programmers typically reimplement the `activate()` and `update()` virtual methods.
```
MaEtUgR, line 27:
```suggestion
Programmers typically override the `activate()` and `update()` virtual methods by calling the base task's minimal implementation and extending with the implementation of the desired behavior.
```

en/concept/flight_tasks.md
```diff
@@ -2,6 +2,158 @@
...
+The `activate()` method is called with the setpoint from the previous task when the task is switched to, while `update()` is called during execution of the control loop to implement the task.
```
MaEtUgR, line 28:
```suggestion
The `activate()` method is called when switching to the task and allows to initialize its state and take over gently from the passed over setpoints the previous task was just applying.

`update()` is called on every loop iteration during the execution and contains the core behavior implementation producing setpoints.
```

en/concept/flight_tasks.md
```diff
@@ -2,6 +2,158 @@
...
+
+By convention tasks are contained in a sub folder of [PX4-Autopilot/src/lib/flight_tasks/tasks/](https://github.com/PX4/PX4-Autopilot/tree/master/src/lib/flight_tasks/tasks) (named after the task), and the source files are named with the prefix "FlightTask".
```
MaEtUgR, line 30:
```suggestion
By convention tasks are contained in a subfolder of [PX4-Autopilot/src/modules/flight_mode_manager/tasks](https://github.com/PX4/PX4-Autopilot/tree/master/src/modules/flight_mode_manager/tasks) named after the task, and the source files are named with the prefix "FlightTask".
```

en/concept/flight_tasks.md
```diff
@@ -2,6 +2,158 @@
...
+
+The instructions below might be used to create a task named *MyTask*:
+
+1. Create a directory for the new flight task in [PX4-Autopilot/src/lib/flight_tasks/tasks/](https://github.com/PX4/PX4-Autopilot/tree/master/src/lib/flight_tasks/tasks).
```
MaEtUgR, line 34:
```suggestion
1. Create a directory for the new flight task in [PX4-Autopilot/src/modules/flight_mode_manager/tasks](https://github.com/PX4/PX4-Autopilot/tree/master/src/modules/flight_mode_manager/tasks).
```

en/concept/flight_tasks.md
```diff
@@ -2,6 +2,158 @@
...
+   By convention the directory is named after the task, so we will call it **/MyTask**.
+   ```
+   mkdir PX4-Autopilot/src/lib/flight_tasks/tasks/MyTask
+   ```
+2. Create empty source code and *cmake* files for the new flight task in the *MyTask* directory using the prefix "FlightTask":
+   - CMakeLists.txt
+   - FlightTaskMyTask.hpp
+   - FlightTaskMyTask.cpp
+3. Update **CMakeLists.txt** for the new task
+   - Copy the contents of the **CMakeLists.txt** for another task - e.g. [AutoMapper/CMakeLists.txt](https://github.com/PX4/PX4-Autopilot/blob/master/src/lib/flight_tasks/tasks/AutoMapper/CMakeLists.txt)
```
MaEtUgR, line 44:
```suggestion
   - Copy the contents of the **CMakeLists.txt** for another task - e.g. [Orbit/CMakeLists.txt](https://github.com/PX4/PX4-Autopilot/blob/master/src/modules/flight_mode_manager/tasks/Orbit/CMakeLists.txt)
```

en/concept/flight_tasks.md
```diff
@@ -2,6 +2,158 @@
...
+   - Update the copyright to the current year
+     ```cmake   
+     ############################################################################
+     #
+     #   Copyright (c) 2021 PX4 Development Team. All rights reserved.
+     #
+     ```
+   - Modify the code to reflect the new task - e.g. replace `FlightTaskAutoMapper` with `FlightTaskMyTask`.
```
MaEtUgR, line 52:
```suggestion
   - Modify the code to reflect the new task - e.g. replace `FlightTaskOrbit` with `FlightTaskMyTask`.
```

en/concept/flight_tasks.md
```diff
@@ -2,6 +2,158 @@
...
+   {
+   public:
+     FlightTaskMyTask() = default;
+     virtual ~FlightTaskMyTask() = default;
+
+     bool update();
+     bool activate(vehicle_local_position_setpoint_s last_setpoint);
+
+   private:
+     float _origin_z = 0.0f;
```
MaEtUgR, line 80:
```suggestion
     float _origin_z{0.f};
```

en/concept/flight_tasks.md
```diff
@@ -2,6 +2,158 @@
...
+   ```
+4. Update the cpp file as appropriate.
+   This example provides as simple implementation of **FlightTaskMyTask.cpp** that simply indicates that the task methods are called.
+   ```cpp
+   #include "FlightTaskMyTask.hpp"
+
+   bool FlightTaskMyTask::activate(vehicle_local_position_setpoint_s last_setpoint)
+   {
+     bool ret = FlightTask::activate(last_setpoint);
+     PX4_INFO("FlightTaskMyTask activate was called! ret:%d", ret); // check if engaged this mode.
```
MaEtUgR, line 91:
```suggestion
     PX4_INFO("FlightTaskMyTask activate was called! ret: %d", ret); // report if activation was succesful
```

en/concept/flight_tasks.md
```diff
@@ -2,6 +2,158 @@
...
+     return ret;
+   }
+
+   bool FlightTaskMyTask::update()
+   {
+     PX4_INFO("FlightTaskMyTask update was called!"); // check updated.
```
MaEtUgR, line 97:
```suggestion
     PX4_INFO("FlightTaskMyTask update was called!"); // report update
```

en/concept/flight_tasks.md
```diff
@@ -2,6 +2,158 @@
...
+     return true;
+   }
+   ```
+5. Add the new task to the list of tasks to be build in [/src/lib/flight_tasks/tasks/CMakeLists.txt](https://github.com/PX4/PX4-Autopilot/blob/master/src/lib/flight_tasks/tasks/CMakeLists.txt):
```
MaEtUgR, line 101:
```suggestion
5. Add the new task to the list of tasks to be built in [PX4-Autopilot/src/modules/flight_mode_manager/CMakeLists.txt](https://github.com/PX4/PX4-Autopilot/blob/master/src/modules/flight_mode_manager/CMakeLists.txt#L40):  
```

en/concept/flight_tasks.md
```diff
@@ -2,6 +2,158 @@
...
+     ManualPositionSmooth
+     ManualPositionSmoothVel
+     AutoLineSmoothVel
+     AutoFollowMe
+     Offboard
+     Failsafe
+     Descend
+     Transition
+     MyTask
+     ${flight_tasks_to_add}
```
MaEtUgR, line 118:
```suggestion
   list(APPEND flight_tasks_to_add
      Orbit
      MyTask
   )
```

en/concept/flight_tasks.md
```diff
@@ -2,6 +2,158 @@
...
+   )
+   ```
+6. Update a flight mode to ensure that the task is called. 
+   Usually a parameter is used to select when a particular flight task should be used.
+   
+   For example, to enable our new `MyTask` in the Manual-Position control sub-mode:
```
MaEtUgR, line 124:
```suggestion
   For example, to enable our new `MyTask` in multicopter Position mode:
```

en/concept/flight_tasks.md
```diff
@@ -2,6 +2,158 @@
...
+   - Update `MPC_POS_MODE` ([mc_pos_control_params.c](https://github.com/PX4/PX4-Autopilot/blob/master/src/modules/mc_pos_control/mc_pos_control_params.c)) to add an option for selecting "MyTask" if the parameter==2:
```
MaEtUgR, line 125:
```suggestion
   - Update `MPC_POS_MODE` ([mc_pos_control_params.c](https://github.com/PX4/PX4-Autopilot/blob/master/src/modules/mc_pos_control/mc_pos_control_params.c#L706-L721)) to add an option for selecting "MyTask" if the parameter has a previously unused value like 5:
```

en/concept/flight_tasks.md
```diff
@@ -2,6 +2,158 @@
...
+     ```
+     ...
+     * @value 2 MyTask
+     * @value 3 Smooth position control (Jerk optimized)
+     * @group Multicopter Position Control
+     */
+     PARAM_DEFINE_INT32(MPC_POS_MODE, 3);
```
MaEtUgR, line 132:
```suggestion
     * @value 4 Acceleration based input
     * @value 5 MyTask position mode implementation
     * @group Multicopter Position Control
     */
    PARAM_DEFINE_INT32(MPC_POS_MODE, 4);
```

en/concept/flight_tasks.md
```diff
@@ -2,6 +2,158 @@
...
+     ``` 
+   - Add a switch in [MulticopterPositionControl.cpp](https://github.com/PX4/PX4-Autopilot/blob/master/src/modules/mc_pos_control/MulticopterPositionControl.cpp) to enable the task when `MPC_POS_MODE==2`"
```
MaEtUgR, line 134:
```suggestion
   - Add a case for your new option in the switch for the parameter [FlightModeManager.cpp](https://github.com/PX4/PX4-Autopilot/blob/master/src/modules/flight_mode_manager/FlightModeManager.cpp#L266-L285) to enable the task when `_param_mpc_pos_mode` has the right value.
```

en/concept/flight_tasks.md
```diff
@@ -2,6 +2,158 @@
...
+     // manual position control
+     ...
+     switch (_param_mpc_pos_mode.get()) {
+       case 1:
+          error =  _flight_tasks.switchTask(FlightTaskIndex::ManualPositionSmooth);
+          break;
+
+       case 2: //Add option for new task: MyTask
+          error =  _flight_tasks.switchTask(FlightTaskIndex::MyTask);
+          break;
```
MaEtUgR, line 146:
```suggestion
     switch (_param_mpc_pos_mode.get()) {
       ...
       case 3:
          error = switchTask(FlightTaskIndex::ManualPositionSmoothVel);
          break;
       case 5: // Add case for new task: MyTask
          error =  _flight_tasks.switchTask(FlightTaskIndex::MyTask);
          break;
	case 4:
	....
```

en/concept/flight_tasks.md
```diff
@@ -2,6 +2,158 @@
...
+     ...
+     ```
+  
+
+##  Test New Flight Task
+
+To test the flight task you need to run the vehicle with the task enabled.
+For the example above, this means setting the parameter `MPC_POS_MODE=2`, taking off, and switching the vehicle to [Position mode](../flight_modes/position_mc.md).
```
MaEtUgR, line 154:
```suggestion
For the example above, this means setting the parameter `MPC_POS_MODE` to 5, taking off, and switching the vehicle to [Position mode](../flight_modes/position_mc.md).
```

en/concept/flight_tasks.md
```diff
@@ -2,6 +2,158 @@
...
+
+:::note
+The task defined above should only be tested on the simulator (the code doesn't actually create setpoints so the vehicle will not fly).
```
MaEtUgR, line 157:
```suggestion
The task defined above should only be tested on the simulator. The code doesn't actually create setpoints so the vehicle will not fly.
```

MaEtUgR (commented):
I think that's a good start for basic understanding. I'll have to add some theory on what's actually happening and what the interfaces are but I'll do that on a new branch.

hamishwillee:
@MaEtUgR Thanks so much for the review - your updates clarify a lot of things for me, so I think that this will be much better for end users now. 

I made a structural modification to separate out your introductory paragraph into an "overview" section, pushing the videos to the bottom (but linked). This is better because it provides a succinct summary, and gives you somewhere to build the explanation of flight tasks when you do your further additions. 

Merging now. @yutaide - thank you too - we needed this!

PR #973, opened 2020-12-17
Group hardware sections

hamishwillee:
This is an experiment in reducing the sidebar. 
- It moves hardware topics under a top level as shown. 
- It also moves the mavlink and serial port configuration into the advanced configuration - my theory here is that those are actually config topics that a peripheral should link to. I am more than happy perhaps to move the mavlink one back if there is disagreement on this.

The main reason for this is to make the sidebar less scary - it is still easy to find all the hardware - where else would it be :-) ? But now you 

You can get a view of it here: https://docs.px4.io/v1.testing/en/complete_vehicles/px4_vision_kit.html

image

If this is OK will improve the entry point a bit before merging.

@bkueng Would appreciate your thoughts on this as a structure.


bkueng (approved):
Looks better than before.
Maybe RTF in the title is not clear to most people?

hamishwillee:
I removed RTF and added a basic intro into the parent.

Probably make sense to also move the companion computer page in here at some point and also modify the parent page to have a diagram showing how all the bit fit together. Maybe something for next week.

Thanks for taking the time to look at this @bkueng - good to get a independent view.

hamishwillee:
@bkueng I'm actually wondering if we should also shove "Basic assembly" in there too.

PR #974, opened 2020-12-18
Update holybro_qav250_pixhawk4_mini.md

Ostefanini:
Update qav 250 pixhawk4 mini doc in correspondance with the kit I've just received this week.

Here is the commercial mail associated:
Sorry, due to the change of the motor supplier, the appearance of the motor in the current package is different from that of the previous version.
The new motor has no DX/SX difference.
All the M5nuts are black.

### DIFF
--- a/en/frames_multicopter/holybro_qav250_pixhawk4_mini.md
+++ b/en/frames_multicopter/holybro_qav250_pixhawk4_mini.md
@@ -115,7 +115,7 @@ The following tools are used in this assembly:
    ![QAV250 Add power module and standoffs](../../assets/airframes/multicopter/qav250_holybro_pixhawk4_mini/5_power_module_on_screws.jpg)
 1. Put the flight controller plate on the standoffs (over the power module)
    ![QAV250 Add flight controller plate](../../assets/airframes/multicopter/qav250_holybro_pixhawk4_mini/6_flight_controller_plate.jpg)
-1. Attach the motors. The motors have an arrow indicating the direction of rotation.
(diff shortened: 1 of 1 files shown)


Ostefanini:
It looks like a very interesting questions ! I haven't finished my build yet, because I will receive some additional components in the next days, but I ''ve added a remind to update the PR right after that.

Since this pull request, I noticed two ther changes:

1. about the radio module wich is now a kit with two module identicals:

http://www.holybro.com/product/transceiver-telemetry-radio-v3-915mhz/

2. about the (optionnal) camera wiring. 

Should I open two other pull requests or just update this one (and change it's title) ?

This frame build tutorial is important as one of the less expensive kit, I will be very happy to participate 👍

hamishwillee:
Hi @Ostefanini 

Feel free to update this PR with the new materials. I agree, this is a good build to keep updated!

hamishwillee:
Hi @Ostefanini 
Any progress on the open questions?

Ostefanini:
Hello, sorry for losing attention Here. You were obviously right, even if the mots are not self oriented, one still needs to ensure clockwise and anti-clockwise spin orientation. It also needs to put propeller in correct orientation.

hamishwillee:
Thanks @Ostefanini. Want to make an update so we can get this in? Something along the lines of "You may get 4 motors with blacks nuts on their top rather than XXXX. In this case you still need to YYYY ?

hamishwillee:
OK. I have decided to close this. You are very welcome to reopen it when you have a moment :-)

PR #983, opened 2020-12-28
Update dev_env_mac.md: Fix invalid brew syntax

canberkdurmus:
`brew cask install xquartz` 

Command gives an error:

`Error: Calling brew cask install is disabled! Use brew install [--cask] instead.`

Correct Sytax added: `brew install --cask xquartz`


hamishwillee:
Thanks very much!

PR #993, opened 2021-01-12
Clarify that rate output limits are applied in the mixer, not controller

hamishwillee:
This comes from a question in [Slack](https://px4.slack.com/archives/C581ST17V/p1609837421266100) about the controller code not showing the output limiting. It clarifies that the rate output limit is in the mixer. I think that is somewhat important to capture since this page is about the controllers.

@bresch
-  Is this useful?
- Is there anywhere else in this page that the information should be applied?
- Anything else that should be said?
- If the answer to any of the above is "Yes" then reminder that we need to continuously update docs as we discover holes :-)

>>> Sander Swart 
>>> I have a question about this page: https://dev.px4.io/master/en/flight_stack/controller_diagrams.html It says in the documentation that the multicopter rate PID controller is limited at the output. I cannot seem to find this limit anywhere in the code however. Looking at the logs, it also doesn't seem to be limited. Is the limiter still used?

>> Mathieu Bresciani 
>> The saturation is actually implemented in the mixer: https://github.com/PX4/PX4-Autopilot/blob/850d0bc588af79186286652af4c8293daafd2c4c/src/lib/mixer/MultirotorMixer/MultirotorMixer.cpp#L323-L326

> Sander Swart 
> That explains why the logged actuator control is not saturated, but the actuator output is. Should have thought of that ^^


bresch (approved):

bresch:
Thanks Hamish.

Regarding the following point:
> A Low Pass Filter (LPF) is used on the derivative path to reduce noise

If we want to add more implementation details, should we also say that the derivative and the filters are in the gyro driver?

hamishwillee:
Hi @bresch 

Re https://github.com/PX4/PX4-user_guide/pull/993#issuecomment-759308439

Yes, if people are going to be looking at the code and thinking that stuff is in the controller code itself we should point them to the right place. 

I think you are saying that the gryo driver provides the derivative value directly to the controller, and this has already been filtered? If so, is the change shown below reasonable?

> A Low Pass Filter (LPF) is used on the derivative path to reduce noise *(the gyro driver provides a filtered derivative to the controller)*

bresch:
@hamishwillee yes, that's correct. We can also say that the gyro data is low-passed in the gyro driver as well.
To summarize, the pipeline is as follows: gyro data -> low-pass filter (`IMU_GYRO_CUTOFF`) -> [filtered angular rate used by the P and I controllers] -> derivative -> low-pass filter (`IMU_DGYRO_CUTOFF`) -> [filtered angular acceleration used by the D controller]

hamishwillee:
Thanks @bresch 
I think that is useful, especially with diagram. Final version looks like this - OK?

image

bresch:
Perfect, thanks @hamishwillee !

PR #997, opened 2021-01-18
Update pmw3901.md

RomanLyskov:
Table with information about PMW3901-based boards displaying correctly now.


hamishwillee:
Thanks @RomanLyskov 
The problem was a missing column marker. I have replaced this with https://github.com/PX4/PX4-user_guide/commit/904c5bb7d0e1e1290cd970c7a0940783b09c71ef because I much prefer the simpler markup to having to make sure columns get aligned everywhere.

PR #998, opened 2021-01-18
VM setup: mention 3D acceleration option for simulation

MaEtUgR:
Not sure if that breaks the point numbering but in any case it's important to mention this fact in the documentation.

Reference:
https://github.com/DrTon/jMAVSim/issues/42
https://discuss.px4.io/t/jmavsim-fails-to-run-under-vmware-ubuntu/6242/5


hamishwillee:
Thanks @MaEtUgR !
I decided to make this its own bullet point. Just FYI, this would have broken indentation - needs to indent 3 spaces to align with text in bullet.

PR #1010, opened 2021-01-23
[NEW] - Crazyflie 2.1 getting started

TheLegendaryJedi:
Documentation for flashing PX4 in Bitcraze Crazyflie 2.1. 

@jinger26

PR #1020, opened 2021-01-28
Move craziflie under complete vehicles

hamishwillee:


hamishwillee:
> I don't think crazyflie 2.0 and 2.1 should be listed under autopilots, can we move them to complete vehicles (like the airmind racers)

@jinger From https://github.com/PX4/PX4-user_guide/pull/1010#issuecomment-769313991

The difference is that the airmind vehicles **USE** an autopilot, while the craziflies, from the point of PX4 software they, **ARE** an autopilot. They absolutely belong in the http://docs.px4.io/master/en/flight_controller/autopilot_experimental.html page.

But they also belong in complete vehicles. So what I have done is moved them there under the sidebar and listed them in the complete vehicle page. In the flight controller page I have made the complete vehicles their own section and explained why they belong here. 

Please review.

jinger26:
Okay, make sense @hamishwillee . Thanks

PR #1031, opened 2021-02-03
De-odroidise the video streaming doc

hamishwillee:
Attempts to fix https://github.com/PX4/PX4-Devguide/issues/666 , the most evil issue on the docs :-).

@bkueng I have never set up Video streaming except for in simulation, so I really wasn't sure what was important what was not important in the Odroid doc. So what I have done is tried to create a more generic "how to do streaming" on vehicle and QGC computer type of doc.

The main point is that setting things up is covered in https://github.com/mavlink/qgroundcontrol/blob/master/src/VideoReceiver/README.md

The bit from Odroid doc is the small section after "If you cannot start the stream with the uvch264s". The reason this is useful is that it shows how you set the output IP address (sink) and the (source). 

What is still to do?
- You review and see if I'm on the right track
- If correct then I think a slight rework of the "If you cannot start the stream with the uvch264s" bit to actually talk about what you're doing here - ie this is  the source, this is the sink. The source being the device for the camera. That to me is what is interesting and new over what is in the QGC Video readme.
- I update the image for QGC.

Rendered view of this https://github.com/PX4/PX4-user_guide/blob/dc6c5f4485e9d23b00e4cd37571447f16e471789/en/tutorials/video_streaming.md


hamishwillee:
Further, where should this live?

I kind of think "Not" in the developer guide section. Maybe in Advanced Features? http://docs.px4.io/master/en/advanced_features/
or somewhere under a companion computer topic?

en/tutorials/video_streaming.md
```diff
@@ -1,199 +1,17 @@
...
-```
-
-# Gstreamer Installation
-
-To install gstreamer packages on the computer and on the Odroid C1 and start the stream, follow the instruction given in the [QGroundControl README](https://github.com/mavlink/qgroundcontrol/blob/master/src/VideoReceiver/README.md).
-
-If you cannot start the stream on the Odroid with the uvch264s plugin, you can also try to start it with the v4l2src plugin:
+If you cannot start the stream with the uvch264s plugin, you can also try to start it with the v4l2src plugin:
 
 ```sh
```
bkueng, line 16: I'd remove this as well and refer to the QGC README so we have highest chances of it being correct.

hamishwillee: @bkueng You know that means there is NOTHING left of the original doc right - except the link to the QGC materials?
Done.

hamishwillee:
@bkueng So now this is nothing more than "this is where GStreamer install and run instructions are" and then some slightly more focused ones on QGC. 
Very ruthless. I think still useful - will show up if someone is interested in video streaming. Arguably it could just as well live in QGC docs, but I like it here because it makes it clear this is a QGC/companion thing rather than a "PX4 thing".

Any ideas where it should live in the hierarchy. I'm happy to leave it here for now, or perhaps under https://docs.px4.io/master/en/advanced_features/ - even though it doesn't feel advanced it is "companion computer" ish, which is advanced by definition.

bkueng:
Yes sorry, it's just that video is particularly easy to setup wrong (and even if you have a stream, you might have a lot of latency, or other problems), which is why I prefer to really only have one common setup page.

Don't we have a companion computer section? I'd put it there if we had one.

hamishwillee:
Hi @bkueng 
Thanks :-). I've given it a further minor subedit and move it under "Advanced Features" (for now). 

I absolutely agree that this and other things related to companion computers should be together. Mostly that would be video streaming and this section on [Computer Vision](https://docs.px4.io/master/en/computer_vision/). The problem is that our companion topic [Pixhawk Companion](https://docs.px4.io/master/en/companion_computer/pixhawk_companion.html) is not correctly homed for all these things to live under it. 

I will be moving that, but for now this is an OK home.

PR #1033, opened 2021-02-04
IDE and toolchain info update

### LINKED ISSUE #1022: Inconsistency in IDE recommendations
The IDEs that are recommended in the user guide are inconsistent. If you go to [Development-->Toolchain Installation](https://docs.px4.io/master/en/dev_setup/dev_env.html) there is a whole section about [VSCode](https://docs.px4.io/master/en/dev_setup/vscode.html). [Building the Code](https://docs.px4.io/master/en/dev_setup/building_px4.html) on the other hand clearly states that

> The PX4 system supports Qt Creator, Eclipse and Sublime Text. Qt Creator is the most user-friendly variant and hence the only officially supported IDE. 

I assume that the information given in the Building the Code section is outdated. However, I'm not really sure about this. A clarification and a fix to the docs would be nice.

hamishwillee:
This makes it clear that VS Code is the "supported/recommended IDE" (not Qt Creator). I pushed the Qt Creator info into a sub doc to capture this.

Also merged the ninja and gcc txt docs that were previously imported into the centos setup doc. The import was not working on vuepress, and this is the only place they were imported.

Fixes #1022


julianoes:
I haven't used either in forever, I'd rather have someone review this that actually uses an IDE.

hamishwillee:
@julianoes Sounds fair, do we know such a paragon? @bkueng @dagar ? 
In fact you could review this knowing nothing about IDEs. There is no new information - it is just a restructure that pushes VSCODE in front of QtCreator.

hamishwillee:
Screw it. This is a good change, and I need it to do my next toolchain fixes. Merging.

en/dev_setup/generic_dev_tools.md
```diff
@@ -6,13 +6,24 @@ This section lists useful additional development tools that are not part of the
...
+:::
+
 ![QGroundControl](../../assets/toolchain/qgc_goto.jpg)
 
-## Editor / IDE
 
-The development team often use:
+## IDE / Code Editors
+
+The development team recommend [Visual Studio Code (VSCode)](../dev_setup/vscode.md), a popular open source IDE.
```
MaEtUgR, line 19: recommend{s}

Typo?

en/dev_setup/generic_dev_tools.md
```diff
@@ -6,13 +6,24 @@ This section lists useful additional development tools that are not part of the
...
 
-* [Visual Studio Code](../dev_setup/vscode.md): Popular new open source IDE
+The following IDEs also work (but are not recommended unless you're an expert):
 * [Eclipse for C/C++](https://www.eclipse.org/downloads/eclipse-packages/): Feature rich Java based IDE
 * [Sublime Text](https://www.sublimetext.com): Fast and lean text editor. 
-* [Qt Creator](https://www.qt.io/download-open-source): Popular cross-platform open-source IDE.
+* [Qt Creator IDE](../dev_setup/qtcreator.md): Popular cross-platform open-source IDE.
+
+:::note
+There is an [Eclipse project](https://github.com/PX4/PX4-Autopilot/blob/master/eclipse.project) and a [Sublime project](https://github.com/PX4/PX4-Autopilot/blob/master/Firmware.sublime-project) in the source tree.
```
MaEtUgR, line 28: Would it be worth noting there's also a VS code config? It just gets used automatically when opening that folder.
https://github.com/PX4/PX4-Autopilot/tree/master/.vscode

en/dev_setup/dev_env_linux_centos.md
```diff
@@ -30,9 +30,62 @@ You may want to also install `python-pip` and `screen`.
 :::
 
 ## GCC Toolchain Installation
-<!-- import GCC toolchain common documentation -->
-{% include "_gcc_toolchain_installation.md" %}
+<!-- GCC toolchain documentation used for all Linux platforms to build NuttX -->
 
+Execute the script below to install GCC 7-2017-q4:
```
MaEtUgR, line 35: The version is quite outdated although it probably still works without issues except maybe fmu-v2 flash size because the old compiler has less optimizations. I just never use CentOS anymore and would not be sure if it's worth testing on there.

MaEtUgR (commented):
@hamishwillee Nice restructuring 👍 I agree that represents the updated use. I have some small comments. Sorry for not bringing them earlier.

en/dev_setup/generic_dev_tools.md
hamishwillee: Nope, this is correct because the development team are a plural i.e. this is equivalent to:  The manager recommends, the managers recommend.

en/dev_setup/generic_dev_tools.md
hamishwillee: I decided against this. The difference being that a user has to know about these ones, but they don't need to know about the VSCode one.

en/dev_setup/dev_env_linux_centos.md
hamishwillee: Thanks. I will fix this as part of https://github.com/PX4/PX4-user_guide/pull/1064 (by marking this stuff as untrusted :-) )

en/dev_setup/generic_dev_tools.md
MaEtUgR: I can certainly learn from you I was assuming it's: One team does whatever.

hamishwillee: A fair assumption. When I went to school we learned very little grammar, so I basically know what is correct from experience and then have to find a rule that supports my opinion :-)

PR #1039, opened 2021-02-08
Update teraranger.md

mamariomiamo:
`distance_sensor/teraranger` should be the correct line instead of `drivers/ll40ls`


hamishwillee:
Thank you! Merged

PR #1042, opened 2021-02-09
ThunderFly TFRPM01 Revolution Counter documentation update

kaklik:
New sections about starting of the driver and sensor function validation.

### DIFF
--- a/en/sensor/thunderfly_tachometer.md
+++ b/en/sensor/thunderfly_tachometer.md
@@ -70,7 +70,10 @@ When using CUAV V5+ or CUAV Nano:
 
 
 ### Function validation
-You can verify functionality in multiple ways:
+You can verify functionality in multiple ways.
+
+
+#### Pomocí PX4 (NuttX) konzole
 
 With [`listener`](../middleware/modules_command.html#listener) command you can monitor RPM messages from the driver. 
 ```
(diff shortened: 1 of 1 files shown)


en/sensor/thunderfly_tachometer.md
```diff
@@ -70,15 +70,24 @@ When using CUAV V5+ or CUAV Nano:
 
 
 ### Function validation
-You can verify functionality in multiple ways:
+You can verify functionality in multiple ways.
+
+
+#### Pomocí PX4 (NuttX) konzole
```
hamishwillee, line 76: Note, this is not English.

hamishwillee:
Ah, I see this is still in draft. Ping me when you're ready for review :-)

en/sensor/thunderfly_tachometer.md
kaklik: Yes, sorry for that.  Corrected and now ready to review.

hamishwillee:
Thanks @kaklik . I've merged and subedited this in #1048.

The only bit that confuses me is:
```
Bus number `-b <bus number>` does not match the bus numbering on the autopilot.
After running the command, the driver shows the bus number corresponding to the label on the box.

When using CUAV V5+ or CUAV Nano:	When using CUAV V5+ or CUAV Nano:
```

It sounds like you run the command and it will show you some number, then perhaps you check the box that the RPM counter came in and you can map that to a number you should use, so you'd re-run the command with that.

Essentially "what is the box" you are referring to? It might be worth you showing the result of the result of calling the driver.

kaklik:
> 
> Essentially "what is the box" you are referring to? It might be worth you showing the result of the result of calling the driver.

The "box" means the autopilot hardware case (cover), where are some connector markings, as `I2C1` in which the TFRPM01 sensor could be connected.  But it, unfortunately, requires an unintuitive command in the form `pcf8583 start -X -b 4`.  

As I could see now, the corresponding table should contain complete bus labels in the "bus label" column. e.g. I2C1, I2C2,  I2C3 etc.  It should increase overall intelligibility.

hamishwillee:
Hi @kaklik So I think I understand you. 

The table looks like this:

bus label | -b number
-- | --
1 | -X -b 4
2 | -X -b 2
3 | -X -b 1

And you think we could improve it by changing to below - is that right?

bus label | -b number
-- | --
I2C1 | -X -b 4
I2C2 | -X -b 2
I2C3 | -X -b 1

When you say

> After running the command, the driver shows the bus number corresponding to the label on the box.

I think you are saying that if you did `pcf8583 start  -X -b 1` the driver might output some string that include `I3C3`. Is that correct? So what you are saying is that you could perhaps try running the command with all the bus numbers to work out the right thing to  plug your component into?

hamishwillee:
@kaklik Can you confirm above ^^^

kaklik:
Yes, Your new table is correct. 
I want to improve this in case of future improvement of the documentation. 
Generally, the table should avoid random trying of different bus numbers to get the sensor working. :)

hamishwillee:
Thanks! Fixed inhttps://github.com/PX4/PX4-user_guide/pull/1075

PR #1036, opened 2021-02-05
Adding information about Rotoye Batmon, editing batt_smbus driver docs

nbelanger99:
This doc update corresponds with a pending [PR](https://github.com/PX4/PX4-Autopilot/pull/16723) for Firmware, adding support for Rotoye Batmon devices and abstracting a generic smbus smart-battery driver

PR #1044, opened 2021-02-10
Fixes to Batmon/Smart battery docs

hamishwillee:
Follows on from #1036

- Adds PX4 build instructions. 
- Adds clear message that you need to build PX4 before you can use this. This can be removed once https://github.com/PX4/PX4-Autopilot/pull/16723 goes in (though we may need to add the driver to firmware I guess; lets see).

@nbelanger99 it would be good if you could characterise exactly what data is updated by the battery monitor - just remaining charge? or does it provide other fields?


echoGee:
@hamishwillee Is this merged change visible at https://docs.px4.io/ ?

hamishwillee:
Yes, but not in the sidebar because it was not added to SUMMARY.md. I'll do that now. 
In the meantime, you can see it at https://docs.px4.io/master/en/smart_batteries/rotoye_batmon.html

echoGee:
Great. Thank you!

PR #1054, opened 2021-02-12
Add missing space character

danslinger:
The missing space character was causing the remainder of the document to render incorrectly,
putting headings and instructions inside code blocks and script commands as plain text.

before
after


hamishwillee:
Thanks very much! Merged.

PR #1058, opened 2021-02-15
Mixing: add output scaler rise time

sfuhrer:
The option to limit the actuator rise time via the output scaler in the mixer was introduced recently via https://github.com/PX4/PX4-Autopilot/pull/16257.

### DIFF
--- a/en/concept/mixing.md
+++ b/en/concept/mixing.md
@@ -250,21 +250,23 @@ Summing mixers are used for actuator and servo control.
 
 A summing (simple) mixer combines zero or more control inputs into a single actuator output.
 Inputs are scaled, and the mixing function sums the result before applying an output scaler.
-In the output scaler, also an actuator raise time limit can be specified (inverse of a slew rate).
+In the output scaler, also a minimal actuator traversal time limit can be specified (inverse of a slew rate).
 
 A simple mixer definition begins with:
 
 ```
 M: <control count>
-O: <-ve scale> <+ve scale> <offset> <lower limit> <upper limit> <raise time>
+O: <-ve scale> <+ve scale> <offset> <lower limit> <upper limit> <traversal time>
 ```
 
 If `<control count>` is zero, the sum is effectively zero and the mixer will output a fixed value that is `<offset>` constrained by `<lower limit>` and `<upper limit>`.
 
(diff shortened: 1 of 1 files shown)


en/concept/mixing.md
```diff
@@ -250,18 +250,21 @@ Summing mixers are used for actuator and servo control.
...
-O: <-ve scale> <+ve scale> <offset> <lower limit> <upper limit>
+O: <-ve scale> <+ve scale> <offset> <lower limit> <upper limit> <raise time>
 ```
 
 If `<control count>` is zero, the sum is effectively zero and the mixer will output a fixed value that is `<offset>` constrained by `<lower limit>` and `<upper limit>`.
 
 The second line defines the output scaler with scaler parameters as discussed above.
-Whilst the calculations are performed as floating-point operations, the values stored in the definition file are scaled by a factor of 10000; i.e. an offset of -0.5 is encoded as -5000.
+Whilst the calculations are performed as floating-point operations, the values stored in the definition file are scaled by a factor of 10000; i.e. an offset of -0.5 is encoded as -5000. 
+The `<raise time>` can be used to limit the rate of change of an actuator.
```
hamishwillee, line 266: 1. Is there a better name that `<raise time>`? I presume this also applies when going from max to min, so just as easily could be "drop time". Perhaps "traversal time"?
2. "Can be used to". Is it optional? Say what happens if it is omitted.
3. Is the rate limiting linear? If so, might be worth saying that - otherwise you might assume some other behaviour.

hamishwillee:
I put a few comments in line 

To help address Lorenz' comment  below ([from here](https://github.com/PX4/PX4-Autopilot/pull/16257#issuecomment-753895320)) you probably want to answer the question of why you might want to use this, and when?

> it is important that it doesn't get abused to "smoothen" transitions. That's however mostly a documentation question.

Essentially a clear direction of when and when you should not use it.

sfuhrer:
@hamishwillee thanks for the feedback! I've tried to address your points, let me know what you think of this version now.

hamishwillee:
@sfuhrer Thanks! Much better. I've subedited the English to my taste, but all your information is still there. Merged!

PR #1061, opened 2021-02-17
Minimal tidy up for frame selection

hamishwillee:
Tidy up frame selection doc, which was very aircraft focussed and clearly incomplete.

This may not be perfect, but at least it looks like the document is finished.

@julianoes @MaEtUgR @bkueng Anything else you'd add?

### DIFF
--- a/en/getting_started/frame_selection.md
+++ b/en/getting_started/frame_selection.md
@@ -1,37 +1,24 @@
-# Frame Selection
+# Vehicle Selection
 
-## Air Vehicles
+PX4 supports air, surface, and submersible vehicles.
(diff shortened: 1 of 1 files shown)


bkueng (approved):
Helicopters or autogyros could be mentioned as well

julianoes (approved):

hamishwillee:
@bkueng So when are helicopters and autogyros compelling?

bkueng:
> @bkueng So when are helicopters and autogyros compelling?

Efficiency, and safety for autogyros, as you can still land with motor failures.

PR #1064, opened 2021-02-18
Rework developer setup

hamishwillee:
Massive simplification of developer setup that focusses on the supported platforms. The other platforms like Archlinux pushed to "Advanced > Unsupported Developer Environments" or similar.

Still a fair bit to do, which I will work on next week. 

FYI @LorenzMeier @dagar As discussed on slack.


en/advanced/windows_cygwin_toolchain_setup.md
```diff
@@ -0,0 +1,148 @@
+# Windows Development Environment Maintenances
```
hamishwillee, line 1: @MaEtUgR In this PR I'm simplifying our developer story so that it is "supported dev setup first". It makes sense to split out THIS bit which is how you modify windows cygwin setup. I was wondering though if perhaps this should be in the cygwin README on the repo instead?

Thoughts?

MaEtUgR: Yeah we can put it there. Developer information see README and a link to it.
I have to do a pass at renewing the toolchain version and checking the docs. Then I could just move it to the repo. Last time I got stuck with file sizes but that might possibly have solved itself if the ingredient that increase the memory footprint that much corrected in between.

hamishwillee: @MaEtUgR Awesome - thanks. Can you do the moving of the content and let me know when it is in the readme? I can tidy up the docs themselves.

The bit that got split out is now here: http://docs.px4.io/master/en/advanced/windows_cygwin_toolchain_setup.html (source - https://github.com/PX4/PX4-user_guide/blob/master/en/advanced/windows_cygwin_toolchain_setup.md)

PR #1066, opened 2021-02-19
Fix dead link in uorb.md

jasonotu:
uORB directory was moved but docs not updated to new location


dagar (approved):
That was fast, thanks @jasonotu.

PR #1074, opened 2021-02-24
Fix more heading anchors

MaEtUgR:
Follow up to https://github.com/PX4/PX4-user_guide/pull/897 with my latest findings in broken anchors and how to fix them.

> The other way is to go through all the links that point here and switch them to point to the auto-generated header anchor.

That seem the better way to me, thanks @hamishwillee !

### DIFF
--- a/en/flying/fixed_wing_takeoff.md
+++ b/en/flying/fixed_wing_takeoff.md
@@ -1,12 +1,11 @@
 # Fixed Wing Takeoff
 
-PX4 supports fixed-wing takeoff in [missions](#mission) and using the [Takeoff](#takeoff) flight mode.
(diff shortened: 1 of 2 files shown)


hamishwillee:
Thanks very much @MaEtUgR . These will all work much better!

FWIW if you do further similar fixups feel free to fix the link but **leave** the anchor. Why? Because that will mean that something else that links to these that you don't know about will still kind-of work.

MaEtUgR:
> Because that will mean that something else that links to these that you don't know about will still kind-of work.

I thought about that but then if we use the new system and change the title we can't support any possible old anchor that was ever used. It's probably worth the effort for entire pages but the anchor if it's not an internal link I guess is not such a bad breakage. Anyways I listen to you and will do it that way next time 👍

hamishwillee:
Hi @MaEtUgR ,
YOu've got a good point there too; I'm really not sure of the "right" way to do things with those URLs. So "do as you want" on these.

PR #1077, opened 2021-02-25
Make UAVCAN user-facing

hamishwillee:
This moves UAVCAN docs into the main documentation rather than having them hidden in the developer section. Some docs to remain in the developer section - mostly debugging info and other development stuff not needed by normal developers.

Replaces #671


hamishwillee:
@bkueng Still very much in draft, but a few questions here (and in-line).

Even though this is part of flight controller peripherals, I put this as a "peer" of all the other peripherals.
 image
 
I was thinking I might move all the CAN peripheral docs under UAVCAN parent - ie.. including the UAVCAN distance sensor, esc pages, and power modules (of course they would still be _linked_ by the parent pages for each type).

My reasoning is that you would go all CAN or No CAN. So if you're interested in CAN you don't want to go see all the power module options, you want to find out about the CAN options. Is this reasonable? Or might you mix and match?

en/peripherals/uavcan_escs.md
```diff
@@ -1,185 +1 @@
...
-Finally the `CAN Baud Rate` must match the value set in [UAVCAN_BITRATE](../advanced_config/parameter_reference.md#UAVCAN_BITRATE).
-
-## Further Information
-
-- [PX4/Sapog](https://github.com/PX4/sapog#px4-sapog) (Github)
-- [Sapog v2 Reference Manual](https://files.zubax.com/products/io.px4.sapog/Sapog_v2_Reference_Manual.pdf)
-- [UAVCAN Device Interconnection](https://kb.zubax.com/display/MAINKB/UAVCAN+device+interconnection) (Zubax KB)
-- [Using Sapog based ESC with PX4](https://kb.zubax.com/display/MAINKB/Using+Sapog-based+ESC+with+PX4) (Zubax KB)
-
+<Redirect to="../uavcan/escs" />
```
hamishwillee, line 1: FYI only. I moved this doc under "/uavcan/. Wherever docs appear in the tree, I think it makes sense to group them under uavcan in the folder structure.

en/uavcan/notes.md
```diff
@@ -1,21 +0,0 @@
-# Various Notes
```
hamishwillee, line 1: FYI, this set of notes is a bit odd. I moved them to better places in the docs.

en/uavcan/node_enumeration.md
```diff
@@ -17,3 +17,4 @@ ESCs running the Sapog firmware will need to be rebooted after enumeration for t
 This step has to be performed only once and does not need to be repeated after firmware upgrades.
```
hamishwillee, line 17: @bkueng I think this doc essentially is a copy of the docs we now have in the esc setup.

What I think needs to happen is that this goes into the parent config doc, which explains there are the three options set by UAVCAN_ENABLE. My assumption
1: Sensors Manual Config - if you set this then you have to configure everything manually using "native" config tools. You probably need this for anything other than Sapog right?
2: Sensors Automatic Config - This basically enumerates SAPOG (only) sensors - right?
3: Sensors and Actuators (ESCs) Automatic Config - This is the ESC config. You set this then you can enumerate the motors.

I have a question though, what does the control highlighted in red below do? (not clickable on my computer)
image

en/uavcan/README.md
```diff
@@ -1,49 +1,55 @@
-# UAVCAN Introduction
+# UAVCAN
 
-![UAVCAN Logo](../../assets/uavcan/uavcan_logo_transparent.png)
+<img style="float:right; width: 200px ; padding: 10px;" src="../../assets/uavcan/uavcan_logo_transparent.png" alt="UAVCAN Logo" /> [UAVCAN](http://uavcan.org) is an onboard network which allows the autopilot to connect to avionics/peripherals.
+In contrast to hobby-grade devices it uses rugged, differential signalling, supports firmware upgrades over the bus and supports status feedback from peripherals.
```
bkueng, line 4: This is a bit strong and inaccurate
- there's also hobby-grade uavcan devices
- there's industrial-grade devices that don't use uavcan (including fw update support and feedback)

en/uavcan/node_enumeration.md
bkueng: 1. This is if all your devices have a fixed node ID - you need to ensure there's no duplicate
2. Any device might support dynamic enumeration, it needs to be checked in the manual
3. yes

> I have a question though, what does the control highlighted in red below do?

No idea

bkueng (commented):
> Is this reasonable? Or might you mix and match?

The tendency is to go all CAN (depending on the setup), so having all in one place makes sense to me as well, but currently there's still some mix.

en/uavcan/README.md
hamishwillee: Thanks, I'll just remove "in contrast to ...".

en/uavcan/pomegranate_systems_pm.md
```diff
@@ -0,0 +1,55 @@
+# Pomegranate Systems Power Module
```
hamishwillee, line 1: FYI Not changed. Just moved under /uavcan. There is a redirect.

en/uavcan/cuav_can_pmu.md
```diff
@@ -0,0 +1,70 @@
+# CAUV CAN PMU
```
hamishwillee, line 1: FYI Not changed. Just moved under /uavcan. There is a redirect.

en/uavcan/node_enumeration.md
hamishwillee: @bkueng Sorry to come back to this. Some more checks on assumptions, in particular when you use the UI and the different options for `UAVCAN_ENABLE`.

1. `UAVCAN_ENABLE=1` means manual enumeration of ids. Am I correct in thinking that you can always set this, but then
   - you will need to manually enumerate all sensors and ESCs? 
   - You would use whatever tools are appropriate/supplied for your ESC and other nodes - e.g. VESC or Sapog or whatever.
2. `UAVCAN_ENABLE=2` means dynamic enumeration. I.e. PX4 will try an enumerate all nodes EXCEPT for your ESCs.
   - Is that correct? I.e. this would enumerate your GNSS, power module, whatever, irrespective of type?
   - What if some of your devices do not support dynamic enumeration? Would you set to this and assignment will happen automatically for those that support it, and then do the others manually?
   - How does the enumeration happen - is it automatic on every boot, or is it a one off and then saved? Do you need to do anything else in the UI or is just setting the value enough?
3. `UAVCAN_ENABLE=3` means dynamic enumeration of sensors and automatic enumeration of ESCS.
   - Does this work on all ESC or just Sapog?
   - What if you have a sensor that doesn't support dynamic allocation - do you use this, do the assignment then manually assign the static allocation using the type-specific tools?
   - If it doesn't work on all ESC, then what do you do for VESC say? Select 2 to do the sensors and then use the VESC tools to assign the ESC.

Hope the question makes sense!

bkueng: These are all good questions, and definitely worth clarifying in detail. But you may have to ask someone else for more details.

> you will need to manually enumerate all sensors and ESCs?

Yes

> You would use whatever tools are appropriate/supplied for your ESC and other nodes - e.g. VESC or Sapog or whatever.

Yes, essentially statically assign them

> PX4 will try an enumerate all nodes EXCEPT for your ESCs.

Presumably, but I'm not 100% sure on the exact details

> What if some of your devices do not support dynamic enumeration?

Good question. :shrug: Probably manual assignment but using a high enough ID to ensure no overlap.

> How does the enumeration happen - is it automatic on every boot, or is it a one off and then saved?

There's a database stored to SD card to ensure it stays consistent.

> Does this work on all ESC or just Sapog?

I don't know, I think it needs special support.

> What if you have a sensor that doesn't support dynamic allocation - do you use this, do the assignment then manually assign the static allocation using the type-specific tools?

I guess so

> If it doesn't work on all ESC, then what do you do for VESC say? Select 2 to do the sensors and then use the VESC tools to assign the ESC.

I don't think you would mix ESCs in the first place, but yes, that could work.

en/uavcan/node_enumeration.md
```diff
@@ -1,19 +0,0 @@
-# UAVCAN Enumeration and Configuration
```
hamishwillee, line 1: FYI Deleted. Is now part of  the ESC docs.

en/uavcan/README.md
```diff
@@ -1,70 +1,92 @@
...
+For more information about proper bus connections see [UAVCAN Device Interconnection](https://kb.zubax.com/display/MAINKB/UAVCAN+device+interconnection) (Zubax KB).
+
+:::note
+- While the connections are the same, the _connectors_ may differ across devices.
+- An second/redundant" CAN interface may be used, as shown above (CAN2).
+  This is optional, but can increase the robustness of the connection.
+:::
+
+
+## PX4 Configuration
```
hamishwillee, line 52: @bkueng This is main change since last time. It makes clear when you select each of the UAVCAN_ENABLE options. Because you will have to manually configure any nodes that don't support auto config, this implies that you only use manual config if NONE of the modes support auto.

en/uavcan/README.md
```diff
@@ -1,70 +1,93 @@
...
+   - `3`: Sensors and Actuators (ESCs) Automatic Config
+
+   Use `1` if _none_ of the connected UAVCAN devices support automatic configuration (check the manual!), `2` or `3` if _some_ of them support automatic configuration, and `3` if you're using UAVCAN ESCs (this assigns motor controls to the UAVCAN bus rather than PWM).
+
+:::note
+You will need to manually allocate static ids for any nodes that don't support automatic configuration.
+When using dynamic configuration, any manually allocated ids should be given a value greater than the number of UAVCAN devices (to avoid clashes).
+:::
+
+Most UAVCAN sensors require no further setup (they are plug'n'play, unless specifically noted in their documentation).
```
hamishwillee, line 73: @bkueng Is this reasonable? My understanding is that all you have to do is enumerate most sensors and they just work. However there are quite a lot of UAVCAN parameters. So it may be that all you have to do is AUTO enumerate most sensors - if you have things that don't support auto setup perhaps there is more configuration to be done.

Can you clarify this point? Or am I OK just saying this?

hamishwillee:
PS You will note I have set to "Ready for review". My feeling is that this is a significant net improvement even without some details nutted out. So possibly we should merge and iterate. Up to you.

en/uavcan/README.md
bkueng: I think that's accurate and the most important thing (node ID's).

bkueng (approved):
LGTM

hamishwillee:
Merging. Thanks for review!

PR #1094, opened 2021-03-09
mc pid tuning: add page for QGC tuning UI

bkueng:
- update from https://github.com/mavlink/qgroundcontrol/pull/9379
- I kept the additional page under advanced largely intact, for in-depth information
- RC_FLT_CUTOFF got removed

@hamishwillee is there a plugin to enlage images when clicking on them?


hamishwillee:
> @hamishwillee is there a plugin to enlage images when clicking on them?

There is *already* a plugin that will expand images to their _actual size_ when you click them. However if the image is small to start with it won't be zoomed/enlarged. 

This is a good thing. Generally you should keep images small, but if you think it is useful to zoom them you have the option for them to be shown at a bigger size.

en/config_mc/pid_tuning_guide_multicopter_basic.md
```diff
@@ -0,0 +1,93 @@
...
+:::
+
+:::warning
+Poorly tuned vehicles are likely to be unstable, and easy to crash.
+Make sure to have assigned a [Kill switch](../config/safety.md#emergency-switches).
+:::
+
+## Introduction
+
+PX4 uses **P**roportional, **I**ntegral, **D**erivative (PID) controllers (these are the most widespread control technique).
```
hamishwillee, line 20: @bkueng This got a significant update. 

This talked about need for tuning the controllers in order, which of felt out of place. I fixed this by moving up the information about the tuning UI and the general remarks. So this is more of a true overview.

I did decide to leave the guidance on how you modify PID down in the procedure.

The general remarks are still _embedded_ in the procedure.

en/config_mc/pid_tuning_guide_multicopter_basic.md
```diff
@@ -0,0 +1,93 @@
...
+  - Remove propellers
+  - Arm the vehicle and lower the throttle to the minimum
+  - Tilt the vehicle to all directions, about 60 degrees
+  - Check that no motors turn off
+- Use a high-rate telemetry link such as WiFi if at all possible (a typical low-range telemetry radio is not fast enough for real-time feedback and plots).
+  This is particularly important for the rate controller.
+- Disable [MC_AIRMODE](../advanced_config/parameter_reference.md#MC_AIRMODE) before tuning a vehicle.
+
+## Tuning Procedure
+
```
hamishwillee, line 60: This is now a true "step by step" procedure. I like doing this even though it is less compact because it makes you think about what order people should do things and what is documented vs not documented. 

Some questions from this
-ie what is thrust curve? What value should be set?
- When are the values saved? 
- What are clipboard values used for?
- When should I clear or start ?
- What does automatic flight mode switching do. When should I use it?

en/config_mc/pid_tuning_guide_multicopter.md
```diff
@@ -1,43 +1,6 @@
...
-  This can be tested in [Acro mode](../flight_modes/acro_mc.md) or in [Manual/Stabilized mode](../flight_modes/manual_stabilized_mc.md):
-  - Remove propellers
-  - Arm the vehicle and lower the throttle to the minimum
-  - Tilt the vehicle to all directions, about 60 degrees
-  - Check that no motors turn off
-- Optionally enable the high-rate logging profile with the [SDLOG_PROFILE](../advanced_config/parameter_reference.md#SDLOG_PROFILE) parameter so you can use the log to evaluate the rate and attitude tracking performance (the option can be disabled afterwards).
-
-:::warning
-Always disable [MC_AIRMODE](../advanced_config/parameter_reference.md#MC_AIRMODE) when tuning a vehicle.
+# Multicopter PID Tuning Guide (in-depth)
```
hamishwillee, line 1: I'm trying to understand when you would use this "in depth" approach. I.e. this could be something you never use - you always use the basic tuning guide. It is for people who want to understand what the basic tuning does "under the hood"? Or it could be that this is a better form of tuning for advanced users?

I guess I'm asking when should someone use this vs the basic doc? We need to make that very clear.

en/config_mc/pid_tuning_guide_multicopter.md
```diff
@@ -53,6 +16,11 @@ Here are some general points to follow when tuning:
 - Land before changing a parameter.
```
hamishwillee, line 16: For the basic tuning we don't mention this. Should all the tuning be done on the ground there too?

en/config_mc/pid_tuning_guide_multicopter_basic.md
```diff
@@ -0,0 +1,93 @@
+# Multicopter PID Tuning Guide
```
hamishwillee, line 1: Lots of change here. Will need your re-review.

en/config_mc/pid_tuning_guide_multicopter_basic.md
```diff
@@ -0,0 +1,93 @@
...
+     - used to reduce steady-state error
+     - if too low, the response might never reach the setpoint (e.g. in wind)
+     - if too high, slow oscillations can occur
+1. Repeat the tuning process above for the pitch and yaw:
+   - Use *Select Tuning* radio button to select the axis to tune
+   - Move the appropriate sticks (i.e. pitch stick for pitch, yaw stick for yaw).
+   - For pitch tuning, start with the same values as for roll.
+1. Repeat the tuning process for the attitude controller on all the axes.
+1. Repeat the tuning process for the velocity and positions controllers (on all the axes).
+   use Position mode when tuning these controllers.
```
hamishwillee, line 93: Do I need to remind them to turn the airmode thingy on again?

en/config_mc/pid_tuning_guide_multicopter.md
bkueng: Yes it's partially for people wanting to learn more, and partially for some more advanced tuning, like the thrust curve

en/config_mc/pid_tuning_guide_multicopter_basic.md
bkueng: Yes we can

en/config_mc/pid_tuning_guide_multicopter_basic.md
```diff
@@ -0,0 +1,93 @@
...
+   - I (integral) gain:
+     - used to reduce steady-state error
+     - if too low, the response might never reach the setpoint (e.g. in wind)
+     - if too high, slow oscillations can occur
+1. Repeat the tuning process above for the pitch and yaw:
+   - Use *Select Tuning* radio button to select the axis to tune
+   - Move the appropriate sticks (i.e. pitch stick for pitch, yaw stick for yaw).
+   - For pitch tuning, start with the same values as for roll.
+1. Repeat the tuning process for the attitude controller on all the axes.
+1. Repeat the tuning process for the velocity and positions controllers (on all the axes).
```
bkueng, line 92: Now we can move the tip from above (`Make sure to switch to the simple input mode, for direct control and possibility to generate step inputs`) down here

en/config_mc/pid_tuning_guide_multicopter_basic.md
bkueng: > -ie what is thrust curve? What value should be set?

I'd point to the advanced page, as it requires using logs if you want to tune it really accurate. But w/o any detailed tuning, 0.3 should generally be a good value (the difficulty is that we can't just change the default, as it would invalidate existing setups)

> When are the values saved?

Immediately

> What are clipboard values used for?

To restore the previous gains

> When should I clear or start ?

Well that's really up to you. When you start zooming/panning, it stops, so you want to start before doing the next iteration (it auto starts on arming).

> What does automatic flight mode switching do. When should I use it?

This is if you have very little flying experience. You can keep it in position control, and when you press start, it will switch to stabilized. This is something we'll have to see how useful it is in practise.

en/config_mc/pid_tuning_guide_multicopter_basic.md
bkueng: Makes sense, more integrated now

bkueng (commented):
Thanks Hamish

> There is already a plugin that will expand images to their actual size when you click them. However if the image is small to start with it won't be zoomed/enlarged.

Didn't work for me then, or the image is too small. I can see it works online though, so it's probably fine.

en/config_mc/pid_tuning_guide_multicopter.md
```diff
@@ -1,44 +1,13 @@
-# Multicopter PID Tuning Guide
+# Multicopter PID Tuning Guide (Advanced/Detailed)
 
-This tutorial explains how to tune the PID loops on PX4 for all [multicopter setups](../airframes/airframe_reference.md#copter) (Quads, Hexa, Octo etc).
+This topic provides detailed information about PX4 controllers, and how they are tuned.
```
hamishwillee, line 3: FYI, I updated this intro to explain why/when you need to read this.

en/config_mc/pid_tuning_guide_multicopter.md
hamishwillee: Hmm. I think perhaps I'll leave this for now.

en/config_mc/pid_tuning_guide_multicopter_basic.md
```diff
@@ -0,0 +1,99 @@
...
+## Tuning Procedure
+
+The tuning procedure is:
+
+1. Open _QGroundControl_ **Vehicle Setup > PID Tuning**
+   ![QGC Rate Controller Tuning UI](../../assets/mc_pid_tuning/qgc_mc_pid_tuning_rate_controller.png)
+1. Select the **Rate Controller** tab.
+1. Confirm that the airmode selector is set to **Disabled**
+1. Set the *Thrust curve* value to: 0.3
+   :::note
```
hamishwillee, line 68: >> -ie what is thrust curve? What value should be set?

> I'd point to the advanced page, as it requires using logs if you want to tune it really accurate. But w/o any detailed tuning, 0.3 should generally be a good value (the difficulty is that we can't just change the default, as it would invalidate existing setups)

OK. Done here.

FMI, I looked at the docs and they seem to indicate that the curve reflects non-linearity in thrust to PWM signals. In other words, if you had a linear relationship you would expect the tuning at hover thrust to work at higher thrust - correct?
Does that imply that a UAVCAN esc would not have a thrust curve? I.e. for UAVCAN we would set this to 0?

en/config_mc/pid_tuning_guide_multicopter_basic.md
hamishwillee: Done. But how do you switch to simple input mode?

hamishwillee:
Thanks @bkueng 

I think this is close now - just a couple of inline questions. The only one impacting whether this is good enough to merge is https://github.com/PX4/PX4-user_guide/pull/1094#discussion_r591987940

en/config_mc/pid_tuning_guide_multicopter_basic.md
bkueng: Yes that is correct. For UAVCAN it depends: for RPM-based control, you don't need to tune it, but for power-based control (used in most cases), it's the same as for PWM.

en/config_mc/pid_tuning_guide_multicopter_basic.md
bkueng: It's in the UI:
image

LorenzMeier:
We can always follow up with an additional PR. Thanks for getting this done!

en/config_mc/pid_tuning_guide_multicopter_basic.md
hamishwillee: Fixed in https://github.com/PX4/PX4-user_guide/pull/1115/files#r595679810

en/config_mc/pid_tuning_guide_multicopter_basic.md
hamishwillee: Thanks. Was not obvious to me. Added in fixes.

PR #1095, opened 2021-03-09
jMAVSim wrong jdk version

romain-chiap:
should we put that here, or in "building jMAVSim" section?

I'm referring to this forum post
https://discuss.px4.io/t/jmavsim-fails-wrong-java-version/21368

### DIFF
--- a/en/dev_setup/dev_env_mac.md
+++ b/en/dev_setup/dev_env_mac.md
@@ -89,6 +89,17 @@ brew install px4-sim-jmavsim
 jMAVSim for PX4 v1.11 and earlier required Java 8.
 :::
 
+### troubleshoot (for macOS Catalina users)
+When compiling jmavsim, if the following error appears
+```
+Exception in thread "main" java.lang.UnsupportedClassVersionError: me/drton/jmavsim/Simulator has been compiled by a more recent version of the Java Runtime (class file version 59.0), this version of the Java Runtime only recognizes class file versions up to 58.0
+```
+it means a more recent version of Java is needed. Class file versions 58 corresponds to jdk14, version 59 to jdk15, etc.
+To fix it, tap
+```sh
+brew install --cask adoptopenjdk15
+```
+
 ## Next Steps
 
 Once you have finished setting up the command-line toolchain:
--- a/en/dev_setup/dev_env_mac.md
+++ b/en/dev_setup/dev_env_mac.md
@@ -73,12 +73,12 @@ brew install px4-sim-gazebo
 
 ## jMAVSim Simulation
 
-To use SITL simulation with jMAVSim you need to install a recent version of Java (e.g. Java 14).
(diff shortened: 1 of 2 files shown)


mrpollo:
Reviewed the openjdk15 fix and works on macOS, I'm going to make some edits to your wording, and I'll merge this.

Thank you @romain-chiap 👏  👏  👏  👏

hamishwillee:
@mrpollo @romain-chiap Perhaps the right place would be alongside all the other troubleshooting docs? https://docs.px4.io/master/en/simulation/jmavsim.html#troubleshooting

@julianoes This is the second report. Perhaps for macOS we could just assume JDK 15 here? https://docs.px4.io/master/en/dev_setup/dev_env_mac.html#jmavsim-simulation

romain-chiap:
I guess this could be put in the troubleshooting section.
Maybe just putting a link to the troubleshooting section, from the toolchain installation page. So the users are aware there is a troubleshooting section.

hamishwillee:
> Maybe just putting a link to the troubleshooting section

Sounds sensible. And/or from wherever we build jmavsim - e.g. the building section

mrpollo:
@romain-chiap @hamishwillee I just made the quick fix please review and merge, we should be ready to go

en/dev_setup/dev_env_mac.md
```diff
@@ -85,8 +85,11 @@ brew install --cask adoptopenjdk14
 brew install px4-sim-jmavsim
 ```
 
-:::note
-jMAVSim for PX4 v1.11 and earlier required Java 8.
+:::warning
+jMAVSim for PX4 v1.11 and beyond we require at least JDK 15.
```
hamishwillee, line 89: @mrpollo The problem, as discussed in related issue you closed, is that this may not be true. 
JDK 14 should work and there is some other problem here. My concern is that this will work for now, but might break for the same reason magically at some other point - just as JDK 14 has ... it used to work.

If no one is going to debug this properly (which it looks like) then yes, we can update. But you will need to update the previous lines from 76 as well to specify JDK15

I have subedit this section

hamishwillee: Actually I have updated the section JDK 14 above as well. I'll add my "concern" as a public comment and merge this.

hamishwillee:
Thanks @mrpollo 

The reason we didn't merge this originally was that we didn't want to "fix the wrong bug". This worked previously then stopped working for _some_ users. What's to stop that happening again? Anyway, I guess this is better than just having people run into the problem. 

I have subedited and will merge.

hamishwillee:
PS, thanks @romain-chiap !

mrpollo:
I can confirm this is an actual problem in Ubuntu and macOS, I recently had to re-install my toolkit, and under macOS I had to install JDK15, and on ubuntu JDK14.

You are totally right, there's an underlying problem we need to figure out and fix

PR #1102, opened 2021-03-12
Fix a command in the installation guide

mohamedsayed18:
Signed-off-by: mohamedsayed18 <mohamed95.a.s@gmail.com>

it is not mentioned that the user should be in the PX4-Autopilot directory, so it can me problem for users not familiar with ubutnu and commandline


LorenzMeier (approved):

LorenzMeier:
Thanks!

PR #1103, opened 2021-03-12
ulog: add parameter default message + compat flag

bkueng:
From https://github.com/PX4/pyulog/pull/56 and https://github.com/PX4/PX4-Autopilot/pull/17106

2. commit is just whitespace removal + formatting fix


LorenzMeier (approved):
Thanks!

PR #1105, opened 2021-03-14
Adding better instructions for payloads

LorenzMeier:
This was long overdue. Overall the whole documentation needs TLC because it is not showing the cool stuff we already have.

### DIFF
--- a/en/payloads/README.md
+++ b/en/payloads/README.md
@@ -1,0 +1,109 @@
+# Payloads and Cameras
+
+PX4 supports a wide range of payloads and cameras.
+
+## Mapping Drones
+
+* [Camera triggering](../peripherals/camera.md) via GPIO out
+* [Camera triggering](../peripherals/camera.md) via PWM out
+* [Camera triggering](../peripherals/camera.md) via MAVLink out
+* [Camera timing](../peripherals/camera.md#camera_capture) feedback via hotshoe input
+
+## Cargo drones and alike: Servos / Outputs
+
+* Servo or GPIO triggering (via RC or via commands)
+
+### Example Mission (in QGC)
+
+Use MAV_CMD_DO_SET_ACTUATOR to trigger one of the payload actuators.
+
+### Example script (MAVSDK)
+
+This script sends a command to set the actuator and trigger the payload release on a servo:
+
+```
+#include <mavsdk/mavsdk.h>
+#include <mavsdk/plugins/mavlink_passthrough/mavlink_passthrough.h>
+#include <mavsdk/plugins/info/info.h>
+#include <chrono>
+#include <cstdint>
+#include <iostream>
+#include <future>
+#include <memory>
+
+using namespace mavsdk;
+
+void send_actuator(MavlinkPassthrough& mavlink_passthrough,
+        float value1, float value2, float value3);
+
+int main(int argc, char **argv)
+{
+    Mavsdk mavsdk;
+    std::string connection_url;
+    ConnectionResult connection_result;
+    float value1, value2, value3;
+
+    if (argc == 5) {
+        connection_url = argv[1];
+        connection_result = mavsdk.add_any_connection(connection_url);
+        value1 = std::stof(argv[2]);
+        value2 = std::stof(argv[3]);
+        value3 = std::stof(argv[4]);
+    } 
+
+    if (connection_result != ConnectionResult::Success) {
+        std::cout << "Connection failed: " << connection_result << std::endl;
+        return 1;
+    }
+
+	bool discovered_system = false;
+    mavsdk.subscribe_on_new_system([&mavsdk, &discovered_system]() {
+		const auto system = mavsdk.systems().at(0);
+
+		if (system->is_connected()) {
+			std::cout << "Discovered system" << std::endl;
+			discovered_system = true;
+		}
+	});
+
+	std::this_thread::sleep_for(std::chrono::seconds(2));
+
+	if (!discovered_system) {
+        std::cout << "No device found, exiting." << std::endl;
+        return 1;
+    }
+
+	std::shared_ptr<System> system = mavsdk.systems().at(0);
+	for (auto& tsystem : mavsdk.systems()) {
+		auto info = Info{tsystem};
+		std::cout << info.get_identification().second.hardware_uid << std::endl;
+		if (info.get_identification().second.hardware_uid == "3762846593019032885") {
+			system = tsystem;
+		}
+	}
+
+	auto mavlink_passthrough = MavlinkPassthrough{system};
+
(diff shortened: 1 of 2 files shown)


davids5 (approved):

hamishwillee:
110% on what @LorenzMeier said. This is great to see. I will have a proper review when I work on PX4 stuff tomorrow, but on the face of it looks great.

en/payloads/README.md
```diff
@@ -0,0 +1,109 @@
...
+### Example Mission (in QGC)
+
+Use MAV_CMD_DO_SET_ACTUATOR to trigger one of the payload actuators.
+
+### Example script (MAVSDK)
+
+This script sends a command to set the actuator and trigger the payload release on a servo:
+
+```
+#include <mavsdk/mavsdk.h>
```
hamishwillee, line 25: @julianoes 
Are there any examples scripts, documentation on using [MAVLinkPassthrough](https://mavsdk.mavlink.io/develop/en/api_reference/classmavsdk_1_1_mavlink_passthrough.html)?

en/payloads/README.md
```diff
@@ -0,0 +1,109 @@
...
+	send_actuator(mavlink_passthrough, value1, value2, value3);
+
+    return 0;
+}
+
+void send_actuator(MavlinkPassthrough& mavlink_passthrough,
+		float value1, float value2, float value3)
+{
+	std::cout << "Sending message" << std::endl;
+	mavlink_message_t message;
```
hamishwillee, line 96: @bkueng This takes three values passed on as arguments, and sends these in [MAV_CMD_DO_SET_ACTUATOR](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_ACTUATOR).

Can you confirm a few things for me:
1. This message is now supported in PX4 master? Will this go into PX4 v1.12?
2. My memory of our discussion is that PX4 would support three assignable actuators (ie the first three assigned here) and that these would be mapped to an actual actuator/servo internally via a parameter. That actuator might be on UAVCAN or PWM or whatever. Can you confirm
   - Three params are supported for this
   - What they are named
   - What busses/outputs they can represent
   - What mitigation exists to prevent outputs being assigned that have an existing purpose - e.g. if I assigned MAIN1 as actuator 1 and that is being used as a flap.
   
@julianoes It would make sense once this is stable to remove WIP on the message and make a MAVSDK action for this. Is that on a roadmap?

en/payloads/README.md
```diff
@@ -0,0 +1,109 @@
...
+	});
+
+	std::this_thread::sleep_for(std::chrono::seconds(2));
+
+	if (!discovered_system) {
+        std::cout << "No device found, exiting." << std::endl;
+        return 1;
+    }
+
+	std::shared_ptr<System> system = mavsdk.systems().at(0);
```
hamishwillee, line 76: @julianoes Can you confirm my understanding of this block of code?
The first line just grabs the first system we connect to. The following lines then check all the connected systems for a specific hardware UID, and if one is found that is used instead. 

So my assumption is that this code allows us to make sure by UID we are connecting to the specific vehicle we want to, if it is connected. Is that the intent? 

If we care that much, shouldn't we pass that in as a value to the code and exit if we don't find it?

Further, do you have similar code in the MAVSDK docs? I see in https://mavsdk.mavlink.io/develop/en/guide/system_information.html#query-system-information we refer to this `.second` attribute but it isn't clear what it is/where it comes from (perhaps it is C++ that is new to me).

en/payloads/README.md
```diff
@@ -0,0 +1,109 @@
...
+{
+    Mavsdk mavsdk;
+    std::string connection_url;
+    ConnectionResult connection_result;
+    float value1, value2, value3;
+
+    if (argc == 5) {
+        connection_url = argv[1];
+        connection_result = mavsdk.add_any_connection(connection_url);
+        value1 = std::stof(argv[2]);
```
hamishwillee, line 49: @julianoes These should set value to `NaN` if no value is passed in the argument - otherwise you're explicitly specifying a value. If not, what's the best way to do that?

en/payloads/README.md
```diff
@@ -0,0 +1,109 @@
...
+## Mapping Drones
+
+* [Camera triggering](../peripherals/camera.md) via GPIO out
+* [Camera triggering](../peripherals/camera.md) via PWM out
+* [Camera triggering](../peripherals/camera.md) via MAVLink out
+* [Camera timing](../peripherals/camera.md#camera_capture) feedback via hotshoe input
+
+## Cargo drones and alike: Servos / Outputs
+
+* Servo or GPIO triggering (via RC or via commands)
```
hamishwillee, line 14: @bkueng For "via RC" is this the [RC_MAP_AUXx](http://docs.px4.io/master/en/advanced_config/parameter_reference.html#RC_MAP_AUX1) values?
My understanding is that you also have to have these AUXx values assigned in the mixer too? Right? So in practise this is a bit of a pain?

en/payloads/README.md
bkueng: > This message is now supported in PX4 master? Will this go into PX4 v1.12?

yes

It's the 3 AUX outputs in https://docs.px4.io/master/en/concept/mixing.html#control-group-3-manual-passthrough so you can configure the mixer to assign it to any output.

en/payloads/README.md
bkueng: Yes RC_MAP_AUX.
Yes, but most airframes have passthrough by default as well.

en/payloads/README.md
julianoes: I will create a proper example once the PR is in, and then I can help document that.

en/payloads/README.md
hamishwillee: @bkueng OK, I thought those controls were for RC passthrough?
- Are you saying that they are also used for the actuator message?
- So in other words, these should already be set up for most vehicles?
- What about contention - if someone has RC and message sent via GCS will this just be "messy"?

If so, great, but does this imply anything about https://docs.px4.io/master/en/concept/custom_mixer_payload.html ? I.e. that explains how to set up a payload mixer, and it uses different control group. So when would you ever need to do this?

en/payloads/README.md
hamishwillee: Great. I've added a section on RC Trigger and merged the other PR. I never really got this before.

en/payloads/README.md
bkueng: > OK, I thought those controls were for RC passthrough?

It's now both. It takes the latest one that got updated/changed.

> So when would you ever need to do this?

You should almost never have to do that anymore. Except when you add a custom module that wants to control an output.

PR #1107, opened 2021-03-14
Add 1.12 release notes

MaEtUgR:
After the discussion started on slack I had the following idea:

> How about one high level point per topic e.g. UAVCAN, Board support, Multicopter, ... and then list the new things in one line but make them a link to the right place in the long description? That's at least what I would like as a user.

> link to where?
PX4 guide with 1.12 release page. which is on the detail level of what I wrote above with links to e.g. prs.

> To me it's more attractive to add description to documentation than to some release blog post. The 1.11 release notes are quite nice but the as short as possible description follows the link to the pr. I think quickly typing down a description on the more in-depth release page in the docs makes you think through what changes and if the description gets too involved it's also a lower barrier to directly add docs with the same pr.

This is the first stab at it and I already posted links to the points I added to https://github.com/PX4/PX4-Autopilot/releases/tag/v1.12.0-beta1

### DIFF
--- a/en/SUMMARY.md
+++ b/en/SUMMARY.md
@@ -388,6 +388,8 @@
     * [ROS Integration Testing](test_and_ci/integration_testing.md)
     * [Docker Containers](test_and_ci/docker.md)
     * [Maintenance](test_and_ci/maintenance.md)
+* [Releases](releases/README.md)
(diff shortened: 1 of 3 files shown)


LorenzMeier (approved):
Great!

PR #1121, opened 2021-03-17
Update dshot.md

jeremyyano:
Added tip that not all ESCs support the [esc_info] command, per Beat Kung's request.

### DIFF
--- a/en/peripherals/dshot.md
+++ b/en/peripherals/dshot.md
@@ -112,3 +112,7 @@ dshot esc_info -m 1
 :::tip
 You may have to configure [MOT_POLE_COUNT](../advanced_config/parameter_reference.md#MOT_POLE_COUNT) to get the correct RPM values.
 :::
+
+:::tip
+Not all DSHOT-capable ESCs support [esc_info] (e.g. APD 80F3x), check with manufacturer for details.
+:::
--- a/en/peripherals/dshot.md
+++ b/en/peripherals/dshot.md
@@ -114,5 +114,10 @@ You may have to configure [MOT_POLE_COUNT](../advanced_config/parameter_referenc
 :::
 
 :::tip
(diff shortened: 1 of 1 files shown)


en/peripherals/dshot.md
```diff
@@ -112,3 +112,7 @@ dshot esc_info -m 1
 :::tip
 You may have to configure [MOT_POLE_COUNT](../advanced_config/parameter_reference.md#MOT_POLE_COUNT) to get the correct RPM values.
 :::
+
+:::tip
+Not all DSHOT-capable ESCs support [esc_info] (e.g. APD 80F3x), check with manufacturer for details.
```
hamishwillee, line 117: @jeremyyano Thanks very much! When you say "support esc_info" do you mean "ESC telemetry" or this command? The reason I ask is that it might make sense for this to be merged into the comment above where I say that the dshots that support telemetry are the ones where ... "These DShot ESCs will have an additional telemetry wire."

hamishwillee: PS 
Or are you saying that the problem that made this question come up was that you tried the `dshot esc_info` command and it didn't work? i.e. you want to explain that this command might not show anything for people who try it and it doesn't work?

What happens in this case when you run the command?

jeremyyano: I'm running APD 80F3x's, which have a telemetry wire, and successfully communicate a telemetry stream when the `listener esc_status` command is run, however, when `dshot esc_info -m #` is run, produce `ERROR [dshot] No data received. If telemetry is setup correctly, try again`.

hamishwillee: Thanks! To me that is all good to know. I have updated the text and will merge. Can you please check that what I have written is still correct?

jeremyyano: Looks good, thanks!

PR #1122, opened 2021-03-17
Add more release notes from past PRs

mrpollo:
Add some features based on the v1.12-Release-Notes label

### DIFF
--- a/en/SUMMARY.md
+++ b/en/SUMMARY.md
@@ -387,8 +387,6 @@
     * [ROS Integration Testing](test_and_ci/integration_testing.md)
     * [Docker Containers](test_and_ci/docker.md)
     * [Maintenance](test_and_ci/maintenance.md)
-* [Releases](releases/README.md)
-  * [1.12](releases/1.12.md)
 * [Robotics](robotics/README.md)
   * [Offboard Control from Linux](ros/offboard_control.md)
   * [ROS](ros/README.md)
@@ -414,3 +412,6 @@
   * [Translation](contribute/translation.md)
   * [Terminology/Notation](contribute/notation.md)
   * [Licenses](contribute/licenses.md)
+* [Releases](releases/README.md)
+  * [1.12](releases/1.12.md)
+  
--- a/en/releases/1.12.md
+++ b/en/releases/1.12.md
@@ -23,8 +23,8 @@
 
 * **MAVLink Ethernet configuration ([PR#14460](https://github.com/PX4/PX4-Autopilot/pull/14460))**
   * MAVLink Ethernet channel settings such as UDP port, remote port and broadcast mode now can be changed dynamically via parameters.
-* **Support for COMPONENT_INFORMATION ([PR#16039](https://github.com/PX4/PX4-Autopilot/pull/16039))**
-  * This new message enables a GCS to understand which commands are supported by any given Autopilot (RFC: [mavlink#1339](https://github.com/mavlink/mavlink/issues/1339))
+* **Support for querying `COMPONENT_INFORMATION` ([PR#16039](https://github.com/PX4/PX4-Autopilot/pull/16039))**
+  * This new message allows a GCS to request rich hierarchical information from an autopilot; i.e. in order to understand which commands are supported in missions, or to get parameter metadata (RFC: [mavlink#1339](https://github.com/mavlink/mavlink/issues/1339))
 
 ### Commander
 
--- a/en/releases/1.12.md
+++ b/en/releases/1.12.md
@@ -7,9 +7,10 @@
 
 * [Common](#common)
 * [MAVLink](#mavlink)
-* [Multicoper](#multicopter)
 * [Commander](#commander)
-* [Control & Navigation](#control
+* [Multicoper](#multicopter)
+* [VTOL](#vtol)
+* [Control & Navigation](#control)
 
 ### Common
 
@@ -31,7 +32,6 @@
   * Consolidates arming logic, and does a better job at keeping the current state
 * **Commander: Separete out manual control setpoint processing ([PR#16878](https://github.com/PX4/PX4-Autopilot/pull/16878))**
   * Adds a new class `ManualControl` for handling `manual_control_setpoint` and handles RC loss, RC override, and RC arming/disarming
-  * 
 
 ### Multicopter
 
@@ -49,12 +49,13 @@
 * **Hover thrust independent velocity control gains**
   * Parameters `MPC_{XY/Z}_VEL_{P/I/D}` were replaced with `MPC_{XY/Z}_VEL_{P/I/D}_ACC`, see:<br/>
    [MPC_XY_VEL_P_ACC](../advanced_config/parameter_reference.md#MPC_XY_VEL_P_ACC), [MPC_XY_VEL_I_ACC](../advanced_config/parameter_reference.md#MPC_XY_VEL_I_ACC), [MPC_XY_VEL_D_ACC](../advanced_config/parameter_reference.md#MPC_XY_VEL_D_ACC), [MPC_Z_VEL_P_ACC](../advanced_config/parameter_reference.md#MPC_Z_VEL_P_ACC), [MPC_Z_VEL_I_ACC](../advanced_config/parameter_reference.md#MPC_Z_VEL_I_ACC), [MPC_Z_VEL_D_ACC](../advanced_config/parameter_reference.md#MPC_Z_VEL_D_ACC)
-  * **Attention:** The gains have a new meaning
+    :::warning The gains have a new meaning
     * Scale from velocity error in m/s to acceleration output in m/s^2
     * Existing gains need to roughly be rescaled by a factor of `* gravitational constant / hover thrust`
     * Automatic parameter transition assumes 50% hover thrust: ~10m/s^2 / 50% = 20 m/s^2. See [question](https://github.com/PX4/PX4-Autopilot/pull/14823#issuecomment-791357646)
+    :::
   * Development: [Logic introduction](https://github.com/PX4/PX4-Autopilot/pull/14749), [Parameter replacement](https://github.com/PX4/PX4-Autopilot/pull/14823)
-* ### Improve Rounded Turns ([PR#16376](https://github.com/PX4/PX4-Autopilot/pull/16376 ))
+* **Improve Rounded Turns ([PR#16376](https://github.com/PX4/PX4-Autopilot/pull/16376))**
   * Creates a more rounded turn in multirotor missions using l1-style guidance logic in corners
 
 ### VTOL
--- a/en/releases/1.12.md
+++ b/en/releases/1.12.md
@@ -17,12 +17,12 @@
 * **Remaining Flight Range based RTL Trigger ([PR#16399](https://github.com/PX4/PX4-Autopilot/pull/16399))**
   * Calculates time to home, on RTL, taking into account vehicle speed, wind speed, and desination distance/direction
 * Pre-emptive geofence breach ([PR#16400](https://github.com/PX4/PX4-Autopilot/pull/16400))
-  * Predicts the motion of the vehicle and triggers a breach if its determined the current trajectory will result in a breach happening, allowing the vehicle to be re-routed to a safe hold position.
+  * Triggers a breach if the _predicted_ current trajectory will result in a breach, allowing the vehicle to be re-routed to a safe hold position.
 
 ### MAVLink
 
 * **MAVLink Ethernet Configuration ([PR#14460](https://github.com/PX4/PX4-Autopilot/pull/14460))**
-  * Mavlink Ethernet channel settings such as UDP port, remote port and broadcast mode now can be changed dynamically via parameters.
+  * MAVLink Ethernet channel settings such as UDP port, remote port and broadcast mode now can be changed dynamically via parameters.
 * **Support for COMPONENT_INFORMATION ([PR#16039](https://github.com/PX4/PX4-Autopilot/pull/16039))**
   * This new message enables a GCS to understand which commands are supported by any given Autopilot (RFC: [mavlink#1339](https://github.com/mavlink/mavlink/issues/1339))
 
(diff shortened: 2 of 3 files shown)


en/SUMMARY.md
```diff
@@ -414,3 +412,6 @@
   * [Translation](contribute/translation.md)
   * [Terminology/Notation](contribute/notation.md)
   * [Licenses](contribute/licenses.md)
+* [Releases](releases/README.md)
```
hamishwillee, line 415: FYI only pushed to end. Certainly shouldn't be "in the middle".

en/releases/1.12.md
```diff
@@ -5,26 +5,65 @@
...
+  * Adds a new class `ManualControl` for handling `manual_control_setpoint` and handles RC loss, RC override, and RC arming/disarming
 
 ### Multicopter
 
-* #### Intuitive stick feel in Position mode
+* **More intuitive stick feel in Position mode**
   * Horizontal stick input mapped to acceleration instead of velocity setpoints
   * Removes unexpected tilt changes upon reaching travel speed velocity
   * Intuitive shunting e.g. when landing
-  * Comparable to Altitude mode but with the familiar benefits
```
hamishwillee, line 16: This is a description of existing position mode when compared to altitude mode, so what is the point?

en/releases/1.12.md
```diff
@@ -5,26 +5,65 @@
...
-  * **Attention:** The gains have a new meaning
-    * Scale from velocity error in m/s to acceleration output in m/s^2
-    * Existing gains need to roughly be rescaled by a factor of `* gravitational constant / hover thrust`
-    * Automatic parameter transition assumes 50% hover thrust: ~10m/s^2 / 50% = 20 m/s^2. See [question](https://github.com/PX4/PX4-Autopilot/pull/14823#issuecomment-791357646)
+   
+    :::warning The gains have a new meaning
+    * Scale from velocity error in $m/s$ to acceleration output in $m/s^2$
+    * Existing gains need to roughly be rescaled by a factor of: $gravitational \_ constant / hover\_thrust$
+    * Automatic parameter transition assumes 50% hover thrust: `~10m/s^2 / 50% = 20 m/s^2`.
```
hamishwillee, line 52: This is confusing. What is automatic parameter transition? What does it affect? All this really says right now is that "something" is double acceleration due to gravity.

en/releases/1.12.md
```diff
@@ -5,26 +5,65 @@
...
    [MPC_XY_VEL_P_ACC](../advanced_config/parameter_reference.md#MPC_XY_VEL_P_ACC), [MPC_XY_VEL_I_ACC](../advanced_config/parameter_reference.md#MPC_XY_VEL_I_ACC), [MPC_XY_VEL_D_ACC](../advanced_config/parameter_reference.md#MPC_XY_VEL_D_ACC), [MPC_Z_VEL_P_ACC](../advanced_config/parameter_reference.md#MPC_Z_VEL_P_ACC), [MPC_Z_VEL_I_ACC](../advanced_config/parameter_reference.md#MPC_Z_VEL_I_ACC), [MPC_Z_VEL_D_ACC](../advanced_config/parameter_reference.md#MPC_Z_VEL_D_ACC)
-  * **Attention:** The gains have a new meaning
-    * Scale from velocity error in m/s to acceleration output in m/s^2
-    * Existing gains need to roughly be rescaled by a factor of `* gravitational constant / hover thrust`
-    * Automatic parameter transition assumes 50% hover thrust: ~10m/s^2 / 50% = 20 m/s^2. See [question](https://github.com/PX4/PX4-Autopilot/pull/14823#issuecomment-791357646)
+   
+    :::warning The gains have a new meaning
+    * Scale from velocity error in $m/s$ to acceleration output in $m/s^2$
+    * Existing gains need to roughly be rescaled by a factor of: $gravitational \_ constant / hover\_thrust$
```
hamishwillee, line 51: 1. When you say "existing gains" you mean "the old gain" right?
2. FYI only, the addition of `$` around equations above allows them to be rendered using mathjax. This is not compatible with github rendering - i.e. if you ever do this you need to make sure you have tested the rendering works by building locally (which I have done in these cases).

en/releases/1.12.md
```diff
@@ -5,26 +5,65 @@
...
+
+* **Commander: use control mode flags and cleanup arm_disarm ([PR#16266](https://github.com/PX4/PX4-Autopilot/pull/16266))**
+  * Consolidates arming logic, and does a better job at keeping the current state
+* **Commander: Separete out manual control setpoint processing ([PR#16878](https://github.com/PX4/PX4-Autopilot/pull/16878))**
+  * Adds a new class `ManualControl` for handling `manual_control_setpoint` and handles RC loss, RC override, and RC arming/disarming
 
 ### Multicopter
 
-* #### Intuitive stick feel in Position mode
+* **More intuitive stick feel in Position mode**
```
hamishwillee, line 38: This needs to go into docs in Position Mode here: http://docs.px4.io/master/en/flight_modes/position_mc.html and possibly also here: http://docs.px4.io/master/en/getting_started/flight_modes.html#position-mode-mc.

I do not understand what these things mean. That's fine here, but they would need to be explained in above docs. For example 
- "what is shunting?", 
- what are these unexpected tilt changes - you mean it is going along at full speed and the tilt of the vehicle - oscillates or something?
- "Horizontal stick input mapped to acceleration instead of velocity setpoints" is just words. Is there some way of explaining why this is more intuitive or do we need to take it on faith?

en/releases/1.12.md
```diff
@@ -5,26 +5,65 @@
...
+
+* **MAVLink Ethernet configuration ([PR#14460](https://github.com/PX4/PX4-Autopilot/pull/14460))**
+  * MAVLink Ethernet channel settings such as UDP port, remote port and broadcast mode now can be changed dynamically via parameters.
+* **Support for COMPONENT_INFORMATION ([PR#16039](https://github.com/PX4/PX4-Autopilot/pull/16039))**
+  * This new message enables a GCS to understand which commands are supported by any given Autopilot (RFC: [mavlink#1339](https://github.com/mavlink/mavlink/issues/1339))
+
+### Commander
+
+* **Commander: use control mode flags and cleanup arm_disarm ([PR#16266](https://github.com/PX4/PX4-Autopilot/pull/16266))**
+  * Consolidates arming logic, and does a better job at keeping the current state
```
hamishwillee, line 32: What does "better job at keeping current state" mean?

Essentially as a user, why do I care about using these flags and this cleanup? Do I have to do anything different? What will I see different? I.e. might I get better information on why things won't arm in the GCS?

en/releases/1.12.md
```diff
@@ -5,26 +5,65 @@
...
+### Common
+
+* **RTL Trigger based on remaining flight range ([PR#16399](https://github.com/PX4/PX4-Autopilot/pull/16399))**
+  * Calculates time to home, on RTL, taking into account vehicle speed, wind speed, and destination distance/direction
+* **Pre-emptive geofence breach ([PR#16400](https://github.com/PX4/PX4-Autopilot/pull/16400))**
+  * Triggers a breach if the _predicted_ current trajectory will result in a breach, allowing the vehicle to be re-routed to a safe hold position.
+
+### MAVLink
+
+* **MAVLink Ethernet configuration ([PR#14460](https://github.com/PX4/PX4-Autopilot/pull/14460))**
```
hamishwillee, line 24: MAVLink Ethernet has NO DOCS. Can we get some. Possibly needs to be added to http://docs.px4.io/master/en/peripherals/mavlink_peripherals.html - someone needs to write these or at least talk to me. 

For now, this needs release note needs list of the affected parameters.

en/releases/1.12.md
```diff
@@ -5,26 +5,65 @@
...
+* **More intuitive stick feel in Position mode**
   * Horizontal stick input mapped to acceleration instead of velocity setpoints
   * Removes unexpected tilt changes upon reaching travel speed velocity
   * Intuitive shunting e.g. when landing
-  * Comparable to Altitude mode but with the familiar benefits
-    * drift less hover
-    * straight fast forward flight
-    * automatic braking when letting go of the stick
   * Opt out possible using [MPC_POS_MODE](../advanced_config/parameter_reference.md#MPC_POS_MODE)
   * Development: [First attempt](https://github.com/PX4/PX4-Autopilot/pull/12072), [Introduction](https://github.com/PX4/PX4-Autopilot/pull/16052), [Improvements](https://github.com/PX4/PX4-Autopilot/pull/16320), [Bugfix zero oscillation](https://github.com/PX4/PX4-Autopilot/pull/16786), [Bugfix position unlock](https://github.com/PX4/PX4-Autopilot/pull/16791), [Bugfix invalid setpoint](https://github.com/PX4/PX4-Autopilot/pull/17078)
```
hamishwillee, line 43: What does all this mean? Are you saying that these parts of the same feature are still under development? Are they intended for this release?

en/releases/1.12.md
```diff
@@ -5,26 +5,65 @@
...
+### VTOL
+
+* **RTL improvements ([PR#16377](https://github.com/PX4/PX4-Autopilot/pull/16377))**
+  * Hardens the RTL safety failsafes taking into consideration the many edge cases when trying to land, depending on the current vehicle mode (Multicopter vs Fixed Wing)
+
+
+### Control
+
+* **Dynamic Notch Filter updated with Gyro FFT ([PR#16385](https://github.com/PX4/PX4-Autopilot/pull/16377))**
+  * Adds dynamic notch filtering to the gyro control data resulting in much smoother control
```
hamishwillee, line 68: This is OK. Is there any relevant information for tuning or whatever? Or is this just magic "goodness"?

en/releases/1.12.md
```diff
@@ -5,26 +5,65 @@
...
+  * MAVLink Ethernet channel settings such as UDP port, remote port and broadcast mode now can be changed dynamically via parameters.
+* **Support for querying `COMPONENT_INFORMATION` ([PR#16039](https://github.com/PX4/PX4-Autopilot/pull/16039))**
+  * This new message allows a GCS to request rich hierarchical information from an autopilot; i.e. in order to understand which commands are supported in missions, or to get parameter metadata (RFC: [mavlink#1339](https://github.com/mavlink/mavlink/issues/1339))
+
+### Commander
+
+* **Commander: use control mode flags and cleanup arm/disarm ([PR#16266](https://github.com/PX4/PX4-Autopilot/pull/16266))**
+  * Consolidates arming logic, and does a better job at keeping the current state
+* **Commander: Separate out manual control setpoint processing ([PR#16878](https://github.com/PX4/PX4-Autopilot/pull/16878))**
+  * Adds a new class `ManualControl` for handling `manual_control_setpoint` and handles RC loss, RC override, and RC arming/disarming
```
hamishwillee, line 34: Same question, is there any impact on end uses, any docs impact? (this is OK, just a query)

en/releases/1.12.md
```diff
@@ -5,26 +5,65 @@
...
+* **RTL Trigger based on remaining flight range ([PR#16399](https://github.com/PX4/PX4-Autopilot/pull/16399))**
+  * Calculates time to home, on RTL, taking into account vehicle speed, wind speed, and destination distance/direction
+* **Pre-emptive geofence breach ([PR#16400](https://github.com/PX4/PX4-Autopilot/pull/16400))**
+  * Triggers a breach if the _predicted_ current trajectory will result in a breach, allowing the vehicle to be re-routed to a safe hold position.
+
+### MAVLink
+
+* **MAVLink Ethernet configuration ([PR#14460](https://github.com/PX4/PX4-Autopilot/pull/14460))**
+  * MAVLink Ethernet channel settings such as UDP port, remote port and broadcast mode now can be changed dynamically via parameters.
+* **Support for querying `COMPONENT_INFORMATION` ([PR#16039](https://github.com/PX4/PX4-Autopilot/pull/16039))**
```
hamishwillee, line 26: FYI I updated this a little. The previous version just mentioned being able to get "supported commands". That's a great use case but we haven't implemented it. 
What we have so far is the infrastructure (nearly done) and the ability to get parameter metadata from the vehicle in a standardised way. 

The difference this adds is that currently QGC has to get parameter info from flight stacks in different formats. I'm not sure exactly how it gets to QGC but often the parameter info gets out of sync. In future QGC will get standardised info irrespective of the flight stack, which means that it can support parameters in other flight stacks much better. The info will always be "in sync", and translations will be able to be dynamically updated after release (in theory).

Generally, again FYI, the key benefit is that this can be used for any kind of metadata, and it is heirarchical. So we can use this without having to have new mavlink messages or transpose complicated nested structures into the flat structures that make sense for mavlink messages.

hamishwillee:
@mrpollo @MaEtUgR @bkueng I have given this a fairly thorough review. This is fine for now. My concern is that a lot of this stuff should have "real" docs and I see no movement to create those docs. I could be wrong, in which case my apologies.

We should go through each item and decide if any docs make sense, and if so start that work NOW. Thoughts?

en/releases/README.md
```diff
@@ -1,5 +1,5 @@
 # Releases
 
-In depth release notes of what changed and how to adapt and make best use.
+Notes explaining changes in each release.
```
hamishwillee, line 3: Note the difference. Feel free to add back "how to adapt to best use" when we have real docs in place :-)

hamishwillee:
PS @mrpollo I don't mind if you merge and we iterate any required fixes from this discussion in new PRs. It isn't that this is wrong, it is just that there are lots of unanswered questions that should naturally be answered in proper docs.

mrpollo:
Your assessment is correct. Most of the features are not documented. Let's start a to-do list. I can lead the primary docs push, but let's try to ping the authors for help. Otherwise, we'll have incomplete or inaccurate information.

hamishwillee:
AFAIK nothing here has docs, so the release notes are the list. Perhaps 10 mins in the dev call so that we can get ownership assigned to each and a bit of forward planning on what is likely to come in the next betas?
If not, lets see what @bkueng and @MaEtUgR say. The job might not be as big as I think.

en/SUMMARY.md
mrpollo: Yeah the middle was weird, i think its best if we keep it towards the end

en/releases/1.12.md
mrpollo: @MaEtUgR can you please chime in, we would like to understand what is the proper way to explain this to developers that wouldn't be 100% knowledgable of the internal working of the changes

en/releases/1.12.md
mrpollo: 1. Yes, this means that they would need to revise their gains after upgrading.

en/releases/1.12.md
mrpollo: Good catch, I created an issue to track this https://github.com/PX4/PX4-user_guide/issues/1133

en/releases/1.12.md
mrpollo: @MaEtUgR what's your intention behind the last list item? I'm wondering if there's a better way to represent this.

en/releases/1.12.md
mrpollo: I think this is not an end-user feature, this is geared towards developers, I will reword to add better detail, explaining the changes

en/releases/1.12.md
mrpollo: Can you raise an issue for this? It's out of my scope, I wouldn't know where to start even creating the issue

en/releases/1.12.md
mrpollo: I pinged a few devs, I will try to add more detail to the release notes, and document properly on the user guides.

https://github.com/PX4/PX4-user_guide/issues/1136

en/releases/1.12.md
mrpollo: Yeah there are consequences here, and should be documented on the guides.

mrpollo: https://github.com/PX4/PX4-user_guide/issues/1137

en/releases/1.12.md
mrpollo: I edited this to mention that this isn't just for GCS, any mavlink system that implements this can take advantage of it (EG: MAVSDK), but did leave a note that this was done with the intention to ease the burden of GCS's

mrpollo:
@hamishwillee let's continue the work in subsequent PRs, I will keep moving fast

en/releases/1.12.md
bkueng: @mrpollo be aware that there are still breaking mavlink message changes coming for this, so it's not yet usable. I don't think it's worth mentioning until we have this working and enabled on both sides.

hamishwillee: @mrpollo I think @bkueng is right, as per usual :-). I have put this in a comment to hide it. We can unhide it and update if it managed to get in and somewhat stable in the final release: https://github.com/PX4/PX4-user_guide/pull/1142

en/releases/1.12.md
```diff
@@ -1,30 +1,69 @@
...
+    * Automatic parameter transition assumes 50% hover thrust: `~10m/s^2 / 50% = 20 m/s^2`.
+      See [question](https://github.com/PX4/PX4-Autopilot/pull/14823#issuecomment-791357646)
+    :::
   * Development: [Logic introduction](https://github.com/PX4/PX4-Autopilot/pull/14749), [Parameter replacement](https://github.com/PX4/PX4-Autopilot/pull/14823)
+* **Improve Rounded Turns ([PR#16376](https://github.com/PX4/PX4-Autopilot/pull/16376))**
+  * Creates a more rounded turn at waypoints in multirotor missions (using L1-style guidance logic in corners)
+
+### VTOL
+
+* **RTL improvements ([PR#16377](https://github.com/PX4/PX4-Autopilot/pull/16377))**
```
hamishwillee, line 61: @mrpollo this needs docs too - i have started discussion here: https://github.com/PX4/PX4-Autopilot/pull/16377#issuecomment-805537699

MaEtUgR:
@hamishwillee Exactly. And having the release notes editable and linkable in the docs at least helps me to keep the overview to not forget to document. I agree we should plan based on the release items. The goal should be that every item has a link to further details.

hamishwillee:
@MaEtUgR Thanks. What would help is if that planning happens "visibly". It is too easy for things to fall of the list, or for me to think there are docs needed, and there are not.

en/releases/1.12.md
hamishwillee: Most of this is not relevant to document other than change from speed to acceleration on sticks. That is done.

en/releases/1.12.md
hamishwillee: @mrpollo FYI following a bit of research and discussion with Mattias there isn't any docs consequence that was not already in PX4 v1.11. I did some tidy on that. Closing.

en/releases/1.12.md
hamishwillee: No answer. Not surprised. RTL behaviour is super complicated and hard to document. This looks like an edge case fix though, so we might be missing something, but what is there probably isn't wrong.

MaEtUgR:
> It is too easy for things to fall of the list, or for me to think there are docs needed, and there are not.

There's no perfect answer and for sure there does not need to be specific documentation for changes but all listed items could at least link to docs about that component/feature and then it gets more obvious if that documentation already contains how it works now or not.

hamishwillee:
@MaEtUgR While I agree there is no perfect answer, there are some ways of working we could all adopt to make this much smoother. The first one being to update the docs immediately at the point that the associated PR is merged. The person writing a PR knows what they just did, and writing the docs is a great sanity check on the usability.

PR #1124, opened 2021-03-17
Add Microhard Serial Radios to Telemetry

AlexKlimaj:
This PR adds documentation detailing Microhard serial radios. Other manufactures are welcome to add theirs to the list as well.


en/telemetry/microhard_serial.md
```diff
@@ -0,0 +1,14 @@
+# Microhard Serial Radio
+
+[Microhard Pico Serial](http://microhardcorp.com/P900.php) radios enable MAVLink communication between a radio on a vehicle and a GCS.  
+Microhard Pico Serial radios are 1 watt output radios that support point to point, point to multi-point, and mesh modes.
+A single ground station radio can be used to communicate with multiple vehicles using point to multi-point or mesh. Vehicles must have different IDs.
+
+![Microhard Radio](../../assets/hardware/telemetry/ark_microhard_serial.jpg)
```
hamishwillee, line 7: Just FMI, why prefix "ark" on this file?

hamishwillee:
@AlexKlimaj Thanks. 

What is missing is setup and configuration information for use with PX4. This typically should be in the same document, since all microhards tend to implement the same API. IF the wiring information can be common, then that should also be in this document.

My experience with Microhard configuration (via Ethernet) is that it is HORRIBLE, mostly because there are many possible configuration architectures. Upshot, really want this in, but not without a turnkey setup of at least one architecture for end users.

en/telemetry/microhard_serial.md
AlexKlimaj: Its a picture of my microhard product. 
https://arkelectron.com/product/1w-900mhz-serial-telemetry-air-radio/

AlexKlimaj:
Thanks, I'll add setup information. The Microhard Serials are much easier to setup than the DDLs.

en/telemetry/microhard_serial.md
```diff
@@ -0,0 +1,35 @@
...
+## Purchase:
+
+* [1W 900MHz Serial Telemetry Radio](https://arkelectron.com/product/1w-900mhz-serial-telemetry-air-radio/)
+* [1W 900MHz USB Serial Telemetry Radio](https://arkelectron.com/product/1w-900mhz-serial-telemetry-ground-radio/)
+* [1W 2.4GHz Serial Telemetry Radio](https://arkelectron.com/product/1w-2400mhz-serial-telemetry-radio/)
+* [1W 2.4GHz USB Serial Telemetry Radio](https://arkelectron.com/product/1w-2400mhz-usb-serial-telemetry-radio/)
+
+### Connecting
+
+Connection on the autopilot is done using a Pixhawk standard 6 pin JST GH Telemetry cable.
```
hamishwillee, line 18: I think you're just saying attach to TELEM2 or some free serial port , and configure for mavlink right (as per https://docs.px4.io/master/en/peripherals/mavlink_peripherals.html) - right? Does the cable come with it?

en/telemetry/microhard_serial.md
```diff
@@ -0,0 +1,35 @@
...
+When set to 100mW output power or less, power from the telemtry cable is sufficient.
+For higher radio output powers above 100mW, a battery connection is needed on the 2 Pin Molex Nano-Fit.
```
hamishwillee, line 20: Cable is supplied?

Not mandatory, but would it be possible to get an image showing your setup with pixhawk, one of the serial radios and the power cable. Picture worth a thousand words (and I'm kind of hoping it shows the BEC powering the MH.

en/telemetry/microhard_serial.md
```diff
@@ -0,0 +1,35 @@
...
+
+Connection on the groundstation is done through USB C. 1W output powers are supported when using USB PD.
+
+## Setup/Configuration
+
+Using [Pico Config](https://arkelectron.com/wp-content/uploads/2020/09/PicoConfig-1.7.zip) on Windows, the radio settings can be adjusted.
```
hamishwillee, line 26: Is this only configurable on windows?

en/telemetry/microhard_serial.md
```diff
@@ -0,0 +1,35 @@
...
+
+![Pico Config](../../assets/hardware/telemetry/pico_configurator.png)
```
hamishwillee, line 28: I think the steps are like this (please confirm):
1. Connect MH to a windows computer using USB-C. 
2. Run Pico Config. The two should connect and you will be able to edit the settings using PicoConfig
3. Do I set a baud rate at this point? If so, what value?
4. Anyway, I then need to create another connection, this time between QGC and the MH. So I add a new connection as per usual using the baud rate I set above.
5. Then I connect vehicle MH and flight controller via TELEM1. 
6. Then I can connect the vehicle unit to the windows PC using the FTDI adapter (ie to serial console). 
   - I presume I disconnect the unit already attached to the PC or QGC? Otherwise wont PicoConfig get confused about which unit to configure.
7. So then I match the baud rate on the vehicle to what I already set in QGC and for the ground unit.
8. Then I can disconnect PicoConfig. Both my ground and air units should have same baud rate and can talk. So does QGC. Everything good

If that is about right, then I might subedit this.

en/telemetry/microhard_serial.md
AlexKlimaj: Yeah, any telemetry or serial port. The cable does come with it.

en/telemetry/microhard_serial.md
AlexKlimaj: Cables are supplied. I'll add a picture.

en/telemetry/microhard_serial.md
AlexKlimaj: PicoConfig only works on windows. If there is enough interest in the future, I could make a command line version for linux. The radios can also be configured through the main serial port using modem AT commands.

en/telemetry/microhard_serial.md
AlexKlimaj: 3. You set the baud rate to the same baud rate used on the flight controller serial port.
6. Yes you need to disconnect the unit already attached to the PC.

Feel free to edit.

hamishwillee: Yes, but for 3 what "should" you use? What do YOU use? 

I ask because the default is pretty low https://docs.px4.io/master/en/peripherals/mavlink_peripherals.html#default-mavlink-ports and - 1200 B/s and the radios might be much more capable.

AlexKlimaj: Since the default baud rate on PX4 is 57600, I would go with that. 
No rush, I appreciate the work.

en/telemetry/microhard_serial.md
```diff
@@ -0,0 +1,53 @@
...
+
+The ground station radio is connected via USB C (for both radio configuration and communication with QGroundControl):
+- *Pico Config* will automatically detect and connect to the configuration port.
+   Adjust the settings so that the baud rate matches PX4.
+- *QGroundControl* should connect automatically. If not, you can add a new serial connection specifying the baud rate use in [Application Settings > Comm Links](https://docs.qgroundcontrol.com/master/en/SettingsView/SettingsView.html).
+
+The vehicle-based radio is connected to the flight-controller's `TELEM1` port.
+For configuration you will also have to connect an FTDI adapter between the 3 pin JST-GH Config port and a Windows PC running Pico Config.
```
hamishwillee, line 51: @AlexKlimaj Is this where the FTDI connection goes: 
image

Do you need to have the telem connected during configuration of the air unit?

en/telemetry/microhard_serial.md
```diff
@@ -0,0 +1,53 @@
+# Microhard Serial Radio
+
+[Microhard Pico Serial](http://microhardcorp.com/P900.php) radios enable MAVLink communication between a radio on a vehicle and a GCS.
+Microhard Pico Serial radios are 1 Watt output radios that support point to point, point to multi-point, and mesh modes.
+A single ground station radio can be used to communicate with multiple vehicles using point to multi-point or mesh.
+Vehicles must have different IDs.
```
hamishwillee, line 6: MAVLink IDs? Or some other ID

People are interested in range these can be used at. Any ball park figures for a single GCS /vehicle system running on 1mW and 1W?

If I change nothing but the baud rate, will this set up a one to one network? 

Does the mesh network stuff "work"? I assume this just means that if you set up a mesh and you had multiple vehicles and GCS they would all see all traffic?

Do you want to talk about other benefits - e.g. encryption. How do you set that up?

en/telemetry/microhard_serial.md
```diff
@@ -0,0 +1,53 @@
...
+:::
+
+The radios are configured using [Pico Config](https://arkelectron.com/wp-content/uploads/2020/09/PicoConfig-1.7.zip) (Windows only).
+
+![Pico Config](../../assets/hardware/telemetry/pico_configurator.png)
+
+The ground station radio is connected via USB C (for both radio configuration and communication with QGroundControl):
+- *Pico Config* will automatically detect and connect to the configuration port.
+   Adjust the settings so that the baud rate matches PX4.
+- *QGroundControl* should connect automatically. If not, you can add a new serial connection specifying the baud rate use in [Application Settings > Comm Links](https://docs.qgroundcontrol.com/master/en/SettingsView/SettingsView.html).
```
hamishwillee, line 48: QGC detects Sik radios out of the box, because I believe they use a matching default baud rate. So I "assume" that out of the box QGC and the radio should just connect if you don't change the PX4 baud. Seem reasonable?

hamishwillee:
I added a few questions. Nearly good to go IMO.

en/telemetry/microhard_serial.md
AlexKlimaj: I haven't done any tests personally, I always use 57600 baud for the serial link. There is also wireless link rate that defaults to 172800bps which corresponds to a 250kHz bandwidth. You need to have the wireless link rate set higher than the serial link rate. I think that might be too much information for this guide.

en/telemetry/microhard_serial.md
AlexKlimaj: Yes that is where the FTDI connection goes. You don't need telem connected when you configure it, but it does need power.

en/telemetry/microhard_serial.md
AlexKlimaj: Yes Mavlink IDs. 
Ballpark for point to point at 1W with the default settings is about 5 miles with no other optimizations. 
Yes the mesh network works on these. Same caveat applies that all vehicles would need a different mavlink ID.
Encryption is an optional feature that is extra at order time. If you have a unit with encryption, the setting appears in PicoConfig.

en/telemetry/microhard_serial.md
AlexKlimaj: QGC has some special code to detect the sik radios. For these, you need to set up the serial link in the settings.

en/telemetry/microhard_serial.md
hamishwillee: @AlexKlimaj 

> I think that might be too much information for this guide.

The only requirement I have for the docs is that the product works and there is a decent turnkey setup provided for users. On top of that I hope that docs fairly position the product so that user can make an informed choice - ie as to the kinds of bandwidth and distance they can use the product, weight, power consumption, and tradeoffs for the above. Within that, I'm really happy for vendors to market the great things they are doing that differentiate their products. 

Upshot, if you think it will help differentiate the product and help people use it better, then providing additional configuration options and tradeoffs information at the end is completely OK :-). 

My impression here though would be that you would want to do some testing to kind of work out the tradeoffs.

hamishwillee:
OK, subedited and pushed. If you want to add more, you can do as a separate PR :-)

PR #1126, opened 2021-03-18
Simplify build instructions/discontinued things

hamishwillee:
This 
- pushes the complicated build instructions for non-nuttx things boards like snapdragon and opcop, and RaPi into the docs for those boards. 
- Moves snapdragon, ocpoc and other discontinued boards into their own sections. 

Result is that build instructions are much less complicated. 

PS This started as tidy small up of  #1100 - fixes for Ocop. Never pull a piece of string.


hamishwillee:
FYI This (merged) PR significantly reduces the build instructions, which were to my mind a big scary. Mostly because they had a lot of stuff on snapdragon and RaPi builds which are either discontinued or not relevant to 99% of users. 
That stuff is now more clearly in the individual docs.

Anyway, it still looks a bit complicated for "first use". I was thinking that most users don't need the stuff in red (below), at least when they are getting started.

image

 Perhaps it could be moved into a couple of sub pages based on the headings - or maybe just one "Advanced Build Information" (probably prefer the former). @bkueng Advice appreciated.

bkueng:
Yes 'Firmware Version & Git Tags' could go to a subpage. The closest fit for 'Binary Size Profiling' I see is https://docs.px4.io/master/en/debug/gdb_debugging.html

PR #1139, opened 2021-03-23
Update airsim.md documentation links

jonyMarino:
There is a page dedicated to AirSim documentation. So it is better to refer to it.


hamishwillee:
Yes, these links are much better. Thank you @jonyMarino

PR #1149, opened 2021-03-26
GPS protocol config

bkueng:
- add note about GPS_x_PROTOCOL
- release notes: airframe scripts require updating

### DIFF
--- a/en/gps_compass/README.md
+++ b/en/gps_compass/README.md
@@ -77,6 +77,11 @@ While these are all software-compatible, there are several different pin orderin
 The "standard" GPS/compass configuration is provided below.
 Additional device-specific configuration may be provided in PX4 or manufacturer device documentation (e.g. [Trimble MB-Two > Configuration](../gps_compass/rtk_gps_trimble_mb_two.md#configuration)).
 
+:::note
+The GPS protocol defaults to u-blox, which can be configured with [GPS_x_PROTOCOL](../advanced_config/parameter_reference.md#GPS_1_PROTOCOL).
+If this is incorrect, the GPS will not be detected.
+:::
+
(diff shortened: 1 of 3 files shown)


en/gps_compass/README.md
```diff
@@ -77,6 +77,11 @@ While these are all software-compatible, there are several different pin orderin
 The "standard" GPS/compass configuration is provided below.
 Additional device-specific configuration may be provided in PX4 or manufacturer device documentation (e.g. [Trimble MB-Two > Configuration](../gps_compass/rtk_gps_trimble_mb_two.md#configuration)).
 
+:::warning
```
hamishwillee, line 80: FYI, I made this a warning to highlight it. Might be easy to miss otherwise.

en/gps_compass/rtk_gps_trimble_mb_two.md
```diff
@@ -49,6 +49,8 @@ The pins on the 28-pin connector are numbered as shown below:
 
 ## Configuration
 
+First set the GPS protocol to Trimble ([GPS_x_PROTOCOL=3](../advanced_config/parameter_reference.md#GPS_1_PROTOCOL)).
```
hamishwillee, line 52: @bkueng Thanks, looks great, merged.

I added this point in the Trimble MB2 config because it is where you are likely to run into problems. I don't think we have stand alone pages for any other non-ublox GPS right - ie. all of the peers of this one are ublox right? http://docs.px4.io/master/en/gps_compass/rtk_gps_trimble_mb_two.html

bkueng: Correct, the vast majority is on u-blox

PR #1153, opened 2021-03-28
Fixed command to return the IP address

PaulStubbs:
Fixed docker command to return only the IP address of the container.


hamishwillee:
Thanks @PaulStubbs ! @julianoes who would be the right person to comment on this change? I can test it of course, but presumably whoever posted the original version might have some insight

julianoes:
Thanks @PaulStubbs. Can you explain what this changes or how it works? I'm not familiar with this at all.

hamishwillee:
@PaulStubbs , can you please explain the change as per https://github.com/PX4/PX4-user_guide/pull/1153#issuecomment-810025167. Otherwise I'll close this as stale, and potentially lose useful information.

hamishwillee:
I am closing this. It is "unverified". Happy for it to be reopened if it can be explained/verified.

julianoes:
I just tested this and the original command does work but only if you remove the space in-between the opening braces as commented below the command.

PR #1157, opened 2021-03-30
Update hitl.md

eldewany:
Fixed launch command for HITL setup for gazebo


hamishwillee:
Thanks @eldewany !

@bperseghetti I'm merging this because it looks very reasonable, but FMI can you please confirm/explain why we need a hitl-specific iris world?

eldewany:
@hamishwillee Since the HITL of iris is separated from the default jinja generation of normal iris vehicle in SITL. it may be ambiguous for a normal user who doesn't dive in GitHub pr's since it's not mentioned in the documentation.

PR #1167, opened 2021-04-01
Improve payload docs

hamishwillee:
This clarifies the mapping between MAV_CMD_DO_SET_ACTUATOR and the outputs (ie. explains it is same outputs as used for RC triggering. 

Further, it does a quick fix on the [Custom Payload Mixer](https://docs.px4.io/master/en/concept/custom_mixer_payload.html) doc which Beat says is not really needed any more. Mostly this is to discuss how we deal with the doc. 

There are some open questions inline.

Follows on from https://github.com/PX4/PX4-user_guide/pull/1105#discussion_r604620274


en/payloads/README.md
```diff
@@ -25,27 +25,41 @@ The following topics show how to *connect* your camera configure PX4:
...
+
+### Mission Triggering
+
+You can use the [MAV_CMD_DO_SET_ACTUATOR](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_ACTUATOR) MAVLink command to set (up to) three actuators values at a time, either in a mission or as a command.
+
+Command parameters `param1`, `param2`, and `param3`, are mapped to the _same outputs_ as are used for [RC triggering](#rc-triggering).
+Usually these are the `AUX1`, `AUX2`, `AUX3` outputs of your flight controller (the RC section above explains how to check).
+The other command parameters (`param4` to `param7`) are unused/ignored by PX4.
+
+The parameters take normalised values between -1 and 1.
```
hamishwillee, line 53: @bkueng "How" are values normalised, and does it matter to the end user/integrator. 

What do I mean? Well an end user will get a setup vehicle and might be told that the particular actuator they are using can be controlled between a resting position of "0" and a full range of "1" and a full negative range of "-1". Presumably they might also be told that the actuator has just a range from 0 to 1, and the setup needs to ensure that negative values are pinned to 0.

So an integrator needs to understand what they have to do in order to make that mapping happen. I am assuming perhaps that the values are just normalised between PWM_AUX_MINx and PWM_AUX_MAXx but I'm not sure then how you'd support the "0 to 1" case. Also what other servo values should be set.

Does that make sense?

en/concept/custom_mixer_payload.md
```diff
@@ -1,5 +1,10 @@
 # Custom Payload Mixers
 
+:::warning
```
hamishwillee, line 3: @bkueng This is basically the quickest way I can mark this as "do not use" without deleting it. 

What I need to understand is when you'd use it in practice. I THINK that this would be when you wanted to implement your own real dedicated command for controlling a "thing". So essentially say I want to add a gripper payload and I want to use a specific MAV_CMD to do that. I'd follow these instructions.

But the story still doesn't feel "complete" for that, firstly because it doesn't map through to RC (which you might also want). Secondly, if you're adding hardware you want an easy way to say "the gripper hardware is on this servo" so that a gripper message can magically map through to that servo. I think that means that we should *reserve* control group ids for specific purposes, or there is no way to know that the output has actually be used/assigned. 

Upshot, I don't understand how this would ever be useful. Can you explain, or perhaps we should delete it?

bkueng: I would just delete it for now, also because the implementation will change quite a bit within the next months or so (there will be a concept of 'functions', where in your example you would just add the gripper as a new function, make sure it's handled in the output driver library, and you're done. The user can then select the function via param).

> I think that means that we should reserve control group ids for specific purposes, or there is no way to know that the output has actually be used/assigned.

Yes that is correct.

en/payloads/README.md
hamishwillee: @bkueng Thanks. I have updated the docs to say this and I will merge

But I think it is a little awkward. In the case where you have an actuator that moves from X to Y, you have to specify values between -1 and 1, while 0 and 1 would be much more natural. I suspect the design of the -1 range was for cases that  have a "center position, that can go positive or negative.

PR #1168, opened 2021-04-01
Binary size profiling in own doc

hamishwillee:
This splits Binary Size profiling into its own doc. It is not particularly relevant to "Building" or to Getting Started, but it is relevant to debugging. So I have moved it into that section.

@bkueng This is a step towards further simplifying the build instructions, as we discussed here: https://github.com/PX4/PX4-user_guide/pull/1126#issuecomment-801561856

The thing about [Building PX4 Software](https://github.com/PX4/PX4-user_guide/pull/1126#issuecomment-801561856) is that it is under Getting Started, but it includes things that are relatively advanced. For example, as a getting started person you basically need to know some basic build commands and troubleshooting. 
Presenting the [PX4 make targets](http://docs.px4.io/master/en/dev_setup/building_px4.html), should be linked for more information, but is kind of irrelevant at this level.

So, what I have done for now is pushed that below troubleshooting to the end. You only see it when you got the basics.

I was wondering if perhaps this could be its own document under "Concepts"? or Concepts > Building, with other more advanced topics related to the toolchain - e.g. cmake files etc. Just an idea. Let me know if you think that is worth exploring.

### DIFF
--- a/en/dev_setup/building_px4.md
+++ b/en/dev_setup/building_px4.md
@@ -137,143 +137,6 @@ Build commands for other boards are given the [board-specific flight controller 
 [VSCode](../dev_setup/vscode.md) is the officially supported (and recommended) IDE for PX4 development.
 It is easy to set up and can be used to compile PX4 for both simulation and hardware environments.
 
-## PX4 Make Build Targets
-
-The previous sections showed how you can call *make* to build a number of different targets, start simulators, use IDEs etc.
-This section shows how *make* options are constructed and how to find the available choices.
-
-The full syntax to call *make* with a particular configuration and initialization file is:
-```sh
-make [VENDOR_][MODEL][_VARIANT] [VIEWER_MODEL_DEBUGGER_WORLD]
-```
-
-**VENDOR_MODEL_VARIANT**: (also known as `CONFIGURATION_TARGET`)
-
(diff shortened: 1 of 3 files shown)


en/dev_setup/building_px4.md
```diff
@@ -137,6 +137,73 @@ Build commands for other boards are given the [board-specific flight controller
 [VSCode](../dev_setup/vscode.md) is the officially supported (and recommended) IDE for PX4 development.
 It is easy to set up and can be used to compile PX4 for both simulation and hardware environments.
 
+
```
bkueng, line 140: Not directly related, but might be worth mentioning `make list_config_targets` already above in `## Other Boards`, so you know what's available.

bkueng (approved):
Looks good, and it's what I had in mind as well.

> I was wondering if perhaps this could be its own document under "Concepts"? or Concepts > Building, with other more advanced topics related to the toolchain - e.g. cmake files etc. Just an idea. Let me know if you think that is worth exploring.

Yes at some point explaining the build system a bit more makes sense.

en/dev_setup/building_px4.md
hamishwillee: Excellent idea. Done. 

I will leave the make target thingy in place for now. I might move it into concepts next week - right now I think this is "OK".

PR #1173, opened 2021-04-03
[Update] Updates and expands the wifi telemetry page to include newer fmu and wifi boards

teyrana:
The old page had a few issues: 
1. It linked to obsolete (out of production) wifi boards
2. It listed advice about supply voltage which is no longer universal 
    - i.e. the Pixhawk 4 (a **reference** board)  supplies 5VDC to  it's telemetry modules.
3. Added some more recent wifi boards
4. Updated instructions for QGC when configuring those less common wifi boards 
    - i.e. how to add a custom connection in QGC

### DIFF
--- a/en/telemetry/esp8266_wifi_module.md
+++ b/en/telemetry/esp8266_wifi_module.md
@@ -1,7 +1,7 @@
 # ESP8266 WiFi Module
 
-The ESP8266 is a low-cost and readily available Wi-Fi module with full TCP/IP stack and microcontroller capability.
-It can be used with any Pixhawk series controller.
+The ESP8266 and its clones are low-cost and readily available Wi-Fi modules with full TCP/IP stack and microcontroller capability.
+They can be used with any Pixhawk series controller.
 
 :::tip
 ESP8266 is the *defacto* default WiFi module for use with [Pixracer](../flight_controller/pixracer.md) (and is usually bundled with it).
@@ -9,21 +9,38 @@ ESP8266 is the *defacto* default WiFi module for use with [Pixracer](../flight_c
 
 ## Where to Buy
 
-The module is readily available.
-A few vendors are listed below.
+The module is readily available, and usually have the firmware already installed and configured to useable defaults.  A few vendors are listed below.
 
+### Accept 3.3v Supply:
 * [Sparkfun](https://www.sparkfun.com/products/13678)
-* [GearBeast](https://us.gearbest.com/boards-shields/pp_009604906563.html)
+* (discontinued) [GearBeast](https://us.gearbest.com/boards-shields/pp_009604906563.html)
+
+### Accept 5.0v Supply:
+* [Banggood]( https://www.banggood.com/Wireless-Wifi-to-Uart-Telemetry-Module-With-Antenna-for-Mini-APM-Flight-Controller-p-1065339.html )
+* [Banggood]( https://www.banggood.com/MAVLink-Wifi-Bridge-2_4G-Wireless-Wifi-Telemetry-Module-with-Antenna-for-Pixhawk-APM-Flight-Controller-p-1428590.html )
 
 
 ## Module Setup
 
-The ESP8266 firmware has these *factory* settings:
+Your board or packaging should include this information.  Often it is simply printed on the reverse side of the board.
 
-* SSID: PixRacer
-* Password: pixracer
-* WiFi Channel: 11
-* UART speed 921600
+
+1. Factory settings -- Example 1:
+
+    - SSID: PixRacer
+    - Password: pixracer
+    - WiFi Channel: 11
+    - UART speed 921600
+
+2. Factory Settings -- Example 2:
+
+    - SSID: IFFRC_xxxxxxxx
+    - Password: 12345678
+    - IP: 192.168.4.1
+    - Port: 6789 (TCP) 
+
+
+
 
 ### Build From Sources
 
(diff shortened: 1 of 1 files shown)


teyrana:
Rebased onto master @ Sat 03 Apr 2021 03:45:37 PM UTC

en/telemetry/esp8266_wifi_module.md
```diff
@@ -58,9 +75,14 @@ Where:
 ### Wiring for Flashing the Firmware
 
 :::warning
-ESP8266 must be powered with 3.3 volts only.
+Most ESP8266 units must only accept 3.3 volts!
```
hamishwillee, line 78: Suggest change to `Most ESP8266 must be powered with 3.3 volts only.`

en/telemetry/esp8266_wifi_module.md
```diff
@@ -112,6 +136,22 @@ On your wifi-enabled *QGroundControl* ground station computer/tablet, find and c
...
+
+QGC supports a variety of other wifi connections, if manually configured: 
+
+https://docs.qgroundcontrol.com/master/en/SettingsView/SettingsView.html
+
+If QGC doesn't automatically connect to your vehicle, add it's network connection as a custom comm link.
+
+1. Go to "Application Settings" > "Comm Links"
+       https://docs.qgroundcontrol.com/master/en/SettingsView/SettingsView.html
+2. add new one matching your settings above.
```
hamishwillee, line 149: You mean these ones? 
```
2. Factory Settings -- Example 2:

    - SSID: IFFRC_xxxxxxxx
    - Password: 12345678
    - IP: 192.168.4.1
    - Port: 6789 (TCP) 
 ```

**Comment by hamishwillee** [comment]
Thanks @teyrana ! Looks useful; especially the note that the setup can be different and to check voltage levels.

@bkueng I will do some minor editorial updates. Any comments "in general"?

File: en/telemetry/esp8266_wifi_module.md (hunk 2 above)
Comment (teyrana) [reply en/telemetry/esp8266_wifi_module.md:149]: well, it might be these ones, it might be something different.... they're not really consistent across products :/ 

I ended up ordering both of the linked boards from banggood, before I figured out how it work.... and even those have a different back text.   Not just different SSID names, but they don't even have the same fields. :/ 

Honestly, it would hurt to add a "advanced users" tag to this page ....

**Comment by bkueng** [comment]
> @bkueng I will do some minor editorial updates. Any comments "in general"?

No, looks good

**Comment by mrpollo** [comment]
@hamishwillee what's blocking the merge here?

**Comment by hamishwillee** [comment]
Mostly the fact I can only work on PX4 2 days per week. I will have a last review now.

File: en/telemetry/esp8266_wifi_module.md (hunk 2 above)
Comment (hamishwillee) [reply en/telemetry/esp8266_wifi_module.md:149]: Thank you for the clarification. The important point is the one you already made - "check the box" :-)

File: en/telemetry/esp8266_wifi_module.md [hunk 3]
```diff
@@ -112,6 +136,22 @@ On your wifi-enabled *QGroundControl* ground station computer/tablet, find and c
 QGC automatically starts its UDP link on boot.
 Once your computer/tablet is connected to the **PixRacer** WiFi Access Point, it will automatically make the connection.
 
+## Configure QGC with non-standard Wifi connections
+
+QGC supports a variety of other wifi connections, if manually configured: 
+
+https://docs.qgroundcontrol.com/master/en/SettingsView/SettingsView.html
+
+If QGC doesn't automatically connect to your vehicle, add it's network connection as a custom comm link.
```
teyrana, line 145: "it's" should be "its".

hamishwillee:
OK, I have given this an extensive subedit. 

The firmware flashing stuff is advanced/not-necessary by default (as I understand it from @teyrana comments). So that is all pushed down to a subsection at the end. 

The setup is now pretty straightforward.

@teyrana I am going to merge this, because it is much better. However I would really appreciate if you could, in a separate PR, add a screenshot of the settings for adding the QGC connection for a non-pixracer module. Would this be possible?

Your general review on my changes also would be appreciated!

en/telemetry/esp8266_wifi_module.md
hamishwillee: Yes, I saw and fixed that :-)

hamishwillee:
@bkueng Would also appreciate it if you could do a final scan of the page http://docs.px4.io/master/en/telemetry/esp8266_wifi_module.html

Assuming that the firmware update is not always necessary then this is much easier to use now.

bkueng:
> and usually have the firmware already installed and configured to useable defaults.

The last ones I've used, and bought separately from a pixracer, I had to flash them. It's been a while though, so it might have changed.

hamishwillee:
Thanks @bkueng. Bugger :-(. @teyrana Was that your experience too?

teyrana:
@hamishwillee @bkueng 
I think it's important to note _which_ board we're talking about.  The banggood ones I have, I did not have to flash. (which is just as well, because they don't have any documentation)

The pixracer ones (which I have no experience with-- because of the supply voltage thing) probably do.  I mean, I truest bkueng here.

The challenge is how to write general documentation that covers both cases. :/  

Would y'all prefer if we have a sub-page for each board?   That might make more sense.

teyrana:
p.s. For sure, not having to flash the firmware is a step forward... but the lack of manufacturer support for my esp8266 board support for my fmu (pixhawk 4) meant I spent several days a false start finding the right cabling and supported boards to even _connect_ these things to my fmu... let alone get 'em talking.

teyrana:
p.s. I might be doing all this wrong.  I'm not _trying_ to turn this into a side project, just trying to jury-rig the simplest working solution I can.

hamishwillee:
@teyrana I'm going to be back on this Tues/Weds. My intent is to change the wording from "you should never have to do this flashing" to you "may have to do this flashing - the instructions with the device should say" or similar. I don't think we need to split out the boards into separate sections.

Thanks a lot for your help here. This is a popular component.

teyrana:
@hamishwillee 
sounds good :) 
I'm not in any hurry.
I agree, splitting it into separate categories for different kinds of boards should make it more clear.

p.s. I verified the settings for both of those banggood boards
p.p.s.  I have some screenshots ready to go, whenever we decide where they should live...

PR #1176, opened 2021-04-06
X500 update

vincentpoont2:
Update Holybro X500 Kit guide


hamishwillee:
@vincentpoont2 This is excellent. Two points:
1. Several of the images are 7Mb+ - do they really need to be that large? Historically we've had complaints, particularly from China for users on slow links having to download such large images. I usually convert to jpg and drop width to something useful on normal screens. Of course there are exceptions.
2. The Tuning section could use an update to point to the new tuning doc: http://docs.px4.io/master/en/config_mc/pid_tuning_guide_multicopter_basic.html

It would be great to include a quality video showing
- Building this
- Flying this

Do you have anything like this? I could use some "open" alternatives like https://www.youtube.com/watch?v=iq_4B1i6mXk or https://www.youtube.com/watch?v=HYd0_Ii9-60 but I figure you guys might have some "product quality" stuff for sales?

mrpollo:
@hamishwillee we have a video incoming that is going to be hosted on the px4 autopilot channel in the next couple of days/weeks

mrpollo:
@hamishwillee @vincentpoont2 I took care of the image sizes, I think we should look into an image squash/minified bot for the docs to do this for us, but for now, I'm more than happy to help. I made sure the largest in size was 600kb, and at max 2048px wide, I can even go lower, but since I didn't really build this locally, I wasn't sure if the size would be an issue.

Screen Shot 2021-04-07 at 9 41 53 AM

hamishwillee:
@mrpollo Thanks for your help - I'll do a minor post process to tidy up a few things shortly.
- I'm a bit nervous about automated compression, but this looks good. Minimally an automated check on file sizes would be great so there is an obvious indicator that files could benefit from this (rather than me checking). Alternatively, if you can find a bot that does it we can trial :-)
- Can you remember/remind me about the video?

PR #1178, opened 2021-04-07
Sidebar grouping folders are not page links

hamishwillee:
Previously the folders like "Getting Started" shown below were clickable to open a grouping page. However there is a bug in Vuepress that grouping pages opened by search do not open the sidebar - so you get the page, but lose the sidebar context of where you are.

As a workaround, the parents in the sidebar tree are no longer pages. Instead the page is displayed immediately underneath as an overview.

image

### DIFF
--- a/.vuepress/get_sidebar.js
+++ b/.vuepress/get_sidebar.js
@@ -101,9 +101,12 @@ module.exports = {
                           //The parent of this is a new child group. So we need to create it as such and make it a parent.
                           let new_parent = {};
                           new_parent.title = last_item.title;
-                          new_parent.path = last_item.path;
+                          new_parent.path = ''; // last_item.path;
                           //new_parent.sidebarDepth = 1;
                           new_parent.children = [];
+                          //Add first child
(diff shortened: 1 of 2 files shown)


.vuepress/get_sidebar.js
```diff
@@ -101,64 +100,68 @@ module.exports = {
                           //The parent of this is a new child group. So we need to create it as such and make it a parent.
                           let new_parent = {};
                           new_parent.title = last_item.title;
-                          new_parent.path = last_item.path;
+                          // new_parent.path = last_item.path;
```
hamishwillee, line 103: @bys1123 FYI, this is the relevant change to the sidebar. To revert you uncomment this line and comment the one below it so code looks like this. 
```
 new_parent.title = last_item.title;
new_parent.path = last_item.path;
```
Then you remove the three new lines below:
```
//Add first child
let entry = new Array(last_item.path, 'Introduction'); //last_item.title+
new_parent.children.push(entry);
```

PR #1181, opened 2021-04-07
Filter tuning - create standalone doc

hamishwillee:
@bkueng @bresch As discussed in slack, it seems a good idea to pull the Filter tuning doc out of Racer setup because it is more generally useful than just for racers..

I have also pulled out the control latency section - actually duplicated it. I think it is useful in both places. Happy to discuss/disagree. 

Assumption for you to confirm:
- This isn't just for MC. I have put it at a higher level. Is this the right place in the hierarchy? Is it MC only? Is it something very few people would do?
- Are there things that aren't mentioned that should be done before you even come to this? E.g. should be mention what tuning should be done first, and if so, what?
- Any other preconditions?


en/config_mc/racer_setup.md
```diff
@@ -112,93 +114,40 @@ Make sure to calibrate the ESCs before doing any tuning.
...
-The plot cannot be compared between different vehicles, as the y axis scale can be different.
-On the same vehicle it is consistent and independent of the flight duration though.
-:::
+Filters trade off control latency and noise filtering, both of which impact performance.
+For information see: [Filter/Control Latency Tuning](../advanced_config/filter_tuning.md)
 
 ### PID Tuning (Second Round)
 
-Now do a second round of PID tuning, this time as tight as possible, and also tune the thrust curve.
+Now do a second round of PID tuning, this time as tight as possible, and also tuning the thrust curve.
```
hamishwillee, line 147: @bkueng Note, this now points to the advanced tuning guide for the thrust curve.

en/config_mc/racer_setup.md
```diff
@@ -112,93 +114,40 @@ Make sure to calibrate the ESCs before doing any tuning.
 
 At this point you should be ready for a first test flight.
 
-If it goes well, do a first pass of [PID tuning](../config_mc/pid_tuning_guide_multicopter.md) (ignore the thrust curve settings).
-The vehicle needs to be **undertuned**, meaning the **P** and **D** gains should be set too low - such that there are no oscillations from the controller that could be interpreted as noise (the default gains might be good enough).
-This is important for the [filter](#filters) tuning.
-There will be a second PID tuning round later.
+Assuming the vehicle is able to fly using the default settings, we then do a first pass of [Basic MC PID tuning](../config_mc/pid_tuning_guide_multicopter_basic.md).
```
hamishwillee, line 117: Note, now points to the basic tuning doc

en/advanced_config/filter_tuning.md
```diff
@@ -0,0 +1,80 @@
...
+  To avoid the IO delay, disable [SYS_USE_IO](../advanced_config/parameter_reference.md#SYS_USE_IO) and attach the motors to the AUX pins instead.
+- PWM output signal: enable One-Shot to reduce latency ([PWM_RATE](../advanced_config/parameter_reference.md#PWM_RATE)=0).
+
+Below we look at the impact of the low pass filters.
+
+## Filters
+
+This is the filtering pipeline for the controllers in PX4:
+- On-chip DLPF for the gyro sensor.
+  The cutoff frequency is set to 98Hz and it is sampled at 1kHz.
```
bkueng, line 30: This one is now disabled on all chips where it can be disabled.

en/advanced_config/filter_tuning.md
```diff
@@ -0,0 +1,80 @@
...
+However this is a trade-off as it will also increase the noise of the signal that is fed to the motors.
+Noise on the motors has the following consequences:
+- Motors and ESCs can get hot, to the point where they get damaged.
+- Reduced flight time because the motors continuously change their speed.
+- Visible random small twitches.
+
+The best filter settings depend on the vehicle.
+The defaults are set conservatively — such that they work on lower-quality setups as well.
+
+## Filter Tuning
```
bkueng, line 49: We could mention here as well that too high gains (particularly D) can show up as vibrations

bkueng (commented):
Thanks Hamish.
We should mention the notch filter as well (essentially if it is set narrow, it comes with a reduced latency compared to a low-pass filter, and is useful if the vehicle has a spike at a fixed frequency).
Then there's also dynamic notch filters based on real-time FFT of the gyro/motor feedback. I have no experience with that yet though.

> This isn't just for MC. I have put it at a higher level. Is this the right place in the hierarchy? Is it MC only? Is it something very few people would do?

It's most relevant for MC, as FW has lower control latency demands.
It's something that should be analyzed if you want a good setup, and in particular if you have problems, but I think it's not often done.

> Are there things that aren't mentioned that should be done before you even come to this? E.g. should be mention what tuning should be done first, and if so, what?

Same as was written before: basic tuning, preferably a bit undertuned.

en/advanced_config/filter_tuning.md
dagar: We should mention that the sampling and filtering is also performed at the full raw sensor rate. https://github.com/PX4/PX4-Autopilot/blob/58b40fbbb921ad306f97da040acd7f141496484c/src/modules/sensors/vehicle_angular_velocity/imu_gyro_parameters.c#L85-L105

dagar:
> Then there's also dynamic notch filters based on real-time FFT of the gyro/motor feedback. I have no experience with that yet though.

I would describe this as currently having the basic building blocks in place (FFT available and running, efficient sensor pipeline with banks of filters we can update, etc), but in need of another few iterations to actually get it working well in practice (better peak frequency tracking, latency considerations, etc). I'm hoping we can find time to explore this over the next release cycle with wider testing.

dagar:
Without getting too detailed it might be worth mentioning the cost (additional latency) of lowering the gyro cutoff (IMU_GYRO_CUTOFF).

```
Filter Group Delay
 Lowpass at 30 Hz  ~= 8 ms
 Lowpass at 60 Hz  ~= 3.8 ms 
 Lowpass at 120 Hz ~= 1.9 ms 
```

en/advanced_config/filter_tuning.md
```diff
@@ -0,0 +1,80 @@
...
+Even one millisecond difference in the latency can have a significant impact.
+:::
+
+The following factors affect control latency:
+- A soft airframe or soft vibration mounting increases latency (they act as a filter).
+- Low-pass filters in software and on the sensor chip trade off increased latency for improved noise filtering.
+- PX4 software internals: the sensor signals need to be read in the driver and then pass through the controller to the output driver.
+- The IO chip (MAIN pins) adds about 5.4 ms latency compared to using the AUX pins (this does not apply to a *Pixracer* or *Omnibus F4*, but does apply to a Pixhawk).
+  To avoid the IO delay, disable [SYS_USE_IO](../advanced_config/parameter_reference.md#SYS_USE_IO) and attach the motors to the AUX pins instead.
+- PWM output signal: enable One-Shot to reduce latency ([PWM_RATE](../advanced_config/parameter_reference.md#PWM_RATE)=0).
```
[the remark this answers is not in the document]
hamishwillee: @bkueng 
- I think you are saying that Dshot is better than oneshot. However doesn't DShot depend on Dshot capable hardware/ESCs? Further, isn't One shot still a viable option if you don't have DShot?
- The `PWM_RATE` parameter doesn't seem to exist??? Is this now [PWM_AUX_RATE_n](http://docs.px4.io/master/en/advanced_config/parameter_reference.html##PWM_AUX_RATE)?

So propose I changed this to:
```
PWM output signal: enable [Dshot](.../en/peripherals/dshot.md) or One-Shot ([PWM_RATE](../advanced_config/parameter_reference.md#PWM_AUX_RATE)=0) to reduce latency .
```
Where the dshot link goes to here: http://docs.px4.io/master/en/peripherals/dshot.html

OK?

en/advanced_config/filter_tuning.md
hamishwillee: I really don't mind if you make the changes directly rather than telling me. Reduces changes of me misunderstanding.

Anyway, I have updated this, please check. @dagar in particular, I have put the comment on sampling after the line below starting `It can be configured with the [IMU_GYRO_CUTOFF]` because I think you're only talking about the IMU_GYRO_CUTOFF

en/advanced_config/filter_tuning.md
hamishwillee: @bkueng Not sure best place for this, so I updated the warning in the 2nd paragraph below to include this as part of "don't use filter tuning to correct for vibration" Work for you?

image

en/advanced_config/filter_tuning.md
```diff
@@ -0,0 +1,94 @@
...
+  Sampling and filtering is always performed at the full raw sensor rate (commonly 8Hz).
+  :::
+- A separate low-pass filter on the D-term.
+  The D-term is most susceptible to noise while slightly increased latency does not negatively affect performance.
+  For this reason the D-term has a separately-configurable low-pass filter, [IMU_DGYRO_CUTOFF](../advanced_config/parameter_reference.md#IMU_DGYRO_CUTOFF).
+- A slewrate filter on the motor outputs ([MOT_SLEW_MAX](../advanced_config/parameter_reference.md#MOT_SLEW_MAX)).
+  Generally not used.
+
+To reduce the control latency, we want to increase the cutoff frequency for the low-pass filters.
+The effect on latency of increasing `IMU_GYRO_CUTOFF` is approximated below.
```
hamishwillee, line 43: FYI @dagar I added your info from https://github.com/PX4/PX4-user_guide/pull/1181#issuecomment-819539718 below.

en/config_mc/filter_tuning.md
```diff
@@ -0,0 +1,98 @@
+# MC Filter Tuning & Control Latency
+
+Filters can be used to trade off [control latency](#control-latency), which affects flight performance, and noise filtering, which impacts both flight performance and motor health.
+
+This topic provides an overview of control latency and PX4 filter tuning.
+
+:::note
```
hamishwillee, line 7: @bkueng So re your comment in https://github.com/PX4/PX4-user_guide/pull/1181#pullrequestreview-635322101

1. As this is more or less MC specific I added this note to repeat the "do basic PID tuning but undertune" message. I also moved the whole doc from advanced config to MC config. 

The notch filters sounds interesting. Where do these fit in the pipeline - ie are they "use this instead of IMU_GYRO_CUTOFF" or might you use all of them?
My very old memory is that a notch filter allows you to filter out a specified band of frequencies, so I guess you are saying that if you do a frequency analysis and see that all your noise is at one frequency you use a notch filter to get rid of that. If it is spread then you might just use the normal IMU_GYRO_CUTOFF.

Anyway, what I think we need here is for **you** to add the information about tuning for the new parameters, including an FFT, list of the parameters affected. Otherwise this will just be a second class citizen in the doc. Can ou do that?

With respect to the dynamic notch filter it sounds premature to fully document it. So perhaps we should just have a note that it exists but you should not yet use it - i.e. don't touch parameters X, Y, Z. 

Make sense?

My thinking is that we would merge this, and you could add update for notch "in the coming month"

en/config_mc/filter_tuning.md
```diff
@@ -0,0 +1,98 @@
...
+
+:::tip
+Lowering latency allows you to increase the rate **P** gains, which results in better flight performance.
+Even one millisecond difference in the latency can have a significant impact.
+:::
+
+The following factors affect control latency:
+- A soft airframe or soft vibration mounting increases latency (they act as a filter).
+- Low-pass filters in software and on the sensor chip trade off increased latency for improved noise filtering.
+- PX4 software internals: the sensor signals need to be read in the driver and then pass through the controller to the output driver.
```
bkueng, line 24: Here we can also mention the update rate, configured with `IMU_GYRO_RATEMAX`

en/config_mc/filter_tuning.md
```diff
@@ -0,0 +1,98 @@
...
+  To avoid the IO delay, disable [SYS_USE_IO](../advanced_config/parameter_reference.md#SYS_USE_IO) and attach the motors to the AUX pins instead.
+- PWM output signal: enable One-Shot to reduce latency ([PWM_RATE](../advanced_config/parameter_reference.md#PWM_RATE)=0).
+
+Below we look at the impact of the low pass filters.
+
+## Filters
+
+This is the filtering pipeline for the controllers in PX4:
+- On-chip DLPF for the gyro sensor.
+  This is disabled on all chips where it can be disabled (if not, the cutoff frequency is set to 98Hz and it is sampled at 1kHz).
```
bkueng, line 35:
```suggestion
  This is disabled on all chips where it can be disabled (if not, the cutoff frequency is set to the highest level of the chip).
```

en/config_mc/filter_tuning.md
```diff
@@ -0,0 +1,98 @@
...
+- Low-pass filter on the gyro sensor data.
+  It can be configured with the [IMU_GYRO_CUTOFF](../advanced_config/parameter_reference.md#IMU_GYRO_CUTOFF) parameter.
+  :::note
+  Sampling and filtering is always performed at the full raw sensor rate (commonly 8Hz).
```
bkueng, line 39:
```suggestion
  Sampling and filtering is always performed at the full raw sensor rate (commonly 8kHz, depending on the IMU).
```

en/config_mc/filter_tuning.md
```diff
@@ -0,0 +1,98 @@
...
+## Filter Tuning
+
+First make sure to have the high-rate logging profile activated ([SDLOG_PROFILE](../advanced_config/parameter_reference.md#SDLOG_PROFILE) parameter).
+[Flight Review](../getting_started/flight_reporting.md) will then show an FFT plot for the roll, pitch and yaw controls.
+
+:::warning
+Do not try to fix a vehicle that suffers from high vibrations with filter tuning!
+Instead:
+- Confirm that PID gains, in particular D, are not set too high as this can show up as vibrations.
+- Fix the vehicle hardware setup.
```
bkueng, line 74: I'd separate like this:
```suggestion
- Do not try to fix a vehicle that suffers from high vibrations with filter tuning!
Instead fix the vehicle hardware setup.
- Confirm that PID gains, in particular D, are not set too high as this can show up as vibrations.
```

en/config_mc/filter_tuning.md
bkueng: You're correct. A notch can be used in addition: you would set the general low-pass quite high (e.g. 120), and then add a lower notch at the rotor blade pass frequency, or your peak, e.g. 80.
Params are IMU_GYRO_NF_BW and IMU_GYRO_NF_FREQ.

This is an example:
bokeh_plot
You would set the freq at 32, and bandwidth to 5 (it's often not quite that narrow).

Does that give you enough info?

> With respect to the dynamic notch filter it sounds premature to fully document it

I think so too.

bkueng (commented):
> Where the dshot link goes to here: http://docs.px4.io/master/en/peripherals/dshot.html
OK?

Yes. Apart from very old ones, probably all ESC's supporting oneshot support dshot as well.

`PWM_AUX_RATE` and `PWM_MAIN_RATE` now (not sure why there's `PWM_AUX_RATE1` as well, @dagar ?)

dagar:
> not sure why there's PWM_AUX_RATE1 as well, @dagar ?

Looks like it slipped into the yaml prematurely, but isn't used anywhere, and possibly will never be if we can expose the rate groups cleanly.

https://github.com/PX4/PX4-Autopilot/blob/3dad16bc40cce9ab9702c1a868ba8bd62b6664bb/src/lib/pwm/pwm_aux_params.yaml#L137-L147

en/advanced_config/filter_tuning.md
dagar: Thanks Hamish.

en/config_mc/filter_tuning.md
hamishwillee: @bkueng So does a higher rate reduce latency? Any reason not to use the highest rate available? / set to 0 and use the sensor default rate?

en/config_mc/filter_tuning.md
hamishwillee: > Does that give you enough info?

@bkueng Nearly. I've taken a first shot at this, but I expect it to need work.

The open questions are:
1.  **when** you do the notch filter 
    - my guess is that you always start by reviewing the plot from with default filter settings. If you saw a spike like this one the FIRST thing you would do is set the notch filter. I.e you wouldn't do the other tuning until any notch was removed right?

2. if you use a notch filter how you determine the Low pass cutoff. 
   - my guess is that for determining the low pass cuttoff you use the **same** process that you used if there was no notch (see next question). The point of the notch is to put you in a state where you don't have to put the low pass filter before that noise spike - after you have it the signal is "as though you never had a spike".

3.  irrespective of whether you use a notch filter or not, at what cuttoff do you start with for IMU_GYRO_CUTOFF (for the d term you said 30).
     - you show plots for IMU_GYRO_CUTOFF of 40, 70, 90, where the mean for the plot is 40, and the signal appears to trail off into noise at about 70. The analysis showed that above 70 noise increases.
     - So my assumption might be that a good stating point for setting the IMU_GYRO_CUTOFF  will be somewhere between the mean and where you see your main frequencies trailing off. Is there a "rule of thumb?

> A notch can be used in addition: you would set the general low-pass quite high (e.g. 120), and then add a lower notch at the rotor blade pass frequency, or your peak, e.g. 80.

Are you saying that the general LP cuttoff might be set at either at 120 or 80? Or are you suggesting IMU_GYRO_CUTOFF =120 and D term cutoff=80? 

Hope that all makes sense.

mcsauder: Hi @hamishwillee , in case it is useful for future questions as well as others in a general understanding, these notch filters pretty much mathematically follow what is called (as perhaps a misnomer), "RPM filtering" in betaflight: https://oscarliang.com/rpm-filter/

mcsauder: In a nutshell, (no PX4 pun, but it is fun anyway...), each airframe will have harmonic resonances which can be seen as yellow bands in the px4 flight review Acceleration PSD graph.  Among other resonance, filtering these out helps reject disturbances that the controller will act upon.  You probably already were aware of this and just clarifying, but to add gravity to what you are documenting I wanted to add a comment.

bkueng: > my guess is that you always start by reviewing the plot from with default filter settings

The defaults are set really low, so you'd increase it at first to see 'how bad it is'.

> I.e you wouldn't do the other tuning until any notch was removed right?

Right

> my guess is that for determining the low pass cuttoff you use the same process that you used if there was no notch (see next question). The point of the notch is to put you in a state where you don't have to put the low pass filter before that noise spike - after you have it the signal is "as though you never had a spike".

Exactly. You set it the next higher peak/where it generally goes up (there could be multiple notch filters, but there's currently only one to configure, as often there is one lower peak).

> So my assumption might be that a good stating point for setting the IMU_GYRO_CUTOFF will be somewhere between the mean and where you see your main frequencies trailing off. Is there a "rule of thumb?

Personally I just start quite high (around 100) and see how it goes (with some experience you can hear it, but also check if the motors get hot), then go either further up or down. You can do the same based on plots.

> Are you saying that the general LP cuttoff might be set at either at 120 or 80? Or are you suggesting IMU_GYRO_CUTOFF =120 and D term cutoff=80?

Here I meant IMU_GYRO_CUTOFF =120 and IMU_GYRO_NF_FREQ = 80, ignoring the D filter for that example.
Might be worth mentioning that the d filter should not be too far off from IMU_GYRO_CUTOFF, as an increased difference in delay starts to have negative impacts at some point. You need to go quite far for that though, e.g. d=15, gyro=80.

> In a nutshell, (no PX4 pun, but it is fun anyway...)

:D

en/config_mc/filter_tuning.md
bkueng: Yes latency. It is really CPU intense, you quickly hit 100%. 2khz is pretty much the limit, on h7 possibly 4khz.

dagar: The danger is if you set the rate too high it will starve things like mavlink and you won't be able to communicate with the board. This is why the param metadata max is 2 kHz (but you can override it).

As a precaution you could mention something simple like 4 kHz+ only recommended for stm32h7 or newer.

hamishwillee: Thanks. I've added this note.

en/config_mc/filter_tuning.md
hamishwillee: > Here I meant IMU_GYRO_CUTOFF =120 and IMU_GYRO_NF_FREQ = 80, ignoring the D filter for that example.

@bkueng That makes no sense. I can see why you might set the IMU_GYRO_CUTOFF =120 (because the noise is still low) but the spike is at 32, so why would you set `IMU_GYRO_NF_FREQ = 80`? Or were you talking about some other hypothetical example where the spike is at 80?

image

hamishwillee: @bkueng Just to be absolutely clear, are the tuning of the notch and IMU_GYRO_CUTOFF from the logs independent? 

I.e. if I do the first flights to tune IMU_GYRO_CUTOFF without tuning the notch I will end up with a particular value for cuttoff based at the point where the noise after 60Hz starts to increase. If I were to run the test once, see I have a low frequency spike and set the notch, THEN run the flights to tune IMU_GYRO_CUTOFF will I get a different value for IMU_GYRO_CUTOFF?

If they are independent then the instructions as they are now are OK. However if you have to do the notch filter first (as I imply from your previous answers) then I will need to reorder the instructions.

----

Further, you describe an experienced approach, which is "set the IMU_GYRO_CUTOFF high (e.g. 100Hz) to assess if there is a problem, which you can hear or detect from the motors getting hot. Then increase/decrease values until you no longer hear the whine or feel excessive heat". Presumably if you set at 1kHz and the motors are still cool/no odd sound then no tuning is needed :-) 
- About right?
- Is there a value for the cuttoff at which you might say "good enough/no point tuning for latency". In other words, if you ran a test at 100Hz and you saw that the motors were cool and there was no bad sound, would you just stop? Or do you always tune until there is a problem?
- How can you tell from this if you need a notch filter?

The existing process is pretty much dumb/methodical - you iterate up your cuttoff from default in 10Hz increments and look at the plots at the end. This is good because no expertise is required. Is there a benefit for these users in doing an initial high test and checking logs in this case? Or is the process pretty OK?

Essentially I'm trying to work out if there is any benefit for the broad base of users in starting from a higher test value for the cuttoff, and what rule of thumb would we use for that. The current process involves quite a few tests, but it is pretty straightforward, and you walk away from it with all the information needed to do your tuning.

hamishwillee: @mcsauder Thanks for commenting - that is certainly useful for me to know.

So here I refer to the low frequency noise spike which we see in the FFTs. The example given is due to rotor blade wash. I infer from your comment that  if I was to look at a corresponding power spectral density plot I might see a yellow band at this frequency, similar to those show in the [Flight Review analysis doc here](https://docs.px4.io/master/en/log/flight_review.html#acceleration-power-spectral-density). Is that about right?

If so, I was kind aware of this. Do you think perhaps we should add a note somewhere also pointing to the Flight review analysis doc? Or is there more you think I could/should say?

In particular, are there other sources of harmonics I might mention?

bkueng: > Or were you talking about some other hypothetical example where the spike is at 80?

Yes unrelated to the example plot.

> I.e. if I do the first flights to tune IMU_GYRO_CUTOFF without tuning the notch I will end up with a particular value for cuttoff based at the point where the noise after 60Hz starts to increase. If I were to run the test once, see I have a low frequency spike and set the notch, THEN run the flights to tune IMU_GYRO_CUTOFF will I get a different value for IMU_GYRO_CUTOFF?

They are independent.

> Presumably if you set at 1kHz and the motors are still cool/no odd sound then no tuning is needed :-)

Indeed, but I'm not aware of any vehicle requiring no filtering at all. Would be nice though, let me know if you find one :)

> Is there a value for the cuttoff at which you might say "good enough/no point tuning for latency". In other words, if you ran a test at 100Hz and you saw that the motors were cool and there was no bad sound, would you just stop? Or do you always tune until there is a problem?

It depends on the vehicle size and expectations. I.e. FPV racers go as high as they can, but for me personally, flying line-of-sight, ~120 is good enough (and never needed a notch), while having the D a bit lower (50-80).
For bigger vehicles you probably don't have to go beyond 80 (as latency is less critical).

> How can you tell from this if you need a notch filter?

For that you really do need to look at the plots.

> Is there a benefit for these users in doing an initial high test and checking logs in this case?

You can start high as well, just need to be aware of the risks. I'd say don't fly for more than 20-30sec. And if the vibrations getting through are really exceedingly high, there's a risk of crashing the vehicle as it becomes unflyable.
Eventually I hope we can simplify this with using dynamic (notch) filters.

hamishwillee: Thanks @bkueng 

Awesome. I've added the additional info mostly as a "tips" section. Can you please do a final sanity check. If it is OK I will merge tomorrow and we can deal with any further comments as a post process.

bkueng (approved):

hamishwillee:
Thanks @bkueng @dagar . Bit of a trudge, but I think this is pretty helpful.

PR #1184, opened 2021-04-08
Add geometry file docs

hamishwillee:
This is just a placeholder with a lot of questions. Has video and useful links. The need was originally raised in https://github.com/PX4/PX4-Devguide/issues/349

@bkueng @bresch Can you help here? Some questions inline, but really I need a practical guide to how to do this.

### DIFF
--- a/en/concept/geometry_files.md
+++ b/en/concept/geometry_files.md
@@ -1,80 +1,100 @@
-# Geometry Files
+# Multicopter Geometry Files
 
-- Associated dev guide issue: https://github.com/PX4/PX4-Devguide/issues/349
-- Geometry files are found here: https://github.com/PX4/PX4-Autopilot/tree/master/src/lib/mixer/MultirotorMixer/geometries
-- They are TOML files
+Geometry files are used by PX4 to generate mixer definitions that map the outputs of PX4 rate controllers to specific motors.
+The files describe the positions, direction of thrust, rotation direction, thrust and drag coefficients of each of the rotors.
 
-@[youtube](https://youtu.be/orvng_11ngQ?t=1080) - Control Allocation - 18 minute in.
+## How to add a New Geometry
 
-[Slide 14 has info](https://static.sched.com/hosted_files/px4developersummitvirtual2020/1b/PX4%20Developer%20Summit%202020%20-%20Overview%20of%20multicopter%20control%20from%20sensors%20to%20motors.pdf)
-
-<!-- stuff just copied from the slide
-In a nutshell
-Desired forces
-and torques (m)
-?
-Actuator
-outputs (u)
-Control allocation
-matrix
-Actuator
-effectiveness
-matrix (B)
-Inverse
-Control
-allocation
-Matrix (P=B
-+
-Moore-Penrose )
--->
-
-
-## How to add a new geometry
-
-1. Create new TOML geometry file in [/src/lib/mixer/MultirotorMixer/geometries](https://github.com/PX4/PX4-Autopilot/tree/master/src/lib/mixer/MultirotorMixer/geometries).
-   For example "foo.toml".
-   with a new key (e.g.: key = "4fo") and add to **CMakeLists.txt** <!-- which one? -->
-
-2. In ROMFS/px4fmu_common/Create new mixer file **mixers/foo.main.mix** with a line containing the new key:
+1. Create new TOML geometry file (e.g. "foo.toml") in [/src/lib/mixer/MultirotorMixer/geometries](https://github.com/PX4/PX4-Autopilot/tree/master/src/lib/mixer/MultirotorMixer/geometries).
+   The file must include a new **key** (e.g.: `key = "4fo"`).
+   See [Geometry File Format](#geometry-file-format) for information on the required fields.
+1. Add the geometry file to [/src/lib/mixer/MultirotorMixer/CMakeLists.txt](https://github.com/PX4/PX4-Autopilot/blob/master/src/lib/mixer/MultirotorMixer/CMakeLists.txt).
+1. Create a new [mixer file](../concept/mixing.md) that uses the new geometry. For example you might create **ROMFS/px4fmu_common/mixers/foo.main.mix** with a line containing the new key (in this case `4fo`):
    ```
-   R 4fo 10000 10000 10000 0  
+   R 4fo
    ```
-   <!-- There was a comment that the numbers are not needed? Why are they needed in quad then -->
-3. Set the new mixer in **init.d/airframes/myconfig**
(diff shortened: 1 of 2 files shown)


en/concept/geometry_files.md
```diff
@@ -0,0 +1,80 @@
...
+
+## How to add a new geometry
+
+1. Create new TOML geometry file in [/src/lib/mixer/MultirotorMixer/geometries](https://github.com/PX4/PX4-Autopilot/tree/master/src/lib/mixer/MultirotorMixer/geometries).
+   For example "foo.toml".
+   with a new key (e.g.: key = "4fo") and add to **CMakeLists.txt** <!-- which one? -->
+
+2. In ROMFS/px4fmu_common/Create new mixer file **mixers/foo.main.mix** with a line containing the new key:
+   ```
+   R 4fo 10000 10000 10000 0
```
hamishwillee, line 40: Presentation indicates that the numbers are just legacy and that everything needed is compiled from TOML . Is that correct?
So what if they are omitted? Or are all values ignored?

en/concept/geometry_files.md
```diff
@@ -0,0 +1,80 @@
...
++
+Moore-Penrose )
+-->
+
+
+## How to add a new geometry
+
+1. Create new TOML geometry file in [/src/lib/mixer/MultirotorMixer/geometries](https://github.com/PX4/PX4-Autopilot/tree/master/src/lib/mixer/MultirotorMixer/geometries).
+   For example "foo.toml".
+   with a new key (e.g.: key = "4fo") and add to **CMakeLists.txt** <!-- which one? -->
```
hamishwillee, line 36: Which cmakelist.txt file?

en/concept/geometry_files.md
```diff
@@ -0,0 +1,80 @@
...
+## Fragment from some files
+
+Looking at the toml files they all have this first. So I am assuming file description follwed by info. 
+- Is key random?
+- Where is the description used?
+
+```
+# Generic Quadcopter in H configuration
+
+[info]
```
hamishwillee, line 58: What are the things that can be in each section? Just these listed? What are axis, ct, cm, position, direction, and how would YOU go about calculating them?

en/concept/geometry_files.md
```diff
@@ -0,0 +1,80 @@
...
+3. Set the new mixer in **init.d/airframes/myconfig**
+   ```
+   set MIXER foo
+   ```
+   
+   
+## Fragment from some files
+
+Looking at the toml files they all have this first. So I am assuming file description follwed by info. 
+- Is key random?
```
hamishwillee, line 52: Where does the key come from?

en/concept/geometry_files.md
bresch: This one: https://github.com/PX4/PX4-Autopilot/blob/master/src/lib/mixer/MultirotorMixer/CMakeLists.txt

en/concept/geometry_files.md
bresch: Yes, it was for legacy compatibility and we now removed those, so you should just write `R: 4fo` now. The rest is ignored

bresch, line 40:
```suggestion
   R: 4fo
```

en/concept/geometry_files.md
bresch: The key is arbitrarily set by the developer in the `info` part of the geometry `toml` file (e.g.: https://github.com/PX4/PX4-Autopilot/blob/58b40fbbb921ad306f97da040acd7f141496484c/src/lib/mixer/MultirotorMixer/geometries/dodeca_bottom_cox.toml#L4). It's usually the number of rotors followed by one or a couple of letters to distinguish it from the others

en/concept/geometry_files.md
```diff
@@ -0,0 +1,80 @@
...
+- Where is the description used?
```
bresch, line 53: I don't think it's actually used, but mainly for the "reader" of this geometry file

en/concept/geometry_files.md
bresch: Those fields are allowed: https://github.com/PX4/PX4-Autopilot/blob/58b40fbbb921ad306f97da040acd7f141496484c/src/lib/mixer/MultirotorMixer/geometries/tools/px_generate_mixers.py#L103
- **name**: name of the rotor, used to make the geometry file more readable
- **position**: a vector describing the location of the rotor in the body frame (Forward-Right-Down) centered on the CG. Can be in meters or other units, only the ratio between the distances is really important since the mixer is normalized.
- **axis**: a vector in the in the body frame (Forward-Right-Down) describing the direction of the thrust produced by the rotor. For example `[1.0, 0.0, -1.0]` means that the rotor produces upward and forward thrust equally (i.e.: angle of 45 degrees).
- **direction**: specifies the direction of rotation of a rotor, `CW` (clockwise) or `CCW` (counter clockwise)
- **Ct**: non-dimensional thrust coefficient. For example, a rotor with a `Ct` of 2.0 produces 2 times the thrust of a rotor with a `Ct` of 1.0
- **Cm**: the non-dimensional drag torque coefficient relates to the axial torque produced by a spinning propeller. This needs to be set relative to `Ct` if some torque can be produced by drag and thrust together. If set to 0, the mixer will assume that the rotor does not produce any axial torque.

en/concept/geometry_files.md
```diff
@@ -0,0 +1,80 @@
...
+description = "Generic Quadcopter in H configuration"
+```
+Then we have rotor default which always has this structure
+```
+[rotor_default]
+axis      = [0.0, 0.0, -1.0]
+Ct        = 1.0
+Cm        = 0.05
+
+direction = "CW" - except I Only see THIS line on Tri
```
bresch, line 69: the `quad_x` mixer for example set `CW` in the default and only specifies two rotors as `CCW`. One could also no set the default and set `CW` or `CCW` on each rotor definition

en/concept/geometry_files.md
```diff
@@ -0,0 +1,80 @@
...
+Then we have one of these for each rotor - so 3 for a tri
+```
+[[rotors]]
+name      = "front_right_top"
+position  = [0.729, 0.684, 0.1]
+direction = "CW"
+```
+
+- How do I work out what each of these fields mean?
+- Are there geometries we can't capture? ie what if roto is on angle say?
```
bresch, line 80: we can specify the angle by setting a `axis` field for each rotor if needed, see https://github.com/PX4/PX4-user_guide/pull/1184/files#r613077240

bresch (commented):
Thanks for the draft, I tried to answer your questions, please tell me if something isn't clear enough.

hamishwillee:
@bresch That is exactly what I needed :-). I've updated with all this. Ready for your review.

bresch (approved):
Looks all good!

PR #1190, opened 2021-04-09
Adding a note to raise awareness for throttle input

max11gen:
It took me a very long time to figure out, why I couldn't use my summing mixers in disarmed state. I knew that throttle signals wouldn't work in disarmed state, but I had assumed that a summing mixer that gets multiple control inputs would still work in disarmed state for all signals that are NOT throttle. Since that is not the case and I assume others might have the same issue, I would like to raise awareness for it. I'm not sure, if this would be the right place to do so, but I decided to just give it a try.

Affects: http://docs.px4.io/master/en/concept/mixing.html

### DIFF
--- a/en/concept/mixing.md
+++ b/en/concept/mixing.md
@@ -284,7 +284,7 @@ The `S:` lines must be below the `O:` line.
 :::
 
 :::note
-Be aware that every Mixer containing an `S:`-line with `<group>=0` and `<index>=3` (throttle) won't work in disarmed state. E.g. a helicopter servo that has four inputs (roll, pitch, yaw and throttle) won't move in disarmed state even with roll/pitch/yaw signals.
+Any mixer output that has a throttle input (an `S:`-line with `<group>=0` and `<index>=3`) won't work in disarmed or prearmed state. For example, a servo that has four inputs (roll, pitch, yaw and throttle) won't move in disarmed state even with roll/pitch/yaw signals.
 :::
 
 The `<group>` value identifies the control group from which the scaler will read, and the `<index>` value an offset within that group.
--- a/en/concept/mixing.md
+++ b/en/concept/mixing.md
@@ -283,6 +283,10 @@ S: <group> <index> <-ve scale> <+ve scale> <offset> <lower limit> <upper limit>
 The `S:` lines must be below the `O:` line.
 :::
 
+:::note
+Be aware that every Mixer containing an `S:`-line with `<group>=0` and `<index>=3` (throttle) won't work in disarmed state. E.g. a helicopter servo that has four inputs (roll, pitch, yaw and throttle) won't move in disarmed state even with roll/pitch/yaw signals.
+:::
+
 The `<group>` value identifies the control group from which the scaler will read, and the `<index>` value an offset within that group.
 These values are specific to the device reading the mixer definition.
 


en/concept/mixing.md
```diff
@@ -283,6 +283,10 @@ S: <group> <index> <-ve scale> <+ve scale> <offset> <lower limit> <upper limit>
 The `S:` lines must be below the `O:` line.
 :::
 
+:::note
+Be aware that every Mixer containing an `S:`-line with `<group>=0` and `<index>=3` (throttle) won't work in disarmed state. E.g. a helicopter servo that has four inputs (roll, pitch, yaw and throttle) won't move in disarmed state even with roll/pitch/yaw signals.
```
hamishwillee, line 287: @max11gen Thanks very much.

My take on this is that this we do need to set the expectation of when control signals are emitted but that this is possibly not the right place for the note. 

Movement of any actuators is controlled by the arm OR prearm state as discussed here: http://docs.px4.io/master/en/advanced_config/prearm_arm_disarm.html.
What this says is that a motor or other "dangerous" control will never move when the vehicle isn't armed, but there is a state pre-arm that allows the actuators to move if they aren't considered dangerous.
So this "may" be incorrect - it may simply be that you're not prearmed.

@bkueng What is it that makes PX4 decide that something can move/not move in prearmed? Does it know that certain output groups are unsafe - so it might allow the yaw to move but not the throttle in prearmed state?

Further, the specific example is not quite in the right place (I "think") because your example uses a helicopter servo - but helicopter servos are defined in helicopter mixer section - so I would expect this comment there.

Anyway, depending on what @bkueng says, proposal is that we remove this and add a note at the [end of the introduction](http://docs.px4.io/master/en/concept/mixing.html).

bkueng: Prearm just sets the throttle input to NaN, so any output using that will be disabled.
So the note is correct.

en/concept/mixing.md
```diff
@@ -283,6 +283,10 @@ S: <group> <index> <-ve scale> <+ve scale> <offset> <lower limit> <upper limit>
 The `S:` lines must be below the `O:` line.
 :::
 
+:::note
```
hamishwillee, line 286: Thanks @bkueng! I've re-written this slightly. Still good.

en/concept/mixing.md
max11gen: @hamishwillee

> @max11gen Thanks very much.
> 
> My take on this is that this we do need to set the expectation of when control signals are emitted but that this is possibly not the right place for the note.

Feel free to move it somewhere else.

> Movement of any actuators is controlled by the arm OR prearm state as discussed here: http://docs.px4.io/master/en/advanced_config/prearm_arm_disarm.html.
> What this says is that a motor or other "dangerous" control will never move when the vehicle isn't armed, but there is a state pre-arm that allows the actuators to move if they aren't considered dangerous.
> So this "may" be incorrect - it may simply be that you're not prearmed.

I was in prearmed state. The thing is, that somewhere else in the docs, it is said that in prearmed state the servo rail is powered while ESCs are NOT powered. But this is not entirely true, because servos won't be powered if there is a throttle input to them, because then they are considered "dangerous". Since this information is all a little ambiguous, I think it can't hurt to mention this once again.

> @bkueng What is it that makes PX4 decide that something can move/not move in prearmed? Does it know that certain output groups are unsafe - so it might allow the yaw to move but not the throttle in prearmed state?
> 
> Further, the specific example is not quite in the right place (I "think") because your example uses a helicopter servo - but helicopter servos are defined in helicopter mixer section - so I would expect this comment there.

The helicopter mixer section is definitely NOT the right place to put this. This explicitely applies to Summing Mixers. The helicopter mixer is not versatile enough for my usecase and mechanics. Therefore I am forced to use a combination of summing mixers to replace the helicopter mixer. The helicopter mixer itself does not process a throttle input.

> Anyway, depending on what @bkueng says, proposal is that we remove this and add a note at the [end of the introduction](http://docs.px4.io/master/en/concept/mixing.html).
 Sure, sounds good. Just make sure that it does not seem like this is an issue appearing with the helicopter mixer.

hamishwillee: Thanks @max11gen . Beat says you were 100% right. I've done a minor subedit, mostly to remove the word helicopter. Now it should all be good and in the correct place.

en/concept/mixing.md
hamishwillee: @bkueng I've merged this because I think the modifications I made look fine. But happy to take late comment if you think I broke it.

hamishwillee:
Merged. Thanks @max11gen !

en/concept/mixing.md
bkueng: yup

PR #1194, opened 2021-04-14
ESP8266 -tidy flash info

hamishwillee:
Follows on from #1173 to make it clear that some modules MAY need to have firmware updates. Also deleted old PX4 v1.82 tip that is no longer relevant.

I'm going to merge this, and do any further updates as a follow on.

### DIFF
--- a/en/telemetry/esp8266_wifi_module.md
+++ b/en/telemetry/esp8266_wifi_module.md
@@ -25,8 +25,8 @@ Modules that accept 5.0V supply:
 ## Pixhawk/PX4 Setup & Configuration
 
 :::tip
-If using PX4 1.8.2 (and earlier) you should connect the ESP8266 to TELEM2 and configure the port by [setting the parameter](../advanced_config/parameters.md) `SYS_COMPANION` to 1921600 (remember to reboot after setting the parameter).
-The following instructions assume you are using PX4 versions after 1.8.2
+You *may* first need to update the ESP8266 module with recent/compatible firmware ([see below](#esp8266-flashing-firmware-advanced)).
+The manufacture instructions should explain if this is needed.
 :::
 
 Connect your ESP8266 to your Pixhawk-series flight controller (e.g. Pixracer) on any free UART.
@@ -39,6 +39,7 @@ Using *QGroundControl*:
   Remember to set the baud rate to 921600 in order to match the value set for the ESP8266.
 
 Once the firmware (port) is set up you can remove the physical USB connection between the ground station and the vehicle.
+
 
 ## Connect via ESP8266 to QGC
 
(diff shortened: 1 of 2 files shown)


hamishwillee:
@teyrana I

> p.p.s. I have some screenshots ready to go, whenever we decide where they should live...

Just attach them here and I can find a place.

en/telemetry/esp8266_wifi_module.md
```diff
@@ -92,30 +93,34 @@ If you have any problem connecting, see [QGC Installation/Configuration Problems
...
 ### Pre Built Binaries
 
 [MavLink ESP8266 Firmware V 1.2.2](http://www.grubba.com/mavesp8266/firmware-1.2.2.bin)
 
-### Updating the Firmware
+### Build From Sources
+
+The [firmware repository](https://github.com/dogmaphobic/mavesp8266) contains instructions and all the tools needed for building and flashing the firmware.
+
+### Updating the Firmware OTA
```
hamishwillee, line 109: @bkueng @teyrana my understanding is that this OTA both uploads and installs the new firmware - is that correct (i.e. if you have a recent firmware, you would always pick this option for an update). If so, then might reorg a little to make this more clear.

hamishwillee: (note, I am merging this anyway, so this would be a post process)

bkueng: I never tried that as I wasn't aware of it. Would make things easier though.

teyrana: I never flashed my radios as I didn't need to.  Both worked with a little configuration.

teyrana:
@hamishwillee 
front & back screenies of the bangood wifi radios.
--------
- https://drive.google.com/file/d/1YxXYUVycBMkD5MkuyN90wEuGc1Af8yaU/view?usp=sharing 
- https://drive.google.com/file/d/1Z1Ny1QUtWEY23WORui1UPLaQq6lX_6W0/view?usp=sharing 
- https://drive.google.com/file/d/1Z3ZMqEuZzpC-KE1LZo0zXf_15zCTH0ti/view?usp=sharing 
- https://drive.google.com/file/d/1Z6Mh28iDs9lzxUWnlY5HfZg1PGDgZZoO/view?usp=sharing

en/telemetry/esp8266_wifi_module.md
teyrana: Also, I think this phrasing is a bit ambiguous... this page is discussing how to flash firmware _for the telemetry_ radio, and in the middle of that, you discuss flashing firmware _for the FMU_.  

At very least, each mention of firmware explicitly mentioned which one you're referring to.

hamishwillee: @teyrana Thanks, good point, agreed. Done in https://github.com/PX4/PX4-user_guide/pull/1203

I also added the images as part of the note explaining that details may be written on the board.

PR #1204, opened 2021-04-15
First time user issue with Lightware Lidar

farhangnaderi:
I experienced that I2C interface is not possible out of the box with Pixhawk and you need to enable this option in Lightware Lidar software prior to being able to use I2C.


en/sensor/sfxx_lidar.md
```diff
@@ -40,6 +40,9 @@ Connect the Lidar the autopilot I2C port as shown below (in this case, for the [
 
 ![SF1XX LIDAR to I2C connection](../../assets/hardware/sensors/sf1xx_i2c.jpg)
 
+:::note
```
mrpollo, line 43: Can you please move this NOTE above the image, please? I think it flows better

en/sensor/sfxx_lidar.md
```diff
@@ -40,6 +40,9 @@ Connect the Lidar the autopilot I2C port as shown below (in this case, for the [
...
+In case it is the first time you are using the sensor with Pixhawk, you have to enable Pixhawk compatibility in [LightWare Studio](https://lightwarelidar.com/pages/lightware-studio).
```
mrpollo, line 44: What do you think if we update the message to something a bit more descriptive and helpful, can you tell our users where to find the ENABLE PIXHAWK setting inside the Lightware Studio application?

See my suggestion below:

> Please be aware that Lightware doesn't ship with Pixhawk compatibility enabled by default. To enable support, you have to download LightWare Studio and ENTER INSTRUCTIONS HERE

mrpollo (changes requested):
@farhangnaderi thanks for the addition, I added a few comments, which I think will make this super helpful for the community.

en/sensor/sfxx_lidar.md
hamishwillee: I hadn't heard this, and the doc has been around for a long time - so thanks for providing an update.

As Ramon says, this should have more detailed instructions than just "set it up using lightware". It might be worth having a separate section as a peer of the "Hardware" heading "Lidar Configuration".

hamishwillee:
Excellent. Thanks @farhangnaderi !

farhangnaderi:
> Excellent. Thanks @farhangnaderi !
> FYI I did minor subedit - it does not need to be a note :-)

Thanks Hamish.

PR #1158, opened 2021-03-30
Position mode stick

hamishwillee:
Fixes #1133

@MaEtUgR Because the docs are versioned, we document the current behaviour rather than the "change". Essentially this means changing the text where we say that we control movement over ground from speed to acceleration. This happens in three places:
- [Getting Started > Flight Modes](https://docs.px4.io/master/en/getting_started/flight_modes.html#position-mode-mc)
- [Flight Modes Overview page](https://docs.px4.io/master/en/flight_modes/)
  - [Position Mode (MC)](https://docs.px4.io/master/en/flight_modes/position_mc.html)

Mostly pretty straightforward, but some clarifications required both related to this change, and generally.

PR #1214, opened 2021-04-21
Position mode image - remove angle/rate control text

hamishwillee:
Fixes to Position mode MC doc following feedback from #1158

Note, this removes the ambiguous text from the diagram.
image


MaEtUgR (approved):
The icing on the cake 🍰

PR #1217, opened 2021-04-22
parameter_reference: updated esc telemetry

Igor-Misic:
This update to this PR: https://github.com/PX4/PX4-Autopilot/pull/16704


hamishwillee:
@Igor-Misic You are a star for taking care of this, but it isn't necessary.

One of the good features of our toolchain is that parameter docs are generated from the source changes.

So if you want different text, you put it direct in the source code.

PR #1220, opened 2021-04-23
parameters: update airframe config & param save behavior

bkueng:
From https://github.com/PX4/PX4-Autopilot/pull/16796


hamishwillee:
Thanks @bkueng - perfect.

PR #1221, opened 2021-04-24
Poor Man's Profiler: Add details

JacobCrabill:
The documentation on how to use the `profile` target / `poor-mans-profiler.sh` is rather incomplete for someone who hasn't tried this before.  Here's a stab at adding some additional details on how to get setup to do profiling.

@dagar / @pavel-kirienko - I think you two may know better than most how to use the profiler; let me know if my notes are incorrect.

### DIFF
--- a/en/debug/profiling.md
+++ b/en/debug/profiling.md
@@ -1,6 +1,6 @@
 # Poor Man's Sampling Profiler
 
-This section describes how to assess performance of the PX4 system by means of profiling.
+This section describes you can use the [Poor Man's Sampling Profiler](https://github.com/PX4/PX4-Autopilot/blob/master/platforms/nuttx/Debug/poor-mans-profiler.sh) (PMSP) shell script to assess the performance of PX4. 
 
 ## Approach
 
@@ -14,7 +14,11 @@ The folded stacks are then fed into the visualization script, for which purpose 
 
 ### Prerequisites
 
-The profiler relies on GDB to run PX4 on the embedded target.  So before profiling a target, you must have the hardware you wish to profile, and you must compile and upload the firmware to that hardware.  You will then need a JTAG debugger, such as the DroneCode Probe, to run the GDB server and interact with the board.  You may also need to modify the contents of `poor-mans-profiler.sh` to point to the debugger.  For example, with a Pixhawk 4 connected via USB, followed by a DroneCode Probe over USB, the following devices are enumerated in Ubuntu:
+The profiler relies on GDB to run PX4 on the embedded target.
+So before profiling a target, you must have the hardware you wish to profile, and you must compile and upload the firmware to that hardware.
+You will then need a JTAG debugger, such as the DroneCode Probe, to run the GDB server and interact with the board.
+You may also need to modify the contents of `poor-mans-profiler.sh` to point to the debugger.
+For example, with a Pixhawk 4 connected via USB, followed by a DroneCode Probe over USB, the following devices are enumerated in Ubuntu:
 
 ```
 usesr@ubuntu:~/PX4-Autopilot$ ls -alh /dev/serial/by-id/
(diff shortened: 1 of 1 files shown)


pavel-kirienko:
@JacobCrabill while you're at it, maybe you could replace that hard-coded `/dev/ttyACM0` with a variable that defaults to `/dev/serial/by-id/usb-*Black_Magic_Probe*-if02`? I could do it but I won't be able to test it anytime soon.

en/debug/profiling.md
```diff
@@ -12,6 +12,28 @@ The folded stacks are then fed into the visualization script, for which purpose
 
 ## Basic Usage
 
+### Prerequisites
+
+The profiler relies on GDB to run PX4 on the embedded target.
+So before profiling a target, you must have the hardware you wish to profile, and you must compile and upload the firmware to that hardware.
+You will then need a JTAG debugger, such as the DroneCode Probe, to run the GDB server and interact with the board.
+You may also need to modify the contents of `poor-mans-profiler.sh` to point to the debugger.
+For example, with a Pixhawk 4 connected via USB, followed by a DroneCode Probe over USB, the following devices are enumerated in Ubuntu:
```
hamishwillee, line 21: What does this mean "with a Pixhawk 4 connected via USB, followed by a DroneCode Probe over USB," - I think maybe "followed" is incorrect, because it implies a daily chain - Pixhawk 4 > USB > Dronecode probe, while you probably mean Pixhawk 4 debug port > Dronecode probe > Usb > laptop  (do you?). Maybe an image?

My thinking is that this section changes now you have removed hard coding for the port right? I think you can say that if you're using the Dronecode probe or black magic probe you can just accept the default. Otherwise then you can use the command (like you've shown below) to locate the port to use. 

Can you also link to http://docs.px4.io/master/en/debug/swd_debug.html#debug-probes because I'm not sure that everyone knows about jtag probes and how to connect them to PX4.

hamishwillee: PS This is definitely needed, so thanks for taking it on.

JacobCrabill: Good point.  I'll update the wording based on the changes I just made.

en/debug/profiling.md
```diff
@@ -1,6 +1,6 @@
 # Poor Man's Sampling Profiler
 
-This section describes how to assess performance of the PX4 system by means of profiling.
+This section describes you can use the [Poor Man's Sampling Profiler](https://github.com/PX4/PX4-Autopilot/blob/master/platforms/nuttx/Debug/poor-mans-profiler.sh) (PMSP) shell script to assess the performance of PX4.
```
pavel-kirienko, line 3:
```suggestion
This section describes how you can use the [Poor Man's Sampling Profiler](https://github.com/PX4/PX4-Autopilot/blob/master/platforms/nuttx/Debug/poor-mans-profiler.sh) (PMSP) shell script to assess the performance of PX4.
This is an implementation of a known method originally invented by [Mark Callaghan and Domas Mituzas](https://poormansprofiler.org/).
```

Add proper credits while we're at it.

en/debug/profiling.md
```diff
@@ -12,6 +13,40 @@ The folded stacks are then fed into the visualization script, for which purpose
...
+lrwxrwxrwx 1 root root  13 Apr 23 18:57 usb-Black_Sphere_Technologies_Black_Magic_Probe_BFCCB401-if00 -> ../../ttyACM1
+lrwxrwxrwx 1 root root  13 Apr 23 18:57 usb-Black_Sphere_Technologies_Black_Magic_Probe_BFCCB401-if02 -> ../../ttyACM2
+```
+
+In this case, the script would automatically pick up the device named `*Black_Magic_Probe*-if00`.
+But if you were using a different device you would be able discover the appropriate id from the listing above. 
+
+Then pass in the appropriate device using the `--gdbdev` argument like this:
+```bash
+./poor-mans-profiler.sh --elf=build/px4_fmu-v4_default/px4_fmu-v4_default.elf --nsamples=30000 --gdbdev=/dev/ttyACM2
```
hamishwillee, line 44: @JacobCrabill Just to confirm, is this correct format `--gdbdev=/dev/ttyACM2` ?

hamishwillee:
Thanks @JacobCrabill .
- This already had credits section at the end - since you added an acknowledgement up the top, I have removed that section from the end.
- You don't have to modify the script because you allow the target to be specified as an argument. I updated it to say that, but would you please confirm that I got the format right: https://github.com/PX4/PX4-user_guide/pull/1221#discussion_r622626650

Otherwise this is good to merge!

hamishwillee (approved):

en/debug/profiling.md
JacobCrabill: Yes that would be correct.  I didn't specifically mention that option here previously because most people will likely rely on the `.cmake`-defined `profile` build option, which calls the script and defines the command-line options for them.

hamishwillee: @JacobCrabill So would you/could you then normally update the profile build option, and if so, where/how?

JacobCrabill: The command you listed would work just fine.

PR #1227, opened 2021-04-27
Change ROS1 command to ROS2 command

max11gen:
I changed the `rostopic` command which only works in ROS1 to `ros2 topic` command. Also I provided a complete example, because in the markdown code here someone had already commented with "how?". Maybe you could also remove this comment.

### DIFF
--- a/en/ros/ros2_comm.md
+++ b/en/ros/ros2_comm.md
@@ -177,7 +177,7 @@ We can do this by running the bridge against PX4 running in the simulator.
    accelerometer_integral_dt: 4739
    ```
 
-You can also verify the rate of the message using `ros2 topic hz`. <!-- how? -->
+You can also verify the rate of the message using `ros2 topic hz`.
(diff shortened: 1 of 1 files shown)


TSC21 (approved):
Thanks! This splipped by.

en/ros/ros2_comm.md
```diff
@@ -177,8 +177,8 @@ We can do this by running the bridge against PX4 running in the simulator.
    accelerometer_integral_dt: 4739
    ```
 
-You can also verify the rate of the message using `rostopic hz`. <!-- how? -->
-For the case of `sensor_combined`:
+You can also verify the rate of the message using `ros2 topic hz`. <!-- how? -->
```
TSC21, line 180: Actually can you remove the `<!-- how? -->` now?

TSC21 (changes requested):

en/ros/ros2_comm.md
max11gen: Done

PR #1228, opened 2021-04-27
Fix typos and improve explanations for multi-ekf tuning

priseborough:


dagar:
Thanks, I didn't realize we had any multi-EKF documentation yet.

dagar (approved):

PR #1213, opened 2021-04-21
added svg for vectored 6 dof uuv airframe

lennartalff:
Followed the [user guide](https://docs.px4.io/master/en/dev_airframes/adding_a_new_frame.html#adding-a-new-airframe-group) to complete the "new" airframe group for the vectored 6 DOF UUV airframe.

Related PRs:
- PX4/PX4-Autopilot#17434
- mavlink/qgroundcontrol#9604

PR #1236, opened 2021-04-29
BlueROV2 basic docs

hamishwillee:
@DanielDuecker @lennartalff 
This follows on from our discussion about BlueROV2 in https://github.com/PX4/PX4-user_guide/pull/1213#issuecomment-825264023
Essentially it creates a UUV/Sub section with a main page that refers to BlueROV2 and the HippoCampus. The BlueROV2 gets its own page which is also linked from the "Complete Vehicles" page.

Can you check that what is there is correct please?

Further, I was wondering if you could answer some questions that would allow me to improve this:
1. Support is in master. Is it going to be in stable PX4 in v1.12 or the next versions?
1. How would you characterize the support level? 
   - Experimental? Really rock solid with lots of features? 
   - If it is very experimental, do you have a roadmap? 
   - Can you list things that people can vs cannot do (that they might expect)?
1. I have noted that you will have to set the firmware and airframe in QGC. What else can you set? What else do you **need** to set? I.e. compass calibration? 
1. What modes are supported? .e.g can you do missions? Position mode? Hold? etc. What do they mean in an underwater context. Will it just reject trying to set another mode?
1. I see it can be used with a gripper. Does this work with PX4? Is it just normal RC passthrough kind of setup?
1. BlueROV2 comes with ArduSub. What can you do with ArduSub that you can't do with PX4 yet. Is the plan to achieve parity?
1. Is there a plan to add a simulation mode in Gazebo? I like that because it allows testing :-)
1. The motor map presumably is the default wiring as suggested in BlueROV setup? (Worth us saying so if that is the case).

Anything else you think would make this more usable for someone? There really isn't any docs on what you can do with a sub on PX4 so being able to characterise what he modes do would be great.


DanielDuecker (approved):
looks good to me,
thanks Hamish!

hamishwillee:
Thanks @DanielDuecker - can you answer the questions in the description "Further, I was wondering ...". Those are the questions I'd like answered if I read this. Offline if you prefer?

DanielDuecker:
@hamishwillee pinged you on slack

PR #1237, opened 2021-04-29
Update instruction for fastrtpsgen on listener example

bastianhjaeger:
As discovered with @TSC21 this should be the correct cmd.


TSC21 (approved):
Thanks @bastianhjaeger!

PR #1207, opened 2021-04-19
simulation: the jMAVSim HITL baudrate has changed

julianoes:
Without this change I could not consistently connect to HITL.

PR #1238, opened 2021-04-29
HITL updates

julianoes:
The setup for Gazebo got a bit easier.

And this also reverts #1207 which didn't actually matter.


hamishwillee (approved):
Looks great thanks. I did think of merging the two first steps, but this good.

PR #1243, opened 2021-05-02
Update frame_selection.md

pxw816:
translated to Chinese


hamishwillee:
Thanks @pxw816 , but we need our translations to go via the crowdin translation service, because that keeps everything in Sync with changes to the English version.

There is some information on the process here: https://docs.px4.io/master/zh/contribute/translation.html

If you have any questions, can you please ask on PX4 slack #translations channel.

PR #1250, opened 2021-05-05
Update ros1.md

potato77:


hamishwillee:
Thanks @potato77 .

Great to have your contribution, but I can't accept this here because it would break our translation model. Essentially translations have to be entered in crowdin and then appear here automatically. 

Please join us on crowdin, as documented here: http://docs.px4.io/master/zh/contribute/translation.html

PR #1261, opened 2021-05-12
Parameter definitition docs updates

### LINKED ISSUE #1256: Parameter C++ API documentation
The documentation at https://docs.px4.io/master/en/advanced/parameters_and_configurations.html#c-api seems outdated. It doesn't match with the module template in the PX4 codebase.

Issues I can see:

- (minor) Is this really needed in the *cpp* file? It seems to be omitted in many modules.

        #include <uORB/topics/parameter_update.h>

- `(...) : ModuleParams(nullptr)` and why you should provide nullptr or something else is missing in the documentation.

- The following code snippet is hardly used in the PX4 codebase and seems no longer needed:

        int parameter_update_sub = orb_subscribe(ORB_ID(parameter_update));

- The code snippet for `Module::parameters_update` is different from `TemplateModule::parameters_update`, which is confusing.

- API docs (nor example code) shows how to use the parameter and how the system reacts (e.g. `.get()`, `.commit()`, ... methods)

- There are no docs on how to add your own parameters

I propose someone who understands the parameter system in and out updates the documentation. I'd be happy to review from my perspective.

hamishwillee:
This attempts to fix #1256.  @bkueng it would be better if you made any needed updates directly. I can see that there will be some around subscription API but I don't want to inject crap.

Further, this is still missing 

1. Info on how to add yaml to the build system
2. Re "how to use the parameter and how the system reacts (e.g. .get(), .commit(), ... methods)" you said best to look at the code: https://github.com/PX4/PX4-Autopilot/blob/master/platforms/common/include/px4_platform_common/param.h#L129 - worth adding that to further reading?
3. How to get the metadata into QGC

### DIFF
--- a/en/advanced/parameters_and_configurations.md
+++ b/en/advanced/parameters_and_configurations.md
@@ -250,8 +250,8 @@ It supports all the same metadata, along with new features like multi-instance d
 - An example of YAML definitions being used can be found in the MAVLink parameter definitions: [/src/modules/mavlink/module.yaml](https://github.com/PX4/PX4-Autopilot/blob/master/src/modules/mavlink/module.yaml).
 - A YAML file is registered in the cmake build system by adding
   ```
-	MODULE_CONFIG
-		module.yaml
+  MODULE_CONFIG
+  	module.yaml
   ```
   to the `px4_add_module` section of the `CMakeLists.txt` file of that module.
 
@@ -331,8 +331,8 @@ The purpose of each line is given below (for more detail see [module_schema.yaml
 ```
 
 
-
 ## Further Information
 
 - [Finding/Updating Parameters](../advanced_config/parameters.md)
 - [Parameter Reference](../advanced_config/parameter_reference.md)
+- [Param implementation](https://github.com/PX4/PX4-Autopilot/blob/master/platforms/common/include/px4_platform_common/param.h#L129) (information on `.get()`, `.commit()`, and other methods)
--- a/en/advanced/parameters_and_configurations.md
+++ b/en/advanced/parameters_and_configurations.md
@@ -117,10 +117,19 @@ You add some "boilerplate" code to regularly listen for changes in the [uORB Top
 Framework code then (invisibly) handles tracking uORB messages that affect your parameter attributes and keeping them in sync. 
 In the rest of the code you can just use the defined parameter attributes and they will always be up to date!
 
-First include **px4_platform_common/module_params.h** in the class header for your module or driver (to get the `DEFINE_PARAMETERS` macro):
-```cpp
-#include <px4_platform_common/module_params.h>
-```
+First include the required needed headers in the class header for your module or driver:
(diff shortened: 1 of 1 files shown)


en/advanced/parameters_and_configurations.md
```diff
@@ -164,7 +173,7 @@ orb_unsubscribe(parameter_update_sub);
 ```
 
 Call `parameters_update(parameter_update_sub);` periodically in code to check if there has been an update (this is boilerplate):
-```cpp
+```cpp
```
hamishwillee, line 176: Does anything need to change here? The application template looks a bit different:
```
void TemplateModule::parameters_update(bool force)
{
	// check for parameter updates
	if (_parameter_update_sub.updated() || force) {
		// clear update
		parameter_update_s update;
		_parameter_update_sub.copy(&update);

		// update parameters from storage
		updateParams();
	}
}
```
The app temp also has a "run" method that is not explained https://github.com/PX4/PX4-Autopilot/blob/master/src/templates/template_module/template_module.cpp#L138

hamishwillee:
Note, this is draft as it is not finished. @bkueng has agreed to look at it when he has bandwidth.

bkueng:
I did the updates, let me know if there's further questions.

> Re "how to use the parameter and how the system reacts (e.g. .get(), .commit(), ... methods)" you said best to look at the code: https://github.com/PX4/PX4-Autopilot/blob/master/platforms/common/include/px4_platform_common/param.h#L129 - worth adding that to further reading?

We can add a link to that.

> How to get the metadata into QGC

I would just mention the comp info API. I know, this also needs docs. Want me to write something?

hamishwillee:
@bkueng Looks very clear thank you. I added the link to param implementation and merged

>> How to get the metadata into QGC

> I would just mention the comp info API. I know, this also needs docs. Want me to write something?

Yes please 

So either of us can update the  "generic" stuff in https://mavlink.io/en/services/component_information.html (but you will need to review since you are aware of changes). 

What I think that only you can do is create a doc that outlines the PX4 integration. i.e. what component information files PX4 supports, how you might modify their contents for your own params (or "whatever") and how to force them to be generated for use in firmware or simulator. 
Probably should be a peer of this doc (i.e. in same section).

bkueng:
Ok will do

pjdewitte:
Thanks for the swift action.

I ran though it, and noticed:
* The link to YAML metadata in the section "Creating/Defining Parameters" is a dead link
* The linked page Module Template includes the snippet `ModuleParams(nullptr)` but I also think it is worth including it in the parameter API docs. I would place the code snippet under "Update the cpp file with boilerplate to check for the uORB message related to parameter updates.".

Again, thanks for the effort you put in this!

hamishwillee:
@pjdewitte Thanks for checking. I've fixed the YAML link. Can you create a PR for other changes you want? I'm away until Wednesday and will otherwise forget this or it may become a lower priority.

PR #1268, opened 2021-05-15
X500 Fullly assembled photo

farhangnaderi:
What user guide mentions for putting the GPS on top (Which in my opinion is the most correct one) contradicts the final assembly photo that has GPS on payload holder. I have replaced it with a correct photo.


hamishwillee:
@farhangnaderi Thanks for this.

OK, so the [docs](http://docs.px4.io/master/en/frames_multicopter/holybro_x500_pixhawk4.html#holybro-x500-pixhawk4-build) don't say anything about where the mast should be mounted - they just say "Use the tape and stick the GPS to the top of the GPS mast and mount the GPS mast."  I think you're saying that it is better up the top than where they have put it? Is that because it is further from the electronics?

image

If so, cool. However this is a 7.6MB image - which is very big for this docset - usually we have images about 100-400KB. Could you shrink the width to 1200px say, and perhaps turn it into a jpg?

farhangnaderi:
@hamishwillee 
Thanks for the comment. I mean the photo below from the docs (https://docs.px4.io/master/assets/img/gpsmount.cc6f450b.jpg) is stating to put the GPS on top. But the final current assembled photo shows the GPS somewhere else. This just does not make sense. 
Besides, the reason for this GPS to be on top is the place already available on the top shell, so why to occupy the payload hanger?! Unless a specific developer wants to use top shell for adding specific components. 
Sure I will update the photo as you requested. Thanks for stating about the size and format.

farhangnaderi:
I was not able to make the image smaller though. The quality would be not acceptable most probably.

hamishwillee:
Good enough thank you! 

@mrpollo FMI, what tool are you using for image size reduction?

mrpollo:
@hamishwillee I'm doing it manually at the moment, but we could totally use one of the following options

https://github.com/marketplace/actions/image-actions

https://github.com/marketplace/imgbot

Happy to help with the set up

hamishwillee:
That would be great thank you. I was thinking that the tools would probably not do a good enough job for this particular case; then I realised that doesn't matter - on average they will do a great job and we can still do our own editing for edge cases.

PR #1273, opened 2021-05-19
fixed the mistake with the sdf codes

canberkgurel:
The .sdf codes in the https://docs.px4.io/master/en/sensor/rangefinders.html#simulation need to be swapped.

So, did that and creating a PR.


hamishwillee:
Thank @canberkgurel !

PR #1275, opened 2021-05-21
parameters: document metadata sync via COMPONENT INFORMATION API

bkueng:
@hamishwillee this is for params, but works the same way for events. Not sure if you want to mention that, and where?

### DIFF
--- a/en/advanced/parameters_and_configurations.md
+++ b/en/advanced/parameters_and_configurations.md
@@ -333,20 +333,25 @@ The purpose of each line is given below (for more detail see [module_schema.yaml
  */
 ```
 
-#### Publishing Parameter Metadata to a GCS
+## Publishing Parameter Metadata to a GCS
 
 Parameter metadata is collected into a JSON or XML file during each PX4 build.
-For all targets with enough FLASH available the JSON file is xz-compressed and stored within the generated binary.
+
+For most flight controllers (as most have enough FLASH available), the JSON file is xz-compressed and stored within the generated binary.
 The file is then shared to ground stations using the [MAVLink Component Information Protocol](https://mavlink.io/en/services/component_information.html).
 This ensures that parameter metadata is always up-to-date with the code running on the vehicle.
 
-Binaries for flight controller targets with constrained memory (`CONSTRAINED_MEMORY`) do not store the *build-specific* parameter metadata in the binary, but instead reference metadata stored on `px4-travis.s3.amazonaws.com`.
-Versions of the metadata exist for the current master and each PX4 release.
-The metadata is uploaded via [github CI](https://github.com/PX4/PX4-Autopilot/blob/master/.github/workflows/metadata.yml) (and hence will only be available once parameters have been merged into master).
+Binaries for flight controller targets with constrained memory do not store the parameter metadata in the binary, but instead reference the same data stored on `px4-travis.s3.amazonaws.com`.
+This applies, for example, to the [Omnibus F4 SD](../flight_controller/omnibus_f4_sd.md).
+The metadata is uploaded via [github CI](https://github.com/PX4/PX4-Autopilot/blob/master/.github/workflows/metadata.yml) for all build targets (and hence will only be available once parameters have been merged into master).
 
 :::note
-The metadata on `px4-travis.s3.amazonaws.com` is used as a fallback if parameter metadata is not present on the vehicle.
-It may also be used as a fallback, for example, to avoid a very slow download over a low-rate telemetry link.
+You can identify memory constrained boards because they specify `CONSTRAINED_MEMORY` in their [cmake definition file](https://github.com/PX4/PX4-Autopilot/blob/release/1.12/boards/omnibus/f4sd/default.cmake#L11)).
+:::
+
+:::note
+The metadata on `px4-travis.s3.amazonaws.com` is used if parameter metadata is not present on the vehicle.
+It may also be used as a fallback to avoid a very slow download over a low-rate telemetry link.
 :::
 
 Anyone doing custom development on a FLASH-constrained board can adjust the URL [here](https://github.com/PX4/PX4-Autopilot/blob/master/src/lib/component_information/CMakeLists.txt#L41) to point to another server.
--- a/en/advanced/parameters_and_configurations.md
+++ b/en/advanced/parameters_and_configurations.md
@@ -330,6 +330,17 @@ The purpose of each line is given below (for more detail see [module_schema.yaml
  */
 ```
 
+#### Metadata Synchronization
+All parameter metadata is collected in an XML or JSON file during build.
+Traditionally the XML file of the master branch was copied into the QGC source tree via CI.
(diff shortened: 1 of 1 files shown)


hamishwillee:
Thanks very much for this @bkueng !

I have subedited and it needs your review. The main change is to unmix the main use case (compressed JSON on targets with enough memory) from the rarer use cases - targets with restricted memory, custom builds, use of XML. Reason being is that the main use case should be "it just works", so best to present that first. 
There are some questions inline that came up as ambiguities to me as I restructured.

> works the same way for events. Not sure if you want to mention that, and where?

Hmmmm. My thinking is that in PX4 we need a document on events much like this parameter doc which explains briefly what events are, how they are created/used in our docs, and how they are generated into firmware. If there are any PX4/non standard events we should document those. We may need to create the mavlink service docs first.

en/advanced/parameters_and_configurations.md
```diff
@@ -86,6 +86,9 @@ Parameters definitions have two parts:
 Several approaches are described below for writing both the metadata and code.
 Where possible code should use newer [YAML metadata](#yaml-metadata) and [C++ API](#c-api) over the older C parameter/code definitions, as these are more flexible and robust.
 
+Parameter metadata is [compiled into the firmware](#publishing-parameter-metadata-to-a-gcs),
```
hamishwillee, line 89: Note that I added links here and comment above to "forward reference" the existence of the information about getting parameter to the GCS. Otherwise it is a bit invisible.

en/advanced/parameters_and_configurations.md
```diff
@@ -330,6 +333,26 @@ The purpose of each line is given below (for more detail see [module_schema.yaml
  */
 ```
 
+#### Publishing Parameter Metadata to a GCS
```
hamishwillee, line 336: Note, new title. I figured this is "more obvious"

en/advanced/parameters_and_configurations.md
```diff
@@ -330,6 +333,26 @@ The purpose of each line is given below (for more detail see [module_schema.yaml
...
+
+Parameter metadata is collected into a JSON or XML file during each PX4 build.
+For all targets with enough FLASH available the JSON file is xz-compressed and stored within the generated binary.
+The file is then shared to ground stations using the [MAVLink Component Information Protocol](https://mavlink.io/en/services/component_information.html).
+This ensures that parameter metadata is always up-to-date with the code running on the vehicle.
+
+Binaries for flight controller targets with constrained memory (`CONSTRAINED_MEMORY`) do not store the *build-specific* parameter metadata in the binary, but instead reference metadata stored on `px4-travis.s3.amazonaws.com`.
```
hamishwillee, line 343: What is `CONSTRAINED_MEMORY`? It doesn't appear to be a parameter. 

Note, this is my interpretation of your original text - ie on constrained target you use the hosted metadata NOT the build specific metadata. If you want to have custom params on resource constrained platform then you have to upload your custom metadata to your own server - below we show how to set your own server.

en/advanced/parameters_and_configurations.md
```diff
@@ -330,6 +333,26 @@ The purpose of each line is given below (for more detail see [module_schema.yaml
...
+Versions of the metadata exist for the current master and each PX4 release.
```
hamishwillee, line 344: @bkueng Note, this is my **assumption** - that each release will be adjusted to find its own appropriate metadata. Otherwise you'd be screwed as time goes by and releases get older. Is this correct?

en/advanced/parameters_and_configurations.md
```diff
@@ -330,6 +333,26 @@ The purpose of each line is given below (for more detail see [module_schema.yaml
...
+The metadata is uploaded via [github CI](https://github.com/PX4/PX4-Autopilot/blob/master/.github/workflows/metadata.yml) (and hence will only be available once parameters have been merged into master).
+
+:::note
+The metadata on `px4-travis.s3.amazonaws.com` is used as a fallback if parameter metadata is not present on the vehicle.
+It may also be used as a fallback, for example, to avoid a very slow download over a low-rate telemetry link.
+:::
+
+Anyone doing custom development on a FLASH-constrained board can adjust the URL [here](https://github.com/PX4/PX4-Autopilot/blob/master/src/lib/component_information/CMakeLists.txt#L41) to point to another server.
+
+The XML file of the master branch is copied into the QGC source tree via CI and is used as a fallback in cases where no metadata is available via the component information service (this approach predates the existence of the component information protocol).
```
hamishwillee, line 354: So this still happens right, and is the only use of the XML files?

en/advanced/parameters_and_configurations.md
bkueng: Example: https://github.com/PX4/PX4-Autopilot/blob/master/boards/omnibus/f4sd/default.cmake#L9

It's hosted, but still build-specific (CI uploads all targets).

en/advanced/parameters_and_configurations.md
bkueng: Correct

en/advanced/parameters_and_configurations.md
```diff
@@ -330,6 +333,26 @@ The purpose of each line is given below (for more detail see [module_schema.yaml
...
+For all targets with enough FLASH available the JSON file is xz-compressed and stored within the generated binary.
+The file is then shared to ground stations using the [MAVLink Component Information Protocol](https://mavlink.io/en/services/component_information.html).
+This ensures that parameter metadata is always up-to-date with the code running on the vehicle.
+
+Binaries for flight controller targets with constrained memory (`CONSTRAINED_MEMORY`) do not store the *build-specific* parameter metadata in the binary, but instead reference metadata stored on `px4-travis.s3.amazonaws.com`.
+Versions of the metadata exist for the current master and each PX4 release.
+The metadata is uploaded via [github CI](https://github.com/PX4/PX4-Autopilot/blob/master/.github/workflows/metadata.yml) (and hence will only be available once parameters have been merged into master).
+
+:::note
+The metadata on `px4-travis.s3.amazonaws.com` is used as a fallback if parameter metadata is not present on the vehicle.
```
bkueng, line 348: Technically, if not on the vehicle, it's not a fallback, but the primary url

en/advanced/parameters_and_configurations.md
bkueng: Yes still in use. At some point we can disable that.
Currently also Flight Review uses the xml. I might change that but don't see immediate need.

bkueng (commented):
Thanks Hamish.

> My thinking is that in PX4 we need a document on events much like this parameter doc which explains briefly what events are, how they are created/used in our docs, and how they are generated into firmware.

Right, more docs... I'll look into that once things are a bit settled.

en/advanced/parameters_and_configurations.md
hamishwillee: Thanks for clarification. I have slightly reworded.

hamishwillee:
Thanks @bkueng - really helpful. I've made some minor changes to capture these details. Also made the heading level higher to make this section visible. 

Merging :-)

PR #1295, opened 2021-06-02
Clarify sub support

hamishwillee:
@DanielDuecker This is first pass at clarifying the support provided by PX4 for UUVs and more specifically by BlueRov2 and Hippocampus. Can you give it a quick look?

I have characterised it as "experimental" and said that this applies to the platform, and hence both the Bluerov and Hippocampus. Can you confirm that this is the case? I.e. that the two vehicles share the same limitations/core behaviour? If not, I can copy this into BlueRov, but would be good to understand what behaviour is enabled for Hippocampus.

Further, I can build the gazebo version using `make px4_sitl gazebo_uuv_bluerov2_heavy` but it won't let me zoom in on the vehicle so I can't get a good screenshot (trying to do this using mousewheel scroll). Can you generate a good gazebo zoomed image for me? (and if so, any idea what I might be doing wrong?

### DIFF
--- a/en/frames_sub/README.md
+++ b/en/frames_sub/README.md
@@ -1,6 +1,15 @@
 # Underwater Vehicles (UUV)
 
-PX4 supports a number of unmanned underwater vehicles (UUV):
+:::warning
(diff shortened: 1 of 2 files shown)


DanielDuecker (approved):
looks good to me =)
I will get a zoomed screenshot

PR #1296, opened 2021-06-02
Update quick_start_cube.md

Ionstream:
Added information on safety switch functionality and links to buzzer tones for troubleshooting.

### DIFF
--- a/en/assembly/quick_start_cube.md
+++ b/en/assembly/quick_start_cube.md
@@ -95,11 +95,12 @@ The *dedicated* safety switch that comes with the Cube is only required if you a
 
 If you are flying without the GPS you must attach the switch directly to the `GPS1` port in order to be able to arm the vehicle and fly (or via a supplied cable if using an old-style 6-pin GPS).
 
+The safety switch also doubles as a means to force bootloader updates. Should there be a need to force a bootloader update after flashing PX4 firmware (e.g. "Program PX4IO" tone), hold down the safety switch, then power the Cube over USB to proceed.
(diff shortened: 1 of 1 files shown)


hamishwillee:
Thanks @Ionstream - really helpful. I have subedited - mostly because the safety switch can be either dedicated OR in the GPS. Further, the bootloader update might not be spotted if in the safety switch. I have therefore put it in configuration.

I'm going to merge this, but one question - can the normal bootloader process be used as well: https://docs.px4.io/master/en/advanced_config/bootloader_update.html#bootloader-update ?

Ionstream:
Cool! I'm not sure, I don't think I've tried the normal bootloader process. 

I have tried forcing a bootloader update in Mission Planner, and that didn't go too well (the bricked cube required an ST-LINK, some JST-SUR cables and a trip to the reseller) to fix the corrupted bootloader. Not sure if the underlying code in the GUI is anything to similar to what's on this page, but I don't have a concrete answer for you yet.

As an aside, the px4-fmu_v6x target doesn't seem to exist on my VS code setup (mine only goes up to v5x). Has it been renamed recently?

hamishwillee:
Thanks, Then this is good for now. If you try the normal process then please let me know so we can update this. 

> As an aside, the px4-fmu_v6x target doesn't seem to exist on my VS code setup (mine only goes up to v5x). Has it been renamed recently?

I don't know. Not sure it has been added yet. I would ask the question in slack, tagging Daniel Agar.

PR #1309, opened 2021-06-03
release notes updates, COMPONENT_INFORMATION is now enabled

bkueng:


en/releases/1.12.md
```diff
@@ -55,15 +55,16 @@ The release includes new hardware support for the following boards, peripherals,
...
+* Removal of discontinued Intel AeroFC
+
 ### MAVLink
 
 * **MAVLink Ethernet configuration ([PR#14460](https://github.com/PX4/PX4-Autopilot/pull/14460))**
   * MAVLink Ethernet channel settings such as UDP port, remote port and broadcast mode now can be changed dynamically via parameters.
-
-<!--
 * **Support for querying `COMPONENT_INFORMATION` ([PR#16039](https://github.com/PX4/PX4-Autopilot/pull/16039))**
+  * Parameter metadata is now automatically synchronized in combination with QGC daily.
```
hamishwillee, line 66: @bkueng Looks good, and I'm going to merge. Not what this might mean. I assume it means that the metadata goes to our server so that it can be downloaded by QGC if the vehicle supplies an HTTPS URL rather than an on-vehicle URL? If so, consider clarifying that.

bkueng: I'm comparing to the status quo where metadata is copied into the QGC tree.

hamishwillee: This is great then. I'd argue for a different wording if QGC released version supported component_information.  Thanks very much.

PR #1312, opened 2021-06-03
PWM_ to PWM_MAIN_xxxn changes

hamishwillee:
PWM_MIN and friends go to PWM_MAIN_MIN. 

@bresch I'm going to merge this. But can you please sanity check. Specifically, we now have PWM_MAIN_MIN which is a default for PWM_MAIN_MIN1, etc. If you think we need to capture that case then please let me know. I have inferred it in a few places using the text `PWM_MAIN_MINn` but still linking to the default item `PWM_MAIN_MIN`

### DIFF
--- a/en/advanced_config/esc_calibration.md
+++ b/en/advanced_config/esc_calibration.md
@@ -48,6 +48,6 @@ To calibrate the ESCs:
 
 :::note
 High-quality controllers come with a factory calibration. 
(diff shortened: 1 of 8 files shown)


bresch:
I think it's all good. Anyway, the description in `PWM_MAIN_MIN1..8` is quite clear

PR #1326, opened 2021-06-09
Preflight check fixes plus a link or two

hamishwillee:


en/flying/pre_flight_checks.md
```diff
@@ -12,53 +12,64 @@ The sections below list the errors, their likely causes and solutions, and any p
...
 * It can fail if the yaw gyro has a large offset or if the vehicle is moved or rotated in the presence of a bad magnetic interference or magnetometer calibration.
 
-`PREFLIGHT FAIL: EKF HIGH IMU ACCEL BIAS`:
+#### PREFLIGHT FAIL: EKF HIGH IMU ACCEL BIAS
+
 * This error is produced when the IMU accelerometer bias estimated by the EKF is excessive. 
-* The check is controlled by the [COM_ARM_EKF_AB](../advanced_config/parameter_reference.md#COM_ARM_EKF_AB) parameter.
-  
-`PREFLIGHT FAIL: EKF HIGH IMU GYRO BIAS`:
+* Excessive in this case means that the bias estimate exceeds half the configured limit.
```
hamishwillee, line 44: @bresch Just this and the next item are actual content changes that need to be looked at.

BTW, the code that generates these messages appears to be the call like `mavlink_log_critical(mavlink_log_pub, "Preflight Fail: High Gyro Bias");`. How does that end up being "`PREFLIGHT FAIL: EKF HIGH IMU ACCEL BIAS`" i.e. all caps etc. The reason I ask is that it would be good to grep these and make sure I have them all - but if they don't quite match what users see it is a little tougher.

bresch: The content is correct. However, it seems that the error messages are a bit outdated... I'll review them.

hamishwillee: Thanks very much @bresch ! 

@bkueng Eventually is it the intent that these warning get published through the events protocol? I hope so, that would allow us to gather all of the published notifications in one place.

bkueng: Yes. We won't have to document them here then, the description will be directly in code.

PR #1327, opened 2021-06-09
CAL_MAG, MAV_BROADCAST, Failure Injection fixes

hamishwillee:
This updates docs for fact that CAL_MAG params changed. Also some questions around MAV_BROADCAST and failure injection (inline)

### DIFF
--- a/en/gps_compass/README.md
+++ b/en/gps_compass/README.md
@@ -132,9 +132,9 @@ Parameter | Setting
 Compass calibration is covered in: [Compass Configuration](../config/compass.md).
 The process is straightforward and will calibrate all connected magnetometers.
 
-Additional configuration can be [performed](../advanced_config/parameters.md) using the [CAL\_MAGx\_](../advanced_config/parameter_reference.md#CAL_MAG0_EN) parameters (where `x=0-3`).
-Generally you will not need to *modify* these as compasses are autodetected, prioritised and are all calibrated at the same time (a possible exception is [CAL\_MAGx\_EN](../advanced_config/parameter_reference.md#CAL_MAG0_EN) which might be used to disable an internal compass).
-You may however wish to read them, as they will let you know which magnetometers are internal or external ([CAL\_MAGx\_EN](../advanced_config/parameter_reference.md#CAL_MAG0_EN)) and which is being uses as the main heading source ([CAL_MAG_PRIME](../advanced_config/parameter_reference.md#CAL_MAG_PRIME)).
+Additional configuration can be [performed](../advanced_config/parameters.md) using the [CAL\_MAGx\_](../advanced_config/parameter_reference.md#CAL_MAG0_ID) parameters (where `x=0-3`).
+Generally you will not need to *modify* these as compasses are autodetected, prioritised and are all calibrated at the same time (a possible exception is [CAL\_MAGx\_PRIO](../advanced_config/parameter_reference.md#CAL_MAG0_PRIO) which might be used to disable an internal compass).
+You may however wish to read them, as they will let you know which magnetometers are internal or external ([CAL_MAGn_ROT=-1](../advanced_config/parameter_reference.md#CAL_MAG0_ROT )) and which is being uses as the main heading source ([CAL_MAG_PRIME](../advanced_config/parameter_reference.md#CAL_MAG_PRIME)).
 
 
 ## Developer Information
--- a/en/gps_compass/README.md
+++ b/en/gps_compass/README.md
@@ -130,11 +130,18 @@ Parameter | Setting
 ### Compass Configuration
 
 Compass calibration is covered in: [Compass Configuration](../config/compass.md).
-The process is straightforward and will calibrate all connected magnetometers.
+The process is straightforward and will autodetect, prioritise and calibrate all connected magnetometers.
 
-Additional configuration can be [performed](../advanced_config/parameters.md) using the [CAL\_MAGx\_](../advanced_config/parameter_reference.md#CAL_MAG0_ID) parameters (where `x=0-3`).
-Generally you will not need to *modify* these as compasses are autodetected, prioritised and are all calibrated at the same time (a possible exception is [CAL\_MAGx\_PRIO](../advanced_config/parameter_reference.md#CAL_MAG0_PRIO) which might be used to disable an internal compass).
-You may however wish to read them, as they will let you know which magnetometers are internal or external ([CAL_MAGn_ROT==-1](../advanced_config/parameter_reference.md#CAL_MAG0_ROT ) indicates internal) and which is being uses as the main heading source ([CAL_MAG_PRIME](../advanced_config/parameter_reference.md#CAL_MAG_PRIME)).
+Further compass configuration should not be required.
+
+#### Compass Parameters
+
+Compass parameters are prefixed with [CAL\_MAGx\_](../advanced_config/parameter_reference.md#CAL_MAG0_ID), where `x=0-3`.
+
+As stated above, generally these should not need to be changed.
+A possible exception is that vehicle developers might choose to disable an internal compass or experiment with adjusting the compass priority:
+- [CAL_MAGn_ROT](../advanced_config/parameter_reference.md#CAL_MAG0_ROT) can be read to determine if a compass is internal (`CAL_MAGn_ROT==1`)
+- [CAL\_MAGx\_PRIO](../advanced_config/parameter_reference.md#CAL_MAG0_PRIO) can be used to disable or change the relative priority of a compass.
 
 
 ## Developer Information
--- a/en/gps_compass/README.md
+++ b/en/gps_compass/README.md
@@ -134,7 +134,7 @@ The process is straightforward and will calibrate all connected magnetometers.
 
 Additional configuration can be [performed](../advanced_config/parameters.md) using the [CAL\_MAGx\_](../advanced_config/parameter_reference.md#CAL_MAG0_ID) parameters (where `x=0-3`).
 Generally you will not need to *modify* these as compasses are autodetected, prioritised and are all calibrated at the same time (a possible exception is [CAL\_MAGx\_PRIO](../advanced_config/parameter_reference.md#CAL_MAG0_PRIO) which might be used to disable an internal compass).
-You may however wish to read them, as they will let you know which magnetometers are internal or external ([CAL_MAGn_ROT=-1](../advanced_config/parameter_reference.md#CAL_MAG0_ROT )) and which is being uses as the main heading source ([CAL_MAG_PRIME](../advanced_config/parameter_reference.md#CAL_MAG_PRIME)).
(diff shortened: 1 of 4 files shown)


en/gps_compass/README.md
```diff
@@ -132,9 +132,9 @@ Parameter | Setting
...
-You may however wish to read them, as they will let you know which magnetometers are internal or external ([CAL\_MAGx\_EN](../advanced_config/parameter_reference.md#CAL_MAG0_EN)) and which is being uses as the main heading source ([CAL_MAG_PRIME](../advanced_config/parameter_reference.md#CAL_MAG_PRIME)).
+Additional configuration can be [performed](../advanced_config/parameters.md) using the [CAL\_MAGx\_](../advanced_config/parameter_reference.md#CAL_MAG0_ID) parameters (where `x=0-3`).
+Generally you will not need to *modify* these as compasses are autodetected, prioritised and are all calibrated at the same time (a possible exception is [CAL\_MAGx\_PRIO](../advanced_config/parameter_reference.md#CAL_MAG0_PRIO) which might be used to disable an internal compass).
+You may however wish to read them, as they will let you know which magnetometers are internal or external ([CAL_MAGn_ROT==-1](../advanced_config/parameter_reference.md#CAL_MAG0_ROT ) indicates internal) and which is being uses as the main heading source ([CAL_MAG_PRIME](../advanced_config/parameter_reference.md#CAL_MAG_PRIME)).
```
hamishwillee, line 137: @bkueng I did minor update to this to replace CAL_MAG0_EN with CAL_MAG0_PRIO, and state that  you might use CAL_MAGn_ROT==1 to indicate that the compass is internal/external. 

However:
1. Not sure it is correct - is it true that prioritisation is done by default, along with selection of the prime compass? 
2. `CAL_MAG_PRIME` is gone. IS there no way to get the prime compass? If not, just delete the last sentence "and which is being uses as the main heading ..."?

en/simulation/flightgear.md
```diff
@@ -132,7 +132,7 @@ INFO  [mavlink] mode: Onboard, data rate: 4000 B/s on udp port 14280 remote port
 INFO  [logger] logger started (mode=all)
 INFO  [logger] Start file log (type: full)
 INFO  [logger] Opened full log file: ./log/2020-04-28/22_03_36.ulg
-INFO  [mavlink] MAVLink only on localhost (set param MAV_BROADCAST = 1 to enable network)
+INFO  [mavlink] MAVLink only on localhost
```
hamishwillee, line 135: Any idea what you would see here?

en/simulation/README.md
```diff
@@ -274,7 +274,7 @@ There are a number of ways to make the UDP packets available on external interfa
 
 ### Enable MAV_BROADCAST
 
-Enable [MAV_BROADCAST](../advanced_config/parameter_reference.md#MAV_BROADCAST) to broadcast heartbeats on the local network.
+Enable [MAV_BROADCAST](../advanced_config/parameter_reference.md#MAV_BROADCAST) to broadcast heartbeats on the local network.
```
hamishwillee, line 277: @bkueng @julianoes So MAV_BROADCAST was deleted in https://github.com/PX4/PX4-Autopilot/pull/14460
It looks like the expectation is that this was something needed for Ethernet, and if you need it you start mavlink module with the `-p` flag.

This particular doc is about setting it up for Simulator. I assume we still want this :-). How do we do it/what do we do?

en/simulation/failsafes.md
```diff
@@ -51,7 +48,25 @@ To control how fast the battery depletes to the minimal value use the parameter
 By changing [SIM_BAT_MIN_PCT](../advanced_config/parameter_reference.md#SIM_BAT_MIN_PCT) in flight, you can also test regaining capacity to simulate inaccurate battery state estimation or in-air charging technology.
 :::
 
-## GPS Loss
-
-To simulate losing and regaining GPS information you can just stop the publication of GPS messages.
-This is done by running the `param set SIM_GPS_BLOCK 1` and `param set SIM_GPS_BLOCK 0` commands on your SITL instance *pxh shell* to block and unblock messages respectively.
+## Sensor/System Failure
+
+System failure injection can be used to simulate different types of failures in many sensors and systems (e.g. GPS, barometer, gyro, avoidance etc.):
```
hamishwillee, line 53: @julianoes Several questions about Failure injection:
1. Is this just for simulation, or can it be used on a real vehicle in flight? My assumption is yes, and that is why we have the `SYS_FAILURE_EN` param (to make sure this doenn't happen by accident)
1. Is the `SYS_FAILURE_EN` enabled by default for simulation and/or HITL?
1. Assuming it can be used in normal flight when would you do this?
1. What are the different failure types. I.e. I get "on and off" but what is the difference between stuck, off and intermittent. What is the difference between wrong and garbage. Basically - when should someone use each?

If this is intended to be generic my thinking is I might move this out of simulation into the Debugging/Logging section and link to this from here.

en/gps_compass/README.md
bkueng: Yes external ones will have a higher prio. You can change that as you want to select a different one as primary.

en/simulation/README.md
bkueng: Yes you need to add the `-p` flag for SITL if you want that.
For the configurable instances (boards with ethernet), it can be configured with `MAV_i_BROADCAST`

en/simulation/flightgear.md
bkueng:
```
INFO  [mavlink] MAVLink only on localhost (set param MAV_{i}_BROADCAST = 1 to enable network)
```
Which isn't correct though for SITL.

en/gps_compass/README.md
hamishwillee: @bkueng Ah, so essentially everything gets done automatically but you can play with these yourself after that? When do they reset? i.e. say I add a new compass, does that cause everything to reshuffle? What if I fix one external mag to highest priority - will that ever be autoshuffled?

en/simulation/flightgear.md
hamishwillee: @bkueng I didn't ignore you last time you mentioned MAV_x_BROADCAST. I just assumed you were wrong because I can't find this parameter in http://docs.px4.io/master/en/advanced_config/parameter_reference.html .
What's going on?

hamishwillee: Ah, looking below I think you're saying this will only appear for a board that supports it - Ethernet etc. Hmmm. So not linkable in docs since not generated.

en/gps_compass/README.md
bkueng: Exactly. Existing ID's don't get changed, new ones will be assigned a new cal param index. What happens on 'overflow' is a question for @dagar.

en/simulation/README.md
bkueng: I don't think so. Arg ordering does not matter.

en/simulation/flightgear.md
bkueng: Correct. Indeed, it's getting more difficult the more board-specific generation we have.
For this case we can pretend there's ethernet, so the generator adds the params if you want.

en/gps_compass/README.md
hamishwillee: Thanks @bkueng . I have reworded a little to make this clear.

MY assumption here is that the auto-calibration is pretty smart - it prioritizes external over internal compasses, and it uses some sort of algorithm to determine which compasses are better than others. Further the internal compass will always be the lowest priority so would never actually get used? So why would you ever need to change the priority or disable an internal compass? Perhaps because you know they'd be better off having no compass than using the internal one?
Similarly, is there a chance they will get the priority better than the internal algorithm?

Perhaps you might want to disable an external compass during testing to work out whether it is actually needed? Doing it in software being easier than physically removing it?

The reason I ask is that if we can be explicit that would be better - e.g. have a heading "disable internal compass" and "changing the priority" makes sense if these are thinks we ever think people need to do - because that will help them find the instructions.

Anyway, merging this for now, but will revisit based on your comment.

en/simulation/flightgear.md
hamishwillee: @bkueng Yes please - can you make the generator changes needed.

en/simulation/README.md
hamishwillee: Thanks @bkueng I have updated this to "Modify Configuration for UDP Broadcast" and moved down near the external broadcast. I guess I could merge the two sections, but I think possibly good to separate the two options.
1. Please check what I wrote here is correct. 
2. Why/why not would you choose to broadcast (p) rather than specific target (t). I.e. do we need to provide a recommendation?

Note module docs out of date?: https://docs.px4.io/master/en/modules/modules_communication.html#usage-2
image

en/simulation/failsafes.md
julianoes: 1. Exactly, it's for both, simulation and real, hence the param.
2. The failure module and therefore param is not built for "real" yet, so it's not available yet. But I should do that soon.
3. To test the behaviour, e.g. to test the RC loss behaviour in flight (without having to switch off RC).
4. Stuck is stuck on a certain value but it keeps publishing it. This could be a malfunctioning sensor.
Off means the sensor does not publish anymore, so it can be detected more easily.
Intermittent is, as the name implies, intermittent, switch between off and on.
Wrong would be a sensor reporting values that are wrong but they might still "look correct". Garbage is just random noice, like reading uninitialized memory.
Most of these are not implemented broadly yet, so they will likely return "Unsupported" when you try them.

julianoes: And you're right it could go in a separate page once it has a bit better support.

Thanks for updating it here though, that's great!

en/simulation/README.md
```diff
@@ -274,7 +272,8 @@ There are a number of ways to make the UDP packets available on external interfa
 
 ### Enable MAV_BROADCAST
```
bkueng, line 273: Enable broadcasting

en/gps_compass/README.md
bkueng: Almost as smart as you would expect: all external ones have the same prio (same for internal). Internal ones are used on failover. So yes, you might prefer to not use that at all.

> Perhaps you might want to disable an external compass during testing to work out whether it is actually needed?

Yes, another possible use case.
Probably in most cases you just want to disable the internal one(s).

en/simulation/README.md
bkueng: Both cases are valid: broadcast makes it easy if you're alone in a network, with possible changing IP's. If you have multiple users (simulations running) broadcast leads to chaos.

en/gps_compass/README.md
hamishwillee: Thanks. FYI only I've captured this use case as the important one in https://github.com/PX4/PX4-user_guide/pull/1350

en/simulation/failsafes.md
hamishwillee: Awesome. You'll be even happier it now has its own page: https://github.com/PX4/PX4-user_guide/pull/1351

en/simulation/README.md
hamishwillee: Thanks. Tidied in https://github.com/PX4/PX4-user_guide/pull/1352

PR #1339, opened 2021-06-11
Update hello_sky.md

Ionstream:
Added links to supporting information on `__EXPORT` and related concepts to help newer developers.

### DIFF
--- a/en/modules/hello_sky.md
+++ b/en/modules/hello_sky.md
@@ -94,7 +94,15 @@ This consists of a single *C* file and a *cmake* definition (which tells the too
      `PX4_INFO` is the equivalent of `printf` for the PX4 shell (included from **px4_platform_common/log.h**).
      There are different log levels: `PX4_INFO`, `PX4_WARN`, `PX4_ERR`, `PX4_DEBUG`.
      Warnings and errors are additionally added to the [ULog](../dev_log/ulog_file_format.md) and shown on [Flight Review](https://logs.px4.io/).
-	 :::
+     :::
+     
+     :::tip `__EXPORT` keyword is an extension of the C/C++ standard that exports the compiled object file to the linker. The object file comes from the .cpp source file that was written; it is C/C++ code translated into machine code by the compiler. 
+     
+     Each .c/.cpp file is translated into it's own object file,typically named as object.o or object.obj. A linker is a program which combines the compiled object file and referenced libraries into a single executable.
+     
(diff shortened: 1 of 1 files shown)


en/modules/hello_sky.md
```diff
@@ -94,7 +94,15 @@ This consists of a single *C* file and a *cmake* definition (which tells the too
      `PX4_INFO` is the equivalent of `printf` for the PX4 shell (included from **px4_platform_common/log.h**).
      There are different log levels: `PX4_INFO`, `PX4_WARN`, `PX4_ERR`, `PX4_DEBUG`.
      Warnings and errors are additionally added to the [ULog](../dev_log/ulog_file_format.md) and shown on [Flight Review](https://logs.px4.io/).
-	 :::
+     :::
```
hamishwillee, line 97: Thanks @Ionstream. Hmmm. While I encourage better explanations for new people wanting to join the domain, I wasn't thinking of "beginner C/C++ programmer". More of somewhat experienced C/C++ programmers who don't necessarily know NuttX and our particular use of Cmake etc. In other words, people who know how to code, but not the domain. 

I'm open to help complete beginners, but they aren't the main audience because coding for dangerous flying robots is perhaps not the best starting point. I'm also a bit wary that this kind of information will irritate people who know what they are doing.

Upshot I don't think we want this here. What it does tell us is that we are missing the context of "what is a developer",  "what do you need to know before starting" in http://docs.px4.io/master/en/dev_setup/getting_started.html. We might consider a set of links for beginner programmers at that point.

hamishwillee: PS by way of example, note that above we say "The main function must be named `<module_name>_main` and exported from the module as shown.". That's a convention that a new programmer to the domain needs to know.

Ionstream: Ah I see, I had a feeling that beginners weren't your target audience, for reasons that I completely understand. If that's case, I think we should drop the changes.

I suppose it's a good thing that I managed to move through the docs at the moment, because it means that there are sufficient starting points for newbies to work from. I'm already keeping personal notes separately over on gitbook, so there's a reference to start from if we want to figure out what's confusing to beginners.

Ionstream: @hamishwillee I've put the link [here](https://app.gitbook.com/@limryan96/s/jetson-nano-docs/px4-development/hello-sky), in case my notes might help you assess the newbie situation better.

hamishwillee:
Thanks @Ionstream - really appreciate your help on this.

PR #1343, opened 2021-06-13
Update cuav_v5.md

hany606:
Change Nuttx to the correct url: https://nuttx.apache.org/

This url is indeed in many pages in the documentation, and if this was really a wrong url in the documentation, is there an easy way to change all the urls instead of manually changing them?


mrpollo:
Perhaps a massive search and replace, but might be dangerous and not something to take on lightly, a one by one approach seems the most sensible to me. What do you think @hamishwillee ?

hamishwillee:
Thanks for this. Yes, I did a search and replace in #1344. It is fairly safe as long as you're careful.

hany606:
Thank you @mrpollo @hamishwillee

PR #1347, opened 2021-06-15
Update building_px4.md

Ionstream:
Added instructions for fixing "Visual Studio Code is unable to watch for file changes in this large workspace".


hamishwillee:
Thanks @Ionstream 

There is nothing wrong with this. However it occurs to me that perhaps people are more likely to expect vs code info in http://docs.px4.io/master/en/dev_setup/vscode.html ?

Do perhaps we should add a troubleshooting section to VSCode with this and any other VSCode specfic issues, and then in this page have a note cross linking to the VSCode troubleshooting section. What do you think?

hamishwillee:
In case you missed my note :-). Please see https://github.com/PX4/PX4-user_guide/pull/1347#issuecomment-861891799

Ionstream:
Ooops, I forgot to reply here. I think your suggestion does add some extra flexibility or means to return a "search hit" for users, so its a good idea to add it in. I'll go update the PR.

hamishwillee:
Thanks!

Ionstream:
@hamishwillee Done!

hamishwillee:
Thanks! Merged.

PR #1351, opened 2021-06-16
Separate out Failure Injection

hamishwillee:
This allows it to be evolved separately of simulation topics, which makes sense as it is intended to be supported on real hardware/flight eventually.

@julianoes I will merge this, but would appreciate your sanity check.

### DIFF
--- a/en/debug/failure_injection.md
+++ b/en/debug/failure_injection.md
@@ -1,0 +1,49 @@
+# System Failure Injection
+
+System failure injection allows you to cause different types of sensor and system failures using the [MAVLink shell](../debug/mavlink_shell.md#mavlink-shell) or other PX4 console.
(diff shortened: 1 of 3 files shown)


en/debug/failure_injection.md
```diff
@@ -0,0 +1,49 @@
+# System Failure Injection
+
+System failure injection allows you to cause different types of sensor and system failures using the [MAVLink shell](../debug/mavlink_shell.md#mavlink-shell) or other PX4 console.
```
julianoes, line 3: Or MAVSDK failure plugin: https://mavsdk.mavlink.io/main/en/cpp/api_reference/classmavsdk_1_1_failure.html

julianoes (commented):
Awesome, thanks @hamishwillee!

en/debug/failure_injection.md
hamishwillee: Thanks. that makes quite a bit of difference :-). Fixed in https://github.com/PX4/PX4-user_guide/pull/1354

PR #1354, opened 2021-06-17
Failure injection improvements and link fixes

hamishwillee:
Updates following feedback.

FYI @julianoes

### DIFF
--- a/en/debug/failure_injection.md
+++ b/en/debug/failure_injection.md
@@ -1,10 +1,9 @@
 # System Failure Injection
 
-System failure injection allows you to cause different types of sensor and system failures using the [MAVLink shell](../debug/mavlink_shell.md#mavlink-shell) or other PX4 console.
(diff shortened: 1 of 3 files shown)


en/debug/failure_injection.md
```diff
@@ -29,12 +53,12 @@ where:
   - `wrong`: Publish invalid values (that still look reasonable/aren't "garbage").
   - `slow`: Publish at a reduced rate.
   - `delayed`: Publish valid data with a significant delay.
-  - `intermittent`: Publish intermittently. 
+  - `intermittent`: Publish intermittently (sometimes correct, sometimes incorrect).
```
julianoes, line 56: This was correct before.

julianoes (commented):
One nitpick, the rest is good.

en/debug/failure_injection.md
hamishwillee: Thanks @julianoes . The thing is, that I rewrote this after looking at Mavsdk docs :-). Publish intermittently is a little ambigous. Does it mean publish right for 20 seconds then stop for 5 seconds, or what?

I guess what I'm saying is that I'm not sure this is all that useful without detail. I'd also want a log entry probably to tell me about each failure, so I could have an easy way to work out where things are failing due to this feature.

hamishwillee: Oh, and I will revert :-)

julianoes: Intermittently means it's publishing, and then it's not, and then it's publishing again. Basically a lose wire.

PR #1364, opened 2021-06-17
Communicating with individual vehicles using ROS2

max11gen:
I added a note on how to publish and subscribe to the topics of a specific instance of PX4 using ROS2. I found the explanation in [this answer](https://discuss.px4.io/t/how-to-isolate-topics-of-different-urtps-agent-if-i-have-multiplel-instance/21576/2) by @TSC21 and thought it might be easier to find for others in the docs.

### DIFF
--- a/en/simulation/multi_vehicle_simulation_gazebo.md
+++ b/en/simulation/multi_vehicle_simulation_gazebo.md
@@ -97,7 +97,9 @@ To build an example setup, follow the steps below:
    micrortps_agent -t UDP -r 2024 -s 2023 &
    micrortps_agent -t UDP -r 2026 -s 2025 &
    ```
-
+   :::note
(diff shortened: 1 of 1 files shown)


TSC21 (approved):
Thanks!

hamishwillee:
Thanks. So if you don't use `n` you can't subscribe to topic individually? I.e. this is a "have to"?

max11gen:
Exactly. If you don't use the `n` option to assign each vehicle its individual namespace, all vehicles publish to the same topic. Then - when you subscribe to e.g. sensors_combined_pubsub topic - you won't be able to tell which vehicle sent which message. Also if you publish e.g. to the vehicle_command topic, all vehicles will to the same thing. So I would go as far as saying that multivehicle control via ROS2 is not possible without the namespace parameter

hamishwillee:
@max11gen Thanks very much. I've modified the text a bit to make it clear this is "must" rather than "can" in https://github.com/PX4/PX4-user_guide/pull/1374

PR #1381, opened 2021-06-24
Update mRo Neo-M8N Dual Compass GPS

mrpollo:
The GPS previously using IST8310 has been deprecated in favor of the mRo GPS u-Blox Neo-M8N IST8308


hamishwillee:
Thanks for the fix!

PR #1394, opened 2021-07-11
Fix link for typhoon_h480 range finder

Jaeyoung-Lim:
**Problem Description**
This fixes the link for the `typhoon_h480` range finder


bkueng (approved):
Thanks

PR #1396, opened 2021-07-12
add events interface page

bkueng:
From https://github.com/PX4/PX4-Autopilot/pull/17479

PX4-specific usage docs, the protocol itself will be documented in MAVLink.

@julianoes fyi

### DIFF
--- a/en/concept/events_interface.md
+++ b/en/concept/events_interface.md
@@ -1,0 +1,128 @@
+# Events Interface
+
+The *Events Interface* provides a system-wide API to inform a user about certain events.
+It is meant to replace existing usage of `mavlink_log_*` calls.
+
+## Usage
+### Basic
+To use the API, add this include:
+```cpp
+#include <px4_platform_common/events.h>
+```
+And then define and send the event from the desired code location:
+```cpp
+events::send(events::ID("mymodule_test"), events::Log::Info, "Test Message");
+```
+
+#### Backward compatibility
+For older GCS versions without events interface support, PX4 currently sends out all events also as `mavlink_log_*` `STATUSTEXT` message.
+In addition, the message must be tagged with an appended tab (`\t`) so that newer GCS's can ignore that and only show the event.
+
+So whenever adding an event, be sure to also add a `mavlink_log_` call. For example:
+```cpp
+mavlink_log_info(mavlink_log_pub, "Test Message\t");
+events::send(events::ID("mymodule_test"), events::Log::Info, "Test Message");
+```
+All such `mavlink_log_` calls will be removed after the next release.
+
+### Detailed
+The above is a minimal example, this is a more extensive one:
+```cpp
+uint8_t arg1 = 0;
+float arg2 = -1.f;
+/* EVENT
+ * @description
+ * This is the detailed event description.
+ *
+ * - value of arg1: {1}
+ * - value of arg2: {2:.1}
+ *
+ * <profile name="dev">
+ * (This paragraph is only meant to be shown to developers).
+ * This behavior can be configured with the parameter <param>COM_EXAMPLE</param>.
+ * </profile>
+ *
+ * Link to documentation: <a>https://docs.px4.io</a>
+ */
+events::send<uint8_t, float>(events::ID("event_name"),
+	{events::Log::Error, events::LogInternal::Info}, "Event Message", arg1, arg2);
+```
+
+Explanations and requirements:
+- `/* EVENT`: this is an indicator for the parser that there is additional metadata following as comment.
+- **event_name**: the event name.
+  - must be unique within the whole source code of PX4.
+    As a general convention, prefix it with the module name, or the source file for larger modules.
+  - must be a valid variable name, i.e. must not contain spaces, colons, etc.
+  - from that name, a 24 bit event ID is derived using a hash function.
+    This means as long as the event name stays the same, the ID is as well.
+- **Log Level**:
+  - valid log levels are (in descending importance):
+    ```
+	Emergency,
+	Alert,
+	Critical,
+	Error,
+	Warning,
+	Notice,
+	Info,
+	Debug,
+	Disabled,
+	```
+  - It is possible to specify an internal log level (for the log file) that is different from the external (that is shown to the GCS user).
+    By default (if only a single log level is passed), both are the same, which is the majority of the use-cases.
+	However in some cases it makes sense to have two different log levels, for example an RTL failsafe action: the user should see it as Warning/Error, whereas in the log, it is an expected system response, so it can be set to Info.
+- **Event Message**:
+  - Single-line, short description of the event. It may contain arguments (see below).
+- **Event Description**:
+  - detailed, optional event description.
+  - can be multiple lines/paragraphs.
+
+#### Arguments and Enums
+Events can have a fixed set of arguments that can be shown in the message or description.
+
+Valid types: `uint8_t`, `int8_t`, `uint16_t`, `int16_t`, `uint32_t`, `int32_t`, `uint64_t`, `int64_t` and `float`.
+
+It is also possible to define and use custom enumerations.
+These are defined in [src/lib/events/enums.json](https://github.com/PX4/PX4-Autopilot/blob/master/src/lib/events/enums.json), and can then be used as event argument in the form of `events::send<events::px4::enums::my_enum_t>(...)`.
+
+#### Text format
+Text format for event message description:
+- characters can be escaped with \\
+
+  These have to be escaped: '\\\\', '\\<', '\\{'.
+- supported tags:
+  - Profiles: `<profile name="[!]NAME">CONTENT</profile>`
+
+    `CONTENT` will only be shown if the name matches the configured profile.
+	This can be used for example to hide developer information from end-users.
+  - URLs: `<a [href="URL"]>CONTENT</a>`.
+    If href is not set, use `CONTENT` as `URL`
+  - Parameters: `<param>PARAM_NAME</param>`
+  - no nested tags of the same type are allowed
+- arguments: following python syntax, with 1-based indexing (instead of 0)
+  - general form: `{ARG_IDX[:.NUM_DECIMAL_DIGITS][UNIT]}`
+
+    UNIT:
+      - m: horizontal distance in meters
+      - m_v: vertical distance in meters
+      - m^2: area in m^2
+      - m/s: speed in m/s
+      - C: temperature in degrees celcius
+
+## Logging
+Events are logged according to the internal log level, and [Flight Review](../log/flight_review.md) displays events.
+
+:::note
+Flight review downloads metadata based on PX4 master, so if a definition is not yet on master, it will only be able to display the event ID.
+:::
+
+
+## Implementation
+During build, only the event ID, log level(s) and arguments (if any) are added to the binary by the compiler.
+At the same time, a python script scans the whole source code for event calls, extracts the metadata and assembles a JSON metadata file.
+
+As for parameter metadata, the events metadata is [compiled into the firmware](../advanced/parameters_and_configurations.md#publishing-parameter-metadata-to-a-gcs),
+and made available to ground stations via the [MAVLink Component Information service](https://mavlink.io/en/services/component_information.html).
+
+
--- a/en/concept/events_interface.md
+++ b/en/concept/events_interface.md
@@ -124,6 +124,7 @@ Text format for event message description:
       - C: temperature in degrees celcius
 
 ## Logging
+
 Events are logged according to the internal log level, and [Flight Review](../log/flight_review.md) displays events.
 
 :::note
@@ -132,9 +133,14 @@ Flight review downloads metadata based on PX4 master, so if a definition is not 
 
 
 ## Implementation
-During build, only the event ID, log level(s) and arguments (if any) are added to the binary by the compiler.
+
+During PX4 build, only the event ID, log level(s) and arguments (if any) are added to the binary by the compiler.
+
 At the same time, a python script scans the whole source code for event calls, extracts the metadata and assembles a JSON metadata file.
 
-As for parameter metadata, the events metadata is [compiled into the firmware](../advanced/parameters_and_configurations.md#publishing-parameter-metadata-to-a-gcs), and made available to ground stations via the [MAVLink Component Information service](https://mavlink.io/en/services/component_information.html).
+### Publishing Event Metadata to a GCS
 
+The event metadata JSON file is compiled into firmware, and made available to ground stations via the [MAVLink Component Information service](https://mavlink.io/en/services/component_information.html).
+This ensures that metadata is always up-to-date with the code running on the vehicle.
 
+This process is the same as for [parameter metadata](../advanced/parameters_and_configurations.md#publishing-parameter-metadata-to-a-gcs)
--- a/en/concept/events_interface.md
+++ b/en/concept/events_interface.md
@@ -95,8 +95,9 @@ Events can have a fixed set of arguments that can be shown in the message or des
 
 Valid types: `uint8_t`, `int8_t`, `uint16_t`, `int16_t`, `uint32_t`, `int32_t`, `uint64_t`, `int64_t` and `float`.
 
-It is also possible to define and use custom enumerations.
-These are defined in [src/lib/events/enums.json](https://github.com/PX4/PX4-Autopilot/blob/master/src/lib/events/enums.json), and can then be used as event argument in the form of `events::send<events::px4::enums::my_enum_t>(...)`.
+You can also use enumerations as arguments:
+- PX4-specific/custom enumerations for events should be defined in [src/lib/events/enums.json](https://github.com/PX4/PX4-Autopilot/blob/master/src/lib/events/enums.json), and can then be used as event argument in the form of `events::send<events::px4::enums::my_enum_t>(...)`.
+- MAVLink "common" events are defined in [mavlink/libevents/events/common.json](https://github.com/mavlink/libevents/blob/master/events/common.json) and can be used as event argument in the form of `events::send<events::common::enums::my_enum_t>(...)`.
 
 #### Text format
 
--- a/en/concept/events_interface.md
+++ b/en/concept/events_interface.md
@@ -15,6 +15,7 @@ events::send(events::ID("mymodule_test"), events::Log::Info, "Test Message");
 ```
 
 #### Backward compatibility
+
 For older GCS versions without events interface support, PX4 currently sends out all events also as `mavlink_log_*` `STATUSTEXT` message.
 In addition, the message must be tagged with an appended tab (`\t`) so that newer GCS's can ignore that and only show the event.
 
@@ -26,6 +27,7 @@ events::send(events::ID("mymodule_test"), events::Log::Info, "Test Message");
 All such `mavlink_log_` calls will be removed after the next release.
 
 ### Detailed
+
 The above is a minimal example, this is a more extensive one:
 ```cpp
 uint8_t arg1 = 0;
@@ -55,7 +57,7 @@ Explanations and requirements:
     As a general convention, prefix it with the module name, or the source file for larger modules.
   - must be a valid variable name, i.e. must not contain spaces, colons, etc.
   - from that name, a 24 bit event ID is derived using a hash function.
-    This means as long as the event name stays the same, the ID is as well.
+    This means as long as the event name stays the same, so will the ID.
 - **Log Level**:
   - valid log levels are (in descending importance):
     ```
@@ -69,16 +71,19 @@ Explanations and requirements:
 	Debug,
 	Disabled,
 	```
-  - It is possible to specify an internal log level (for the log file) that is different from the external (that is shown to the GCS user).
-    By default (if only a single log level is passed), both are the same, which is the majority of the use-cases.
-	However in some cases it makes sense to have two different log levels, for example an RTL failsafe action: the user should see it as Warning/Error, whereas in the log, it is an expected system response, so it can be set to Info.
+  - Above we specify a separate external and internal log level, which are the levels displayed to GCS users and in the log file, respectively: `{events::Log::Error, events::LogInternal::Info}`.
+    For the majority of cases you can pass a single log level, and this will be used for both exernal and internal cases.
+	There are cases it makes sense to have two different log levels.
+	For example an RTL failsafe action: the user should see it as Warning/Error, whereas in the log, it is an expected system response, so it can be set to `Info`.
 - **Event Message**:
-  - Single-line, short description of the event. It may contain arguments (see below).
+  - Single-line, short description of the event.
+    It may contain arguments (see below).
 - **Event Description**:
   - detailed, optional event description.
   - can be multiple lines/paragraphs.
 
 #### Arguments and Enums
+
 Events can have a fixed set of arguments that can be shown in the message or description.
 
 Valid types: `uint8_t`, `int8_t`, `uint16_t`, `int16_t`, `uint32_t`, `int32_t`, `uint64_t`, `int64_t` and `float`.
@@ -87,6 +92,7 @@ It is also possible to define and use custom enumerations.
 These are defined in [src/lib/events/enums.json](https://github.com/PX4/PX4-Autopilot/blob/master/src/lib/events/enums.json), and can then be used as event argument in the form of `events::send<events::px4::enums::my_enum_t>(...)`.
 
 #### Text format
+
 Text format for event message description:
 - characters can be escaped with \\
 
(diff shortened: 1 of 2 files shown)


en/concept/events_interface.md
```diff
@@ -0,0 +1,128 @@
...
+#include <px4_platform_common/events.h>
+```
+And then define and send the event from the desired code location:
+```cpp
+events::send(events::ID("mymodule_test"), events::Log::Info, "Test Message");
+```
+
+#### Backward compatibility
+For older GCS versions without events interface support, PX4 currently sends out all events also as `mavlink_log_*` `STATUSTEXT` message.
+In addition, the message must be tagged with an appended tab (`\t`) so that newer GCS's can ignore that and only show the event.
```
julianoes, line 19: Aha! :smile:

julianoes (approved):
Awesome!

hamishwillee:
Is this in PX4 v1.12?

en/concept/events_interface.md
```diff
@@ -0,0 +1,128 @@
...
+ * </profile>
+ *
+ * Link to documentation: <a>https://docs.px4.io</a>
+ */
+events::send<uint8_t, float>(events::ID("event_name"),
+	{events::Log::Error, events::LogInternal::Info}, "Event Message", arg1, arg2);
+```
+
+Explanations and requirements:
+- `/* EVENT`: this is an indicator for the parser that there is additional metadata following as comment.
```
hamishwillee, line 52: "the parser" - what parser?

en/concept/events_interface.md
```diff
@@ -0,0 +1,128 @@
...
+- **event_name**: the event name.
+  - must be unique within the whole source code of PX4.
+    As a general convention, prefix it with the module name, or the source file for larger modules.
+  - must be a valid variable name, i.e. must not contain spaces, colons, etc.
+  - from that name, a 24 bit event ID is derived using a hash function.
+    This means as long as the event name stays the same, the ID is as well.
```
hamishwillee, line 58: "the ID is as well." > "the ID will not change".

en/concept/events_interface.md
```diff
@@ -0,0 +1,133 @@
...
+
+For older GCS versions without events interface support, PX4 currently sends out all events also as `mavlink_log_*` `STATUSTEXT` message.
+In addition, the message must be tagged with an appended tab (`\t`) so that newer GCS's can ignore that and only show the event.
+
+So whenever adding an event, be sure to also add a `mavlink_log_` call. For example:
+```cpp
+mavlink_log_info(mavlink_log_pub, "Test Message\t");
+events::send(events::ID("mymodule_test"), events::Log::Info, "Test Message");
+```
+All such `mavlink_log_` calls will be removed after the next release.
```
hamishwillee, line 27: So we will have to remove this section in next release?

en/concept/events_interface.md
```diff
@@ -0,0 +1,133 @@
...
+- **Event Message**:
+  - Single-line, short description of the event.
+    It may contain arguments (see below).
+- **Event Description**:
+  - detailed, optional event description.
+  - can be multiple lines/paragraphs.
+
+#### Arguments and Enums
+
+Events can have a fixed set of arguments that can be shown in the message or description.
```
hamishwillee, line 87: "numeric arguments"?

en/concept/events_interface.md
```diff
@@ -0,0 +1,133 @@
...
+#### Text format
+
+Text format for event message description:
+- characters can be escaped with \\
+
+  These have to be escaped: '\\\\', '\\<', '\\{'.
+- supported tags:
+  - Profiles: `<profile name="[!]NAME">CONTENT</profile>`
+
+    `CONTENT` will only be shown if the name matches the configured profile.
```
hamishwillee, line 103: What is a profile? When will this information be shown? 

My guess is that the information in the event description gets somehow parsed "somewhere" and that you can configure various parts that will go to various audiences. I assume this is different parts of docs? Or is is GCS? 

I guess what I'm saying is that I THINK this whole section is about the documentation part, but I'm not sure. I think perhaps a real example here would be useful.

en/concept/events_interface.md
```diff
@@ -0,0 +1,133 @@
...
+:::note
+Flight review downloads metadata based on PX4 master, so if a definition is not yet on master, it will only be able to display the event ID.
+:::
+
+
+## Implementation
+During build, only the event ID, log level(s) and arguments (if any) are added to the binary by the compiler.
+At the same time, a python script scans the whole source code for event calls, extracts the metadata and assembles a JSON metadata file.
+
+As for parameter metadata, the events metadata is [compiled into the firmware](../advanced/parameters_and_configurations.md#publishing-parameter-metadata-to-a-gcs), and made available to ground stations via the [MAVLink Component Information service](https://mavlink.io/en/services/component_information.html).
```
hamishwillee, line 131: In the same way as for parameter metadata

en/concept/events_interface.md
```diff
@@ -0,0 +1,133 @@
...
+- `/* EVENT`: this is an indicator for the parser that there is additional metadata following as comment.
+- **event_name**: the event name.
+  - must be unique within the whole source code of PX4.
+    As a general convention, prefix it with the module name, or the source file for larger modules.
+  - must be a valid variable name, i.e. must not contain spaces, colons, etc.
+  - from that name, a 24 bit event ID is derived using a hash function.
+    This means as long as the event name stays the same, so will the ID.
+- **Log Level**:
+  - valid log levels are (in descending importance):
+    ```
```
hamishwillee, line 63: This is good. Do we have any guidance on which you should use and when?

en/concept/events_interface.md
```diff
@@ -0,0 +1,133 @@
+# Events Interface
+
+The *Events Interface* provides a system-wide API to inform a user about certain events.
```
hamishwillee, line 3: Perhaps a little more background and preparation for the existence of the MAVLink docs. For example, some examples would be nice. Maybe something like the below.

---

The *Events Interface* provides a system-wide API for notification of events, which are published to GCSs via the MAVLink Events Service [LINK WHEN AVAILABLE] (to GCSs) and also stored in system logs [LINK TO LOGS DOCS].

The interface can be used for publishing events for state changes or any other type of occurrence, including things like arming readiness, [WHAT ELSE DO WE USE STATUS TEXT FOR?]

:::note
The events interface will replace the use of `mavlink_log_*` calls in PX4 code, (and `STATUS_TEXT` messages in MAVLink) for event notification following (???) PX4 v1.12.
There will be an intermediate period where both approaches are supported [LINK TO SECTION BELOW ON BACKWARDS COMPAT]
:::

hamishwillee: @bkueng Does the MAVLink spec cover the concept of common/managed enums? Specifically, if it is likely that a GCS will configure itself based on events (say display "ready to ARM" when a readiness event is received) then it makes sense that this kind of event is shared between GCS so that nothing needs to be coded twice. 

@julianoes Perhaps madness, but could/should we reasonably use events as a generic mavlink notifier "in future"? What I mean is that for command protocol we're tdiscussing mechanisms to tell us completion of long running events. We could instead bypass completely and define a Takeoff complete event. I appreciate this depends on events interface being broadly implemented, but is the idea "sane"?

en/concept/events_interface.md
```diff
@@ -0,0 +1,133 @@
...
+It is meant to replace existing usage of `mavlink_log_*` calls.
+
```
hamishwillee, line 5: @bkueng Does the MAVLink spec cover the concept of common/managed enums? Specifically, if it is likely that a GCS will configure itself based on events (say display "ready to ARM" when a readiness event is received) then it makes sense that this kind of event is shared between GCS so that nothing needs to be coded twice.
 
@julianoes Perhaps madness, but could/should we reasonably use events as a generic mavlink notifier "in future"? What I mean is that for command protocol we're tdiscussing mechanisms to tell us completion of long running events. We could instead bypass completely and define a Takeoff complete event. I appreciate this depends on events interface being broadly implemented, but is the idea "sane"?

en/concept/events_interface.md
hamishwillee: Or to put it another way, it isn't clear what events are generated into. I know we do generate the ids and metadata, but perhaps also docs.

bkueng: Would `metadata parser` be clearer for you? The output could be different things, docs could also be generated in future.

en/concept/events_interface.md
bkueng: Correct, after the next release.

en/concept/events_interface.md
bkueng: The concept needs a bit of explanation (on the events interface level, but could be repeated here): It's used to show certan text to specific audiences (on the GCS). For example you would only show 'this check can be disabled via parameter xy' to a developer, but not an end-user. The GCS then has a config option if the user is a developer or end-user.
The is an example already above.

en/concept/events_interface.md
bkueng: Right

en/concept/events_interface.md
bkueng: No, but certainly wouldn't hurt. It's the same levels as mavlink. Does that have any guidance?

en/concept/events_interface.md
bkueng: Yes, certainly. Ideally we'll link to the (not-yet-existing) mavlink docs

en/concept/events_interface.md
bkueng: This is on the level of events, not mavlink. And yes there is a shared set of enumerations/messages: https://github.com/mavlink/libevents/blob/master/events/common.json

> @julianoes Perhaps madness, but could/should we reasonably use events as a generic mavlink notifier "in future"? What I mean is that for command protocol we're tdiscussing mechanisms to tell us completion of long running events. We could instead bypass completely and define a Takeoff complete event. I appreciate this depends on events interface being broadly implemented, but is the idea "sane"?

I'm not Julian, but my answer: I would not tie it to certain mavlink commands, but something like a 'takeoff complete' event makes sense and is something I intended to use it for. Such events will be defined in the common namespace: https://github.com/mavlink/libevents/blob/master/events/common.json

bkueng:
> Is this in PX4 v1.12?

No, this is for the next release.

en/concept/events_interface.md
hamishwillee: That's great - even though you are not Julian :-).

en/concept/events_interface.md
hamishwillee: OK, I have updated this section, please check.

en/concept/events_interface.md
hamishwillee: Not really. At this point in the doc you'd either link to information about the parser OR you might modify the text from

> this is an indicator for the parser that there is additional metadata following as comment.

To

> This tag indicates the start of a metadata definition section.

The problem here is that when I saw parser, I wanted to know what the parser does and can do/cannot do. We need to document things like:
- What the parser generates
- we know component info file - is the format anywhere that can be linked to
- what about translation information? 
- docs aren't generated yet - we should say that, unless we plan to generate docs for the events in near future.

Does that make sense?

en/concept/events_interface.md
hamishwillee: Yes, I will add them https://mavlink.io/en/messages/common.html#MAV_SEVERITY

en/concept/events_interface.md
```diff
@@ -0,0 +1,140 @@
...
+  - detailed, optional event description.
+  - can be multiple lines/paragraphs.
+
+#### Arguments and Enums
+
+Events can have a fixed set of arguments that can be shown in the message or description.
+
+Valid types: `uint8_t`, `int8_t`, `uint16_t`, `int16_t`, `uint32_t`, `int32_t`, `uint64_t`, `int64_t` and `float`.
+
+It is also possible to define and use custom enumerations.
```
hamishwillee, line 98: Given your answer to https://github.com/PX4/PX4-user_guide/pull/1396/files#r669777075 don't we need to refer to the standard enums too? How do users link to those, or are they somehow part of src/lib/events/enums.json ?

en/concept/events_interface.md
```diff
@@ -0,0 +1,140 @@
...
+	Info,
+	Debug,
+	Disabled,
+	```
+  - Above we specify a separate external and internal log level, which are the levels displayed to GCS users and in the log file, respectively: `{events::Log::Error, events::LogInternal::Info}`.
+    For the majority of cases you can pass a single log level, and this will be used for both exernal and internal cases.
+	There are cases it makes sense to have two different log levels.
+	For example an RTL failsafe action: the user should see it as Warning/Error, whereas in the log, it is an expected system response, so it can be set to `Info`.
+- **Event Message**:
+  - Single-line, short description of the event.
```
hamishwillee, line 86: is this `@description`. Might be worth mentioning that.

en/concept/events_interface.md
hamishwillee: > The is an example already above.

The problem is that it feels ambiguous. For example `* - value of arg1: {1}` could mean lots of things - i.e. perhaps you mean literal text `value of arg1:`, or perhaps it is placeholder text meaning `arg_1_is_a_foo`. A real example would make it clear which of them. 

With respect to profiles - fair enough. I guess we can link/revisit once those docs exist.

en/concept/events_interface.md
```diff
@@ -0,0 +1,146 @@
...
+Events are logged according to the internal log level, and [Flight Review](../log/flight_review.md) displays events.
+
+:::note
+Flight review downloads metadata based on PX4 master, so if a definition is not yet on master, it will only be able to display the event ID.
+:::
+
+
+## Implementation
+
+During PX4 build, only the event ID, log level(s) and arguments (if any) are added to the binary by the compiler.
```
hamishwillee, line 137: Technically the metadata is also added to the binary ... as a JSON file :-). Is it worth modifying to something like ...

---

During PX4 build, only the code is added directly to the binary by the compiler (i.e. the event ID, log level(s) and any arguments).

The metadata for all events is compiled into a separate JSON metadata file (using a python script that scans the whole source code for event calls).

en/concept/events_interface.md
bkueng: Ok, updated

en/concept/events_interface.md
bkueng: I don't follow your ambiguity, but feel free to add a real example (or change as you see fit).

en/concept/events_interface.md
bkueng: Looks good, thanks

en/concept/events_interface.md
bkueng: We can if you want to (I don't expect there to be a lot of use for that though, at least initially).
The change is simple: `events::common::enums::my_enum_t`.

en/concept/events_interface.md
bkueng: No, `Event Description` is the description.

en/concept/events_interface.md
bkueng: > Technically the metadata is also added to the binary ... as a JSON file :-)

Yes but not by the compiler :). Ok, your version is fine with me.

hamishwillee: Fair enough - its a bit pedantic. Made the change anyway :-)

en/concept/events_interface.md
hamishwillee: Thanks very much. I've added this, because even though you're right and it might not be useful "yet", best to capture while it is fresh in my mind.

en/concept/events_interface.md
hamishwillee: Catch 22 - how can I add a real example if I don't understand the syntax. I searched and there aren't any in the codebase where there are descriptions for arguments.
The only example is one that has just the single line description.
```
/* EVENT
 * @description
 * Land immediately and check the system.
 */
events::send<events::px4::enums::sensor_type_t, uint8_t, events::px4::enums::sensor_failover_reason_t>(
  events::ID("sensor_failover"), events::Log::Emergency, 
  "{1} sensor #{2} failure: {3}", sensor_type, failover_index,
						failover_reason);
```

So for the above, what would the arguments look like - e.g. what should be in place in this event for these in your pseudo example?:
```
 * - value of arg1: {1}
 * - value of arg2: {2:.1}
 ```

**Comment by hamishwillee** [comment]
@bkueng I'm merging this because it is definitely useful - and "my" open questions should not block it. I hope we can iterate with information on support for translation etc - let's do a round after the MAVLink docs go in. 

Did you have a plan/timeframe for putting those docs into mavlink.io, or are you hoping I will do first draft?

File: en/concept/events_interface.md (hunk 6 above)
Comment (bkueng) [reply en/concept/events_interface.md:111]: Sorry I still don't follow.
In the example, the user would see this:
```
 * - value of arg1: 0
 * - value of arg2: -1.0
```

**Comment by bkueng** [comment]
> Did you have a plan/timeframe for putting those docs into mavlink.io, or are you hoping I will do first draft?

I will do that, but I wanted to first get a bit of field-testing, so if changes are required, I don't need to do docs twice. So maybe 1-2 months?

File: en/concept/events_interface.md (hunk 6 above)
Comment (hamishwillee) [reply en/concept/events_interface.md:111]: Thanks!

Comment (bkueng) [reply en/concept/events_interface.md:111]: So you think we should have a different example?

Comment (hamishwillee) [reply en/concept/events_interface.md:111]: @bkueng No, I completely don't get this, but I'm assuming I'm just being thick, and it will be obvious. Perhaps @julianoes could read it and see if does?

What I had hoped for was a real event in our codebase that had argument documentation - basically used the entire syntax in a real case. With that the docs could link to the real example so even if I don't get it, others might. 
I think perhaps this will be more clear once the mavlink docs are present - my confusion is probably indicative that I don't fully understand/remember what we're trying to do.

For example, I'm see from the [proposal](https://docs.google.com/document/d/18qdDgfML97lItom09MJhngYnFzAm1zFdmlCKG7TaBpg/edit#heading=h.l8v2vka1vs88) and the [EVENT](https://mavlink.io/en/messages/common.html#EVENT) message that the event has a 40 byte buffer for the arguments. My original assumption was that this means 40 arguments, but now I think maybe the buffer is "packed" with the arguments, and it is up to the event definitions to define how this is done.

IF that is correct, then shouldn't the event code be dynamically generated from the event definition file rather than explicitly written in our code? Then if we want to send an event or receive and event we use handy methods autogenerated for us - rather than having to hand code everything like this? I.e. the same way we unpack messages?

ANyway, still confused. Sorry.

Comment (hamishwillee) [reply en/concept/events_interface.md:111]: Let me try again using the real example ....
```
/* EVENT
 * @description
 * Land immediately and check the system.
 */
events::send<events::px4::enums::sensor_type_t, uint8_t, events::px4::enums::sensor_failover_reason_t>(
  events::ID("sensor_failover"), events::Log::Emergency, 
  "{1} sensor #{2} failure: {3}", sensor_type, failover_index,
						failover_reason);
```

Here
- we just have the one line description with the text "Land immediately and check the system.". 
  - The GCS presumably has a shared understanding based on the event id (`sensor_failover`) and can associate the above message from component information metadata with the event.
- We have three arguments, sensor_type, failover_index, failover_reason, we know they are "bytes" from the `send` signature. 
- We have the message itself  `{1} sensor #{2} failure: {3}` (so `{sensor_type1} sensor #{failover_index} failure: {failover_reason}`). I assume this is the format of the message in the metadata which is logged - i.e. this goes into metadata to allow reconstruction of the message, but is not actually sent.
- So there is no argument information in this definition.  Looking at your original example my guess is that this provides information about how the argument is supposed to be rendered at the other end and in text - it isn't actually about a _particular_ value. So this tells us that arg 1 is just rendered as whatever value is passed into the argument. It tells us that the second argument is presented as a number with `NUM_DECIMAL_DIGITS` (i.e. it is a number like 2.1, not 2.11). Is that right? So the essentially in this real example above we're just saying that this is the default - there is no formatting - it is just a number without decimal points or units?
  ```
  * - value of arg1: {1}
  * - value of arg2: {2:.1}
 ```

bkueng: > now I think maybe the buffer is "packed" with the arguments, and it is up to the event definitions to define how this is done.

Yes that is correct.

> This tells us that arg 1 is just rendered as whatever value is passed into the argument. It tells us that the second argument is presented as a number with NUM_DECIMAL_DIGITS (i.e. it is a number like 2.1, not 2.11). Is that right?

Correct

> so if the value is rendered without decimal points or units then you just omit the arguments.

No, the decimal points is just to configure the rendering. For int's you would not need it.

In general there's 2 use-cases:
- event definition in px4 source: in that case you don't typically care about the arguments, but just about the message (which may or may not print the argument(s)). In that case the parser will just fill in the args automatically, no special code generation needed.
- for events in common namespace, there is code generation, and you have the equivalent as in mavlink to get the arguments (not documented here, as not yet used). But you can still just use it as in the previous case when only caring about the message.

Does that explain it a bit better?

hamishwillee: > Does that explain it a bit better?

Definitely getting there! This is part of the story that I was missing - there is are two ways to define events. I had assumed that we would follow a MAVLINK dialect-like model for defining dialect/non-standard events. In other words you'd define the events in JSON or XML or whatever for your dialect and that would generate the methods to both send and extract your event and also the metadata.

Instead, my understanding is now that the common **events** are predefined and some PX4 **enums** are predefined. However for PX4 (say) events are generated into the required metadata at build time. If the value is a PX4 enum then the enum itself defines the size of the data for the argument, if not then this can presumably be extracted from the method signature. Either way, this is turned into event metadata a bit like https://github.com/mavlink/libevents/blob/master/events/common.json#L291 for the PX4 event (presumably if this uses enums in common or PX4 they are cross linked magically). Right?

Why not just define the whole event in the PX4 .json file? I can see it might make things faster to define, but presumably QGC would have to write custom code to extract the PX4 events - which seems a little harder?

> > so if the value is rendered without decimal points or units then you just omit the arguments.

> No, the decimal points is just to configure the rendering. For int's you would not need it.

Not sure how this clarification differs from what I said.
- The `* - value of arg2:` defines how the arg2 will be rendered. So imagine this is  `* - value of arg2:` {2:.2m}`, that says whenever a message has the placeholder `{2}` it will render the argument in the format indicated. So assuming arg 2 is a real number, 5.4556 and we had a message definition "The distance is: {2}"it will be rendered as "The distance is: 5.45m`
- You only need decimal for reals. YOu might need the units for any of them. 
- Right?

I think perhaps "value of arg" was a bit unhelpful for the identifier - it wasn't even clear to me at first this was actually an identifier from the point of view of the parser. Might have been more intuitive as "@arg1_format_string:`
---

Is there any more context for params and URLs? Sometimes we use the term param meaning arguments? What are URLS and params for?/how are they used?/how do they map to the generated event information? 

Above tells us they exist, but not sure why they are needed/when they are needed, how they map to a common event (say).

bkueng: > Right?

Correct.

> Why not just define the whole event in the PX4 .json file?

Because of maintainability & compatibility: I've seen it myself already, if I had to edit an external file for 100+ events, to then use them in code, it's just too cumbersome. And also harder to trace an error message in code.
Then we want to avoid QGC knowing about as many events as possible, so we can freely update events on the PX4 side.

> You only need decimal for reals. YOu might need the units for any of them.

Yes

>  "value of arg" was a bit unhelpful for the identifier

It's not the identifier. `{x}` is, where x is the argument index.

bkueng: > Is there any more context for params and URLs? Sometimes we use the term param meaning arguments? What are URLS and params for?/how are they used?/how do they map to the generated event information?

The idea is that you can have urls and px4 params in descriptions. And there's a special syntax so that the parser can generate a link out of it for displaying (this is application-specific). E.g. clicking on a param will open the param editor.

hamishwillee: >> "value of arg" was a bit unhelpful for the identifier

> It's not the identifier. {x} is, where x is the argument index.

In a real example, might the actual text that appears be `* - value of arg2: {2:.1}`? What I mean by saying "identifier" is that the text `value of arg` is something the parser looks for in order to tell that the format of the string used for presenting arguments is what follows. If that is incorrect then I reiterate, need a pointer to code that shows an event that includes use of a URL. 

Can you add an actual case of URL and param in your example that we can refer to?

hamishwillee: > Because of maintainability & compatibility: ...

Makes a lot of sense. Does/can QGC still automatically ease extraction of the actual arguments and strings from the metadata or do you still have to create code to do that in QGC every time you add/update a new PX4 side event?

bkueng: > What I mean by saying "identifier" is that the text value of arg is something the parser looks for in order to tell that the format of the string used for presenting arguments is what follows.

Yes that is incorrect. The parser only cares about `{...}` and tags `<...>...</...>`.
The url example is also in there above: `<a>https://docs.px4.io</a>`, which would just turn into plain text `https://docs.px4.io` by the default parser config.

> Does/can QGC still automatically ease extraction of the actual arguments and strings from the metadata or do you still have to create code to do that in QGC every time you add/update a new PX4 side event?

It's all automatic. Meaning you can use any version of QGC that supports the events interface.

hamishwillee: Finally I get it. The description is whatever text you like with some placeholders for urls that will get generated in by the parser. Some tags are also interpreted in a particular way - e.g. `<a>`

PR #1403, opened 2021-07-14
docs: tri airframe setup. 

### LINKED ISSUE #725: Airframe config for tricopter configuration not correct
Inoticed that the airframe config for the tricopter airframe have incorrect prop directions. 

As in: https://docs.px4.io/master/en/airframes/airframe_reference.html#tricopter-y

@hamishwillee How is the image generated? The prop layout is only marked on the image and I am not sure how we can change this

DanAbara:
### Issue
The docs for airframe config for the tricopter airframe had all propellers rotating in same direction.

### Solution
- Added a comment that the software does not take the propeller directions into account.
- Updated images in airframe config. so that front propellers turn in opposite directions.

Fixes #725


.idea/.gitignore
```diff
@@ -0,0 +1,8 @@
+# Default ignored files
```
hamishwillee, line 1: @DanAbara Please remove all the changes under .idea folder from this PR.

en/airframes/airframe_reference.md
```diff
@@ -733,6 +733,15 @@ This page lists all supported airframes and types including
 </tbody></table>
 
 ### Tricopter Y+
+> **Note:**
```
hamishwillee, line 736: Please remove all changes to this file. It is not that they are "wrong", it is that this content is autogenerated from source, so we have to make the changes in the source. 

Unfortunately we can't - there isn't any provision in the [airframe format](https://github.com/PX4/PX4-Autopilot/blob/master/ROMFS/px4fmu_common/init.d/airframes/14001_tri_y_yaw%2B) to add more information about a single airframe, let alone comparative stuff about a class of airfames. The best I think we might do right now is use the  `# @url` in the tricopter airframe definitions (e.g. https://github.com/PX4/PX4-Autopilot/blob/master/ROMFS/px4fmu_common/init.d/airframes/14001_tri_y_yaw%2B) and link to this information on a page in docs about tricopters. 

@bkueng Does that make sense - using the #url on the "generic" airframe of each type to link to generic documentation about each airframe type?
If so, do you think it makes sense to make this explicit in the airframe reference - if there is a generic link add comment to the top of the section?
image

That said, whatever we do this comment will be easy to miss, so is there a way we could put it in the source, ideally shared by both (but if not, at least in one). I'm thinking an `@description` tag or  `@note` tag or similar.

DanAbara: I have done as requested. Please review.

hamishwillee:
Looks great! Thanks very much @DanAbara !

PR #1408, opened 2021-07-16
Correct image for FC wiring for Holybro x500 airframe

ykompis:
Uploaded a new photo and reverted the changes of the last commit (except for the correct connection names)


hamishwillee:
Thank you very much. The image is quite big for the information conveyed. Do you think it would be possible to reduce it to below 200KB?

Generally we try to keep images small to reduce the cost for people on slower links.

ykompis:
Sure. Reduced the size to ~170kb.

hamishwillee:
Thanks very much!

PR #1409, opened 2021-07-16
Clarified rate requirements for external vision fusion

ykompis:
Explicitly stating that EKF2 will not fuse external vision data if the rate is not high enough. While this will not be an issue for most Mocap systems, VIO framerates of 10-20Hz are not uncommon.


hamishwillee:
Excellent. Thanks for the clarification

PR #1410, opened 2021-07-16
Clarifications on External Vision EKF2 Delay Tunig

ykompis:
Explicitly state that SDLOG_PROFILE needs to be correctly set to log data form external vision. The plot shown contains data that is not logged by default.
Explicitly mention that neither the online Flight Review tool nor MAVGCL are able to plot the necessary data. This is not apparent from the flight log analysis page and caused some minor headaches.
This list is not complete, just the first two loggers I tried (and failed) with.


TSC21 (approved):

hamishwillee:
Thanks very much!

PR #1413, opened 2021-07-20
Add instructions to run ignition gazebo

Jaeyoung-Lim:
**Problem Description**
This PR adds instructions on how to run `ignition` gazebo, which is a new simulation that is supported for SITL testing in PX4.

The PR was merged in: https://github.com/PX4/PX4-Autopilot/pull/17889

A demonstration of the simulator can be found in the following video
[ignition gazebo sitl](https://youtu.be/38UJqrNQChg "Circular trajectory tracking")

### DIFF
--- a/en/simulation/ignition_gazebo.md
+++ b/en/simulation/ignition_gazebo.md
@@ -1,13 +1,11 @@
 # Ignition Gazebo Simulation
 
 :::warning
-Ignition Gazebo has just been added to PX4 (July 2020) and only has basic support in PX4.
-It supports a single frame and world, and cannot be run in headless mode.
+Ignition Gazebo supports a single frame (quadcopter) and world, and cannot be run in headless mode (July 2021).
 :::
 
-
 [Ignition Gazebo](https://ignitionrobotics.org/libs/gazebo) is an open source robotics simulator from the [Ignition Robotics Project](https://ignitionrobotics.org/home).
-Ignition Gazebo is derived from the popular robotics simulator [Gazebo](./gazebo.md), featuring more advanced rendering, physics and sensor models.
+It is derived from the popular robotics simulator [Gazebo](./gazebo.md), featuring more advanced rendering, physics and sensor models.
 
 **Supported Vehicles:** Quadrotor
 
--- a/en/simulation/ignition_gazebo.md
+++ b/en/simulation/ignition_gazebo.md
@@ -1,5 +1,11 @@
 # Ignition Gazebo Simulation
 
+:::warning
+Ignition Gazebo has just been added to PX4 (July 2020) and only has basic support in PX4.
+It supports a single frame and world, and cannot be run in headless mode.
+:::
+
+
 [Ignition Gazebo](https://ignitionrobotics.org/libs/gazebo) is an open source robotics simulator from the [Ignition Robotics Project](https://ignitionrobotics.org/home).
 Ignition Gazebo is derived from the popular robotics simulator [Gazebo](./gazebo.md), featuring more advanced rendering, physics and sensor models.
 
@@ -18,7 +24,7 @@ These instructions were tested on Ubuntu 18.04 and Ubuntu 20.04
 :::
 
 1. Install the usual [Development Environment on Ubuntu LTS / Debian Linux](../dev_setup/dev_env_linux_ubuntu.md).
-1. Install Ignition Gazebo following the installation [instructions]()
+1. Install Ignition Gazebo following the [installation instructions](https://github.com/Auterion/px4-simulation-ignition#readme) (`sudo` may be required):
    ```sh
    sh -c 'echo "deb http://packages.osrfoundation.org/gazebo/ubuntu-stable `lsb_release -cs` main" > /etc/apt/sources.list.d/gazebo-stable.list'
    wget http://packages.osrfoundation.org/gazebo.key -O - | sudo apt-key add -
@@ -29,7 +35,7 @@ These instructions were tested on Ubuntu 18.04 and Ubuntu 20.04
 ## Running the Simulation
 
 Ignition Gazebo SITL simulation can be conveniently run through a `make` command as shown below:
-```sh
+```bash
 cd /path/to/PX4-Autopilot
 make px4_sitl ignition
 ```
--- a/en/simulation/ignition_gazebo.md
+++ b/en/simulation/ignition_gazebo.md
@@ -11,7 +11,6 @@ Ignition Gazebo is derived from the popular robotics simulator [Gazebo](./gazebo
 See [Simulation](../simulation/README.md) for general information about simulators, the simulation environment, and simulation configuration (e.g. supported vehicles).
 :::
 
-<a id="installation"></a>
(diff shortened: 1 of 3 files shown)


en/simulation/ignition_gazebo.md
```diff
@@ -0,0 +1,50 @@
...
+## Installation (Ubuntu Linux)
+
+:::note
+These instructions were tested on Ubuntu 18.04 and Ubuntu 20.04
+:::
+
+1. Install the usual [Development Environment on Ubuntu LTS / Debian Linux](../dev_setup/dev_env_linux_ubuntu.md).
+1. Install Ignition Gazebo following the installation [instructions]()
+   ```sh
+   sh -c 'echo "deb http://packages.osrfoundation.org/gazebo/ubuntu-stable `lsb_release -cs` main" > /etc/apt/sources.list.d/gazebo-stable.list'
```
hamishwillee, line 23: 1. For me all of this needed prefix of `sudo`. Should we add that?
2. Will we be testing ignition gazebo regularly? If so, can we get this added to the setup script?

hamishwillee:
This looks cool. 
- Can we also get information updated in http://docs.px4.io/master/en/simulation/ ?
- Can this live in parallel with normal gazebo (I assume so based on the installation instructions)? I
- Can we get the installation as part of https://github.com/PX4/PX4-Autopilot/blob/master/Tools/setup/ubuntu.sh ? Ideally all the SIMS should be separately installable.
- What is the long term story here? 
  - Is it that ignition gazebo will replace gazebo and where? i.e. is PX4 eventually moving to this version? Or will the gazebo project move to this version? Will ROS move to this version.
  - Why would I choose to use it now?
  - Can I run video etc in it, change worlds etc, or is it just what you have documented - e.g. first steps?

hamishwillee:
FYI, I tried this in a fresh Ubuntu 20.04 in a VM. The installation instructions work. The UI renders with horrible flicker (unusable). It is much faster to start than normal Gazebo though. Does not support HEADLESS=1.


There are various warnings in the build like
```
In file included from /home/ubuntu/github/PX4-Autopilot/mavlink/include/mavlink/v2.0/common/common.h:2368,
                 from /home/ubuntu/github/PX4-Autopilot/mavlink/include/mavlink/v2.0/common/mavlink.h:32,
                 from /home/ubuntu/github/PX4-Autopilot/Tools/simulation-ignition/include/gazebo_mavlink_interface.h:75,
                 from /home/ubuntu/github/PX4-Autopilot/Tools/simulation-ignition/src/gazebo_mavlink_interface.cpp:22:
/home/ubuntu/github/PX4-Autopilot/mavlink/include/mavlink/v2.0/common/./mavlink_msg_landing_target.h: In function ‘uint16_t mavlink_msg_landing_target_encode(uint8_t, uint8_t, mavlink_message_t*, const mavlink_landing_target_t*)’:
/home/ubuntu/github/PX4-Autopilot/mavlink/include/mavlink/v2.0/common/./mavlink_msg_landing_target.h:218:349: warning: taking address of packed member of ‘__mavlink_landing_target_t’ may result in an unaligned pointer value [-Waddress-of-packed-member]
  218 |     return mavlink_msg_landing_target_pack(system_id, component_id, msg, landing_target->time_usec, landing_target->target_num, landing_target->frame, landing_target->angle_x, landing_target->angle_y, landing_target->distance, landing_target->size_x, landing_target->size_y, landing_target->x, landing_target->y, landing_target->z, landing_target->q, landing_target->type, landing_target->position_valid);
      |                                                                                                                                                                                                                                                                                                                                             ~~~~~~~~~~~~~~~~^
```
At runtime, get these fairly common warnings too

```
WARN  [logger] Too many subscriptions, failed to add: obstacle_distance_fused 0
WARN  [logger] Too many subscriptions, failed to add: vehicle_mocap_odometry 0
WARN  [logger] Too many subscriptions, failed to add: vehicle_trajectory_waypoint 0
WARN  [logger] Too many subscriptions, failed to add: vehicle_trajectory_waypoint_desired 0
```

Jaeyoung-Lim:
> Can we also get information updated in http://docs.px4.io/master/en/simulation/ ?

Done!

> Can this live in parallel with normal gazebo (I assume so based on the installation instructions)? I

Yes!

> Can we get the installation as part of https://github.com/PX4/PX4-Autopilot/blob/master/Tools/setup/ubuntu.sh ? Ideally all the SIMS should be separately installable.

I don't think this should be installed by default, since by default we are installing gazebo.

> What is the long term story here?
Is it that ignition gazebo will replace gazebo and where? i.e. is PX4 eventually moving to this version? Or will the gazebo project move to this version? Will ROS move to this version.

According to OSRF, ignition will replace gazebo. However, gazebo is currently the most commonly used simulator, and for this reason PX4 will not be replacing gazebo. I imagine we would need to support gazebo as long as OSRF is releasing it. :smile:

> Why would I choose to use it now?
Can I run video etc in it, change worlds etc, or is it just what you have documented - e.g. first steps?

Only what I have documented is supported.

> There are various warnings in the build like

This is something coming from mavlink. I will take a look on how we can suppress the warnings. 

> FYI, I tried this in a fresh Ubuntu 20.04 in a VM. The installation instructions work. The UI renders with horrible flicker (unusable). It is much faster to start than normal Gazebo though. Does not support HEADLESS=1.

As the instructions do not state the `HEADLESS=1` option, this is not supported. (The current implementation does not have a detached simulation available) . About the flicker, I think this is related to the GPU not being available to the VM? I have tested on my laptop both for fresh ubuntu 18.04 and ubuntu 20.04 and didn't have any issues.

hamishwillee:
Thanks! I'm going to be OOF until next week but will look at this again then. What I do note though:

>> Can we get the installation as part of https://github.com/PX4/PX4-Autopilot/blob/master/Tools/setup/ubuntu.sh ? Ideally all the SIMS should be separately installable.

> I don't think this should be installed by default, since by default we are installing gazebo.

I just tested and you can install both at the same time. So why not install both? 

But actually what I would like to do is change our script a bit so that you can rather than install everything and do (examples, not final call :-) ):  --NO-SIM you can also do --SIM-GAZ, --SIM-JMAV, --SIM-IGNITGAZ, This would disable installation of all sim and do just the sims you specify.

Jaeyoung-Lim:
> But actually what I would like to do is change our script a bit so that you can rather than install everything and do (examples, not final call :-) ): --NO-SIM you can also do --SIM-GAZ, --SIM-JMAV, --SIM-IGNITGAZ, This would disable installation of all sim and do just the sims you specify.

@hamishwillee I agree. I am expecting less people running this sim compared to gazebo and this will create unnecessary conflicts with other projects people have. Since this project doesn't have much additional dependencies (non actually) IMHO it would be better to just tell people who want to use this sim to install it.

hamishwillee:
Thanks very much. I've added a warning up the top that it only has basic support for one frame, world and not headless. I was thinking we should provide context as to why it was added, and isn't more complete. But that is perhaps not necessary as this isn't put forward as a major option. As we add more features we can remove the note.

PR #1415, opened 2021-07-20
Debug improvements

julianoes:
I recently used the Segger debugger instead of the Dronecode probe, and I was surprised that I couldn't find the information to connect GDB to it without setting up an IDE and all. So I added that here, and I cleaned up a few small things that caught my mind reading through the debug docs.


en/debug/gdb_debugging.md
```diff
@@ -24,52 +18,24 @@ top
 
 Stack usage is calculated with stack coloring and thus is not the current usage, but the maximum since the start of the task.
 
-### Heap allocations
```
hamishwillee, line 27: YOu moved this section into Simulation debugging, which is great. Does this topic need to be covered for the hardware case too - or is that done by the console commands you added?

en/debug/gdb_debugging.md
```diff
@@ -1,14 +1,8 @@
 # Embedded Debugging
```
hamishwillee, line 1: Should this file be renamed? Seems to be no longer about GDB.

hamishwillee:
@julianoes Excellent improvements. I'm going to merge, but I have a few questions that may warrant follow up. Essentially they come down to "do you think we can/should provide some structure to the Debugging/Logging topic.

The "Embedded Debugging" starts with "The autopilots running PX4 support debugging via GDB or LLDB." and then contains topics on hard fault debugging and identifying large memory consumers.

Were thinking of this as like a "Parent" doc for topics related to debugging on hardware (hence the first line). That seems logical, but then you have a mixed bag of other topics that fit inside it?
 
The reason I ask is because we name a bunch of these topics debugging, but then logging, profiling etc are also forms of debugging. Actually might be best off making it clear that we mean interactive debugging (is there a better name?). 

- I think that gathering all the debugging on hardware together under a parent might be useful.
- I think perhaps adding an overview on your options for debugging might be worth doing inhttp://docs.px4.io/master/en/debug/. I think it would be useful for people to understand when you might use interactive debugging, what you can use the simulator for vs when hardware must be used - where logging helps etc.

Thoughts?

en/debug/gdb_debugging.md
julianoes: I would not know how it can be done on hardware.

en/debug/gdb_debugging.md
julianoes: Yes, should be.

julianoes:
> The reason I ask is because we name a bunch of these topics debugging, but then logging, profiling etc are also forms of debugging. Actually might be best off making it clear that we mean interactive debugging (is there a better name?).

I don't know of a better name :thinking:.

> - I think that gathering all the debugging on hardware together under a parent might be useful.

Yes, these edits already went in that direction.

> - I think perhaps adding an overview on your options for debugging might be worth doing inhttp://docs.px4.io/master/en/debug/. I think it would be useful for people to understand when you might use interactive debugging, what you can use the simulator for vs when hardware must be used - where logging helps etc.

Yes, that would be nice.

PR #1417, opened 2021-07-21
Add disclaimer about Telega FOC tuning

hamishwillee:
This adds a note that Zubax telega sensorless FOC ESC require non-trivial tuning. I have not generalised this because 
- we only have support details for Zubax
- While all FOC ESC require more tuning, Telega appear (anecdotally) to be more sensitive to this issue.

If more data is provided on this issue in future we might split out an FOC ESC tuning section


hamishwillee:
FYI @bresch @pavel-kirienko  - using slight modification of the text provided.

PR #1422, opened 2021-07-21
Added ESP32 WiFi telemetry hardware option

seeul8er:
Translations in EN & DE added.
DroneBridge for ESP32 is presented as an option


hamishwillee:
@seeul8er Thank you, but we can't accept translations in this way :-(.

To keep everything in sync with the English version we use a translation service called Crowdin. Crowdin takes the English versions, you translate the strings in crowdin, and it spits out the translated version. This is great, because if the English version changes, the translated version doesn't immediately become wrong - only lose the specific strings that were changed. 

The process is described here: https://docs.px4.io/master/en/contribute/translation.html

Would very much welcome you joining the translation teams.

PR #1423, opened 2021-07-22
Added ESP32 WiFi telemetry option

seeul8er:
Just the EN version then for now :)

DroneBridge for ESP32 as firmware

### DIFF
--- a/en/telemetry/esp32_wifi_module.md
+++ b/en/telemetry/esp32_wifi_module.md
@@ -1,0 +1,142 @@
+# ESP32 WiFi Module
+
+The ESP32 are readily available Wi-Fi modules with full TCP/IP stack and microcontroller capability. They offer 
+dedicated UART, SPI and I2C interfaces. 
+They can be used with any Pixhawk series controller.
+
+## DroneBridge for ESP32
+
+**DroneBridge for ESP32 offers a transparent and bi-directional serial to WiFi bridge.**  
+Despite the original DroneBridge (for Raspberry Pi) 
+it uses the WiFi protocol. Therefore, it can not offer the same range as the other DroneBridge implementations. 
+Typical WiFi range is ~50m-200m depending on the antennas. High gain directional antennas might offer even more range.
+![DroneBridge for ESP32 connection concept](https://raw.githubusercontent.com/DroneBridge/ESP32/master/wiki/db_ESP32_setup.png)
+## Recommended Hardware
+Almost every ESP32 development board is capable to run DroneBridge for ESP32. Boards and modules with an external 
+antenna connector are recommended, since those will offer more range.
+**Most modules support 3.3V input (only), while some flight controllers (e.g. Pixhawk 4) output at 5V (you will need to 
+check compatibility and step down the voltage if needed).**
+Modules and DevKits that accept 3.3V supply:
+* ESP32-WROOM-32UE
+* ESP32-WROOM-32E
+* AZDelivery DevKit C
+* NodeMCU style DevKit
+Boards with an IPEX port for an external antenna often also offer an onboard antenna that is activated by default. You 
+  may need to resolder a resistor to activate the external antenna port.
+## Downloading and Flashing the Firmware
+[Download the firmware from the GitHub repository](https://github.com/DroneBridge/ESP32/releases) and
+[follow the flashing instructions there](https://github.com/DroneBridge/ESP32#installationflashing-using-precompiled-binaries). 
+They are always up to date. 
+:::tip
+[Follow the flashing instructions inside the GitHub Repository.](https://github.com/DroneBridge/ESP32#installationflashing-using-precompiled-binaries) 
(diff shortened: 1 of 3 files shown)


hamishwillee:
Yes :-). Can you make your repo editable so that I can subedit this?

hamishwillee:
Just to summarise, you're saying that ESP32 is a wifi module you can use with Pixhawk/PX4. It runs software on board called "DroneBridge for ESP32" that allows you to route between PX4 and QGC (or something else). The range of this module is pretty low. Is that about right?

- What does Dronebridge for Android do for you? Say I'm running this and QGC on my android device - what difference does this software make?
- Can you provide some links for buying this module?
- Do you always have to update the firmware? What do they come with by default? I assume your answer is that you need Dronebridge to make it work, in the same way that ESP8266 requires custom firmware (but may already have some onboard, depending on vendor).
- Does the "Default Configuration" work out of the box for connecting? i.e. when do you want to change this?
- When you talk about wiring, does this mean, soldering or are their preexisting cables to do the connection?

Looks pretty good thanks. Just making sure I fully understand it so that we can set the scene for users. I would actually show an example with Pixhwwk 4 say - then you can show both the wiriing and how you might do step down voltage.

seeul8er:
Thank you! I added you to the repos members. You should now be able to edit things.

> Just to summarise, you're saying that ESP32 is a wifi module you can use with Pixhawk/PX4. It runs software on board called "DroneBridge for ESP32" that allows you to route between PX4 and QGC (or something else). The range of this module is pretty low. Is that about right?

Yes!

> What does DroneBridge for Android do for you? Say I'm running this and QGC on my android device - what difference does this software make?

DroneBridge for Android as an alternative GCS for android that can be used instead of QGC. It currently lacks mission planning capabilities but besides that it is fully integrated into the "DroneBridge eco system" - allowing you to change DroneBridge settings, USB wired connection to the GND Station and seeing an HD video stream. DroneBridge usually requires Raspberry Pis and high power WiFi cards but does not use a traditional WiFi connection. That setup offers many km of range and allows for a low latency HD video stream.
DroneBridge for ESP32 does not do any of that. I basically just use the name DroneBridge but the connection is basic Wifi (Access point and clients). That way it works with any GCS that can receive TCP/UDP MAVLink telemetry.

> Do you always have to update the firmware? What do they come with by default? I assume your answer is that you need Dronebridge to make it work, in the same way that ESP8266 requires custom firmware (but may already have some onboard, depending on vendor).

Yes. The ESP32 DevKits and modules come with no firmware and you can flash DroneBridge for ESP32 to make it do telemetry. I am currently not aware of any vendors supplying custom ESP32 boards that already come with DB for ESP32. So it is sort of a DIY project with no plug&play hardware available. But it is stupid simple and cheap to get it running.
Regarding suitable boards (now you got me searching ;) ): I just found some on Adafruit that will work great on drones so I'll add them to the list. 

> Does the "Default Configuration" work out of the box for connecting? i.e. when do you want to change this?

If you are using a PX4 or any kind of MAVLink device there is no need to change the settings. The defaults will work out of the box. You only need to make sure the configured baud rates of the ESP32 and flight controller match.
You will want to change settings if you want to use different pins on the ESP32, a different WiFi configuration or to tweak the packet size. Lower packet size means more overhead and load on the system but also less latency and faster recovery from a lost packet.

> When you talk about wiring, does this mean, soldering or are their preexisting cables to do the connection?

No pre-existing cables existing. You can use 2.54mm pitch header connectors or solder the PX4 telemetry cables directly to the board.

I'll see what I can come up with regarding the wiring picture.

Thank you very much for your feedback!

hamishwillee:
Thanks very much. I'll look at this again next week and integrate any of the above that is needed.

@bkueng Is there anything other than those questions that we'd need to consider here?

bkueng:
Looks good. Might be worth comparing to the ESP8266

hamishwillee:
Thanks @seeul8er 
The offer to edit your repo got withdrawn so I have merged this and will edit as a post process.
Note that there is a problem with our build system so I can't tell you when this will go live.

en/telemetry/esp32_wifi_module.md
```diff
@@ -0,0 +1,149 @@
...
+Wiring is very simple and mostly the same for all devices connected to the TELEM1/2 ports of a Pixhawk. That is why this 
+guide does not go into detail here.
+
+* Connect UART of ESP32 to a UART of your flight controller (e.g. TELEM 1 or TELEM 2 port). Make sure the voltage levels 
+  match! Most ESP32 DevKits can only take 3.3V!
+  * TX to RX
+  * RX to TX
+  * GND to GND
+  * Stable 3.3V or 5V power supply to the ESP32 (depending on the available inputs of your DevKit)
+* Set the flight controller port to the desired protocol.
```
hamishwillee, line 77: @seeul8er Do you mean enable mavlink on the port as per ... http://docs.px4.io/master/en/peripherals/mavlink_peripherals.html ?

PR #1436, opened 2021-07-29
Updated message IDs and required messages

iwishiwasaneagle:
- Offboard control with just `offboard_control_mode` and `trajectory_setpoint` enabled didn't work. ` vehicle_local_position_setpoint` had to be set to `receive: true` as well.
- `trajectory_setpoint` had `id: 186` in the docs, but has been changed to `id: 196` in updates without docs reflecting this.

Links: 
 - https://github.com/PX4/PX4-Autopilot/blob/607be59fd5b47160326052dea8eea7edc7619d66/msg/tools/uorb_rtps_message_ids.yaml#L401

### DIFF
--- a/en/ros/ros2_offboard_control.md
+++ b/en/ros/ros2_offboard_control.md
@@ -32,8 +32,12 @@ Besides that:
        id: 44
        receive: true
      ...
+     - msg: vehicle_local_position_setpoint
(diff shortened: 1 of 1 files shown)


TSC21 (approved):
Thanks

PR #1446, opened 2021-08-02
Update Rangefinder

vincentpoont2:
Add Holybro ST VL53L1X Lidar

### DIFF
--- a/en/sensor/rangefinders.md
+++ b/en/sensor/rangefinders.md
@@ -9,6 +9,10 @@ More detailed setup and configuration information is provided in the topics link
 
 ## Supported Rangefinders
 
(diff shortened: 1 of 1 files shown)


hamishwillee:
@vincentpoont2 Thanks. Do you plan on adding a page for this - like http://docs.px4.io/master/en/sensor/leddar_one.html ?

I ask because it makes it easier for people if we state how it is connected - I2C, serial, with pinouts. Information on what connectors it comes with, and how you configure it in PX4.

vincentpoont2:
Yes, I plan to do it when I have time. I'm still not that familiar with the structure and git so it takes some time.

hamishwillee:
Thank you.

PR #1447, opened 2021-08-02
uorb graph: fix and re-enable

bkueng:
Separate js code into new file

@hamishwillee the build works for me locally with this.


hamishwillee:
Works great. Thank you

PR #1457, opened 2021-08-05
Update Holybro_xbp9x Radio.md

vincentpoont2:
Move photo to first line.


hamishwillee:
@vincentpoont2 Thanks. I would move this to the second line, after "between devices with 256-bit AES encryption.".
Also trim all the whitespace from the top and bottom of the image - this will make it render more compactly.

PR #1462, opened 2021-08-09
concept: Change the path

muramura:
Incorrect flight task path.
I'll change the path.


dagar:
Thanks @muramura. This is going to be in flex for a while as we slowly evolve FlightTasks into unified flight modes.

dagar (approved):

hamishwillee (approved):
@muramura Thanks very much. I'll merge, but it might be worth holding off on changes for a little if things are changing a lot. Then work through the whole tutorial in one go to make sure that it still works.

PR #1471, opened 2021-08-11
Precision landing - note can use landing target protocol

hamishwillee:
This is minimal changes to make it clear that Precision landing supports the MAVLink [Landing Target protocol](https://mavlink.io/en/services/landing_target.html) as well as the IRLock. Attempts to address https://github.com/PX4/PX4-Autopilot/issues/18025

Next step is a more complete "how do you implement that" but I have some questions/statements - see notes.

### DIFF
--- a/en/advanced_features/precland.md
+++ b/en/advanced_features/precland.md
@@ -1,17 +1,108 @@
 # Precision Landing
 
-PX4 supports precision landing for *multicopters* using a landing beacon/target.
-
-This requires an [IR-LOCK Sensor](https://irlock.com/products/ir-lock-sensor-precision-landing-kit) and downward facing [distance sensor](../sensor/rangefinders.md) connected to the flight controller, and an IR beacon as a target (e.g. [IR-LOCK MarkOne](https://irlock.com/collections/markone)).
-This enables landing with a precision of roughly 10 cm (GPS precision, by contrast, may be as large as several meters).
-
-Alternatively, precision landing can be supported "offboard", using a positioning system that implements the MAVLink [Landing Target Protocol](https://mavlink.io/en/services/landing_target.html).
+PX4 supports precision landing for *multicopters* on either stationary or moving targets.
+The target may be provided by an onboard IR sensor and a landing beacon, or by an offboard positioning system.
  
-Precision landing can be [started/initiated](#initiating-a-precision-landing) by entering the *Precision Land* flight mode, or as part of a [mission](#mission).
+Precision landing can be [started/initiated](#initiating-a-precision-landing) as part of a [mission](#mission), in a [Return mode](#return-mode-precision-landing) landing, or by entering the [*Precision Land* flight mode](#precision-landing-flight-mode).
 
-## Setup
+:::note
+Precision landing is only possible with a valid global position (due to a limitation in the current implementation of the position controller).
+:::
 
-### Hardware Setup
+## Overview
+
+### Land Modes
+
+A precision landing can be configured to either be "required" or "opportunistic".
+The choice of mode affects how a precision landing is performed.
+
+#### Required Mode
+
+In *Required Mode* the vehicle will search for a target if none is visible when landing is initiated.
+The vehicle will perform a precision landing if a target is located.
+
+The search procedure consists of climbing to the search altitude ([PLD_SRCH_ALT](../advanced_config/parameter_reference.md#PLD_SRCH_ALT)).
+If the target is still not visible at the search altitude and after a search timeout ([PLD_SRCH_TOUT](../advanced_config/parameter_reference.md#PLD_SRCH_TOUT)), a normal landing is initiated at the current position.
+
+:::note
+If using an offboard positioning system PX4 assumes that the target is visible when it is recieving MAVLink [LANDING_TARGET](https://mavlink.io/en/messages/common.html#LANDING_TARGET) messages.
+:::
+
+#### Opportunistic Mode
+
+In *Opportunistic Mode* the vehicle will use precision landing *if* (and only if) the target is visible when landing is initiated.
+If it is not visible the vehicle immediately performs a *normal* landing at the current position.
+
+### Landing Phases
+
+A precision landing has three phases:
+
+1. **Horizontal approach:** The vehicle approaches the target horizontally while keeping its current altitude.
+   Once the position of the target relative to the vehicle is below a threshold ([PLD_HACC_RAD](../advanced_config/parameter_reference.md#PLD_HACC_RAD)), the next phase is entered.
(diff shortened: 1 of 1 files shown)


hamishwillee:
Say I use an offboard system for my positioning. I can see in the [source code](https://github.com/PX4/PX4-Autopilot/blob/master/src/modules/mavlink/mavlink_receiver.cpp#L2387) that from the `LANDING_TARGET` PX4 confirms that we're using the local NED frame, and then sets just the target position, x, y, z.

My assumptions are (please confirm):
- x, y, z are the the NED position of the target from the current vehicle position.
- The vehicle doesn't care how that position is obtained - it could be a map position for all it cares, or a transponder, or a beacon or a visual marker.
- The vehicle navigates to land on that position using whatever sensors it has enabled/fused to establish its own position and velocity - in other words, it will do its best if you just have GPS but if you have a rangefinder and optical flow enabled it will use those? [i.e. the implication I am trying to confirm here is that users probably still want to install a rangefinder and maybe optical flow? Further, if they do, they will need to enable them in estimates.
- If you use the offboard approach you still want to set `LTEST_MODE` but you don't want to set the `LTEST_SCALE_X` and `LTEST_SCALE_Y`? Or is it that these are applied if set, but generally if you were using the offboard approach you assume the scaling is already done?

en/advanced_features/precland.md
```diff
@@ -70,29 +74,28 @@ If the beacon is still not visible at the search altitude and after a search tim
...
 Precision landing can be initiated as part of a [mission](../flying/missions.md) using [MAV_CMD_NAV_LAND](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_LAND) with `param2` set appropriately:
 
-- `param2` = 0: Normal landing without using the beacon.
-- `param2` = 1: *Opportunistic* precision landing.
-- `param2` = 2: *Required* precision landing.
+- `0`: Normal landing without using the beacon.
+- `1`: [Opportunistic](#opportunistic-mode) precision landing.
+- `2`: [Required](#required-mode) precision landing.
 
+### Command Line
```
hamishwillee, line 92: On further thought, this should probably be NuttX or SITL console. 

This is a bit poor usability. Yes I can go in through the mavlink console. Presumably I could also use MAVLink Set_Mode message (but what would I specify for the custom and main mode?). But much better would be to enable the MAV_CMD_NAV_LAND mission item as a command rather than just mission item. Any chance of this?

en/advanced_features/precland.md
```diff
@@ -70,29 +74,28 @@ If the beacon is still not visible at the search altitude and after a search tim
 In *Opportunistic Mode* the vehicle will use precision landing *if* (and only if) the beacon is visible when landing is initiated.
 If it is not visible the vehicle immediately performs a *normal* landing at the current position.
 
-## Performing a Precision Landing
+## Initiating a Precision Landing
 
 :::note
-Due to a limitation in the current implementation of the position controller, precision landing is only possible with a valid global position.
+Precision landing is only possible with a valid global position (due to a limitation in the current implementation of the position controller).
```
hamishwillee, line 80: 1. I assume this is still true?
2. We're missing the option to do this in RTL using `RTL_PLD_MD` too right?

en/advanced_features/precland.md
```diff
@@ -1,9 +1,13 @@
 # Precision Landing
 
-PX4 supports precision landing for *multicopters* (from PX4 v1.7.4) using the [IR-LOCK Sensor](https://irlock.com/products/ir-lock-sensor-precision-landing-kit), an IR beacon (e.g. [IR-LOCK MarkOne](https://irlock.com/collections/markone)), and a downward facing range sensor.
+PX4 supports precision landing for *multicopters* using a landing beacon/target.
```
hamishwillee, line 3: Perhaps " using a landing beacon/target." should be " using a landing target (beacon, visual marker, or any other positioning mechanism)."?

en/advanced_features/precland.md
potaito: 1. Good point. Yes, this should still be the case, but might change in the future after we complete further refactoring of the navigator module.
2. Actually the RTL method using `RTL_PLD_MD` is the only method I know if. I have not seen a standalone "Precision Landing" flight mode as described in the docs.

potaito: Just checked in the latest QGC and there I don't see the precision landing flight mode :sweat_smile: 
image

potaito: But then again `commander mode auto:precland` works just fine. For now I will focus on what I know and add this information. Perhaps I can figure out how the flight modes work later.

en/advanced_features/precland.md
potaito: Man this is confusing. I know for a fact that if the precision landing mode paramter `RTL_PLD_MD` is set to 1 or 2, it will perform the precision landing also during a mission in place of a standard RTL.

I'm actually not sure about the existing text here with mavlink and param2...

potaito: The simplest way of using precision landing is really to configure parameter `RTL_PLD_MD` and then just execute "RTL", which will internally do the precision landing. That's why the above mention of a "Precision Landing" flight mode also confused me.

potaito:
I think the major confusion on slack came from the fact, that the guide only refers to IR beacons and locks as a viable option, when in fact any landmark + detection algorithm can work.

en/advanced_features/precland.md
hamishwillee: The list of flight modes is hard coded, and indicates the set of modes that you can actually *set* - which is not true for precland. However if you put into this mode QGC will display it correctly (where it says "Ready to Fly" above).

You can set the modes using the MAV_CMD_SET_MODE command (I'm 99% sure) but this requires you to set the base and custom modes appropriately. This is probably not be as simple as setting [the values as 4 and 9](https://github.com/PX4/PX4-Autopilot/blob/master/src/modules/commander/px4_custom_mode.h#L66). It would be much better to abstract these by supporting precision landing in the landing commands.


Yes, and if you put into this mode you'll see

hamishwillee:
@potaito Thanks very much for your help - really useful.

I've restructured this. I'm happy enough with it that I have also merged. Essentially it makes things much more clear that both beacons and other targets are supported and gives better hints as how to do those. It also starts with the "how it works" and then follows on with the "how you implement it".
Last of all it adds the RTL enabling option and makes it clear that you can set the mode, but it isn't easy. 

On that last point, we should make it easy by supporting MAV_CMD_NAV_LAND as a command.

Anyway, would be great if you could sanity check as a post process: https://docs.px4.io/master/en/advanced_features/precland.html#precision-landing

potaito:
Looks good to me and the new structure is much nicer to read and digest. Thanks for your work :smile:  :+1:

hamishwillee:
Glad you like it. Thanks for all the help!

PR #1474, opened 2021-08-12
Start adding more refined config guidance

hamishwillee:
This is part of the work on making it easier to work out what is relevant if you're on a particular frame.

### DIFF
--- a/en/config_vtol/README.md
+++ b/en/config_vtol/README.md
@@ -1,3 +1,11 @@
 # VTOL Configuration
 
-This section contains topics related to VTOL configuration and tuning.
+First perform the [Standard Configuration](../config/README.md).
+As part of this you should calibrate the [Airspeed sensor](../config/airspeed.md) (optional, but highly recommended) and [assign a VTOL transition switch](../config/flight_mode.md#what-flight-modes-and-switches-should-i-set) to your RC controller.
+
+Then perform VTOL-specific configuration and tuning:
+
+* [QuadPlane Configuration](../config_vtol/vtol_quad_configuration.md)
+* [Back-transition Tuning](../config_vtol/vtol_back_transition_tuning.md)
+* [VTOL w/o Airspeed Sensor](../config_vtol/vtol_without_airspeed_sensor.md)
+* [VTOL Weather Vane](../config_vtol/vtol_weathervane.md)
--- a/en/config/README.md
+++ b/en/config/README.md
@@ -1,4 +1,4 @@
-# Basic Configuration
+# Standard Configuration
 
 *QGroundControl* is used to install [firmware](../config/firmware.md) onto the flight controller hardware, specify an [airframe](../config/airframe.md), and configure the core sensors that PX4 needs to be present on every vehicle (compass, GPS, gyro etc.).
 
--- a/en/config/airspeed.md
+++ b/en/config/airspeed.md
@@ -1,9 +1,10 @@
 # AirSpeed Calibration
 
-The airspeed calibration needs to read a stable baseline with 0 airspeed in order to determine an offset. Cup your hands over the pitot to block any wind (if calibrating the sensor indoors this is not needed) and then blow into the tube using your mouth (to signal completion of the calibration).
+The airspeed calibration needs to read a stable baseline with 0 airspeed in order to determine an offset.
+Cup your hands over the pitot to block any wind (if calibrating the sensor indoors this is not needed) and then blow into the tube using your mouth (to signal completion of the calibration).
 
 :::note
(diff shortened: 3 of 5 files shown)


en/config_vtol/README.md
```diff
@@ -1,3 +1,11 @@
 # VTOL Configuration
 
-This section contains topics related to VTOL configuration and tuning.
\ No newline at end of file
+First perform the [Standard Configuration](../config/README.md).
+As part of this you should calibrate the [Airspeed sensor](../config/airspeed.md) (optional, but highly recommended) and [assign a VTOL transition switch](../config/flight_mode.md#what-flight-modes-and-switches-should-i-set) to your RC controller.
+
+Then perform VTOL-specific configuration and tuning:
+
+* [QuadPlane Configuration](../config_vtol/vtol_quad_configuration.md)
```
hamishwillee, line 8: @sfuhrer IMO this is actually "Standard Plane" configuration - right - ie it doesn't have to be a quad plane, it is just that it's a multicopter grafted onto a fixed wing?

en/config_vtol/README.md
```diff
@@ -1,3 +1,11 @@
 # VTOL Configuration
 
-This section contains topics related to VTOL configuration and tuning.
\ No newline at end of file
+First perform the [Standard Configuration](../config/README.md).
+As part of this you should calibrate the [Airspeed sensor](../config/airspeed.md) (optional, but highly recommended) and [assign a VTOL transition switch](../config/flight_mode.md#what-flight-modes-and-switches-should-i-set) to your RC controller.
+
+Then perform VTOL-specific configuration and tuning:
```
hamishwillee, line 6: @sfuhrer There are 4 config docs below. The first one is clearly for Quadplane. Can you confirm whether/not the other three apply to quadplane, tiltrotor, tailsitter?

en/config_vtol/README.md
sfuhrer: Do you mean "Standard VTOL"? TBH I don't like "Standard VTOL" too much, but as you say, QuadPlane doesn't work to be a generic name of this type as there can be more than 4 hover motors. And yes, the category defines itself with having separate propulsion systems for hover and cruise flight (opposed to tilt-body (aka tailsitter), tilt-wing, tilt-rotor). As I don't have a better proposal for this category we should stick to "Standard VTOL" consistently.

sfuhrer: In "Standard Plane" I would miss the VTOL notion - but maybe "Standard Plane VTOL" to make it clear that it has a wing (opposed to all multirotors or helicopters that actually are just as much VTOLs). This categorization is really not easy.

sfuhrer: The first two are mainly for standard VTOLs. 

I'd propose to have the following subcategories in VTOL Config/Tuning:
- Transition tuning
    - Standard VTOL
    - Tiltrotor
    - Tailsitter
- weather vane
- Safety (in here we can talk about Quadchutes, permanent stabilization, ..)

sfuhrer (approved):
Independent to my comments I would bring this in like this and then I would create a PR for my ideas for structuring the VTOL tuning docs better and add also the steps for tiltrotor and tailsitters.

en/config_vtol/README.md
hamishwillee: Sorry, yes I do mean Standard VTOL. Tired! I've used that elsewhere. I'd refer to QuadPlane for the Quad motor Standard VTOL. Yes, it is a crap name, but I don't have anything better either.

hamishwillee: I am happy enough with that organisation, but I need to map it to the docs we have.

So I think you are saying
- Transition tuning
   - Standard VTOL
      * [QuadPlane Configuration](../config_vtol/vtol_quad_configuration.md)
      * [Back-transition Tuning](../config_vtol/vtol_back_transition_tuning.md)
   - Tiltrotor
      - Nothing?
   - Tailsitter
      - Nothing?
- weather vane
   - [VTOL Weather Vane](../config_vtol/vtol_weathervane.md)
- Safety

What about this one: * [VTOL w/o Airspeed Sensor](../config_vtol/vtol_without_airspeed_sensor.md)

hamishwillee:
> Independent to my comments I would bring this in like this and then I would create a PR for my ideas for structuring the VTOL tuning docs better and add also the steps for tiltrotor and tailsitters.

Ah, missed this. Merging. Then you can answer https://github.com/PX4/PX4-user_guide/pull/1474#discussion_r687547344 in your PR :-)

en/config_vtol/README.md
hamishwillee: Ignore. I see you're going to PR a structural update.

PR #1510, opened 2021-08-26
[WIP] Kconfig/boardconfig documentation

PetervdPerk-NXP:
Replaces cmake board configuration documentation with Kconfig px4board configuration.

Do not merge before https://github.com/PX4/PX4-Autopilot/pull/17100

Right now only existing Cmake documentation has been replaced by px4board.

Have to create a seperate chapter to go into depth of boardconfig tool and Kconfig language guidelines.

@hamishwillee what would be a good location for the kconfig documentation?

### DIFF
--- a/en/hardware/porting_guide_config.md
+++ b/en/hardware/porting_guide_config.md
@@ -10,7 +10,9 @@ The configuration options (often referred as "symbols" by the *kconfig* language
 By convention, symbols for modules/drivers are named based on the module folder path.
 For example, the symbol for the ADC driver at `src/drivers/adc/board_adc` must be named `DRIVERS_ADC_BOARD_ADC`.
 
-To add symbols for driver/module specific options, the naming convention is the module name followed by the option name for example `UAVCAN_V1_GNSS_PUBLISHER` which is an option `GNSS_PUBLISHER` for the `UAVCAN_V1` module. The options have to be guarded behind an `if` statement to ensure that the options are only visible when the module itself is enabled.
+To add symbols for driver/module specific options, the naming convention is the module name followed by the option name.
+Fo example `UAVCAN_V1_GNSS_PUBLISHER` which is an option `GNSS_PUBLISHER` for the `UAVCAN_V1` module.
+The options have to be guarded behind an `if` statement to ensure that the options are only visible when the module itself is enabled.
 
 Full example:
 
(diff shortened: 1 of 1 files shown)


en/modules/hello_sky.md
```diff
@@ -154,21 +154,31 @@ This consists of a single *C* file and a *cmake* definition (which tells the too
    You can then run your command by loading the file at runtime using the `dyn` command: `dyn ./examples__px4_simple_app.px4mod`
    :::
 
+1. Create and open a new *Kconfig* definition file named **Kconfig**.
+   Copy in the text below:
+   ```menuconfig EXAMPLES_PX4_SIMPLE_APP
```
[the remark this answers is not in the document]
PetervdPerk-NXP: Added PX4 Kconfig chapter that contains naming convention

PetervdPerk-NXP:
Hi Hamish,

Thanks for the review, I've added a separate chapter elaborating the basics of Kconfig and the naming convention used. And I've processed the feedback you gave. 

If there's anything else that has to be elaborated further, just notify me here.

hamishwillee:
Thanks very much! I'll have a look tomorrow. 

Note, I no longer have time to track the firmware repo for changes, so when the associated PR merges, would appreciate you removing the WIP up the top so I get reminded to merge.

en/hardware/porting_guide_config.md
```diff
@@ -0,0 +1,46 @@
+# PX4 Board Configuration (Kconfig)
```
hamishwillee, line 1: Note, I renamed this. This is board configuration. It might be used by a Porting Guide, but it isn't specific to porting. We might want to rename the file too if you are OK with that?

en/hardware/porting_guide_config.md
```diff
@@ -0,0 +1,46 @@
...
+This configuration is handled through *Kconfig*, which is the same [configuration system used by NuttX](../hardware/porting_guide_nuttx.md#nuttx-menuconfig-setup).
+
+The configuration options (often referred as "symbols" by the *kconfig* language) are defined in `Kconfig` files under the **/src** directory. 
+
+## PX4 Kconfig Symbol Naming Convention
+
+By convention, symbols for modules/drivers are named based on the module folder path.
+For example, the symbol for the ADC driver at `src/drivers/adc/board_adc` must be named `DRIVERS_ADC_BOARD_ADC`.
+
+Symbols for driver/module specific options should omit the path, use the name of the module, and depend on the symbol.
```
hamishwillee, line 13: What do you mean by "depend on the symbol"? I THINK from looking at the example that you mean that UAVCAN_V1_GNSS_PUBLISHER is a symbol for an option used in DRIVERS_UAVCAN_V1 - so `UAVCAN_V1` is the module name.

I find this a little confusing because while I get the above, and I can see that UAVCAN_V1_GNSS_PUBLISHER is conditional on `DRIVERS_UAVCAN_V1` due to the `if` statement, I don't understand what the example I am looking at belongs to - though presumably it is pat of the config file for uavcan v1

hamishwillee:
I like what you've done structurally - particularly in linking to the new document to show how symbol is set etc. I've given it a minor subedit to make it a little more readable. The main change is naming of the new guide document.

Please check it out. I will give another quick review before we send it out.

en/hardware/porting_guide_config.md
PetervdPerk-NXP: Indeed makes sense it can also be used for reconfiguration as well.

en/hardware/porting_guide_config.md
PetervdPerk-NXP: The `UAVCAN_V1_GNSS_PUBLISHER` key is an example of an option for the `UAVCAN_V1` application. Thus you've to guard these options behind the `DRIVERS_UAVCAN_V1` key otherwise you end up with lots of options which aren't applicable if the application itself isn't enabled (thus the dependency).

I do agree it might be a bit confusing right now, would it help to remove the wording "depend" and replace it by conditional?

hamishwillee: I think it might help. But what I'd do is spell it out just as you have for me-

So something like (feel free to adjust if I was not correct):

----

Symbols for driver/module specific options should omit the path, use the name of the module, and are conditionally defined based on the module symbol. For example, below the UAVCAN_V1 module (in driver/uavcan_v1/?) with module symbol `DRIVERS_UAVCAN_V1` defines the option `UAVCAN_V1_GNSS_PUBLISHER` (but only if DRIVERS_UAVCAN_V1 exists).

hamishwillee:
Thanks @PetervdPerk-NXP  - I think this is pretty good to go when the flight stack PR merges.

hamishwillee:
That is awesome. Congrats!

PR #1514, opened 2021-08-30
Update airspeed.md

sfuhrer:
- add airspeed sensor start params  (necessary since https://github.com/PX4/PX4-Autopilot/pull/17810). Given that it's a quite big UX change, this info maybe also needs to be at a more prominent place.
- remove "test" section, as this is not valid for sdp3x sensors, and in general I have personally never had the case where this check was actually necessary
- some wording changes


hamishwillee:
@sfuhrer Are docs the right way to "fix" this problem for an end user? You might, for example, have QGC auto detect that none of the know sensors are enabled and offer option to the user to start one. Or something less clunky like attempt at auto start.

Anyway, I've subedited to my taste. The enable driver setup is now a "section", which is just as prominent. The main takeaway for you is that I have linked the parameters. This is helpful for end users, but the main reason I do it is that if parameters get removed, I can find things that have modified using a link checker.

I haven't merged in case you object to me reverting the page title :-)

sfuhrer:
@hamishwillee that's ofc a valid point. We could e.g. on the px4 side check if airspeed mode is enabled (CBRK_AIRSPD_CHK param), and if yes, then complain about no airspeed driver is started. 

@dagar did you actually have a particular workflow in mind when you made the change to only start the airspeed drivers with the corresponding param being set (for context, that was needed as we have I2C address clashes, and e.g. baros detected as airspeed sensors)?

And thanks for pushing the improvements, they make a lot of sense.

hamishwillee:
@sfuhrer I made a few more minor changes - duplicated some of this info in the hardware > peripherals > airspeed sensor page, and turned it in this page into a warning. Nothing serious.

Re 
> @dagar did you actually have a particular workflow in mind when you made the change to only start the airspeed drivers with the corresponding param being set (for context, that was needed as we have I2C address clashes, and e.g. baros detected as airspeed sensors)?

Perhaps as this issue has merged, you could create an issue and assign it to @dagar ? I'm not sure he monitors docs repo issues very closely.

FWIW isn't this "fix" a bit bogus - i.e. won't it just hide the clash until someone tries to use the baro? Some of these sensors work with SPI bus - would a new driver fix the problem?

PR #1516, opened 2021-08-31
Update rangefinders.md

MuhammadBilal1:
Screenshots are updated with the latest version of the QGroundControl flow.
Old flows aren't helpful for the new users.


hamishwillee:
Thank very much @MuhammadBilal1 - absolutely agree. 

I have modified this slightly because it is better to have the images locally stored than link to images on github. You can't easily add images using the github client, but you can with the git tool. Should you wish to do this yourself in future, there is some information in http://docs.px4.io/master/en/contribute/docs.html#changes-using-git-new-pages-and-images

PR #1521, opened 2021-09-01
- fix flightgear protocol path

profff:
after following this I found a miss spelling in protocol flight gear path


hamishwillee:
Thanks very much!

PR #1527, opened 2021-09-02
Document FW_T_CLMB_R_SP and FW_T_SINK_R_SP tecs parameters

RomanBapst:


hamishwillee:
@RomanBapst I recreated this in #1529 in the English path. 

If you look closely you will notice that you edited a file in the /de/ path - that's the German translation (though untranslated). 

The main fix was to move your warning about "o not use [FW_T_CLMB_MAX ..." to the top so that it is very obvious - don't set these values as you might desire to set limits - follow the tuning. Please check it is good. 

Anyway, thanks for this. Made sense to me!

PR #1530, opened 2021-09-03
Update crazyflie2.md

richtong:
Note that the Crazyflie 2.0 is out of stock and you should buy the 2.1 instead.


hamishwillee:
Thanks for this! I have replaced this PR with a more thorough fix in #1543. Essentially since this is no longer available this is the first thing people find out now. I also moved it out of complete vehicles section to discontinued section to ensure you only find it if you are looking for it.

PR #1531, opened 2021-09-03
The Airdog II discontinued and Sentera PXH new link

richtong:
And the px4.io link used for the Sentera points to a dead page, replaced with the store link and data page


hamishwillee:
Thanks. I decided to remove Airdog II from the docs altogether. For sentera used https://sentera.shop/products/sentera-phx-complete-system link instead as it is more direct (also got this link fixed on PX4.io. Cheers

PR #1535, opened 2021-09-06
Fix to commander takeoff, qgroundcontrol must be running

richtong:
As noted in https://github.com/PX4/PX4-Autopilot/issues/18142


hamishwillee:
Thanks @richtong !!
I've subedited
- The datalink issue is actually PX4 configuration and not specific to the simulator - the system expects a datalink by default but you can turn that off using parameters. 
- This is generic building topic without any platform specifics. 

Upshot, I have replaced the text with a note that explains you may need QGC and where to download it. The download page has setup information for different platforms.

Much appreciated that you took the time to post this.

PR #1539, opened 2021-09-07
Update dev_env_mac.md to document gazebo tbb@2020 workaround

richtong:
Document current use of tbb@2020 for make px4_sitl gazebo as noted in https://github.com/PX4/PX4-Autopilot/issues/17644


hamishwillee:
Thanks very much! I have subedited "to my taste" (i.e. nothing wrong with what you wrote, I just prefer to have the instructions people have to run in order. I also don't think that readers need to know the particular reason for this fix in these docs if we provide the link to the bug report.

PR #1547, opened 2021-09-08
Update for multisim-jmavsim on debugging

richtong:
Notes on how to restart the system and how to use with qgroundcontrol for noobs. Also how to run restart

### DIFF
--- a/en/simulation/multi_vehicle_jmavsim.md
+++ b/en/simulation/multi_vehicle_jmavsim.md
@@ -18,21 +18,23 @@ To start multiple instances (on separate ports):
    ```
    make px4_sitl_default
    ```
-1. Run **sitl_multiple_run.sh**, specifying the number of instances to start (e.g. 2):
+1. Run **sitl_multiple_run.sh**, specifying the number of instances to start (e.g. 2). T
    ```
    ./Tools/sitl_multiple_run.sh 2
    ```
-1. Start the first instance:
+1. Start the first instance, this will run in the foreground
    ```
    ./Tools/jmavsim_run.sh -l
    ```
-1. Start subsequent instances, specifying the *simulation* TCP port for the instance:
+1. Create another terminal window for each subsequent instances, specifying the *simulation* TCP port for the instance:
    ```
    ./Tools/jmavsim_run.sh -p 4561 -l
    ```
-   The port should be set to `4560+i` for `i` in `[0, N-1]`.
-
-Ground stations such as *QGroundControl* listen for all vehicle instances on the PX4's remote UDP port: `14550` (all GCS traffic is sent to the *same* remote port).
+   The port should be set to `4560+i` for `i` in `[0, N-1]`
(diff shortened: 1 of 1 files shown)


en/simulation/multi_vehicle_jmavsim.md
```diff
@@ -22,21 +22,25 @@ To start multiple instances (on separate ports):
...
-1. Start the first instance:
+1. Start the first instance in the same terminal (this will run in the foreground):
    ```
    ./Tools/jmavsim_run.sh -l
    ```
-1. Start subsequent instances, specifying the *simulation* TCP port for the instance:
+1. Open a new terminal for each subsequent instance, specifying the *simulation* TCP port for the instance:
    ```
-   ./Tools/jmavsim_run.sh -p 4561 -l
+   ./Tools/jmavsim_run.sh -p 4560 -l
```
hamishwillee, line 31: Note, I think this means it starts at 4560 right? i.e. did you start the first one at 4560 or 4561 ?

en/simulation/multi_vehicle_jmavsim.md
```diff
@@ -22,21 +22,25 @@ To start multiple instances (on separate ports):
...
    ```
-   The port should be set to `4560+i` for `i` in `[0, N-1]`.
+   The port should be set to `4560+i` where `i` iterates for each instance (from `0` to `N-1`)
+ 
+*QGroundControl* should automatically connect to all the new vehicle instances (all GCS traffic is sent to PX4's remote UDP port: `14550` from all instances).
```
hamishwillee, line 35: I removed the numbering from here. The instructions/list are for "how to start the instances". That finished once the starting part is done. Now we go onto "how you test it". It doesn't have to be done this way, but it shouldn't be a continuation of the old list.

en/simulation/multi_vehicle_jmavsim.md
```diff
@@ -22,21 +22,25 @@ To start multiple instances (on separate ports):
...
+The vehicle that is currently being controlled is displayed in the application to bar; you can select this vehicle text to display a selection list of all of the (simulated) connected vehicle instnaces (`Vehicle 1`, `Vehicle 2`, etc.) and choose a new vehicle to control.  
 
-Ground stations such as *QGroundControl* listen for all vehicle instances on the PX4's remote UDP port: `14550` (all GCS traffic is sent to the *same* remote port).
-
-Developer APIs such as *MAVSDK* or *MAVROS* listen on sequentially allocated PX4 remote UDP ports from `14540` (first instance) to `14549`.
+Developer APIs such as *MAVSDK* or *MAVROS* can connect to individual instances by listening on sequentially allocated PX4 remote UDP ports from `14540` (first instance) to `14549`.
 Additional instances *all* connect to port `14549`.
 
+> **Tip** The **sitl_multiple_run.sh** script starts a separate process for each vehicle.
```
hamishwillee, line 41: I made this a note at the end. Is it correct? I think you're saying if you kill on simulation you must restart the first script and all the instance scripts. 

Are you sure this is true? I THINK if you kill one of the terminals you can create a new one and start just that instance. YOu do have to run all of them again if you want to change the number of simulations.

hamishwillee: @richtong When you have moment, can you please check my review comments, including this one?

PR #1552, opened 2021-09-20
Add further information on vehicle_local_position

max11gen:
It wasn't clear to me, where the origin of vehicle_local_position is. When I asked in slack @bresch confirmed, that it is the vehicle's position at boot time (start of EKF2 mdoule). I thought the information might be helpful to others, too.


bresch (approved):

hamishwillee:
Thanks @max11gen . Useful information to include I think. The file is actually auto-generated from source. I have updated this in https://github.com/PX4/PX4-Autopilot/pull/18274

Note, that the update might not be appear in the summary, but it will appear in the actual file: http://docs.px4.io/master/en/msg_docs/vehicle_local_position.html

PR #1554, opened 2021-09-21
Removed link to inactive webshop

ykompis:
Reason: The linked webshop still seems to operate (based on the website), but the shop itself is not. I ordered through the webshop but nothing ever shipped and emails are not being answered. According to a forum posts ( https://community.shopify.com/c/shopify-discussions/how-to-reach-out-a-non-performing-vendor-how-to-understand-a/m-p/1266655/highlight/true#M252691 ) on the web this is not the first instance of not delivered products.


hamishwillee:
Thank you. I am investigating.

hamishwillee:
Updating, still investigating. If I don't get a response soon I will accept this PR.

Will also update to provide other alternatives:
- Off the shelf ftdi cable and a JST SH to male header adapter
- Or the com port on the black magic probe or jlink

PR #1567, opened 2021-10-01
Pixracer telemetry documentation

potaito:
As explained on slack, I found the configuration of the telemetry on the pixracer non-trivial when using any other port than the default port. 

Here's the original slack message:
> 
> I just struggled for half an hour to setup the Pixracer with an ESP8266 WiFi module. I had the module connected to the TELEM2 port, but the documentation I saw did not mention what I'd have to do afterwards.
> 
> The solution was to change parameter MAV_2_CONFIG from WiFi Port, to TELEM 2 . After a reboot, PX4 will then successfully open a mavlink stream on /dev/ttyS2 instead of /dev/ttyS0 on the pixracer.
> 
> I was particularly confused because the Pixracer page https://docs.px4.io/master/en/flight_controller/pixracer.html even mentions (text in the image), that when plugging a wifi module into telem2, it should work out of the box.
> The page on ESP8266 Wifi modules https://docs.px4.io/master/en/telemetry/esp8266_wifi_module.html also does not mention how to configure mavlink streams.
> The page on mavlink telemetry https://docs.px4.io/master/en/peripherals/mavlink_peripherals.html#mavlink-instances explains exactly the step I had to perform.

### DIFF
--- a/en/flight_controller/pixracer.md
+++ b/en/flight_controller/pixracer.md
@@ -60,6 +60,12 @@ Setup and telemetry are already available, firmware upgrade is already supported
 * [Custom ESP8266 MAVLink firmware](https://github.com/dogmaphobic/mavesp8266)
 
 
+### Configuring Wifi telemetry
+The pictures in the wiring diagram below suggest, that a Wifi module can optionally also be plugged into TELEM2. This is possible, but requires further configuration, for example by modifying the parameter [MAV_2_CONFIG](../advanced_config/parameter_reference.md#MAV_2_CONFIG) to specify that second mavlink instance should be running on the TELEM2 port.
+
+For more information on configuring mavlink instances, see the page on [Mavlink peripherals](../peripherals/mavlink_peripherals.md#mavlink-instances). And for additional information on configuring serial ports in general, see the instructions for [configuring serial ports](../peripherals/serial_configuration.md).
+
+
 ## Wiring Diagrams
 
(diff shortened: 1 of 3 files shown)


en/flight_controller/pixracer.md
```diff
@@ -52,13 +52,24 @@ The Pixracer is designed to use a separate avionics power supply. This is necess
...
 
+:::note
+Firmware upgrade is not yet enabled over WiFi (it is supported by the default bootloader but not yet enabled).
+Setup and telemetry are supported.
+:::
+
+
+## External Telemetry
+
+The flight controller also supports telemetry via external Wi-Fi or radio telemetry modules connected to the `TELEM1` or `TELEM2` ports.
```
hamishwillee, line 66: @potaito Thanks very much. I'm going to merge this so I can add an image below here.

Note, the change from a WiFi and Telem2 focus to an External telemetry focus. Looking at the diagram this refers to how you can connect either radios and Wifi to Telem1 OR Telem2. 

The only error in the diagram is that it says everything will be plug and play, while for TELEM2  you discovered it isn't - so that is the bit we need to surface. But I don't want to just focus on that issue.

I have not confirmed that TELEM1 is plug and play :-)

potaito: Thanks for the clarification. You're right, it's odd to focus on this one example. Thanks for reformatting the text.

PR #1571, opened 2021-10-04
dev_env_mac.md note on brew update breaking make

richtong:
If you do a brew update after you initially run make, then all the hard links that CMAke has into /usr/local/Cellar may point to older versions like as an example
recently, a brew update bumped Cmake to ../Cellar/cmake/3.21.3/bin/cmake and so the hard link that PX4 make has to /usr/local/Cellar/camke/3.21.2/cmake will become invalid and the make will fail.the 

The fix is to run make clean when you get not found messages. (You could also do a brew install cmake@3.21.2 if you want to hard install and pin the cmake build

### DIFF
--- a/en/dev_setup/dev_env_mac.md
+++ b/en/dev_setup/dev_env_mac.md
@@ -29,6 +29,11 @@ To build other targets you will need to use a [different OS](../dev_setup/dev_en
 
 The installation of Homebrew is quick and easy: [installation instructions](https://brew.sh).
 
+Note that with PX4 make, the exact versions of tools like CMake are cached in the build. If you do a `brew update` it will automatically 
+delete older versions and your build can fail searching for executables in `/usr/local/Cellar` the fix is to run `make clean` which
+will update all the links into Homebrew with the correct versions. Alternatively if you want to pin the Cmake or other version, you can do a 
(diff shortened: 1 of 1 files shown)


en/dev_setup/dev_env_mac.md
```diff
@@ -29,6 +29,11 @@ To build other targets you will need to use a [different OS](../dev_setup/dev_en
 
 The installation of Homebrew is quick and easy: [installation instructions](https://brew.sh).
 
+Note that with PX4 make, the exact versions of tools like CMake are cached in the build. If you do a `brew update` it will automatically
```
hamishwillee, line 32: Thanks @richtong 

80%+ of users will not need to know this when they first start using the build toolchain - they just install and everything works. They only need this after doing `brew update`, which will happen later, if at all.

My suggestion is to add a troubleshooting section at the end of the document and then have this note under its own heading.

What exactly do you see in the failure case of "your build can fail searching for executables in `/usr/local/Cellar` "?

hamishwillee: Ping

hamishwillee:
I'm closing this as the author has lost interest, and I see it as a "limited use case"

PR #1593, opened 2021-10-14
update range finder readme

vincentpoont2:
update range finder Holybro


hamishwillee:
Thanks very much. I made minor modification to make it clear that this cable is pixhawk standards compliant.

PR #1624, opened 2021-10-28
Update microhard serial entry page

hamishwillee:
This updates the entry page for microhard based serial radios to include common configuration information and an overview of the range vs configuration tradeoffs.

This common information could, in theory, be referenced by sub pages instead of including their own configuration. In particular, the images come from Holybro page. I took these because they show both remote and master configuration.

@AlexKlimaj Thanks for the information you provided in #1614 , which is integrated here. Can you please sanity check this. 

IF you think works well, we might then modify the https://docs.px4.io/master/en/telemetry/arkflow_microhard_serial.html#setup-configuration section to look something like this (I won't do until/unless you are happy with the idea).

----

## Setup/Configuration

PX4 is configured to use TELEM1 or TELEM2 for telemetry radios, with a default baud rate of 57600 (recommended). No further PX4 or QGroundControl configuration is required if you are using this port and baud rate.

The ground radio, air radio, PX4, and QGroundControl must all be set to the same baud rate. Instructions for this are provided in [LINK TO OVERVIEW CONFIG SECTION]. 

[THIS IS YOUR TEXT ON CONNECTING] For vehicle radio configuration you will have to connect an FTDI adapter between the radio's 3 pin JST-GH Config port and a Windows PC running Pico Config (the radio must be powered, which you can do ...


hamishwillee:
@AlexKlimaj Thanks - doc renamed and merge!  I have updated this a little and also updated both docs to point to the common config.

en/telemetry/holybro_microhard_p900_radio.md
```diff
@@ -88,11 +74,20 @@ You can purchase a [Holybro UART to USB Converter](https://shop.holybro.com/gps-
 *Pico Config* will automatically detect and connect to the configuration port.
 Adjust the settings so that the baud rate matches PX4 (and the ground station).
 
-Once the radios and PX4 are configured to use the same baud rate you can connect *QGroundControl* to the vehicle via the radio.
-*QGroundControl* will not automatically detect the radio for anything other than the default baud rate, so first create a new "serial connection" in [Application Settings > Comm Links](https://docs.qgroundcontrol.com/master/en/SettingsView/SettingsView.html) (set the same baud rate as used by PX4/radios).
-
 Holding the **Config** button while powering-up the radio will boot the unit into COMMAND mode: the default serial interface will be active and temporarily set to operate at its default serial settings of 9600/8/N/1.
 
+Note that it is also possible to configure the radios using AT commands through the data port.
```
hamishwillee, line 79: @vincentpoont2 Note, that I have updated the docs to point to the common configuration in the parent page. The general bit that is related to connecting up a particular device remains of course.

The reason this note is here is because originally you said we can configure the device in two ways, and AT commands through the data port is one of them. However there are no instructions on that and the general setup that is explained is the one to do with using Picoconfig. So I have moved this to be a note at the end, and you can provide extra information on how to use the data port if you think it is useful.

PR #1639, opened 2021-11-03
Improve quadchute docs

ThomasRigi:
I specified the different ways of triggering a quadchute. 

I also took out the note that you can retransition to FW mode again as it wasn't really true the way it was written. The real situation is more complex as you need to reset the `transition_failure` flag either by switching to MC mode on your transmitter if in manual mode or by sending a transition-to-MC mavlink command if in auto mode.

To note that QGC to my knowledge doesn't send transition-to-MC commands if the drone is already in MC mode. 

So instead of mentioning all of this I thought the page is clearer if we just don't mention it at all.

If you prefer to add it ok for me too.

### DIFF
--- a/en/config/safety.md
+++ b/en/config/safety.md
@@ -250,7 +250,7 @@ If triggered, the vehicle will immediately switch to multicopter mode.
 If the vehicle was in [Mission mode](../flight_modes/mission.md) it enters failsafe [Return mode](../flight_modes/return.md).
 
 :::note
-The quadchote can also be triggered by sending a MAVLINK [MAV_CMD_DO_VTOL_TRANSITION](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_VTOL_TRANSITION) message with `param2` set to `1`.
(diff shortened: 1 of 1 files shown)


hamishwillee:
@ThomasRigi Yes, this is much better thanks. In particular the first sentence only it highlighted just one failure trigger and forward linking to the different triggers as defined in the table is better.

I have subedited to my taste. This includes pulling the comment about the mavlink command into a note - because that's not a failsafe (though of course it is is still a quadchute and there is nowhere better to put the information). 

Thanks for removing the incorrect note. I would "quite like" to capture the correct information but we can do that as a post-process. Perhaps something like (?)

> Quadchute sets the `transition_failure` flag, which prevents the transition back to forward flight even if you exit Return mode.
> The flag is cleared when the vehicle lands and disarms, or when you _explicitly_ send a command to set MC mode (e.g. using a transmitter switch or by sending `MAV_CMD_DO_VTOL_TRANSITION`).
> Note that if the condition that caused the failsafe still exists, it may trigger again!

PR #1645, opened 2021-11-06
ROS2 install additional warning update

spes2357:
https://github.com/PX4/px4_ros_com/issues/83#issuecomment-962495445

Hi, I am Joonmo. I am new to PX4, and this is my first time contribution.

### DIFF
--- a/en/ros/ros2_comm.md
+++ b/en/ros/ros2_comm.md
@@ -81,7 +81,9 @@ You won't be able to continue with this guide until the correct versions of **Fa
 :::note
 This install and build guide covers ROS 2 Foxy in Ubuntu 20.04.
 :::
-
+:::warning
+If ROS_DOMAIN_ID is set in environment variables from ROS2 tutorial, you need to unset ROS_DOMAIN_ID for connection between ROS2 and microRTPS-agent.
+:::
 To install ROS 2 and its dependencies:
 
 1. [Install ROS 2 Foxy](https://index.ros.org/doc/ros2/Installation/Foxy/Linux-Install-Debians/)


hamishwillee:
@spes2357 Thanks for posting this, but I don't think it makes sense to include this "as is".

If you follow the instructions in this document you will not set `ROS_DOMAIN_ID` and you will in fact never see this variable. That means that this note is confusing for people who are following the instructions. 

It would be good to help people understand that they should not be running custom tutorials. We could do this at the top with a warning:

> **Warning** Only run official tutorials from OSRF: custom tutorials have been known to break this installation".

It might also be reasonable to have a troubleshooting section at the end, but again, troubleshooting is supposed to cover things that might go wrong when you do everything right, not things resulting from random setup.

@TSC21 What do you think about this. Would the warning above make sense, or a troubleshooting section, or is this just something we can't really expect to try and counter?

spes2357:
@hamishwillee Hi,

Yes, I agree it could custom tutorial, but I followed the link, "Install ROS2 Foxy", in 
https://docs.px4.io/master/en/ros/ros2_comm.html#install-ros-2. The ROS2 install page leads to "Next steps after installing" which makes the user set ROS_DOMAIN_ID. I wanted to warn new Px4 users following the link or users from ROS2.

https://docs.ros.org/en/foxy/Tutorials/Configuring-ROS2-Environment.html

hamishwillee:
Ah, thank you for the clarification.

@TSC21 So it is reasonable if we point to https://index.ros.org/doc/ros2/Installation/Foxy/Linux-Install-Debians/ that a user will follow through any and all of that doc to the end, including next steps - which in turn does https://docs.ros.org/en/foxy/Tutorials/Configuring-ROS2-Environment.html

The right thing to do to be very clear about what of that document they need to do or do not need to do. Ideally copy the relevant parts into our docs. But if not, add a note about which bits of the installing Roxy link we want them to do.
image

I can't answer what they should and should not do, so your advice appreciated.

hamishwillee:
@TSC21 Can you comment on my note above ^^^

TSC21 (approved):
I am ok with this. Thanks!

hamishwillee:
Arrgh. I didn't want to merge this quite as is. Good enough for now though. Might move things around later.

PR #1648, opened 2021-11-10
OneShot docs

hamishwillee:
This is first draft of docs for oneshot. It is essentially first step on a more general ESC comparison topic.


en/peripherals/oneshot.md
```diff
@@ -0,0 +1,44 @@
...
+### Wiring
+
+Wiring is exactly the same as for [PWM ESCs](../peripherals/pwm_escs_and_servo.md).
+
+### PX4 Configuration
+
+
+To enable OneShot use the following parameters:
+
+- DSHOT_CONFIG...= 0
```
hamishwillee, line 40: @bkueng 

This is draft docs, but you can see the idea. The most important bits are:
- Wiring same as PWM
- How you set up config - hence this question

I "think" you'll need to set up the frame rate / duty cycle and the max, min pulse lengths for the esc you use, possibly separately for MAIN and AUX buses.  But I can't find the parameters that you need to set . Can you point me to them - ie.. I don't see the RATE params: https://docs.px4.io/master/en/advanced_config/parameter_reference.html#pwm-outputs.

bkueng: Hmm I'll look into it why there're missing, but it's PWM_MAIN_RATE (IO) and PWM_AUX_RATE (FMU). Then you would use PWM_MAIN_{MIN,MAX} as normal, so e.g. to 1000,2000 (they are scaled internally).

(with the new configuration UI you can directly set it per output group: https://github.com/mavlink/qgroundcontrol/pull/9952).

hamishwillee: Would be good if you could!

In https://github.com/PX4/PX4-Autopilot/blob/master/src/lib/pwm/pwm_main_params.yaml#L24 it says for main rate to set:

> Set to 400 for industry default or 1000 for high frequency ESCs. Set to 0 for Oneshot125.

If the pulse width for PWM is 2000us, the theoretical maximum rate is 500Hz (assuming infinitely small "off" cycle could be measured). So 400 makes sense, but what would 1000 mean?

On the net the range for normal PWM seems to be between 50Hz and 490Hz.

hamishwillee:
OK @bkueng I think this is getting there - though I probably need to answer https://github.com/PX4/PX4-user_guide/pull/1648#discussion_r747211930

What this does now is add a high level overview of the different types of ESC protocols to the ESC & MOtors page, with links to more information in the main page. Essentially someone reading this should know the protocols we support and that dshot is the one to use if you can.
It does feel a little two dimensional - I get the impression a plane using a 50Hz rate esc would not care.

This could be extended with info about the different open source esc firmware etc, and types of motors - brushless etc. Hopefully by someone else :-)

The UAVCAN section is thin. I have used the comment "use this if you're using the UAVCAN bus" and said it shares similar benefits to dshot. Hope that seems reasonable.

Anyway, do you think it does the job now of making it clear how you connect and configure each type, and differentiate the types that are supported?

en/peripherals/oneshot.md
```diff
@@ -0,0 +1,44 @@
...
+Both PWM and OneShot communicate using a periodic pulse, where the _length_ of the pulse indicates the desired power level.
+For PWM the pulse length typically ranges between 1000uS (zero) and 2000uS (full power), while for OneShot 125 the pulse widths are 8 times shorter, ranging from 125us (zero power) to 250us (full power).
+
+The maximum rate at which pulses can be sent, and hence the responsiveness, depends on the width of the largest pulse.
+For PWM the theoretical maximum is close to 500 Hz while for OneShot it approaches 4 kHz.
+
+:::note
+The actual frame rate/duty cycle depends on the capabilities of the ESC.
+PWM ESC commonly have a rate between 50Hz and 490Hz.
+OneShot 125 ESCs are generally higher than this, though not necessarily approaching the theoretical maximum.
```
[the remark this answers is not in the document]
bkueng: Typically 1-2khz.
The only use-case where I would use that over DShot is HW limitations (not enough pins available for DShot, or very old ESCs w/o DShot).

en/peripherals/oneshot.md
bkueng: I found it and will fix it. It's getting a bit difficult to generate a general parameter reference due to increasing board-specific parameters.

> If the pulse width for PWM is 2000us, the theoretical maximum rate is 500Hz (assuming infinitely small "off" cycle could be measured). So 400 makes sense, but what would 1000 mean?

Yes, I'm not aware of any real use for 1000.

en/peripherals/oneshot.md
```diff
@@ -0,0 +1,46 @@
+# OneShot Servos and ESCs (Motor Controllers)
+
+PX4 support OneShot 125 ESCs.
+These are typically faster and more responsive than [PWM ESCs](../peripherals/pwm_escs_and_servo.md) but share the same wiring setup (all you need to do is set some different parameters)
+
+:::note
+[DShot](../peripherals/dshot.md) ESCs are even more responsive, and hence recommended.
```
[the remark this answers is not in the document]
bkueng: With my upcoming changes you can choose more fine-grained. E.g. have DShot on some pins and PWM on others.

en/advanced_config/esc_calibration.md
```diff
@@ -1,7 +1,7 @@
 # ESC Calibration
 
 :::note
-These instructions are only relevant to [PWM ESCs](../peripherals/pwm_escs_and_servo.md).
+These instructions are only relevant to [PWM ESCs](../peripherals/pwm_escs_and_servo.md) and [OneShot ESCs](../peripherals/oneshot.md).
```
hamishwillee, line 4:
```suggestion
These instructions are only relevant to [PWM ESCs](../peripherals/pwm_escs_and_servo.md) and [OneShot ESCs](../peripherals/oneshot.md) ([DShot](../peripherals/dshot.md) and [UAVCAN](../uavcan/escs.md) ESCs do not require calibration).
```

en/peripherals/esc_motors.md
```diff
@@ -1,12 +1,80 @@
...
+## ESC Protocols
+
+### PWM
+
+[PWM ESCs](../peripherals/pwm_escs_and_servo.md) are commonly used for fixed wing vehicles and ground vehicles (vehicles that require a higher response rate like multicopters typically use oneshot or dshot ESCs).
+
+PWM ESCs communicate using a periodic pulse, where the _width_ of the pulse indicates the desired power level.
+The pulse wdith typically ranges between 1000uS for zero power and 2000uS for full power.
+The periodic frame rate of the signal depends on the capability of the ESC, and commonly ranges between 50Hz and 490 Hz (the theoretical maximum being 500Hz for a very small "off" cycle).
+A higher rate is better, in particular where a rapid response to setpoint changes is needed.
```
hamishwillee, line 27:
```suggestion
```suggestion
A higher rate is better for ESCs, in particular where a rapid response to setpoint changes is needed.
For PWM servos 50Hz is usually sufficient, and many don't support higher rates.
``` 
```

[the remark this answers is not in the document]
Comment (hamishwillee) [reply en/peripherals/esc_motors.md:27]: Will do! Thanks

File: en/peripherals/esc_motors.md [hunk 6]
```diff
@@ -1,12 +1,80 @@
...
+- [ESC Calibration](../advanced_config/esc_calibration.md)
+
+### DShot
+
+[DShot](../peripherals/dshot.md) is a digital ESC protocol that is highly recommended for vehicles that can benefit from reduce latency, in particular racing multicopters, VTOL vehicles, and so on.
+
+It has reduced latency annd is more robust than both [PWM](#pwm) and [OneShot](#oneshot).
+In addition it does not require ESC calibration, telemetry is available from some ESCs, and you can revers motor spin directions
+
+PX4 configuration is done by setting the desired rate .
```
hamishwillee, line 65:
```suggestion
PX4 configuration is done by setting the desired rate using [DSHOT_CONFIG](../advanced_config/parameter_reference.md#DSHOT_CONFIG).
```

en/peripherals/esc_motors.md
```diff
@@ -1,12 +1,80 @@
...
+The parameter name indicates the data rate in kbaud - e.g. DShot150 has a rate of 150kbaud.
```
hamishwillee, line 66:
```suggestion
The parameter value indicates the data rate in kbaud - e.g. DShot150 has a rate of 150kbaud.
```

en/peripherals/esc_motors.md
```diff
@@ -1,12 +1,80 @@
...
+Generally higher rates are better, but lower rates may be needed for larger aircraft (with longer leads).
```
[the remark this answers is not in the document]
hamishwillee:
```suggestion
Higher rates result in lower latency, but higher rates are more robust (and hence more suitable for large aircraft with longer leads); some ESCs only support lower rates (see datasheets for information).
```

en/peripherals/esc_motors.md
```diff
@@ -1,12 +1,80 @@
...
+
+Setup:
+- [ESC Wiring](../peripherals/pwm_escs_and_servo.md) (same as for PWM ESCs)
+- [PX4 Configuration](../peripherals/dshot.md#configuration) (using [DSHOT_CONFIG](../advanced_config/parameter_reference.md#DSHOT_CONFIG))
+- [DShot](../peripherals/dshot.md) also contains information about how to send commands etc.
+
+### UAVCAN
+
+[UAVCAN ESCs](../peripherals/uavcan_escs.md) are highly recommended when UAVCAN is the primary bus used for your vehicle.
```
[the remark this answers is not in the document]
hamishwillee: Thanks. Good to know. I will modify to include this information about rate.

```suggestion
[UAVCAN ESCs](../peripherals/uavcan_escs.md) are recommended when UAVCAN is the primary bus used for your vehicle.
The current PX4 implementation has an update rate of 200Hz.
```

en/peripherals/oneshot.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Overview
+
+OneShot is essentially a version of [PWM](../peripherals/pwm_escs_and_servo.md) that can be, in theory, up to 8 times faster.
+
+Both PWM and OneShot communicate using a periodic pulse, where the width of the pulse indicates the desired power level.
+For PWM the pulse length typically ranges between 1000uS (zero) and 2000uS (full power), while for OneShot 125 the pulse widths are 8 times shorter, ranging from 125us (zero power) to 250us (full power).
+
+The theoretical maximum rate at which pulses can be sent, and hence the responsiveness, depends on the width of the largest pulse.
+For PWM this rate is close to 500 Hz while for OneShot it approaches 4 kHz.
```
hamishwillee, line 19:
```suggestion
For PWM this rate is close to 500 Hz while for OneShot it approaches 4 kHz.
In practice the actual maximum rate for OneShot ESCs is typically between 1 kHz and 2 kHz, depends on the ESC used.
```

en/peripherals/oneshot.md
```diff
@@ -0,0 +1,42 @@
+# OneShot Servos and ESCs (Motor Controllers)
+
+PX4 support OneShot 125 ESCs (only).
+These are typically faster and more responsive than [PWM ESCs](../peripherals/pwm_escs_and_servo.md) but share the same wiring setup (all you need to do is set some different parameters)
+
+:::note
+They are not as highly recommended as [DShot](../peripherals/dshot.md) ESCs, which are more responsive, robust, do not required calibration, and may support telemetry.
```
hamishwillee, line 7: This updates to make it very clear to use Dshot, using @bkueng comment below.
```suggestion
[DShot](../peripherals/dshot.md) should always be used instead of OneShot where possible, as it is more responsive, more robust, does not required calibration, and may support telemetry.
The only reason not to use DShot would be hardware limitations (insufficient DShot pins available or using an ESC that does not support DShot).
```

en/peripherals/oneshot.md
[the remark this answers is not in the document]
hamishwillee: Great. Resolved above.

en/peripherals/esc_motors.md
```diff
@@ -1,12 +1,81 @@
...
+Higher rates result in lower latency, but lower rates are more robust (and hence more suitable for large aircraft with longer leads); some ESCs only support lower rates (see datasheets for information).
+
+Setup:
+- [ESC Wiring](../peripherals/pwm_escs_and_servo.md) (same as for PWM ESCs)
+- [PX4 Configuration](../peripherals/dshot.md#configuration) (using [DSHOT_CONFIG](../advanced_config/parameter_reference.md#DSHOT_CONFIG))
+- [DShot](../peripherals/dshot.md) also contains information about how to send commands etc.
+
+### UAVCAN
+
+[UAVCAN ESCs](../peripherals/uavcan_escs.md) are recommended when UAVCAN is the primary bus used for your vehicle.
```
hamishwillee, line 77:
```suggestion
[UAVCAN ESCs](../peripherals/uavcan_escs.md) are recommended when UAVCAN is the primary bus used for your vehicle.
The PX4 implementation is currently limited to update rates of 200Hz.
```

hamishwillee:
@bkueng Thanks so much for the review. I have integrated your comments and I am happy enough with this to merge.

I will ignore the issue with PWM rate of 1000 as I don't mention it - users should specify whatever rate their ESC datasheet says if they select PWM.

> It's getting a bit difficult to generate a general parameter reference due to increasing board-specific parameters.

Is there anything we can/should do about this?

My thinking is that it is perhaps useful to identify the parameters that are board specific. Worse case can we run this generator multiple times, one for general case and once for each board. If you can generate files for each case I could write a parser to strip out duplicates and combine them.

bkueng:
> My thinking is that it is perhaps useful to identify the parameters that are board specific. Worse case can we run this generator multiple times, one for general case and once for each board. If you can generate files for each case I could write a parser to strip out duplicates and combine them.

Yes we might have to at some point. Things like `PWM_AUX_MAX1` is already board-specific.
For now I'm trying to include as much as we can until it's not possible anymore.

hamishwillee:
> Yes we might have to at some point. Things like PWM_AUX_MAX1 is already board-specific. For now I'm trying to include as much as we can until it's not possible anymore.

Your call. No one seems to have complained that we don't highlight the parameters which are board-specific.

Any progress on working out why PWM_AUX_MAX1  and co aren't being generated in docs? I like being able to link to them :-)

bkueng:
Yes I did it in https://github.com/PX4/PX4-Autopilot/pull/18563/commits/8b619acb0d28c373b4436587e2b9568e20d95d44

hamishwillee:
> Yes I did it in PX4/PX4-Autopilot@8b619ac

Ah, not merged yet. Thanks!

bkueng:
fyi params are there again: https://docs.px4.io/master/en/advanced_config/parameter_reference.html#PWM_AUX_MAX1

hamishwillee:
@bkueng Thanks very much - excellent.

That tells me that actuators metadata must have merged, and hence we need to do a sweep through docs and tidy up our mixing story etc.
E.g. 
- update this to refer to parameters and state that the mixers no longer used - https://docs.px4.io/master/en/concept/mixing.html#mixing-and-actuators
- update airframes to make it clear these are default and how they are linked.
- update https://docs.px4.io/master/en/payloads/#cargo-drones-actuator-payloads to talk about how things assigned 

Probably more. Do you plan on doing first shot at this or should we have a brief chat and I do it?

bkueng:
Currently this is all still hidden from the normal user, as we have not switched over yet. Some things are still in the works, others need a bit more testing (MC is getting in shape, I'm going through VTOL currently). The param to switch is `SYS_CTRL_ALLOC`, and we probably do it in stages.
But you're right, this will need quite a bit of docs updating. I can take a first stab.

For the generated airframes I'm not quite sure yet what to do. For example now you can just select an MC airframe and assign any number of motors to it. For generic airframes, we probably won't have any default assignments for AUX or MAIN.

hamishwillee:
Great. Well ping me when you need any support.

> For generic airframes, we probably won't have any default assignments for AUX or MAIN.

You're the expert. My first reaction would be that the less that changes the better; which implies to me setting the same default everything, including assignments, as there are now.

PR #1651, opened 2021-11-11
Update dev_env_mac.md

dlech:
https://github.com/PX4/PX4-Autopilot/issues/17644 has been closed, so this can be removed as stated in the note.


dlech:
OK. maybe the issue is not fixed yet

hamishwillee:
@richtong @dlech I'm confused - is the text still needed or not?

dlech:
Yes, it is still needed. I guess they closed the issue since the note with the workaround was added in the docs and not because the root issue was actually fixed.

hamishwillee:
Thanks. I've reopened the other issue so we can explore whether the instructions just need to live in the docs forever.

PR #1663, opened 2021-11-21
Additional Python packages for building PX4 on Mac

dbaldwin:
After going through the documentation I found that to get a build of PX4 it was necessary to add the following packages:

pip3 install kconfiglib
pip3 install future
pip3 install jsonschema

After installing these packages I was able to build successfully.

### DIFF
--- a/en/dev_setup/dev_env_mac.md
+++ b/en/dev_setup/dev_env_mac.md
@@ -57,9 +57,9 @@ Install the required Python packages
 
 ```sh
 # install required packages using pip3
-python3 -m pip install --user pyserial empy toml numpy pandas jinja2 pyyaml pyros-genmsg packaging
+python3 -m pip install --user pyserial empy toml numpy pandas jinja2 pyyaml pyros-genmsg packaging kconfiglib future jsonschema
 # if this fails with a permissions error, your Python install is in a system path - use this command instead:
-sudo -H python3 -m pip install --user pyserial empy toml numpy pandas jinja2 pyyaml pyros-genmsg packaging
+sudo -H python3 -m pip install --user pyserial empy toml numpy pandas jinja2 pyyaml pyros-genmsg packaging kconfiglib future jsonschema
 ```
 
 ## Gazebo Simulation


dagar (approved):

dagar:
Thanks Dennis.

dbaldwin:
No problem @dagar. I just realized there are other locales and I only edited the English version. Happy to make edits to the others.

hamishwillee:
Thanks very much @dbaldwin - just FYI, if you ever need to find a changed dependency, the best list to use is https://github.com/PX4/PX4-Autopilot/blob/master/Tools/setup/requirements.txt (since it is used in our CI system).

No need to update the locales, in fact much better if you don't. The translations are automatically updated for any strings that change in the English version.

dbaldwin:
@hamishwillee thanks for the explanation. I'm curious where the documentation says to use these shell scripts:

https://github.com/PX4/PX4-Autopilot/tree/master/Tools/setup

I did not know those existed and would certainly have used them :) I'm about to go through this process on Ubuntu so will give the script a try.

Finally, I hope to contribute more to the docs as I'm sifting through a bunch of info related to my latest build:

https://www.youtube.com/watch?v=5O_WOCUkqNM

I have it well tuned in hover mode and am getting prepared for transition, but want to make sure I have everything tested and ready. It's based off the Covergence VTOL airframe. Thanks for all that you guys do for our community. :100:

hamishwillee:
Hi @dbaldwin 
Love it! I've been thinking about buying a convergence clones (since original is no longer made). Found a couple of good alternatives here: http://docs.px4.io/master/en/frames_vtol/vtol_tiltrotor_eflite_convergence_pixfalcon.html#purchase

The scripts are linked in the ubuntu setup page right near the top: http://docs.px4.io/master/en/dev_setup/dev_env_linux_ubuntu.html#bash-scripts
We only recommend the scripts for Ubuntu, so you wouldn't have found them for Mac.

Note that the ROS setup ones may be broken. I'd probably set that up from scratch.

dbaldwin:
Nice. I had looked at the X450 but figured it was too small given the Convergence is a 650 and the conversion looks pretty tight. Maybe I'll give it another look after I get this build dialed in.

The Ubuntu script worked a dream today. Thanks for getting back to me.

hamishwillee:
Yeah, I'm really not sure if the bits will fit. I just got myself a JJRC M02 Brushless RC Plane which is much the same size to try out. I probably need a new RC receiver - the one I normally use is too big.

PR #1668, opened 2021-11-23
Update with FMUv5X info

vincentpoont2:
Update with FMUv5X info, add pixhawk 5X


hamishwillee:
Thanks @vincentpoont2 - especially for adding the entry for fmu v5x.

PR #1671, opened 2021-11-25
Autotune video and cross links

hamishwillee:
This adds a video for autotuning. It also updates the manual PID tuning topics with links to the autotuning topic and information on when you would want to use the manual topic.

Essentially it says "use autotune" unless you can't.


hamishwillee:
@bresch I'm pulling this in as "in line with our discussion". A sanity check at your convenience would be great.

PR #1674, opened 2021-11-28
Remove Incorrect Info

vincentpoont2:
Remove Incorrect Info on Power Module and add connector info.


hamishwillee:
I'll merge, but what do you have to configure for the power module now?

PR #1704, opened 2021-12-15
basic config: add actuators page (replacing motors)

bkueng:
Initial docs for dynamic control allocation, containing tips and conventions on how to use the new UI (https://github.com/mavlink/qgroundcontrol/pull/9952).

- I'll probably add a direct link from QGC to this page (will need a metadata extension)
- We could add airframe-type specific sections, if there's more specific things to be explained
- Adding more screenshots for different types might be useful, but I did not want to add too many already as they're still in flux


en/config/actuators.md
```diff
@@ -0,0 +1,116 @@
...
+It replaces the existing mixer files and allows configuration via parameters instead.
+:::
+
+This is a screenshot for a multicotor:
+
+![Actuators MC (QGC)](../../assets/config/qgc_actuators_mc_aux.png)
+
+
+:::note
+If you find a setting missing (e.g. reversable motors), make sure to enable the **advanced** checkbox in the top right corner (not visible above).
```
hamishwillee, line 18: Sometimes worth "cheating" on the image. I.e. it is not possible to display the checkbox and the important parts of the image in the same view, so you take several shots and build a composite. Not sure in this case. Other option is to snapshot just the bit of interest.

en/config/actuators.md
```diff
@@ -0,0 +1,116 @@
+# Actuator Configuration and Testing
+
+After the airframe is setup you should configure the geometry, assign actuators to outputs, and test the actuator response.
```
hamishwillee, line 3: Just FMI What do you mean by "is setup". I think you mean after you have wired up your airframe and selected the airframe type? I.e. I'm trying to confirm the order in which you can do this w.r.t. other config sections and I think it can be done after firmware and airframe selection

en/config/actuators.md
```diff
@@ -0,0 +1,116 @@
...
+![Actuators MC (QGC)](../../assets/config/qgc_actuators_mc_aux.png)
+
+
+:::note
+If you find a setting missing (e.g. reversable motors), make sure to enable the **advanced** checkbox in the top right corner (not visible above).
+:::
+
+## Geometry
+The airframe pre-selects the geometry type, which can provide additional configuration options, such as the motors positions on a multicotor, or the control surface types on a plane.
+The UI reflects that and displays a customized view for each type.
```
hamishwillee, line 23: Ah, so you're saying that the geometry options you are offered will depend on the airframe - so you'll just be offered the multicopter option here?

What I'm getting is where the available geometry is configurable - ie at this point it is too late to select a + config right - the airframe selection means here you already have to have chosen X.

en/config/actuators.md
```diff
@@ -0,0 +1,116 @@
...
+
+
+### Conventions
+The following sections contain the conventions and explanations for configuring the geometry.
+
+#### Coordinate system
+The coordinate system is NED, where the X axis points forward, the Y axis to the right and the Z axis down.
```
hamishwillee, line 30: NED relative to airframe body or to earth tangent? Might not matter here but if we can be clear, we should be.

en/config/actuators.md
```diff
@@ -0,0 +1,116 @@
...
+- Control: depending on the airframe, tilt servos can be used to control torque on one or more axis (it's possible to only use a subset of the available tilts for a certain torque control):
+  - Yaw: the tilts are used to control yaw (generally desired).
+    If four or more motors are used, the motors can be used instead.
+  - Pitch: typically differential motor thrust is used to control pitch, but some airframes require pitch to be controlled by the tilt servos. Bicopters are among those.
+- Tiltable motors are then assigned to one of the tilt servos.
+
+![Tilt Axis](../../assets/config/tilt_axis.png)
+
+### Reversible Motors
+If reversible motor(s) are used (i.e. motor spins in direction 1 for lower output range and in direction 2 for the upper half), make sure to select the **Reversible** checkbox for those motor(s).
```
hamishwillee, line 66: Might be worth showing a screenshot example since there is no reversible motor shown in the screenshots

en/config/actuators.md
```diff
@@ -0,0 +1,116 @@
...
+  Adjust the minimum output value such that the motors spin at that slider position.
+  :::note
+  VTOL's will automatically turn off motors pointing upwards during forward flight.
+  :::
+- Servos move into the direction of the convention described above.
+  :::note
+  A trim value can be configured for control surfaces, which is also applied to the test slider.
+  :::
+
+Note the following behaviour:
```
hamishwillee, line 97: Should we perhaps have at least some of this stuff first as safety  warnings before the guidance? e.g. kill switch, safety button, and perhaps removing props?

en/config/actuators.md
```diff
@@ -0,0 +1,116 @@
...
+- If a safety button is used, it must be pressed before actuator testing is allowed.
+- The kill-switch can still be used to stop motors immediately.
+- Servos do not actually move until the corresponding slider is changed.
+- The parameter [COM_MOT_TEST_EN](../advanced_config/parameter_reference.md#COM_MOT_TEST_EN) can be used to completely disable actuator testing.
+- On the shell, [actuator_test](../modules/modules_command.md#actuator-test) can be used as well.
+
+### Reversing Motors
+:::note
+The *spin direction* checkbox for the geometry does not actually change the motor spin direction. The setting must match with the physical spin direction of the motor.
```
hamishwillee, line 106:
```suggestion
The *spin direction* checkbox for the geometry must be set to reflect the actual spin direction of the motor (it does not change the motor spin direction).
```

en/config/actuators.md
```diff
@@ -0,0 +1,116 @@
...
+:::
+
+If one or more of the motors do not turn in the correct direction according to the configured geometry, they must be reversed.
+There are several options:
+- If the ESCs are configured as [DShot](../peripherals/dshot.md) you can reverse the direction via UI (**Set Spin Direction** buttons). Note that the current direction cannot be querried, so you might have to try both options.
```
hamishwillee, line 111: Typo. Again  we need to show somewhere setting spin directions.

```suggestion
- If the ESCs are configured as [DShot](../peripherals/dshot.md) you can reverse the direction via UI (**Set Spin Direction** buttons). Note that the current direction cannot be queried, so you might have to try both options.
```

hamishwillee:
@bkueng Good start. Some questions in line which will help me nitpick a bit more.

1. Given this is not the default I suggest "for now" we keep both this and the motors page (rather than replacing). On the motors page add a note that this is being superseded - with link.
2. There are questions inline about what airframe selection actually configures. I tried to test but my VM is screwed. Reinstalling.
3. Yes, this will need more images for other airframes, but as you say, might be too early. 
4. A video would be great, but could wait too. 
5. Might be too early, but are there any other sections in config that will need to change because of this. Specifically I am thinking things like RC / actuator passthrough setup and in http://docs.px4.io/master/en/config/radio.html#additional-radio-setup

My main concern is that it isn't quite clear how much of this you have to do for what airframe. It's a bit like the tuning stuff - we used to say you can just fly a matching airframe without "having" to do this. Now it sounds like we might be having fewer airframes and we expect people to do this.

en/config/actuators.md
bkueng: It was more lazyness of just using an existing screenshot. I updated it.

en/config/actuators.md
bkueng: Updated.

en/config/actuators.md
bkueng: I added one

en/config/actuators.md
bkueng: Yes we could. Although the switch in the UI already says it.

en/config/actuators.md
bkueng: Applied

en/config/actuators.md
bkueng: Applied

en/config/actuators.md
bkueng: > I think you mean after you have wired up your airframe and selected the airframe type?

Correct

> Further, is all of this a "have to". Specifically are their airframes where you will arrive here, having selected your airframe, and all you need to do is test that the motors spin in the right way?

Yes for fully preconfigured off-the shelf vehicles that is correct. For generic airframes the outputs are most likely not assigned. In that case the UI tells you that you still need to assign them:
image

en/config/actuators.md
bkueng: Yes. We have some flexibility where that line exactly is. For now it is: when selecting an MC you can fully reconfigure the motors: switch between X/+/h, add/remove motors, change motor positions. What you cannot do is change it to a vtol for example.

bkueng:
> My main concern is that it isn't quite clear how much of this you have to do for what airframe. It's a bit like the tuning stuff - we used to say you can just fly a matching airframe without "having" to do this. Now it sounds like we might be having fewer airframes and we expect people to do this.

Assigning the outputs will be the main (and only) change for the user, if it's not an exotic one. I still forsee that we have generic airframes for different MC types (i.e. quad, hexa, coax, ...), where the geometry is preconfigured appropriately.

en/config/actuators.md
```diff
@@ -0,0 +1,118 @@
...
+- mixed (e.g. V-Tail): up = positive deflection
+
+![Plane Deflections](../../assets/config/plane_servo_convention.png)
+
+:::note
+If a servo does not move into the expected direction, reverse it via checkbox on the Actuator Output.
+:::
+
+
+### Tilt Servos
```
sfuhrer, line 46: Should we name them "Motor Tilt Servos"? Or just mention somewhere in the beginning that it's about tilting motors, and not control surfaces or anything else.

en/config/actuators.md
```diff
@@ -0,0 +1,118 @@
...
+Tilt servos are configured as following:
+- The reference direction is upwards (negative Z direction).
+- Tilt direction: **Towards Front** means the servo tilts towards positive X direction, whereas **Towards Right** means towards positive Y direction.
+- Minimum and maximum tilt angles: specify the physical limits of the tilt at minimum control and maximum respectively. An angle of 0 means to point upwards, then increases towards the tilt direction.
```
sfuhrer, line 50:
```suggestion
- Minimum and maximum tilt angles: specify the physical limits in degrees of the tilt at minimum control and maximum respectively. An angle of 0° means to point upwards, then increases towards the tilt direction.
```

en/config/actuators.md
```diff
@@ -0,0 +1,118 @@
...
+The AUX pins have additional configuration options for camera capture/triggering.
+Selecting these requires a reboot before they are applied.
+
+
+## Actuator Testing
+When testing actuators, make sure that:
+- Motors spin at the minimum. The sliders snap into place at the lower end: at the bottom, the motor is turned off (disarmed), whereas the next slider position commands the minimum thrust.
+  Adjust the minimum output value such that the motors spin at that slider position.
+  :::note
+  VTOL's will automatically turn off motors pointing upwards during forward flight.
```
sfuhrer, line 92:
```suggestion
  VTOLs will automatically turn off motors pointing upwards during fixed-wing flight. For Standard VTOLs that's the motors defined as MC motors, for Tiltrotors the ones that have no tilt servo associated to. Tailsitters use all motors in fixed-wing flight. 
```

en/config/actuators.md
bkueng: Right, we can do that

en/SUMMARY.md
```diff
@@ -41,7 +41,8 @@
   * [Flight Modes](config/flight_mode.md)
   * [Battery](config/battery.md)
   * [Safety](config/safety.md)
-  * [Motors/Servos](config/motors.md)
+  * [Motors](config/motors.md)
```
hamishwillee, line 44: Note, added back.

en/config/actuators.md
```diff
@@ -0,0 +1,140 @@
+# Actuator Configuration and Testing
+
+:::note
```
hamishwillee, line 3: Moved up top so people immediately know why it isn't displayed. We just delete this note when the thing goes default enabled.

en/config/actuators.md
```diff
@@ -0,0 +1,140 @@
...
+
+The multicopter configuration screen looks like this.
+
+![Actuators MC (QGC)](../../assets/config/actuators/qgc_actuators_mc_aux.png)
+
+Note that some settings are hidden unless you select the **Advanced** checkbox in the top right corner.
+
+## Geometry
+
+The geometry section is used to configure any additional geometry-related settings for the selected [airframe](../config/airframe.md).
```
hamishwillee, line 21: I reworked this section and added a screenshot. Key point here is that geometry is the "expectation". Also that it might display different things for different frames. 

NOTE, I cannot get this image to display myself - i.e. no images of the frame etc.

en/config/actuators.md
hamishwillee: How does this work? I mean elsewhere we talk about motors having a single direction and reversing is used to make the motor go in that correct direction.

Here it sounds like you're saying that it is part of the vehicle geometry that it must have a reversible motor/servo and this is how you define it. Is that right? I guess that makes sense for a ground vehicle in particular so it can go backwards.

This only appears when relevant for your selected frame?

en/config/actuators.md
```diff
@@ -0,0 +1,140 @@
...
+
+
+## Actuator Testing
+
+When testing actuators, make sure that:
+- Motors spin at the "minimum thrust" position.
+
+  The sliders snap into place at the lower end, and motors are turned off (disarmed).
+  The "minimum thrust" position is the next slider position, which commands the minimum thrust.
+  Adjust the minimum output value such that the motors spin at that slider position.
```
hamishwillee, line 109: How do we do this? On this screen we can configure max/min value for some servos but not for motors.

en/config/actuators.md
```diff
@@ -0,0 +1,140 @@
...
+PWM outputs are grouped according to the hardware groups of the autopilot.
+Each group allows to configure the PWM rate or DShot/Oneshot (if supported).
+
+:::note
+For boards with MAIN and AUX, prefer the AUX pins over the MAIN pins for motors, as they have lower latency.
+:::
+
+The AUX pins have additional configuration options for camera capture/triggering.
+Selecting these requires a reboot before they are applied.
+
```
hamishwillee, line 100: Do you want to say something about "Identify and assign motors" button here?

en/config/actuators.md
bkueng: It's not a must, it's a feature. Yes for rovers it's most relevant, but there's various other use-cases, like a vtol pusher to go back and forwards.
It's generally an advanced option.

en/config/actuators.md
bkueng: We can mention that this is mostly relevant when PWM is used. For DShot it's not required.

en/config/actuators.md
bkueng: This is one of the airframe-specifics we can expand on, yes.

en/config/actuators.md
hamishwillee: I can add that, but can you confirm that, for PWM, this is something you have to do outside of this screen/configuration? i wonder if we should provide PWM range setting controls for motors using PWM?

en/config/actuators.md
hamishwillee: Thanks. What I mean is, this will only be an option in geometry that is offered for frames where it makes sense - so you can't just decide that one motor in an MC is reversible because you haven't specified that in the actuator config file as an option. Right?

I'll update this section a bit.

en/config/actuators.md
hamishwillee: OK. Probably worth waiting. When you say this is all good to go, let's go through the doc again.

en/config/actuators.md
```diff
@@ -0,0 +1,150 @@
...
+  - Yaw: the tilts are used to control yaw (generally desired).
+    If four or more motors are used, the motors can be used instead.
+  - Pitch: typically differential motor thrust is used to control pitch, but some airframes require pitch to be controlled by the tilt servos.
+    Bicopters are among those.
+- Tiltable motors are then assigned to one of the tilt servos.
+
+![Tilt Axis](../../assets/config/actuators/tilt_axis.png)
+
+
+### Bidirectional Motors
```
hamishwillee, line 82: @bkueng Not, I have updated this section from reversible to bidirectional motors. Why? Because we were also using reversible to mean a motor that can configure to go in either direction.

Do you think perhaps we could also change QGC to use the term "bidirectional"?

hamishwillee:
There are some comments above, but I think this is probably OK to go in now. We can update again when FW and VTOL are fully supported or when control allocation is fully enabled.

Let me know if you want me to merge or not.

en/config/actuators.md
hamishwillee: @bkueng Did you see this? Are you otherwise good for this to merge?

en/config/actuators.md
bkueng: No, min/max will show up if you have PWM selected (as in the screenshot), and only be hidden if DShot is selected.

en/config/actuators.md
bkueng: Yes bidirectional is fine.
- instead of 'required', can you use something like 'you have the option to...'? Even for rovers you don't need it, if you don't care about only going forwards.
- `the option will only be displayed if relevant for the frame type` -> you always have the option under 'advanced'

en/config/actuators.md
```diff
@@ -0,0 +1,150 @@
...
+
+The geometry section is used to configure any additional geometry-related settings for the selected [airframe](../config/airframe.md).
+The UI displays a customised view for the selected type; if there are no configurable options this may display a static image of the frame geometry, or nothing at all.
+
+The screenshot below shows the geometry screen for a multicopter frame, which allows you to select the number of motors, their relative positions on the frame, and their expected spin directions (select "**Direction CCW**" for counter-clockwise motors).
+This particular frame also includes an image showing the motor positions, which dynamically updates as the motors settings are changed.
+
+![Geometry MC (QGC)](../../assets/config/actuators/qgc_actuators_mc_geometry.png)
+
+A fixed wing airframe would instead display the parameters that define control surfaces, while a VTOL airframe would display both motors and control surfaces.
```
hamishwillee, line 33:
```suggestion
A fixed wing airframe would instead display the parameters that define control surfaces, while a VTOL airframe could display both motors and control surfaces.
```

en/config/actuators.md
```diff
@@ -0,0 +1,150 @@
...
+    Bicopters are among those.
+- Tiltable motors are then assigned to one of the tilt servos.
+
+![Tilt Axis](../../assets/config/actuators/tilt_axis.png)
+
+
+### Bidirectional Motors
+
+Some geometries require bidirectional motors (i.e. motor spins in direction 1 for lower output range and in direction 2 for the upper half).
+For example, ground vehicles that need to move forwards and backwards, or VTOL vehicles that have pusher motors that go in either direction.
```
hamishwillee, line 85:
```suggestion
Some vehicles may use bidirectional motors (i.e. motor spins in direction 1 for lower output range and in direction 2 for the upper half).
For example, ground vehicles that want to move forwards and backwards, or VTOL vehicles that have pusher motors that go in either direction.
```

en/config/actuators.md
```diff
@@ -0,0 +1,150 @@
...
+
+If bidiectional motors are required, make sure to select the **Reversible** checkbox for those motors (the option will only be displayed if relevant for the frame type).
```
hamishwillee, line 87:
```suggestion
If bidiectional motors are used, make sure to select the **Reversible** checkbox for those motors (the option will only be displayed if relevant for the frame type).
```

en/config/actuators.md
hamishwillee: > Yes bidirectional is fine.

It would be even better if you changed it in QGC :-).

I've added as (accepted) suggestions below. Note that when I pressed advanced checkbox in Sim last time I did not see the reversible option, which is why I assumed this is a feature of the airframe. Nevertheless I'll assume the sim is incorrect, not the intent.

hamishwillee:
@bkueng This is good initial docs for now. Thanks. Merging.

en/config/actuators.md
bkueng: Already in the pipeline: https://github.com/PX4/PX4-Autopilot/pull/18990/commits/d23b5fdc0ea7f6afa47b9715e3d4723c523397b2 :-)
Hmm did you check on the geometry side?

image

hamishwillee: This is simulated iris from 23 December, with advanced checked.

image

hamishwillee: Fixed in todays build:
image
So we can forget this

PR #1711, opened 2021-12-21
removed wrong comma

dippa-1:


hamishwillee (approved):
Good catch. Thanks very much.

PR #1720, opened 2022-01-04
Fixed wing manual pid - remove PSP_OFFSET

### LINKED ISSUE #1709: Unclear Fixed-wing PID Tuning Guide
While going through the [fixed-wing PID tuning guide](https://docs.px4.io/master/en/config_fw/pid_tuning_guide_fixedwing.html#tune-pitch) I was not sure, wheter I need to correctly set the pitch offset (orange box) or set it to zero (blue box).

image

Do I only need to set it to zero while tuning the feedforward gain and use the correct pitch offset for the P and I gains?

Am I missing something?

hamishwillee:
Fixes #1709

@sfuhrer I am not at all sure this is correct. Your response in https://github.com/PX4/PX4-user_guide/issues/1709#issuecomment-1004701509 was

> ... let's remove this line. In my eyes the PSP_OFFSET is independent of the tuning gains, so I would set it first thing and then don't though it again.

So, line removed. I also removed `FW_RSP_OFF ` when tuning roll.

But you said "I would set it first thing". How do we set it? This does not appear to be covered in the document?

### DIFF
--- a/en/config_fw/pid_tuning_guide_fixedwing.md
+++ b/en/config_fw/pid_tuning_guide_fixedwing.md
@@ -56,7 +56,6 @@ To tune this gain, set the other gains to zero.
 
 - FW_RR_I = 0
 - FW_RR_P = 0
-- FW_RSP_OFF = 0
 
 
 #### Gains to tune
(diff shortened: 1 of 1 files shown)


sfuhrer (approved):
>But you said "I would set it first thing". How do we set it? This does not appear to be covered in the document?

It's covered in the Trimming part: https://docs.px4.io/master/en/config_fw/trimming_guide_fixedwing.html. I think that's enough, it's mentioned that the trimming should be done first.

hamishwillee:
Thanks! I added a link to trims in the preconditions.

PR #1724, opened 2022-01-07
Add a description of wiring requirements in UAV

kaklik:
The section [Basic assembly](https://docs.px4.io/master/en/assembly/) is missing a section for _Cable Wiring_.  Where should be described: 

1. How to avoid EMI troubles by the cable installation (high power RF emissions). 
2. How to correctly wire and **twist** I²C cables 
3. How to wire CAN bus cables correctly

This information is mandatory to have proper working I2C sensors in unmanned vehicles.


en/assembly/cable_wiring.md
```diff
@@ -0,0 +1,186 @@
...
+|Signal | Pixhawk Colors | ThunderFly colors | CUAV colors (I2C/CAN) |
+|--------|------------------|---------------------|-----------------------|
+| +5V  | Red             |   ![red](../../assets/hardware/cables/red.png) Red                   | Red                   |
+| SCL  | Black          |  ![yellow](../../assets/hardware/cables/yellow.png) Yellow            | White                |
+| SDA  | Black          |  ![green](../../assets/hardware/cables/green.png) Green               | Yellow               |
+| GND | Black          | ![black](../../assets/hardware/cables/black.png) Black                 | Black                 |
+
+The [Dronecode standard](https://github.com/pixhawk/Pixhawk-Standards/blob/master/DS-009%20Pixhawk%20Connector%20Standard.pdf) assumes a 1.5k ohm pull-up resistor on SDA and SCL signals in autopilot.
+
+#### Cable twisting
```
hamishwillee, line 60: Does the benefit of cable twisting only apply to I2C? As this only covers I2C perhaps put that into the heading.

hamishwillee: Ignore, sorry, didn't read ahead to see you cover this in each section

hamishwillee:
Looks great - a topic that has been missing for a long time. I'll give it a proper review on Wednesday. 

@bkueng @MaEtUgR Could you guys provide technical feedback on this? Not just what is correct/incorrect, but what might be missing. For example, does it matter where or how you lay your cables - i.e. can all cable jumble together. Are there any rules about  where you might choose to not put your cables. Advice on length. Advice on cable terminations etc.

At some point we should also add a topic on common connector types.

bkueng:
Nice. I'm certainly no expert here, but in case we don't have this already:
- GPS is generally very sensitive to EMI, might need shielding depending on the setup
- Preferrably using CAN for longer lines instead of I2C (relevant for bigger airframes, e.g. airspeed sensor on a larger vtol)

kaklik:
> Nice. I'm certainly no expert here, but in case we don't have this already:
> 
>     * GPS is generally very sensitive to EMI, might need shielding depending on the setup

Yes, that is one of the reasons, forced me to make that page, although the general GPS sensitivity should be mentioned on "GPS mounting instructions". Those instructions could be useless in the case where EMI is too large, and generally, the main source of EMI radiation are cables or signal sources connected by wires. PCB boards themselves usually radiate only a fraction of all radiated energy. 

> 
>     * Preferrably using CAN for longer lines instead of I2C (relevant for bigger airframes, e.g. airspeed sensor on a larger vtol)

Yes, that mention could improve the guide.

en/assembly/cable_wiring.md
```diff
@@ -0,0 +1,186 @@
+# Cable Wiring Basics
+
+Generally the cables are the main sources of [EMI](https://en.wikipedia.org/wiki/Electromagnetic_compatibility) troubles. Therefore many electrical hardware related failures could be avoid by proper cabling in the UAV. There are a few basics concepts related to RF induced failures to which are mandatory to mind during design of cabling in the drone.
+
+  * The High-Power and signaling cabling should be separated as much as is practical    
+  * The lengths of cabling should be reduced to amount necessary to easy handling of wired components. Tension of wires should be adequate to possible airframe deformations even in a crash landing. (The wires must not be the first to break)  
+  * Cable loops to reduce excess length needs to be avoided
```
hamishwillee, line 7: IMO it is worth identifying what problems might be caused. As far as I know the main one is EMI affecting the compasses resulting in poor directional control.

```suggestion
Cables are a common source of [electromagnetic interference (EMI)](https://en.wikipedia.org/wiki/Electromagnetic_interference), which can cause problems including flyways, "toilet bowling" and generally poor flight.
These problems can be avoided by using appropriate cabling
 in the UAV. 
 
The following basic concepts should be kept in mind when designing drone cabling:
* High-Power and signal cables should be separated as much as is practical    
* Cable lengths should be the minimum needed to enable easy handling of wired components.
  The wire tension should be adequate to survive possible airframe deformations even in a crash landing
  (wires must not be the first thing to break).
* Cable loops to reduce excess length should be avoided: use shorter lengths
```

hamishwillee, line 7: IMO it is worth identifying what problems might be caused. As far as I know the main one is EMI affecting the compasses resulting in poor directional control.

```suggestion
Cables are a common source of [electromagnetic interference (EMI)](https://en.wikipedia.org/wiki/Electromagnetic_interference), which can cause problems including flyways, "toilet bowling" and generally poor flight.
These problems can be avoided by using appropriate cabling
 in the UAV. 
 
The following basic concepts should be kept in mind when designing drone cabling:
* High-Power and signal cables should be separated as much as is practical    
* Cable lengths should be the minimum needed to enable easy handling of wired components.
  The wire tension should be adequate to survive possible airframe deformations even in a crash landing
  (wires must not be the first thing to break).
* Cable loops to reduce excess length should be avoided: use shorter lengths
```

hamishwillee:
@kaklik I would like to edit this PR but your settings prevent it. That makes this difficult since I don't want to have to put in a suggestion for every typo. Can you open your repo for me to edit? Note, tomorrow is my last day for a few weeks so I hope you are able to respond today :-)

In general this looks great. With respect to: https://github.com/PX4/PX4-user_guide/pull/1724#issuecomment-1008850235

>> Nice. I'm certainly no expert here, but in case we don't have this already:
>> 
>>     * GPS is generally very sensitive to EMI, might need shielding depending on the setup

> Yes, that is one of the reasons, forced me to make that page, although the general GPS sensitivity should be mentioned on "GPS mounting instructions". Those instructions could be useless in the case where EMI is too large, and generally, the main source of EMI radiation are cables or signal sources connected by wires. PCB boards themselves usually radiate only a fraction of all radiated energy. 

IMO we **should** mention compass sensitivity here anyway,  and cross link to the mounting instructions for more information. 
You could even note that mounting the compass (+GPS) away is a good idea, but may be insufficient if cabling is badly designed.

If there are other things that are significantly affected by EMI then we should mention those too up the very top (see my suggestion above). 

>>     * Preferably using CAN for longer lines instead of I2C (relevant for bigger airframes, e.g. airspeed sensor on a larger vtol)

> Yes, that mention could improve the guide. But I do not know how to set the I2C based airframe limits. I know about I2C network installation even in size comparable to buildings. 

@bkueng Can you give us a broad rule of thumb on frame sizes/line lengths at which it makes sense to start using CAN and for what components? 

What I mean by "what components" is I recall you talking about Dshot ESC and saying that you could choose a lower rate for a longer cable length, and that UAVCAN is robust but our implementation is not very high rate. We don't use I2C for motors but I'm wondering if there are other I2C components or conditions where the decision to use I2C vs UAVCAN might be based on something other than line length.

bkueng:
> Note, tomorrow is my last day for a few weeks

:scream:  

>  @bkueng Can you give us a broad rule of thumb on frame sizes/line lengths at which it makes sense to start using CAN and for what components?

Hmm somewhere around 50cm cable length. I'm sure @kaklik has a better answer though.

> What I mean by "what components" is I recall you talking about Dshot ESC and saying that you could choose a lower rate for a longer cable length, and that UAVCAN is robust but our implementation is not very high rate.

Yes this is a good point to mention as well: lower baudrate -> more robust. You typically need to measure the signal to have a good understanding if lowering is required.

kaklik:
> @kaklik I would like to edit this PR but your settings prevent it. That makes this difficult since I don't want to have to put in a suggestion for every typo. Can you open your repo for me to edit? Note, tomorrow is my last day for a few weeks so I hope you are able to respond today :-)

Ok, sorry for that. I did not find how to apply [repo opening](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork), the option I'm unable to see. I send you an invitation as an "outside collaborator"

I had tried to solve every main point you mentioned in the comments. @bkueng  @hamishwillee. In the meantime, I found that in some cases the oscilloscope images of correct signals might be helpful, but it is probably over the current scope of that pull request.

hamishwillee:
I have given this a subedit and made a few other changes:
- The coloured dots are now used for all cases, not just thunderfly. I used a different markdown syntax for that to improve the layout.
- I moved the cable design guide down the bottom - it is useful but you can jump just to the cable information in the signaling section. I also reframed this as "general guidance contributed by Thunderfly" 
- I rebased and squashed. We lose a bit of history, but it is worth it for the shrunken images.

This is really useful, so I am going to merge.

However @bkueng would really appreciate your post-merge final review.

Thanks very much @kaklik - I've been hoping for something like this for a long time.

PR #1734, opened 2022-01-14
Remove section about px4_task_spawn_cmd()

ivodre:
Seems not to be used anymore.


bkueng:
This is still used, just not for work queue tasks.

PR #1741, opened 2022-01-22
Out-dated line about mavlink headers

ivodre:
For your own use/testing you can just copy the generated headers into **PX4-Autopilot/mavlink/include/mavlink/v2.0**.

### DIFF
--- a/en/middleware/mavlink.md
+++ b/en/middleware/mavlink.md
@@ -24,7 +24,7 @@ Once you've [installed MAVLink](https://mavlink.io/en/getting_started/installati
 python -m pymavlink.tools.mavgen --lang=C --wire-protocol=2.0 --output=generated/include/mavlink/v2.0 message_definitions/v1.0/custom_messages.xml
 ```
 
-For your own use/testing you can just copy the generated headers into **PX4-Autopilot/mavlink/include/mavlink/v2.0**.
+For your own use/testing you can just copy the generated headers into **PX4-Autopilot/mavlink/include/mavlink/v2.0**. // is this outdated?
 
 To make it easier for others to test your changes, a better approach is to add your generated headers to a fork of https://github.com/mavlink/c_library_v2.
 PX4 developers can then update the submodule to your fork in the PX4-Autopilot repo before building.


bkueng:
You are right, this needs to be updated. You can now just update the xml, and the headers get generated at build-time.

hamishwillee:
@bkueng @julianoes By inspection, it looks to me that PX4 automatically builds development.xml (includes common.xml). It does not build the whole tree. Is that correct?

Just testing a proposal locally is pretty easy - you could add your changes in development.xml and they would be built. But what is the workflow for **contribution**?

Specifically, say I'm in the dev team and I want to create a PR with a new feature. I think the workflow ends up having to be something like:
1. Fork mavlink and create a PR with my changes to development.xml
2. Fork PX4-Autopilot and create a PR that references my fork of the submodule instead of the one in master (not sure how you would do that).
3. Do all your code and testing, updating both PRs as needed. 
4. When happy submit changes to development.xml in main repo and hope it gets in. That will depend on mavlink team.
5. Modify your PX4-Autopilot PR back to pointing to the mavlink master development.xml once the mavlink PR goes in
6. Get PX4-Autopilot PR merged.

That sounds a bit horrible, in particular because you keep on having to change the mavlink repo you point at. Or do I miss something?

What about the case of someone who has a fork of PX4. Are they expected to maintain their changes in a modified development.xml/common.xml? Is there any expectation that they will have their own custom_dialect file as shown in the current docs (and if not, why did we originally suggest a custom dialect file?).

It strikes me that development.xml doesn't quite work for ... development. Things would be much easier if we had a px4.xml file so we didn't need to get permission for contribution from anyone outside PX4. Alternatively a local-dev.xml file that lived in PX4-autopilot tree that could be used during development by anyone.

Anyway, the point is that we can't update the doc properly until we know the flow. Can you let me know how you think this should work?

bkueng:
> Is that correct?

Correct, and you can change the dialect: https://github.com/PX4/PX4-Autopilot/blob/master/src/modules/mavlink/CMakeLists.txt#L34

> Fork QGC and get your development.xml in there too - is that possible? Does QGC include development.xml - if not, perhaps then the theory is that you should be making changes in common.xml?

QGC is still using the ardupilot xml.

> That sounds a bit horrible, in particular because you keep on having to change the mavlink repo you point at everywhere. Or do I miss something?

You don't have to change the repo's, you can just open the PX4 PR and reference the mavlink PR. And then update the submodule after the mavlink PR is merged. Only if you want to check if PX4 CI passes before merging mavlink, you need to update the repo.

> Are they expected to maintain their changes in a modified development.xml/common.xml? Is there any expectation that they will have their own custom_dialect file as shown in the current docs (and if not, why did we originally suggest a custom dialect file?).

They can do either way, the cleanest is to create a new dialect though, just so it's clear which messages are custom.

ivodre:
> 

Sorry for not reacting. I wasn't sure how to proceed.

hamishwillee:
No worries, @ivodre - highlighting the issue is a good start. 

@bkueng Thanks. Re "They can do either way, the cleanest is to create a new dialect though, just so it's clear which messages are custom.". That said, we probably need to show where you update QGC too and how to build it. Otherwise it would be hard to test - can you point me to where the XML file is set in QGC?

bkueng:
You can set [MAVLINK_CONF](https://github.com/mavlink/qgroundcontrol/blob/master/QGCExternalLibs.pri#L52) when running `qmake`.

hamishwillee:
Thanks. Closed - I think addressed by https://github.com/PX4/PX4-user_guide/pull/1765

PR #1747, opened 2022-01-31
WSL2 instructions to build, flash and simulate on Windows

MaEtUgR:
I added instructions on how to employ WSL2 to do PX4 development on Windows. I'm successfully using this environment for several months now. It works seamlessly, fast and allows for efficient work. The biggest improvement that brought me to switch to this environment is the native graphics support with 3D acceleration through the built-in [WSLg](https://github.com/microsoft/wslg). I suggest we add these instructions as even a preferred option to the Cygwin toolchain and keep both options for now. I would honestly also surface the VMWare instructions as a valid option because for certain developers it might be the prefered way.

I replaced the BashOnWindows instructions because those are deprecated.


en/dev_setup/dev_env_windows_wsl.md
```diff
@@ -0,0 +1,82 @@
...
+2. Check what IP address it received using the command "ip address" and copy it to clipboard by selecting the IP and using the shortcut [Ctrl] + [Shift] + [C].
+3. In QGC go to Q > Application Settings > Comm Links
+4. Add a UDP Link called "WSL" to port 18570 of the IP address copied before.
+5. Save it and connect to it.
+Note: This has to be repeated when WSL restarts because it gets a dynamic IP address.
+
+## Flash a Flight Control Board
+Flashing a custom built PX4 bianry has to be done using QGroundcontrol for Windows since WSL 2 does not natively offer direct access to serial devices like pixhawk boards. Do the following steps to flash your custom binary built in WSL.
+
+1. If you haven't allready build the binary in WSL e.g. with a [WSL shell](dev_env_windows_wsl.md#opening-a-wsl-shell) and by running:
```
dagar, line 69:
```suggestion
1. If you haven't already build the binary in WSL e.g. with a [WSL shell](dev_env_windows_wsl.md#opening-a-wsl-shell) and by running:
```

hamishwillee:
I will try this out tomorrow. Thanks!

There should be only one recommended main approach for each platform - what would it take for WSL2 to become that option?

MaEtUgR:
> what would it take for WSL2 to become that option?

For my workflow it is already the main option. The caveats are:
- No direct serial flashing with the upload script
- No automatic connection of native QGC to the sim running in WSL
- ~Win 11 requirement for graphics support

This might as well be a complete deal-breaker for some but on the other side the size problems preventing to generate a new Cygwin toolchain with the existing script makes the current version more and more outdated.

hamishwillee:
Thanks @MaEtUgR!

Are any of the issues you listed - direct serial flashing, auto connection of native QGC, graphics support amenable to fixing via configuration in some way or are they likely "forever" problems?

My take on this is that we can live with "No direct serial flashing with the upload script", and "No automatic connection of native QGC to the sim running in WSL." (if we have to, very undesirable) but it is much too early to insist on Win 11.

The graphics support, was not present in older WSL, but I think it is part of WSL2 rather than Windows 11 right? If so then that should also be present on Windows 10?

hamishwillee:
I got the error `WARNING: An illegal reflective access operation has occurred` running JMAVSim. Anything special I need to do?
```
Feb 02, 2022 4:35:32 PM java.util.prefs.FileSystemPreferences$1 run
INFO: Created user preferences directory.
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by javax.media.j3d.JoglPipeline (rsrc:j3dcore.jar) to method sun.awt.AppContext.getAppContext()
WARNING: Please consider reporting this to the maintainers of javax.media.j3d.JoglPipeline
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
Exception in thread "main" java.lang.reflect.InvocationTargetException
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
```

en/SUMMARY.md
```diff
@@ -397,7 +397,7 @@
       * [CentOS Linux](dev_setup/dev_env_linux_centos.md)
       * [Arch Linux](dev_setup/dev_env_linux_arch.md)
       * [Windows VM Toolchain](dev_setup/dev_env_windows_vm.md)
-      * [Windows WSL Toolchain](dev_setup/dev_env_windows_bash_on_win.md)
```
hamishwillee, line 400: Yes, removing this was a good decision.

en/dev_setup/dev_env_windows.md
```diff
@@ -19,23 +19,23 @@ Once you have finished setting up the command-line toolchain:
...
 A comparison of the options is provided below.
 
 :::note
 The [Cygwin Toolchain](../dev_setup/dev_env_windows_cygwin.md) is the only one that is supported by the PX4 dev team.
-It is regularly tested as part of our continuous integration system and is known to be better performing than the other alternatives. 
+It is regularly tested as part of our continuous integration system and is known to be better performing than the other alternatives.
 :::
 
-| | [Cygwin Toolchain](../dev_setup/dev_env_windows_cygwin.md) **(Supported)**  | [Virtual Machine Toolchain](../dev_setup/dev_env_windows_vm.md) | [Bash on Windows Toolchain](../dev_setup/dev_env_windows_bash_on_win.md) |
+| | [Cygwin Toolchain](../dev_setup/dev_env_windows_cygwin.md) **(Supported)**  | [Virtual Machine Toolchain](../dev_setup/dev_env_windows_vm.md) | [WSL Toolchain](../dev_setup/dev_env_windows_wsl.md) |
```
hamishwillee, line 38: There is a table below here. Do we need to update it. I note that the instructions only cover Pixhawk, Gazebo, Jmavsim, so we should perhaps at least highlight JMAVsim too?

Note that I expect some of the other cases will work too.

en/dev_setup/dev_env_windows_wsl.md
```diff
@@ -0,0 +1,161 @@
...
+   
+   :::note
+   This installs tools to build PX4 for Pixhawk, Gazebo and JMAVSim targets.
+   - You can use the `--no-nuttx` and `--no-sim-tools` options to omit the NuttX and/or simulation tools.
+   - Other Linux build targets are untested (you can try these by entering the appropriate commands in [Ubuntu Development Environment](../dev_setup/dev_env_linux_ubuntu.md) into the WSL shell.
+     Note that `git` is already installed in WSL.
+   
+1. Restart the "WSL computer" on completion.
+   ```
+   wsl --shutdown
```
hamishwillee, line 82: FYI `wsl` and `wsl --shutdown` don't work for me:
```
Command 'wsl' not found, but can be installed with:
sudo apt install wsl
```
Doesn't seem to matter - I just exit and restart.

en/dev_setup/dev_env_windows_wsl.md
```diff
@@ -0,0 +1,161 @@
...
+
+
+## Visual Studio Code Integration
+
+1. Download and install Visual Studio Code on Windows from https://code.visualstudio.com/
+2. Open _Visual Studio Code_.
+3. Install the extension called [Remote - WSL](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-wsl) (marketplace) 
+4. [Open a WSL shell](dev_env_windows_wsl.md#opening-a-wsl-shell)
+5. Switch to the PX4 folder: `cd ~/PX4-Autopilot`
+6. Open this folder with WSL integration in Visual Studio Code `code .`
```
hamishwillee, line 107: How? I presume you mean find the folder in the windows C drive? Note that the Linux penguin doesn't show up in VS Code so you can't select the WSL version in the same way as you would in Explorer?

en/dev_setup/dev_env_windows_wsl.md
```diff
@@ -0,0 +1,161 @@
...
+7. Now you have the full IDE integration with Terminal, make sure you always open the PX4 repository in the Remote WSL mode.
```
hamishwillee, line 108: What does that actually mean? I.e. does this mean I should be able to see a build target after this?

en/dev_setup/dev_env_windows_wsl.md
```diff
@@ -0,0 +1,161 @@
...
+   It will be available in your recently opened locations whenever you start VS Code.
+
+## QGroundControl
+
+You can run QGroundControl in either WSL or Windows to connect to the running simulation.
+If you need to [flash a flight control board](#flash-a-flight-control-board) with new firmware you can only do this from the QGroundControl for Windows.
+
+### QGroundControl in WSL
+
+The easiest way to set up and use QGroundControl is to download the Linux version into your WSL.
```
hamishwillee, line 118: What approach do you use for this - `wget` linux version into the home directory? Do you need to set it as executable?

en/dev_setup/dev_env_windows_wsl.md
```diff
@@ -0,0 +1,161 @@
...
+
+
+### QGroundcontrol on Windows
+
+Install QGroundControl on Windows if you want to be able to update hardware with firmware created within PX4.
+
+These steps describe how you can connect to the simulation running in the WSL.
+
+1. [Open a WSL shell](dev_env_windows_wsl.md#opening-a-wsl-shell)
+2. Check what IP address it received using the command `ip address` and copy it to clipboard by selecting the IP and using the shortcut [Ctrl] + [Shift] + [C].
```
hamishwillee, line 131: 1. Does this ever change? I.e. do I have to get new address every time I run the WSL?
   If so, yuk. Might be better to suggest use QGC in WSL for flying and ANOTHER QGC in windows, just for uploading. Thoughts?
   
2. This is what I see. Is the IP address the one listed for Ethernet? 172.19.7.31/20 

```
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host
       valid_lft forever preferred_lft forever
2: bond0: <BROADCAST,MULTICAST,MASTER> mtu 1500 qdisc noop state DOWN group default qlen 1000
    link/ether 2e:0f:f9:cc:bd:77 brd ff:ff:ff:ff:ff:ff
3: dummy0: <BROADCAST,NOARP> mtu 1500 qdisc noop state DOWN group default qlen 1000
    link/ether ae:09:c7:b6:44:e7 brd ff:ff:ff:ff:ff:ff
4: tunl0@NONE: <NOARP> mtu 1480 qdisc noop state DOWN group default qlen 1000
    link/ipip 0.0.0.0 brd 0.0.0.0
5: sit0@NONE: <NOARP> mtu 1480 qdisc noop state DOWN group default qlen 1000
    link/sit 0.0.0.0 brd 0.0.0.0
6: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000
    link/ether 00:15:5d:be:1b:8f brd ff:ff:ff:ff:ff:ff
    inet 172.19.7.31/20 brd 172.19.15.255 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::215:5dff:febe:1b8f/64 scope link
       valid_lft forever preferred_lft forever
```

en/dev_setup/dev_env_windows_wsl.md
```diff
@@ -0,0 +1,161 @@
+# Windows WSL-Based Development Environment
```
hamishwillee, line 1: I have done a very rough update of this. I want to get it closer to the very simple instructions for Windows currently or for macOS. Not easy since there are lots of caveats. 

However I think we should iterate quickly, get it in, then think about next steps to make it "the way".

MaEtUgR:
> are they likely "forever" problems?

Theoretically there can be workarounds. E.g. serial flashing could maybe call a pyton script outside of WSL and for the automatic connection to native QGC we could in theory run MAVLink router in WSL to forward the conntection to Windows. So easy to fix no, possibly yes with some cumbersome workarounds.

> but it is much too early to insist on Win 11

I thought so too. It's only for graphics support that the Windows version has to be so recent.

> You state in "Requirements for graphics support" that this is in Windows insider preview builds, so assuming it will go into Windows 10 soon

I expected the same since I had it running on the Windows 10 insider build before I switched to Windows 11 stable. But when I checked https://github.com/microsoft/wslg#pre-requisites again it switched from Windows 10 insider build to Windows 11 insider build 🤔 Maybe it's just to tease or not up to date. In any case Windows 11 stable works for sure. Windows 10 I'd need to test again.

> That said, if we can somehow route the comms to localhost on the host, that would be even better.

Do you mean such that e.g. jMAVsim could run natively? That's also theoretically possible but with multiple workarounds to start and connect it all.

> I got the error WARNING: An illegal reflective access operation has occurred running JMAVSim.

I got that as well. Was assuming it's the same for Ubuntu since it's basically running on Ubuntu. I remember jMAVsim worked with this setup earlier on. So I was assuming it's a general issue.

en/dev_setup/dev_env_windows.md
MaEtUgR: I have to check if the table is still up to date 👀

en/dev_setup/dev_env_windows_wsl.md
MaEtUgR: Sorry if it wasn't written clearly. The commands `wsl --shutdown` and `wsl` are to be ran in a native Windows cmd shell. So if the WSL shell was accessed through cmd after typing `exit` on the WSL bash and returning to the original shell.

en/dev_setup/dev_env_windows_wsl.md
MaEtUgR: You install the extension in the VS code gui, then open a WSL shell, navigate to the folder with the command `cd ~/PX4-Autopilot` and in this folder execute the command `code .` as if you would be on Linux and the repository opens in VS code but natively. They do some trickery WSL magic to make that happen.

en/dev_setup/dev_env_windows_wsl.md
MaEtUgR: You see everything the same as if you would be using VS code on Ubuntu with the repository opened. So you can also use the integrate terminal to execute command which is what I use in my workflow but you'll also see the configured targets and so on in the GUI.

MaEtUgR: image

en/dev_setup/dev_env_windows_wsl.md
MaEtUgR: Inside WSL execute:
```
cd ~
wget https://d176tv9ibo4jno.cloudfront.net/builds/master/QGroundControl.AppImage
chmod +x QGroundControl.AppImage
./QGroundControl.AppImage
```

en/dev_setup/dev_env_windows_wsl.md
MaEtUgR: 1. Every time WSL is restarted either because you reboot the computer or because you do that manually using `wsl --shutdown`. I know it's not great, not sure how this IP could be made static. I didn't find an easy solution.
2. Exactly `172.19.7.31` in your case.

en/dev_setup/dev_env_windows_wsl.md
MaEtUgR: Yes, I thought the same. Rather have the not yet polished way documented but not as the main way until it's further than just not documenting that it's usable at all.

hamishwillee:
>> That said, if we can somehow route the comms to localhost on the host, that would be even better.

> Do you mean such that e.g. jMAVsim could run natively? That's also theoretically possible but with multiple workarounds to start and connect it all.

@MaEtUgR No, I mean setting things up such that you can automatically connect to simulator from QGC on Windows (without having to work out the IP address of your virtual machine). You answered here https://github.com/PX4/PX4-user_guide/pull/1747#discussion_r801070979 as "maybe, but I'm not sure how".

>> I got the error WARNING: An illegal reflective access operation has occurred running JMAVSim.

> I got that as well. Was assuming it's the same for Ubuntu since it's basically running on Ubuntu. I remember jMAVsim worked with this setup earlier on. So I was assuming it's a general issue.

You would be right/it is: http://docs.px4.io/master/en/simulation/jmavsim.html#an-illegal-reflective-access-operation-has-occured

en/dev_setup/dev_env_windows_wsl.md
hamishwillee: The problem here is that it changes every time - so you would have to update QGC every time. That's a real pain. I guess for now we document it, and think about possible alternatives. I am sure it is possible for a VM to be allocated a static address.

en/dev_setup/dev_env_windows.md
hamishwillee: OK. Update at will.

en/dev_setup/dev_env_windows_wsl.md
hamishwillee: So this fails
```
wsl
wsl --shutdown
```
but this works
```
wsl
exit
wsl --shutdown
```
So I think you are saying you can only run wsl-shutdown after exiting? in the raw cmd prompt?

en/dev_setup/dev_env_windows_wsl.md
```diff
@@ -0,0 +1,238 @@
...
+   ```
+   cd ~/PX4-Autopilot
+   ```
+6. In the WSL shell, start VS Code:
+   ```
+   code .
+   ```
+   This will open the IDE fully integrated with the WSL shell.
+   Make sure you always open the PX4 repository in the Remote WSL mode.
+   It will be available in your recently opened locations whenever you start VS Code.
```
hamishwillee, line 150: I get items 5, 6 now - after running VS Code using `code .` it is associatd with the WSL shell. Further, you're saying if you open recent the option will be present to start this again. Is that right?

To be very clear, if I turned by computer on and off and came back to VS code WSL would have closed down. Are you saying that selecting recent would reopen WSL with the correct integration? What about things like the IP address - even if I reopen WSL2 I no longer have knowledge of what that is. Is it that if I open WSL in a prompt again it will open the same WSL as VS Code does?

hamishwillee:
Awesome. Nearly there. Would be good if you could sanity check my changes.

en/dev_setup/dev_env_windows_wsl.md
MaEtUgR: > But what if I turn my computer on/off and came back to VS code. WSL would have closed down.

Yes, it shuts down the WSL kernel when you turn off your computer (or run `wsl --shutdown` in a Windows shell).

> Are you saying that selecting the option would reopen WSL with the correct integration?

Yes, as soon as you access one of the repositories you're working on through WSL e.g.
image
it will automatically start WSL and open the repository with the same setup you had open when you last left.

> What about things like the IP address - even if I reopen WSL2 I no longer have knowledge of what that is.

The IP address of WSL changes with every restart. VS code does all the magic in the background but for the ground station that is not running within WSL you have to manually check the IP and add it to the "comm link" of QGC.

> Is it that if I open WSL in a prompt again it will open the same WSL as VS Code does?

Yes, if you run the `wsl` command on a Windows shell it will enter a shell on the same WSL shell as VSCode is using.

Note: You can in principle have multiple WSLs and even different distributions but I never bothered. Then you have to always specify what opens where.
Note2: You can also open the repository which is in WSL directly in VSCode (just the folder) but then you don't have all the advantages of the linux like experience within the IDE. That's why it matters how you open it all the way at the begining and afterwards you can just "open recent" and reuse the setup.

en/dev_setup/dev_env_windows_wsl.md
```diff
@@ -13,70 +25,137 @@ Alternatively QGC for Linux can be run within WSL
...
+1. Run the **ubuntu.sh** installer script and acknowledge any prompts as the script progresses:
+   ```bash
+   bash ./PX4-Autopilot/Tools/setup/ubuntu.sh
+   ```
+   
+   :::note
+   This installs tools to build PX4 for Pixhawk, Gazebo and JMAVSim targets.
+   - You can use the `--no-nuttx` and `--no-sim-tools` options to omit the NuttX and/or simulation tools.
+   - Other Linux build targets are untested (you can try these by entering the appropriate commands in [Ubuntu Development Environment](../dev_setup/dev_env_linux_ubuntu.md) into the WSL shell.
+     Note that `git` is already installed in WSL.
```
MaEtUgR, line 102: I would say that one step above because it asks to use git.

en/dev_setup/dev_env_windows_wsl.md
```diff
@@ -13,70 +25,137 @@ Alternatively QGC for Linux can be run within WSL
...
-5. Switch to the PX4 folder `cd ~/PX4-Autopilot`
+5. Switch to the PX4 folder: `cd ~/PX4-Autopilot`
 6. Open this folder with WSL integration in Visual Studio Code `code .`
-7. Now you have the full IDE integration with Terminal, make sure you always open the PX4 repository in the Remote WSL mode. It will be available in your recently opened locations whenever you start VS Code.
+7. Now you have the full IDE integration with Terminal, make sure you always open the PX4 repository in the Remote WSL mode.
+   It will be available in your recently opened locations whenever you start VS Code.
+
+## QGroundControl
+
+You can run QGroundControl in either WSL or Windows to connect to the running simulation.
```
MaEtUgR, line 154: You can either run the Linux version of QGC within WSL or the Windows version of QGC outside of WSL.
If QGC is ran inside WSL it will automatically connect to your simulation but show through the WSLg graphics layer and is not be able to flash a flight control board. If you run QGC for Windows it runs native and can flash flight control boards but you have to manually set up the communication link to the simulation running in WSL.

Does that make sense?

en/dev_setup/dev_env_windows_wsl.md
MaEtUgR: > The problem here is that it changes every time - so you would have to update QGC every time. That's a real pain.

I know I have to deal with this all the time...

> I guess for now we document it, and think about possible alternatives.

Sure, it might be possible to automate this in some way. The only solution I know of so far is from @lyraik. He ran MAVLink router with the right configuration in WSL and that forwards the MAVLink to native Windows so it will auto connect. It's additional dependencies and would probably need to be scripted up but it's the easiest I know of.

> I am sure it is possible for a VM to be allocated a static address.

I don't think so, see https://superuser.com/a/1619390

hamishwillee:
@MaEtUgR One more idea for "static IP address" workaround. Not sure it is "scalable" or something we can support, but perhaps something we could mention.

A similar problem comes up virtual websites - every time they boot they get a new IP. The solution as I understand it is Dynamic DNS. Yes your IP address changes, but on boot you find the address and send it to the DNS system which then updates. So you end up doing your addressing via DNS. https://www.noip.com/support/knowledgebase/installing-the-linux-dynamic-update-client-on-ubuntu/

Something to think about.

en/dev_setup/dev_env_windows.md
hamishwillee: Actually, don't bother. This page is not actually in the sidebar, or built. I magically redirect it (https://docs.px4.io/master/en/dev_setup/dev_env_windows.html) to the cygwin page 

We should delete it, but I kept it in case we reverted the decision to have multiple supported versions.

en/dev_setup/dev_env_windows_wsl.md
```diff
@@ -13,70 +25,137 @@ Alternatively QGC for Linux can be run within WSL
...
-   Can be done by pressing the start key, typing "cmd" and pressing enter.
+1. Open `cmd` again as a normal user (not as administrator).
+   This can be done by pressing the **Start** key, typing "cmd" and pressing **Enter**.
 1. Execute the command `wsl` to access the WSL shell.
-1. Fill out the prompt for user name and password for the Ubuntu in your WSL. Be sure to remember the credentials for later on.
+1. WSL will prompt you for a user name and password for the Ubuntu installation.
+   **Note:** _Remember the credentials for later on!_
+
+:::note
+If you have any problems with your setup, check the current microsoft WSL installation documentation.
```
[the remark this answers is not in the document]
hamishwillee:
```suggestion
If you have any problems with your setup, check the current [Microsoft WSL installation documentation](https://docs.microsoft.com/windows/wsl/).
```

en/dev_setup/dev_env_windows_wsl.md
```diff
@@ -13,70 +25,137 @@ Alternatively QGC for Linux can be run within WSL
...
+:::
+
+### Install PX4 Toolchain
+
+To install the development toolchain for Gazebo simulation, JMAVSim simulation and Pixhawk/NuttX harware:
+
 1. Execute the command `cd ~` to switch to the home folder of WSL for the next steps.
```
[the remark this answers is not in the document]
hamishwillee: Not completely. Fixed now.

en/dev_setup/dev_env_windows_wsl.md
hamishwillee: Crystal clear. I moved the open command earlier and I only refer to this way of doing things now.

en/dev_setup/dev_env_windows_wsl.md
hamishwillee: Excellent. Clear now.

en/dev_setup/dev_env_windows_wsl.md
```diff
@@ -0,0 +1,270 @@
...
+   ```
+   cd ~/PX4-Autopilot
+   ```
+6. In the WSL shell, start VS Code:
+   ```
+   code .
+   ```
+   This will open the IDE fully integrated with the WSL shell.
+   
+   Make sure you always open the PX4 repository in the Remote WSL mode.
```
hamishwillee, line 173: FYI, @MaEtUgR Thanks for your explanation of the WSL integration.

To make this clear to others, I have added this line here - so that make it very clear that "this" is what it mean to open the PX4 repository in the Remote WSL mode.

Then in step 7 I add your image of how you "Open Recent" and not that it is also " opening in Remote WSL mode "

hamishwillee:
@MaEtUgR Thanks for your patience. I've integrated your comments. Also done a cross link from Cygwin recommending this if you're building against master. There is a suggestion here https://github.com/PX4/PX4-user_guide/pull/1747#issuecomment-1070329764 but should not block merging.

Upshot, merging now as "useful starting point". We can iterate.

If you find out that this runs on windows 10 with the GUI, let me know. That would trigger it replacing Cygwin IMO.

en/dev_setup/dev_env_windows_wsl.md
lyraik: Im currently using  a .sh script to start everything:
with this command i get the ip of the wsl instance: 
`ip=$(tail -1 /etc/resolv.conf | cut -d' ' -f2)`
and then I just reroute everything:
`./mavlink-routerd -e $ip:14550 0.0.0.0:14550 & `

afterwards you can just start the simulator normaly in wsl and mavlink gets routed to windows

hamishwillee: Thanks @lyraik - interesting idea, but requires install/setup of additional dependencies. @MaEtUgR is aware of this and will doubtless evaluate in the coming weeks.

FWIW I think any setup where we only need to install one version of QGC and then everything just works is "a good thing".

en/dev_setup/dev_env_windows_wsl.md
MaEtUgR: Perfect, thanks a lot. I just didn't see how to tell that in a clear way 😇

PR #1753, opened 2022-02-02
X500-v2 build fixups

hamishwillee:
This improves the x500 v2 frame docs slightly:
- Defines ARF
- Compressed and reduced images.

@farhangnaderi Doc is really good thanks. This was mostly about reducing page load time on slow links by reducing image size. It is also generally better to trim whitespace from images so that they display larger when the page is printed.


farhangnaderi:
Thanks @hamishwillee , I appreciate it.

PR #1761, opened 2022-02-04
Remove deprecated information on Iris lockstep

max11gen:
The Xacro file doesn't seem to exist anymore, neither does the `models/rotors_model`-path. If you want to turn off lockstep for Iris, you can do that in the SDF file as well. So this additional information is deprecated and confusing.


hamishwillee:
Thanks @max11gen - appreciate you providing the reasoning in the commit as well.

PR #1767, opened 2022-02-11
added FMU_CAP1 mapping to camera capture (PI0)

ashu41:
With reference to this discussion on slack: 
https://px4.slack.com/archives/C0V533X4N/p1644442549357989?thread_ts=1644228980.662649&cid=C0V533X4N


hamishwillee:
Thanks very much! I've subedited this slightly, mostly because I thought that it would help to note the port that the pin was on as well as the name on the diagram. This structure also makes it easy to add more notes as they come along.

As I commented in Slack, it would be even better to convert the pinouts to markdown, but not a priority.

PR #1769, opened 2022-02-12
Issue with :::caution :::

ivodre:
Somehow the :::caution ::: seem not to be rendered properly (also in other parts of the guide)


hamishwillee:
Thanks @ivodre 
That's because caution is not supported - the equivalent is "warning". Feel free to modify as shown here in any case you see it appears.

PR #1770, opened 2022-02-12
Update ros2_comm.md

ivodre:


en/ros/ros2_comm.md
```diff
@@ -50,7 +51,7 @@ In the near future the intention is to:
 
 :::warning
 You cannot use an agent generated as part of a "normal" PX4 build with ROS 2 (e.g. if the user uses `BUILD_MICRORTPS_AGENT=1 make px4_sitl_rtps`).
-While microRTPS client is the same, the IDL files used by ROS 2 are slightly different from the custom ones with generate in PX4.
+While microRTPS client is the same, the IDL files used by ROS 2 are slightly different from the [ROS-independent files generated in PX4 builds](../middleware/micrortps.md).
```
hamishwillee, line 54: @TSC21 I'm going to merge this, but can you please sanity check.

The text was confusing because it isn't clear what a "custom on with generate in PX4" might mean. I think it just means the IDL files generated by "raw" DDS are different from the IDL files generated for ROS2. 

The fix I used was to call these ROS-independent. 
**But should I call them instead** "normal DDS IDL files generated in ..." ??

hamishwillee (approved):

hamishwillee:
Thanks @ivodre - that is confusing. I have updated using terminology that is "more correct" but perhaps not optimal. I'll merge and perhaps update further if there are comments from @TSC21

PR #1779, opened 2022-02-16
pix4hawk5x: document use of ADS1115

bkueng:


en/flight_controller/pixhawk5x.md
```diff
@@ -172,6 +172,10 @@ Under these conditions the system will not draw any power (will not be operation
 1. **USB** input (operational range 4.1V to 5.7V, 0V to 6V undamaged)
 1. Servo input: VDD_SERVO pin of **FMU PWM OUT** and **I/O PWM OUT** (0V to 42V undamaged)
 
+:::note
+If you need analog battery voltage monitoring, you can connect an ADS1115 ADC via I2C, enable it via [ADC_ADS1115_EN](../advanced_config/parameter_reference.md#ADC_ADS1115_EN) and pull down the nPOWER_IN_A gpio input.
+:::
+
```
hamishwillee, line 178: Proposal, we mention I2C monitoring here too.
```suggestion
:::note
I2C battery monitoring is enabled by default.
If analog battery voltage monitoring is needed, you can connect an ADS1115 ADC via I2C, enable it via [ADC_ADS1115_EN](../advanced_config/parameter_reference.md#ADC_ADS1115_EN), and pull down the `nPOWER_IN_A` GPIO input.
:::

```
Questions:
-  "pull down the `nPOWER_IN_A` GPIO input" means attach that input to ground right?
- where is the nPOWER_IN_A input? I can't see it on the above diagrams.

bkueng: I would say `Digital I2C battery monitoring`

> "pull down the nPOWER_IN_A GPIO input" means attach that input to ground right?

Yes

> where is the nPOWER_IN_A input? I can't see it on the above diagrams.

Not sure for this board. @AntonioAguilera?

> Is there a good "canonical link for ADS1115? I couldn't see one, but if we can find a good description it might be good to link

It is common, but I don't know if there's a canonical one.

hamishwillee: @AntonioAguilera Can you tell me where the `nPOWER_IN_A` input is on the Holybro Pixhawk 5X?

@bkueng If Antonio can't answer, is there someone else we can ask? Hint is not much use without this information.

AntonioAguilera:
Hi guys... sorry for the delay.

image

in this case we designed a carrier board to feed voltage to the 5x, so as we don't have the power feed by the power connector.

According to https://github.com/pixhawk/Pixhawk-Standards/blob/master/DS-010%20Pixhawk%20Autopilot%20Bus%20Standard.pdf, the X1 pinout shows 3 enables to tell 5x where the power is coming from, in our case we pulled down nPOWER_IN_A to tell the FC that was coming from Battery 1.

image

in this case you can see the power selection on the original carrier board, this inputs on the FC are pulled down to led 5X know where the power is coming from.

What i assume is that if you provided power over a different source without any of those enable signals properly pulled to GND then you don't see the power as a valid source, that was our case.

AntonioAguilera:
Also this is the same if you are not using not using an external ADC, that is our case now. 
You can use the 3.3V internal ADC on the 5X as voltage monitor.

image

hamishwillee:
Thanks @AntonioAguilera . @bkueng Can you translate that for the hardware-ignorant. As an end user, I'd just want to know where I plug in my converter - port and pin.

bkueng:
@hamishwillee this is an internal pin (to the carrier) that's not exposed. Apparently on Skynode this works w/o having to pull down the signal. We'll need some software changes to ignore it for that setup to cleanly support it. Until then it can still be used like that for integrators.

hamishwillee:
@bkueng Thanks. So my understanding is that this just doesn't work for Holybro Pixhawk 5X - right? 

> Digital I2C battery monitoring is enabled by default.
> Analog battery monitoring via an ADC is not supported on this board.

bkueng:
Yes, assuming you use the full board as it is.

en/flight_controller/pixhawk5x.md
hamishwillee:
```suggestion
**Voltage monitoring**

Digital I2C battery monitoring is enabled by default (see [Quickstart > Power](../assembly/quick_start_pixhawk5x.md#power)).

:::note
Analog battery monitoring via an ADC is not supported on this particular board, but may be supported in variations of this flight controller with a different baseboard.
:::

```

hamishwillee:
Thanks @bkueng , in https://github.com/PX4/PX4-user_guide/pull/1779/files#r823246274 I went with

> **Voltage monitoring**
> 
> Digital I2C battery monitoring is enabled by default (see [Quickstart > Power](../assembly/quick_start_pixhawk5x.md#power)).
> :::note
> Analog battery monitoring via an ADC is not supported on this particular board, but may be supported in variations of this flight controller with a different baseboard.
> :::

PR #1783, opened 2022-02-17
Precland: Update to use new parameter to enable irlock

hamishwillee:
A parameter was added to enable IRLOCK to be started on boot in https://github.com/PX4/PX4-Autopilot/pull/19175/
This change adds the parameter to the configuration section.

I also noted that the modules needed for IR lock ARE included on most boards by default (now), so the docs are now incorrect. This updates them to say that the modules are present by default in most cases, and then shows how you set them if they are not present.


dagar:
Thanks @hamishwillee

PR #1785, opened 2022-02-17
Update tunes to match current status

hamishwillee:
tunes were updated in https://github.com/PX4/PX4-Autopilot/commit/08bf71b73d8eabd9cffca6dd24583fa9d586fff1

Some were renumbered, but the actual name and tune remained the same. EKF and Baro tunes had previously not been used, and were deleted.

New tune for power off was added. Placeholder here, but we need new tune.

### DIFF
--- a/en/getting_started/tunes.md
+++ b/en/getting_started/tunes.md
@@ -6,7 +6,9 @@ The set of standard sounds are listed below.
 
 :::note
 **Developers:** Tunes are defined in [/lib/tunes/tune_definition.desc](https://github.com/PX4/PX4-Autopilot/blob/master/src/lib/tunes/tune_definition.desc) and can be tested using the [tune-control](../modules/modules_system.md#tune-control) module.
+You can search for tune use using the string `TUNE_ID_name`(e.g. `TUNE_ID_PARACHUTE_RELEASE)
 :::
+
 
 ## Boot/Startup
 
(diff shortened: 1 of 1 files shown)


en/getting_started/tunes.md
```diff
@@ -118,6 +132,7 @@ Your browser does not support the audio element.
   <source src="../../assets/tunes/4_notify_neutral_tone.mp3" type="audio/mpeg">
 Your browser does not support the audio element.
 </audio>
+<!-- 4, NOTIFY_NEUTRAL -->
```
hamishwillee, line 135: FYI, these added to make tracking change easier next time.

en/getting_started/tunes.md
```diff
@@ -226,8 +229,11 @@ Your browser does not support the audio element.
   <source src="../../assets/tunes/15_home_set_tune.mp3" type="audio/mpeg">
 Your browser does not support the audio element.
 </audio>
+<!-- 13, HOME_SET (previously id 15) -->
 
 - Home position initialised (first time only).
 
+#### Power Off Tune
```
hamishwillee, line 236: @julianoes Note, missing tune I "think" you added. I can do this if you can't, but not until next week.

julianoes: It currently only applies to Mantis.

hamishwillee: Sure, but one day it will apply to more. Added in #1791

PR #1809, opened 2022-03-10
PREFLIGHT FAIL: EKF HIGH IMU ACCEL BIAS - better docs

hamishwillee:
This updates [Preflight Sensor/Estimator Checks](https://docs.px4.io/master/en/flying/pre_flight_checks.html#preflight-fail-ekf-high-imu-accel-bias) with better information about the `PREFLIGHT FAIL: EKF HIGH IMU ACCEL BIAS` error, as provided by @bresch 

A few questions inline.


en/flying/pre_flight_checks.md
```diff
@@ -40,10 +41,39 @@ The following errors (with associated checks and parameters) are reported by the
...
+This bias may change when the sensor is turned on (“turn-on bias”) and over time due to noise and temperature differences (“in-run bias”).
+The number should generally be very small (near zero), indicating that measurements from different sources all agree on the acceleration.
+
+The warning indicates that the bias is higher than some arbitrary threshold.
+It is most likely a sign that accelerometer or thermal calibration are required:
+
+- If you _sometimes_ get the warning: [re-calibrate the accelerometer](../config/accelerometer.md).
+- If you get _regularly_ get the warning: Perform a [thermal calibration](../advanced_config/sensor_thermal_calibration.md).
+- If you still get the warning after thermal calibration (or you can't perform thermal calibration): 
+  - Verify that the issues does not come from the sensor, or autopilot hardware
```
hamishwillee, line 58: @bresch How might you verify this? If we give an instruction, we should provide hints or link to next steps.

en/flying/pre_flight_checks.md
```diff
@@ -40,10 +41,39 @@ The following errors (with associated checks and parameters) are reported by the
...
+However it may be required if there are problems with the sensor that cannot be fixed by other means (i.e you can tune the EKF for better performance, but there is no way you can calibrate the accelerometer "better").
+
+:::warning
+Tuning these parameters is a last resort.
+It should only be attempted if you have data showing it will improve the performance of the estimator.
+:::
+
+Parameter | Description
+--- | ---
+<a id="EKF2_ABL_LIM"></a>[EKF2_ABL_LIM](../advanced_config/parameter_reference.md#EKF2_ABL_LIM) | The maximum value that the EKF is allowed to learn. The autopilot will report a “high accel bias” if the estimated bias exceeds 75% of this parameter during a preflight check. The current value of 0.4m/s2 is already quite high and increasing it would make the autopilot less likely to detect an issue.
```
hamishwillee, line 71: What is meant by the term "learn" and "learning parameters" here? I think this might be saying that EKF does the estimation based on multiple sources and will attempt to compensate for a certain amount of poor general calibration or  changes due to the turn-on and run-in biases. So when the term "learn" is used here, I think you mean "amount of variation before it decides enough is enough and falls over in some way?

If so, what happens then?

bresch: It's simpler than that, "learn" is just a synonym of "estimate" here. "The EKF estimates the accelerometer bias" == "The EKF learns the accelerometer bias", nothing more.

en/flying/pre_flight_checks.md
bresch: The "simpler" check would be to try with another autopilot for example. Otherwise do some bench test logging runs and compare the output of all the accelerometers. (To log all the accelerometers, enable "sensor comparison" in [SDLOG_PROFILE](https://docs.px4.io/master/en/advanced_config/parameter_reference.html#SDLOG_PROFILE))

en/flying/pre_flight_checks.md
hamishwillee: @bresch Thanks! Does that mean that the maximum acceleration that the estimator can detect is 0.4m/s^2? Seems low - won't it detect 10m/s^2 if the engines stop?
Why is there a limit on the maximum it can estimate?

Otherwise I think it makes sense - there is a maximum acceleration that the estimator can report. If the difference between your sensor-measured and estimated acceleration is a significant amount of this then probably the sensor is uncalibrated.

Anyway, assuming this makes sense, so just change one work as below then?:

```suggestion
<a id="EKF2_ABL_LIM"></a>[EKF2_ABL_LIM](../advanced_config/parameter_reference.md#EKF2_ABL_LIM) | The maximum value that the EKF is allowed to estimate. The autopilot will report a “high accel bias” if the estimated bias exceeds 75% of this parameter during a preflight check. The current value of 0.4m/s2 is already quite high and increasing it would make the autopilot less likely to detect an issue.
```

hamishwillee:
@bresch Thanks very much. I have decided to merge this as "much better than what was here before".

I would appreciate it if you could have a look at my question in  https://github.com/PX4/PX4-user_guide/pull/1809/files#r827641331 - probably a stupid one, but just want to make sure I'm not way off track on how this works.

bresch:
> I would appreciate it if you could have a look at my question in https://github.com/PX4/PX4-user_guide/pull/1809/files#r827641331

Which question? Le link just redirects me to the changes of this PR...

en/flying/pre_flight_checks.md
bresch: Sorry, I wasn't clear: it's the maximum value of acceleration bias, not just acceleration.

Loosely speaking, the accelerometer bias estimation works as follows:
1. get the accelerometer data (called "specific force", i.e: """acceleration but including gravity""")
2. rotate it into the NED (North-East-Down) frame
3. remove gravity vector from the specific force, giving the acceleration
4. integrate the acceleration to get the velocity vector (still NED)
5. Compare it with a velocity aiding measurement (e.g.: GPS velocity): from the difference between the velocity measurement and the velocity from integrated acceleration, you can "estimate" the bias in the acceleration data
6. rotate it back to the body frame to obtain the accelerometer bias

hamishwillee: Great! Makes sense. Still more questions which I sent you offline.

PR #1821, opened 2022-03-22
Update Kconfig doc to fix build error and clarify Kconfig usage

junwoo091400:
I was adding a [custom module](https://nxp.gitbook.io/hovergames/developerguide/px4-tutorial-example-code/hg-px4-example-lab-5) and noticed that PX4 now uses Kconfig architecture for module configuration.

However, `autoreconf` command was missing from the doc. Without this, user can not build the menuconfig successfully. Original issue is referenced here : https://github.com/jerryscript-project/jerryscript/pull/2225

Also, while manually editing the `default.px4` board file, I noticed that when I mistype the module's `menuconfig` property (a.k.a, it's name), it doesn't show any error messages & just doesn't include it for the build. This can happen to any user who isn't so aware of Kconfig architecture. I added an helper description for that.

### DIFF
--- a/en/hardware/porting_guide_config.md
+++ b/en/hardware/porting_guide_config.md
@@ -31,7 +31,8 @@ endif #DRIVERS_UAVCAN_V1
 ```
 
 :::note
-The `menuconfig` property in Kconfig file is the name used for configuration. If you edit the `*.px4board` configuration file manually, make sure you didn't misspell the module name! Otherwise, it won't include the module for the build.
+The `menuconfig` property in Kconfig file is the name used for configuration.
+If you edit the `*.px4board` configuration file manually, make sure you don't misspell a module name (or it won't be included in the build).
 :::
 
 ## PX4 Kconfig Label Inheritance
--- a/en/hardware/porting_guide_config.md
+++ b/en/hardware/porting_guide_config.md
@@ -31,7 +31,6 @@ endif #DRIVERS_UAVCAN_V1
 ```
 
 :::note
-The `menuconfig` property in Kconfig file is the name used for configuration.
 Builds will silently ignore any missing or miss-spelled modules in the  `*.px4board` configuration file.
 :::
 
--- a/en/hardware/porting_guide_config.md
+++ b/en/hardware/porting_guide_config.md
@@ -32,7 +32,7 @@ endif #DRIVERS_UAVCAN_V1
 
 :::note
 The `menuconfig` property in Kconfig file is the name used for configuration.
-If you edit the `*.px4board` configuration file manually, make sure you don't misspell a module name (or it won't be included in the build).
+Builds will silently ignore any missing or miss-spelled modules in the  `*.px4board` configuration file.
 :::
 
 ## PX4 Kconfig Label Inheritance
--- a/en/hardware/porting_guide_config.md
+++ b/en/hardware/porting_guide_config.md
@@ -30,6 +30,10 @@ if DRIVERS_UAVCAN_V1
 endif #DRIVERS_UAVCAN_V1
 ```
 
+:::note
+The `menuconfig` property in Kconfig file is the name used for configuration. If you edit the `*.px4board` configuration file manually, make sure you didn't misspell the module name! Otherwise, it won't include the module for the build.
+:::
+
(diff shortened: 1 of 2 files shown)


en/hardware/porting_guide_config.md
```diff
@@ -30,6 +30,11 @@ if DRIVERS_UAVCAN_V1
 endif #DRIVERS_UAVCAN_V1
 ```
 
+:::note
+The `menuconfig` property in Kconfig file is the name used for configuration.
+If you edit the `*.px4board` configuration file manually, make sure you don't misspell a module name (or it won't be included in the build).
```
hamishwillee, line 35: This seems a bit redundant - you'd be expected to spell names correctly.

en/hardware/porting_guide_config.md
```diff
@@ -30,6 +30,11 @@ if DRIVERS_UAVCAN_V1
 endif #DRIVERS_UAVCAN_V1
 ```
 
+:::note
+The `menuconfig` property in Kconfig file is the name used for configuration.
```
hamishwillee, line 34: This is a bit confusing. Perhaps we should move this line as a note next to the above comment "This configuration is handled through *Kconfig*, ...." ?

hamishwillee:
@PetervdPerk-NXP Would appreciate your review, in particular of the autoreconf change.

en/hardware/porting_guide_config.md
hamishwillee: Ah, I see your point. Perhaps better as:
```suggestion
Builds will silently ignore any missing or miss-spelled modules in the  `*.px4board` configuration file.
```

en/hardware/porting_guide_config.md
hamishwillee, line 34: This just isn't right here
```suggestion
```

hamishwillee:
@PetervdPerk-NXP I think you're saying that the  `autoreconf --force --install` isn't needed right? 

The other comment is OK I think - though I have reworded it simply to say that missing modules are silently ignored.

en/hardware/porting_guide_config.md
```diff
@@ -30,6 +30,10 @@ if DRIVERS_UAVCAN_V1
 endif #DRIVERS_UAVCAN_V1
 ```
 
+:::note
+Builds will silently ignore any missing or miss-spelled modules in the  `*.px4board` configuration file.
```
junwoo091400, line 34: @hamishwillee Shouldn't we add the *if edited manually part here? As it wouldn't happening using kconfig gui utilities?

I spent my fair share of time just troubleshooting this because of this, since I didn't expect the menuconfig name to be the one included in the .px4board file.

(I thought CONFIG_*) ones were parsed automatically to access the modules in certain directories, and didn't know I had to manually edit the Kconfig to match the directory naming convention exactly 🙊

hamishwillee: The note highlights the result of both missing and miss-spelled modules.
So if this is not sufficient then I think that we'd be better off with a statement like "Don't edit the files manually".

en/hardware/porting_guide_nuttx.md
```diff
@@ -33,6 +33,7 @@ Run the following commands from any directory:
 ```sh
 git clone https://bitbucket.org/nuttx/tools.git
 cd tools/kconfig-frontends
+autoreconf --force --install
```
hamishwillee, line 36: Reverting
```suggestion
```

hamishwillee:
@PetervdPerk-NXP It wasn't clear to me whether you were saying `autoreconf --force --install` is not needed ... or it is. I have therefore reverted that. If it is needed, please comment here/chase up with @junwoo091400 

Junwoo, I've kept just the note that silently editing the file can cause problems. You can add back a separate PR on `autoreconf --force --install` if you can confirm it is needed with Peter.

PetervdPerk-NXP:
See the official NuttX documentation.
https://nuttx.apache.org/docs/latest/quickstart/install.html#kconfig-frontend
There are 2 options "Ubuntu 20.04 LTS or Later" or "MacOS, Ubuntu 18.04 LTS and earlier"
I think we just can that description.

PR #1813, opened 2022-03-12
Adding Documentatoin for PX4 Ethernet setup

farhangnaderi:
I need reviews if possible. 

Thanks to @mrpollo and [this link](https://github.com/PX4/PX4-Autopilot/pull/16330) for making it happen. 

Hopefully could help [553](https://github.com/PX4/PX4-user_guide/issues/553 ) as well.

PR #1823, opened 2022-03-23
Ethernet setup - add to sidebar

hamishwillee:
This adds the ethernet setup doc from #1813 into the sidebar. It also updates the MAVLink doc to add default port info.

I'm merging. There may be more to say, but this is good enough for now.


farhangnaderi:
Oh, I forgot this. 🤦🏻‍♂️Thanks Hamish.

PR #1826, opened 2022-03-25
Adding a page about I2C from the user's point of view and solving common i2c problems. 

roman-dvorak:
Hi, we have written a page that provides basic information about using the I2C bus. It also contains information about I2C address translators and bus boosters. 

At the same time, it tries to unify the structure with the pages about CAN, where one page is for FW developers and the second is for PX4 users.


hamishwillee:
Thanks - I'll check this out next week when I'm back on this. Can you please crop the whitespace around the images. We try reduce the size of images to reduce cost of download for readers.

en/sensor_bus/i2c_general.md
```diff
@@ -0,0 +1,153 @@
...
+    ADT -- > RPM5(Out2: TFRPM01C 0x50 - 0x2f)
+    ADT -- > RPM6(Out2: TFRPM01C 0x52 - 0x2e)
+-->
+
+
+:::note
+TFI2CADT01 does not contain any I2C buffer or accelerator.
+It itself creates an additional capacity on the bus.
+Therefore it is advisable to combine TFI2CADT01 with some bus booster, e.g. TFI2CEXT01.
+:::
```
hamishwillee, line 130: I was thinking all of this should go in a sub page

hamishwillee:
I have given this a review and a restructure. Firstly I think it is a good idea to have the "General" page - in particular as a place to talk about the wiring, address clashes and reduced capacity.

My only issue with this is that PX4 does not promote single solutions - and this looks like advertising for your products. To be clear, love the products, we have a place for them. But not in a structure which makes them look like the only option.

So what I have done is made this generic, where your products look like options. However others can create their own products and link them if they like. 

What I would like you to do is read this and check it is OK. Perhaps split out the TFI2CADT01 stuff into its own document as a sub page?

en/sensor_bus/i2c_general.md
```diff
@@ -0,0 +1,153 @@
...
+TFI2CADT01 performs an XOR operation on the called address.
+Therefore, a new device address can be found by taking the original address and applying an XOR operation with the value specified on the module.
+By default, the output 1 performs XOR with 0x08 value and the second port with 0x78.
+By short-circuiting the solder jumper you can change the XOR value to 0x0f for the first and 0x7f for the second port.
+
+If you need your own value for address translation, changing the configuration resistors makes it possible to set any XOR value.
+
+#### Example of use
+
+The tachometer sensor [TFRPM01]() can be set to 2 different addresses using a solder jumper.
```
hamishwillee, line 97: broken link

roman-dvorak:
Thanks @hamishwillee for edits. 

I will separate the address translator information into a separate document. And I'll go through the whole text again

> My only issue with this is that PX4 does not promote single solutions - and this looks like advertising for your products. To be clear, love the products, we have a place for them. But not in a structure which makes them look like the only option.

All the mentioned modules that I present there are open-source design. Source data (kicad) and production data (gerber) can be found on github. It also contains information about components, etc. So anyone can download and produce them.

And the second thing is that I don't know of any alternative that would be ready to connect to dronecode devices (autopilot).

But I agree that it should be like options.

hamishwillee:
Thanks! Not an issue, I just wanted to be clear why I modified this - not "just" because I'm annoying :-)

en/sensor_bus/i2c_general.md
roman-dvorak: Moved to custom page. Added information about open-source license, added link to datasheet of translator IC.

kaklik:
I made a few additional improvements to the documentation pages. I think that is ready to be merged.

hamishwillee:
Thanks very much!!! Looks good.

PR #1829, opened 2022-03-31
Fix SDLOG_PROFILE in Mag Comp

AlexKlimaj:
This has always been wrong. High Rate does not log the mag at a high rate, but Sensor Comparison does.

### DIFF
--- a/en/advanced_config/compass_power_compensation.md
+++ b/en/advanced_config/compass_power_compensation.md
@@ -28,7 +28,7 @@ Performing this power compensation is advisable only if all the following statem
 1. Make sure your drone runs a Firmware version supporting power compensation (current master, or releases from v.1.11.0).
 1. Perform the [standard compass calibration](../config/compass.md#compass-calibration).
 1. Set the parameter [SDLOG_MODE](../advanced_config/parameter_reference.md#SDLOG_MODE) to 2 to enable logging of data from boot. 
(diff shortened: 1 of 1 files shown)


en/advanced_config/compass_power_compensation.md
```diff
@@ -28,7 +28,7 @@ Performing this power compensation is advisable only if all the following statem
 1. Make sure your drone runs a Firmware version supporting power compensation (current master, or releases from v.1.11.0).
 1. Perform the [standard compass calibration](../config/compass.md#compass-calibration).
 1. Set the parameter [SDLOG_MODE](../advanced_config/parameter_reference.md#SDLOG_MODE) to 2 to enable logging of data from boot. 
-1. Set the parameter [SDLOG_PROFILE](../advanced_config/parameter_reference.md#SDLOG_PROFILE) checkbox for *high rate* (bit 2) to get more data points.
+1. Set the parameter [SDLOG_PROFILE](../advanced_config/parameter_reference.md#SDLOG_PROFILE) checkbox for *Sensor comparison* (bit 6) to get more data points.
```
hamishwillee, line 31: I'm happy to change bit 2, since that is marked as being for logging "Thermal Calibration", but bit 4 is marked as "High Rate" - why "6" and not that?
https://docs.px4.io/master/en/advanced_config/parameter_reference.html#SDLOG_PROFILE

AlexKlimaj: Sensor comparison is the only mode that will log the mag at a high rate for the mag compensation routine.

hamishwillee: Thanks for confirming.

PR #1831, opened 2022-04-02
Added some Chinese translations

wrk-mk:


hamishwillee:
Thank you, but this is not how we add translations in PX4. Essentially we use a service called crowdin that helps us keep the English and translated version in sync. 

Everything you need to know about this is here: http://docs.px4.io/master/en/contribute/translation.html#translation

But if you have more questions, feel free to ask on Slack.

PR #1836, opened 2022-04-06
Update holybro dual f9p heading info

vincentpoont2:
Update holybro dual f9p heading info. Holybro H-RTK F9P Helical & Base version can do dual f9p heading.


hamishwillee:
Thanks. Smart idea to ship a smaller GPS for the dual GPS heading setup!

PR #1837, opened 2022-04-06
Update parameter_reference.md

CamronMyers:
When Ainstein released the latest version of our standard altimeter we renamed it the US-D1 Radar Altimeter as https://docs.px4.io/master/en/sensor/ulanding_radar.html shows. I wanted to rename the uLanding radar sections to US-D1.
We can keep the ULAND_CFG parameter the same for the sake of simplicity.
Thanks!


hamishwillee:
Sorry, you can't do it this way as the parameter reference is auto-generated. You will need to change things in the source code.

PR #1848, opened 2022-04-15
Update korean modules_command 

sju0924:
1. Add actuator_test paragraph
2. Fixed the source link which had been pushed back
3. Add description subhead for paragraphs that already have been translated
4. Add description detail for dumpfile
5. Update translation for failure paragraph


hamishwillee:
Thank you, we don't translate directly in sources, but via the crowdin tool: http://docs.px4.io/master/en/contribute/translation.html#translation

Would very much welcome your contribution there!

PR #1855, opened 2022-04-26
add kakute h7.md

vincentpoont2:
Add holybro kakute h7


hamishwillee (approved):
Thanks @vincentpoont2 

I did a minor subedit and shrunk the images. The main change was to remove the comment about having a barometer and hence being ready for autonomous flight. On PX4 you need position lock for autonomous flight: having altitude is not enough.

PR #1858, opened 2022-04-27
Actuators Document Improvement

junwoo091400:
## Explanation
As part of [Control Allocation User experience Improvement](https://github.com/PX4/PX4-Autopilot/issues/19444) that I am working on, I am submitting this PR draft to get Document Improvement going.

## Main Changes Proposed
1. Create Actuators setup document for each vehicle type (airframe), to give more specific instructions
2. Add diagrams to explain important concepts (assumptions) in Control Allocation that can be missed by the user
3. Add explanations along with the diagrams

## TODOs
* Updated in the [main Control Allocation Improvement Issue](https://github.com/PX4/PX4-Autopilot/issues/19444)


hamishwillee:
@junwoo091400 As discussed, the existing document is generic because it was created when the only thing that "worked" was the iris frame. The way it would work in practice for specific cases was unknown.

If it has now evolved such that it is clear how this will be used on all frames then yes, this is very timely and needs to be documented. I can't comment _how_ without knowing how it works myself. 

You actually need to also much earlier than these docs with the fundamentals of how the integration of the mixer/geometry works. I know that the actuator screens are generated from actuator definitions with this schema: https://github.com/mavlink/mavlink/blob/master/component_metadata/actuators.schema.json

But the rest isn't clear, and should be:

1.  How do I [add a new airframe](http://docs.px4.io/master/en/dev_airframes/adding_a_new_frame.html)?
      - Previously I would have a geometry file and a mixer file. The mixer would be referenced from the airframe:
        ```
        # Set mixer
        set MIXER wingwing
        # Configure PWM outputs (specify the outputs to drive/activate, and the levels).
        set PWM_OUT 4
        ```
        So the question now is HOW does an airframe that you load know that the geometry it is defining is a vtol etc?
2. How does this get generated out into an acutators metadata file
3. What does our current metadata file look like?

That will then impact the behaviour in the front end, for which I have these questions:

1. There is a frame for iris testing that you can simulate: `make px4_sitl gazebo_iris_ctrlalloc` - I also see there is now `gazebo_typhoon_h480_ctrlalloc` but nothing to demo VTOL or FW right?
    - Previously you couldn't modify the [SYS_CTRL_ALLOC](http://docs.px4.io/master/en/advanced_config/parameter_reference.html#SYS_CTRL_ALLOC) on simulator, so you couldn't try this for any other frame. 
    - Can I try set SYS_CTRL_ALLOC on an arbitrary frame now to configure it?
    - Upshot, what do the screens look like for VTOL and FW - how is setup different.
2. Existing airframes:
   - Will the same frame set be available for selection?
   - Do these arrive all set up with same default mapping as currently?
      - If so, good, because the airframe reference then is accurate, and just needs to be updated to indicate that the output > motor mappings can all be changed if needed.
      - If not, then how will the mapping work?
3. With the iris and the typhoon gazebo _ctrlalloc sims they arrive with 4 and 6 motors configured respectively, which makes sense. When would you even select a different number of motors in the geometry?
   - What I'm really asking is how do I set up a new airframe and what doe airframe mean now days - if the number of motors can be selected that implies that the configuration is variable.
   
   
My guess is that we define a generic geometry for a copter, which defines a bunch of parameters. When we define an airframe we associate the airframe to that "broad" geometry definition. We couldthen set specific parameters for the airframe to set default values for the parameters that specifically apply to the selected geometry.

So it is possible that we might supply just one generic airframe for all copters, and let people assign their own values for everything. Or we could supply a generic frame for each main type of copter with some basic settings. Or we could supply settings exactly as we do now so that we mimic the existing airframes. 
These are choices - and what I don't know is how they will be made.

Anyway, next step is for you to help me understand some of the above. We can always ask @bkueng . He knows everything :-)

hamishwillee:
@junwoo091400 As discussed, the existing document is generic because it was created when the only thing that "worked" was the iris frame. The way it would work in practice for specific cases was unknown.

If it has now evolved such that it is clear how this will be used on all frames then yes, this is very timely and needs to be documented. I can't comment _how_ without knowing how it works myself. 

You actually need to also much earlier than these docs with the fundamentals of how the integration of the mixer/geometry works. I know that the actuator screens are generated from actuator definitions with this schema: https://github.com/mavlink/mavlink/blob/master/component_metadata/actuators.schema.json

But the rest isn't clear, and should be:

1.  How do I [add a new airframe](http://docs.px4.io/master/en/dev_airframes/adding_a_new_frame.html)?
      - Previously I would have a geometry file and a mixer file. The mixer would be referenced from the airframe:
        ```
        # Set mixer
        set MIXER wingwing
        # Configure PWM outputs (specify the outputs to drive/activate, and the levels).
        set PWM_OUT 4
        ```
        So the question now is HOW does an airframe that you load know that the geometry it is defining is a vtol etc?
2. How does this get generated out into an acutators metadata file
3. What does our current metadata file look like?

That will then impact the behaviour in the front end, for which I have these questions:

1. There is a frame for iris testing that you can simulate: `make px4_sitl gazebo_iris_ctrlalloc` - I also see there is now `gazebo_typhoon_h480_ctrlalloc` but nothing to demo VTOL or FW right?
    - Previously you couldn't modify the [SYS_CTRL_ALLOC](http://docs.px4.io/master/en/advanced_config/parameter_reference.html#SYS_CTRL_ALLOC) on simulator, so you couldn't try this for any other frame. 
    - Can I try set SYS_CTRL_ALLOC on an arbitrary frame now to configure it?
    - Upshot, what do the screens look like for VTOL and FW - how is setup different.
2. Existing airframes:
   - Will the same frame set be available for selection?
   - Do these arrive all set up with same default mapping as currently?
      - If so, good, because the airframe reference then is accurate, and just needs to be updated to indicate that the output > motor mappings can all be changed if needed.
      - If not, then how will the mapping work?
3. With the iris and the typhoon gazebo _ctrlalloc sims they arrive with 4 and 6 motors configured respectively, which makes sense. When would you even select a different number of motors in the geometry?
   - What I'm really asking is how do I set up a new airframe and what doe airframe mean now days - if the number of motors can be selected that implies that the configuration is variable.
   
   
My guess is that we define a generic geometry for a copter, which defines a bunch of parameters. When we define an airframe we associate the airframe to that "broad" geometry definition. We couldthen set specific parameters for the airframe to set default values for the parameters that specifically apply to the selected geometry.

So it is possible that we might supply just one generic airframe for all copters, and let people assign their own values for everything. Or we could supply a generic frame for each main type of copter with some basic settings. Or we could supply settings exactly as we do now so that we mimic the existing airframes. 
These are choices - and what I don't know is how they will be made.

Anyway, next step is for you to help me understand some of the above. We can always ask @bkueng . He knows everything :-)

hamishwillee:
Or to summarise, what's the story for:
- someone creating a new airframe
  - from scratch - say a chinook. New geometry new flight behaviour
  - based on an existing kind of geometry
  - other cases?  
- a new airframe from an established geometry - a quad + or a tailsitter quad vtol.
-a prepurchased frame of some kind and wants to set it up with PX4.

hamishwillee:
Or to summarise, what's the story for:
- someone creating a new airframe
  - from scratch - say a chinook. New geometry new flight behaviour
  - based on an existing kind of geometry
  - other cases?  
- a new airframe from an established geometry - a quad + or a tailsitter quad vtol.
-a prepurchased frame of some kind and wants to set it up with PX4.

bkueng:
> Can I try set SYS_CTRL_ALLOC on an arbitrary frame now to configure it?

Yes you can set the param on most existing sim targets to test them, they will have the geometry preconfigured (including vtols and plane).

> Or we could supply a generic frame for each main type of copter with some basic settings.

My intention is to go with this, so that people don't have to change motor numbers & positions.

> He knows everything :-)

haha, most certainly not

hamishwillee:
Thanks @bkueng - that means we can pretty much address this UI for an end user. 

This tells me that:
- if there is a matching airframe you just assign it.
- if there is a matching geometry you'd take the closest "generic airframe" and modify that geometry using the UI - for example if you had a big Quad X with slightly offset props you could start with generic quad X  (or +) and make the appropriate changes.

Note sure how you create a new geometry definition and tie it into PX4 and assign it to each airframe (along with associated updates to get it into the actuator.json). Are you planning on doing those docs?

hamishwillee:
> Yes you can set the param on most existing sim targets to test them, they will have the geometry preconfigured (including vtols and plane).

@bkueng FYI There is still a bug here. You have to "manually" reboot the sim, which I expected. What I did not expect is that you have to turn QGC on and off to detect the config change and see the Actuators setup section.

bkueng:
>  @bkueng FYI There is still a bug here. You have to "manually" reboot the sim, which I expected. What I did not expect is that you have to turn QGC on and off to detect the config change and see the Actuators setup section.

I'm aware, the problem is more generic: qgc needs to detect a vehicle reboot and then reconnect.

> I assume it looks correct?

Yes

junwoo091400:
I have made the following diagram to show an example of how the quadcopter's motor geometry coordinates are calculated. Do you have any comments on this? @bkueng @hamishwillee 

Quadcopter_MotorPositions_v4_CoordinateColors

hamishwillee:
@junwoo091400 I could be wrong, but I **think** the values are unitless - ie it is not 23cm but 0.23 relative/normalized to the other values. 

IF so, this is definitely something to explain. It might be better actually to have a diagram that is "wrong" to explain this - i.e. that has a value of 2 in there somewhere. 

@bkueng Is that right? That implies that it is only geometry that matters - a big drone with the same geometry has the same values as a small drone.

hamishwillee:
>> I assume it looks correct?

> Yes

Thanks. Why no motor mapping image? 

Which ones have an image, and which do not?

bkueng:
The values are in meters. For now the scaling does not matter, but it might eventually. @junwoo091400 can you move the CG a bit more off, so it's more obvious?

> Why no motor mapping image?

Because there's no servo info, so it would be incomplete. This is probably the 3rd time you're asking that, can we settle that? If you want we can show the same image (MC part), but I think it would be weird showing a wrong geometry.
So only MC shows it.

hamishwillee:
> Because there's no servo info, so it would be incomplete. This is probably the 3rd time you're asking that, can we settle that? 

Sorry to irritate you. I'm looking at your response "there no servo info" and thinking WTF? We know that a vtol has certain controls, why can't we show them in the image?

In any case, 
- geometries are more intuitive with an image. 
- if it can't be done, then OK
- I will try remember this in future. 

> If you want we can show the same image (MC part), but I think it would be weird showing a wrong geometry.
> So only MC shows it.

Fair enough. If we can't show something accurate we shouldn't show anything.

hamishwillee:
>> We know that a vtol has certain controls, why can't we show them in the image?

> Because they are configurable, but we don't know the positions, tilt vs control surface, and shape of the vehicle. With assumptions and for the simpler cases we could still render something. Not sure how well this would work.
> I would certainly also prefer to display the geometry. I think at some point we will be able to.

Cool. Though "I think at some point we will be able to." tells me that you have a vision of how this might be possible :-)

I was hoping we might display image only for known/canned airframe and "grey out" anything that can no longer be considered reliable if a parameter is changed. That way you could indicate motor geometry while making it clear that changes to an actuator are no longer reflected in the image. The reason you'd bother is that it makes sense to display the motor order diagram if there is any way to do so.

hamishwillee:
As an aside, do we have a way to indicate a copter motor/props that are aligned at an angle rather than parallel to the ground?

bkueng:
Well, 'vision' is a bit much :)

> I was hoping we might display image only for known/canned airframe and "grey out" anything that can no longer be considered reliable if a parameter is changed. That way you could indicate motor geometry while making it clear that changes to an actuator are no longer reflected in the image. The reason you'd bother is that it makes sense to display the motor order diagram if there is any way to do so.

That was my idea at first (and it's still possible), but then we went for the fully generic versions. Which makes more sense as there are just too many different ones, even FW for example. An in-between solution as you suggest is possible, I think best in the form of:
- extend metadata so that the GCS knows more (e.g. control surface type)
- let the GCS select the image by checking the geometry in the form of 'Nr motors=1, Nr servos=3, servo types = (left aileron, right aileron, rudder)' -> select image xy.

This still leaves the additional complexity for configurable motor positions on vtol. It can be handled, just a question of how much effort you want to put in.

> As an aside, do we have a way to indicate a copter motor/props that are aligned at an angle rather than parallel to the ground?

Yes the axis can be configured, see https://github.com/PX4/PX4-Autopilot/pull/19557 for an extreme example. I still want to expose that on the UI (via advanced).

hamishwillee:
Your approach looks pretty flexible and IMO is worth doing. But I don't pay you :-)

Given its hard work, I "wonder" if an even more hacky approach is worth doing in the interrim. A "clearly placeholder frame" that just shows the motors, if present. Again the idea is to give the motor mapping and make it clear that the rest of the geometry does not work. 

> I still want to expose that on the UI (via advanced).

OK. @junwoo091400 This would be a case you would add a note in the docs that the geometry can't yet be supported. Ideally with a PR or issue link for users to track.

junwoo091400:
Quadcopter_MotorPositions_v6_CG_more_Dramatic
The CG Diagram has been updated. Any comments are welcome :hand:

junwoo091400:
image

Actually this edit-disabled screen looks pretty nice in terms of preventing users from editing the motor geometry. Could use this!
This is currently what happens when we click "identify & assign motors".

junwoo091400:
More details and diagrams has been added that will be helpful for users in setting up Actuators. Will be a nice improvement, and after a review, it would be nice to get it merged first. So that existing users can benefit from this change early on.

en/config/actuators.md
```diff
@@ -33,50 +33,87 @@ This particular frame also includes an image showing the motor positions, which
...
+
+* Example : Having a V or A shaped tail in the back, positive deflection in Rightwards direction will Yaw the vehicle to the right.
+
+
+### Motor Tilt Servos (Conceptual)
+![Tilt Servo Geometry Setup Example](../../assets/config/actuators/tilt_servo_geometry_config.png)
+
+> Example Tilt Servo Geometry setup for a Tiltrotor VTOL
+
+Similar to the Control Surfaces, Servos can also rotate the motors themselves! This is often the case in a VTOL like [Tiltrotors](/en/frames_vtol/tiltrotor.md).
```
bkueng, line 98: How is this similar?

en/config/actuators.md
```diff
@@ -33,50 +33,87 @@ This particular frame also includes an image showing the motor positions, which
...
-### Motor Tilt Servos
+![Tilt Axis](../../assets/config/actuators/tilt_axis.png)
 
-Tilt servos are configured as follows:
-- The reference direction is upwards (negative Z direction).
-- Tilt direction: **Towards Front** means the servo tilts towards positive X direction, whereas **Towards Right** means towards positive Y direction.
-- Minimum and maximum tilt angles: specify the physical limits in degrees of the tilt at minimum control and maximum respectively.
-  An angle of 0° means to point upwards, then increases towards the tilt direction.
+* The reference direction is upwards, indicating that a positive commanded motor thrust will point upwards (negative Z direction).
+* Tilting **Front** means the servo tilts towards Front (Positive X direction).
```
bkueng, line 103:
```suggestion
* Tilting towards **Front** means the servo tilts towards Front (Positive X direction).
```

en/config/actuators.md
```diff
@@ -33,50 +33,87 @@ This particular frame also includes an image showing the motor positions, which
...
+Similar to the Control Surfaces, Servos can also rotate the motors themselves! This is often the case in a VTOL like [Tiltrotors](/en/frames_vtol/tiltrotor.md). 
 
-### Motor Tilt Servos
+![Tilt Axis](../../assets/config/actuators/tilt_axis.png)
 
-Tilt servos are configured as follows:
-- The reference direction is upwards (negative Z direction).
-- Tilt direction: **Towards Front** means the servo tilts towards positive X direction, whereas **Towards Right** means towards positive Y direction.
-- Minimum and maximum tilt angles: specify the physical limits in degrees of the tilt at minimum control and maximum respectively.
-  An angle of 0° means to point upwards, then increases towards the tilt direction.
```
bkueng, line 65: Why did you remove this?

en/config/actuators.md
```diff
@@ -109,31 +146,104 @@ Selecting these requires a reboot before they are applied.
...
-- Servos move into the direction of the convention described above.
-  :::note
-  A trim value can be configured for control surfaces, which is also applied to the test slider.
-  :::
+![](../../assets/config/actuators/identify_motors_button.png)
+
+QGC Supports a feature of identifying the motors by sending the motor command to different output pins and letting you click on the motor to map the Motor to each Output ports.
+
+To use this mode:
+1. Setup the Motors Geometry correctly to include all the motors in your vehicle you want to configure
```
bkueng, line 156: Add: 2. select the output tab where you want to assign the motors, if not assigned already

en/config/actuators.md
```diff
@@ -109,31 +146,104 @@ Selecting these requires a reboot before they are applied.
...
-  For Tiltrotors these are the motors that have no associated tilt servo.
-  Tailsitters use all motors in fixed-wing flight.
-  :::
-- Servos move into the direction of the convention described above.
-  :::note
-  A trim value can be configured for control surfaces, which is also applied to the test slider.
-  :::
+![](../../assets/config/actuators/identify_motors_button.png)
+
+QGC Supports a feature of identifying the motors by sending the motor command to different output pins and letting you click on the motor to map the Motor to each Output ports.
```
bkueng, line 153:
```suggestion
QGC Supports a feature of identifying the motors by sending the motor command to different outputs and letting you click on the motor to map the Motor to each Output ports.
```

en/config/actuators.md
```diff
@@ -109,31 +146,104 @@ Selecting these requires a reboot before they are applied.
...
+
+To use this mode:
+1. Setup the Motors Geometry correctly to include all the motors in your vehicle you want to configure
+2. Click the 'Identify & Assign Motors' Button
+3. Each motor will start spinning. Select the motor that spun up in the GUI.
```
bkueng, line 158:
```suggestion
3. One motor will start spinning. Select the motor that spun up in the GUI.
```

en/config/actuators.md
```diff
@@ -109,31 +146,104 @@ Selecting these requires a reboot before they are applied.
...
+   1. For PWM output, adjust the Disarmed value in case the motor spins at the disarmed command
+2. Slowly move the Motor slider up, and it will snap in at the '**minimum** command'. Verify that the motor is spinning at the minimum speed.
+   1. For **PWM output, adjust the minimum output** value in the 'Actuator Outputs' tab on the right such that the motors barely spin
+    ![PWM Minimum Output](../../assets/config/actuators/pwm_minimum_output.png)
+   :::note
+   For DShot output, this is not required
+   :::
+3. Move the slider up even more, and verify that the Motor is spinning in the correct direction and that it would give a positive thrust in the expected direction
+   1. The expected thrust direction can vary by vehicle type, for example in multicopters the thrust should always point upwards
+   2. For Fixed Wings, the thrust should point to forward direction
```
bkueng, line 200: Isn't that too basic?

en/config/actuators.md
```diff
@@ -109,31 +146,104 @@ Selecting these requires a reboot before they are applied.
...
+   4. Or, an easier way to test without using the slider would be to set the [`COM_PREARM_MODE`](../advanced_config/parameter_reference.md#COM_PREARM_MODE) parameter to `Always`.
+      1. Which will enable the control of Servos even when the vehicle is disarmed, and will constantly be applying the Trim setting to the Control Surfaces
+      2. Here you can try setting different values for the Trim and check the alignment, and then settle on the value you are happy with.
+
+#### Tilt Servo setup
+![Tilt Servo Setup](../../assets/config/actuators/tilt_servo_setup.png)
+
+> Setup of Tilt Servo is similar to Control Surface, as both of them uses the Servo hardware (in most cases)
+
+1. **Set the 'Disarmed' value for Control Surface outputs to '1500'** in case of using a PWM output
```
bkueng, line 228: This depends on the vtol, some might want to use 1000 or 2000.

bkueng (commented):
Thanks for improving the docs.
Could you use the dark qgc theme so the screenshots are consistent?

en/config/actuators.md
junwoo091400: What do you mean by basic?

hamishwillee:
@bkueng @junwoo091400 I've just pushed in a significant (partial) restructure. Wish I'd got it in before @bkueng reviewed. I think it fixes some of Beat's problems. Beat, it is **not ready** to review yet again.

The approach I have taken is to clearly split geometry into motor, control surface and tiltrotor geometry. So for motors I methodically outline what the fields mean. I've done motors. I started tilts, but I don't understand the information. I haven't started the control surface bit - perhaps Junwoo might follow the pattern.

The other thing I plan to to is move most of the testing stuff back into outputs section - that is what it is "about" - setting your outputs.

I might not have time for this until next week. Lets see.

en/config/actuators.md
```diff
@@ -1,94 +1,251 @@
...
-A fixed wing airframe would instead display the parameters that define control surfaces, while a VTOL airframe could display both motors and control surfaces.
+The **Motors** setting lets you choose the number of motors.
 
+For each multicopter motor you can then set:
+- `Position X`: [X-position](#motor-position-coordinate-system), in metres.
+- `Position Y`: [Y-position](#motor-position-coordinate-system), in metres.
+- `Position Z`: [Z-position](#motor-position-coordinate-system), in metres.
+- `Direction CCW`: Checkbox to indicate motor spins counter-clockwise (uncheck for clockwise).
+- `Bidirectional`: Checkbox to indicate motor is [bidirectional](#bidirectional-motors) 
+- `Slew Rate`: ?
```
hamishwillee, line 82: What is slew rate?

en/config/actuators.md
```diff
@@ -1,94 +1,251 @@
...
 
-### Conventions
+The X, Y, Z positions are in [FRD relative to the centre of gravity](#motor-position-coordinate-system).
```
hamishwillee, line 84: Here I forward link to the position information

en/config/actuators.md
```diff
@@ -1,94 +1,251 @@
...
+Motors have most of the same configuration fields as for the [multicopter geometry](#motor-geometry-multicopter).
+There is an additional field to indicate the direction in which the motor moves the vehicle (for a standard VTOL, the hover motors are usually set "upwards" and the pusher motor is set to "forwards").
+
+- `Axis`: One of `Upwards`, `Downwards`, `Forwards`, `Backwards`, `Leftwards`, `Rightwards`, `Custom`
+  - If `Custom` is selected, then the UI displays three additional fields for setting the motor orientation.
+
+#### Motor Geometry: Other Vehicles
 
-![Plane Deflections](../../assets/config/actuators/plane_servo_convention.png)
+Other vehicle types will define an appropriate motor geometry for their frame type.
```
hamishwillee, line 121: This is just a kind of placeholder. If we want we could add boat, etc here too. We might in future, or split out instructions for the different vehicle types.

en/config/actuators.md
```diff
@@ -1,94 +1,251 @@
...
+
+![Reversible](../../assets/config/actuators/qgc_geometry_reversible.png)
+
+Note that you will need to also ensure that the ESC associated with bidirectional motors is configured appropriately (e.g. 3D mode enabled for DShot ESCs, which can be achieved via [DShot commands](../peripherals/dshot.md#commands)).
+
+
+<!-- For example, the roll scale, pitch scale, slew rate and trim can be set (where appropriate) for a control surface, but the positions of these surfaces to each other and to motors cannot be specified. -->
+
+
+### Control Surfaces Geometry
```
hamishwillee, line 148: This bit below is very hard to read. @junwoo091400 I was thinking of making it more "real" like the way I did for multicopters. So none of this "conceptual" heading stuff - use a real example and then extend it. 

As it is now all the information is there, but I think people will find it hard to map to what they are actually doing in the section.

So actually I'd start with a standard vtol (say) and show its actuators and list all the possible options. I might then link down to explain the specific types.

I will look at this tomorrow.

en/config/actuators.md
```diff
@@ -1,94 +1,251 @@
...
+'Pitch UP' control input should command positive deflection to both V-tails
+
+* Example : To pitch in positive direction (pitching upwards),  'Both V-tail surfaces' need a positive deflection.
+
+> Rightwards movement equals Positive Deflection
+'Yaw RIGHT' control input should command positive deflection to both V-tails.
+
+* Example : Having a V or A shaped tail in the back, positive deflection in Rightwards direction will Yaw the vehicle to the right.
+
+### Motor Tilt Servo Geometry
```
hamishwillee, line 199: OK, so here I updated your Motor tilt bit into its own "geometry" section. It cross links back to the tilt motor section.

@bkueng Arguably, it might be good for this section in the geometry to appear just below the motor configuration, since they are closely linked.

This section not quite complete. In particular, I don't understand the reference frame, and how the vehicles tilt.

en/config/actuators.md
hamishwillee: Either way I don't understand the way this works/coordinate system. 

My understanding of tiltrotors comes from the eflight convergence. This has a rotor that points up to hover and forward to fly forward. My understanding is that this would be set to "Towards Front". It points either forward or straight up but it isn't clear to me what the min and max tilt angle will be - I know one will be 0, but not which position, and also whether the other value will be 90 or -90.
Also, what does it mean if the motor goes to the right? Does that mean you can't use it as a pusher?

bkueng: There's a lot of degrees of freedom here, hence it's complicated. You can think of the tilt rotating within a vertical plane, with the 'yaw' of the plane given by the tilt direction (front or right in that case).
Then within the plan you have a starting angle and an end angle. You can think of a clock: 12=upwards=0 deg=always aligned with Z axis, then the angle increases clockwise (if you were to look at the plane from the right side).
Not sure if that helps or makes it even more confusing.

hamishwillee: For the eflight I can kind of see this. The min angle is 0 when in hover, and the max angle is 90 and the tilt is forward. From the right I can see the motor moving from its spindle being pointed up to being forwards, a move from 12 oclock (0) to 3oclock (90). 

But if the motor was moving right then I would see the props coming from invisible to visible and facing me. I guess the move is still 0 to 90. But what use is it? When facing me it is pulling the VTOL "plane" sideways?

IF I had an eflight that used a prop like a pusher below the vehicle and then moved forward, the angle would change from -180 to 90 anticlockwise (from right). 

Nup. I guess I don't get it :-(. I do know this does not pass the "idiot gets it test"

junwoo091400:
> The other thing I plan to to is move most of the testing stuff back into outputs section - that is what it is "about" - setting your outputs.

Thanks for the refactor! I was a bit torn here as the Actuator testing part & Actuator Output mapping part is tightly coupled & should be used together to get maximum value out of them. But I can def move it to Output section, and shrink the Actuator Testing part :wink:

en/config/actuators.md
```diff
@@ -1,65 +1,170 @@
...
+Note that only the most common settings are displayed by default.
+Select the **Advanced** checkbox in the top right corner to display all settings.
+
 
 ## Geometry
 
-The geometry section is used to configure any additional geometry-related settings for the selected [airframe](../config/airframe.md).
-The UI displays a customised view for the selected type; if there are no configurable options this may display a static image of the frame geometry, or nothing at all.
+The geometry section is used to set any configurable geometry-related parameters for the selected [airframe](../config/airframe.md).
+This includes the number and position of [motors](#motor-geometry), and the number, function, and properties of [control surfaces](#control-surfaces-geometry).
```
junwoo091400, line 44: We could omit the 'function' part of the Motors part, I am not sure what you mean by 'function'?

en/config/actuators.md
```diff
@@ -1,65 +1,170 @@
...
 
-:::note
-**Positions are relative to the Center of Gravity of the vehicle, and NOT the Autopilot location!**
-:::
+#### Motor Geometry: VTOL Tailsitter
+
+The default tailsitter motor geometry is shown below.
+Motors have the same configuration fields as for the [multicopter geometry](#motor-geometry-multicopter).
+
+![Geometry motor: tailsitter vtol](../../assets/config/actuators/qgc_geometry_tailsitter_motors.png)
```
junwoo091400, line 92: I think you are referring to the Quad Tailsitter (https://docs.px4.io/master/en/airframes/airframe_reference.html#vtol-quad-tailsitter). But there also is the Duo Tailsitter (https://docs.px4.io/master/en/airframes/airframe_reference.html#vtol-duo-tailsitter), which is what I first thought of when I read the name "VTOL Tailsitter".

It would be better to clarify that this is a 'Quad' Tailsitter!

en/config/actuators.md
```diff
@@ -1,65 +1,170 @@
...
+
+#### Motor Geometry: VTOL Tiltrotor
+
+The default VTOL tiltrotor motor geometry is shown below.
+
+![Geometry motor: tiltrotor vtol](../../assets/config/actuators/qgc_geometry_tiltrotor_motors.png)
```
junwoo091400, line 98: This doesn't really match with the Tiltrotor VTOL configuration (with 3 motors). I will replace it with the actual screenshot from the Converence setup :+1:

en/config/actuators.md
```diff
@@ -1,65 +1,170 @@
...
+The coordinate system for motor positions is FRD (in body frame), where the X axis points forward, the Y axis to the right and the Z axis down.
+The origin of the system is the vehicle centre-of-gravity (COG) (**not** the autopilot location).
 
-### Control Surface direction (Conceptual)
+![Actuators CG reference diagram](../../assets/config/actuators/quadcopter_actuators_cg_reference.png)
+
+#### Bidirectional Motors
+
+Some vehicles may use bidirectional motors (i.e. motor spins in direction 1 for lower output range and in direction 2 for the upper half).
+For example, ground vehicles that want to move forwards and backwards, or VTOL vehicles that have pusher motors that go in either direction.
```
junwoo091400, line 136: What is this 'VTOL vehicle with pusher that can go in either direction' use case? Is this a real use case?

en/config/actuators.md
hamishwillee: Sure!

en/config/actuators.md
hamishwillee: Yes, according to Beat. I guess the pusher can act as an airbrake in that case.

en/config/actuators.md
```diff
@@ -1,65 +1,170 @@
...
+For VTOL tiltrotor vehicles, it will also include the number and properties of [tilt servos](#motor-tilt-servo-geometry)
+
+The UI is customised for the selected airframe.
+Only motors and/or control surfaces that are relevant to the airframe type are displayed or available to assign.
+If there are no configurable geometry options for a particular frame, this section could conceivably be empty.
+
+:::note
+Not all aspects of a geometry are necessarily configurable through this interface.
+Factors that cannot be configured are hard coded for the current airframe. 
+:::
```
junwoo091400, line 54: What do you mean by this? Any examples?

en/config/actuators.md
```diff
@@ -1,65 +1,170 @@
...
+
+### Motor Geometry
+
+The motor geometry section lets you set the number motors, and the relative position, function, and other properties of each motor.
+The motor properties are similar for all frames, but there are some exceptions.
```
junwoo091400, line 59: What kind of 'exceptions' are there?

en/config/actuators.md
hamishwillee: We know for a fact that you can't use a multicopter definition with a vtol, so there are some aspects of the geometry which are not captured. I assume some of them are irrelevant or can be tuned away. 

Or I could just be wrong. @bkueng ?

en/config/actuators.md
hamishwillee: The ones that have different behaviours, outlined in the following sections. So for example, tiltotors additionally set the associated tilt servo, and standard quads have a direction applied to the motors.

en/config/actuators.md
```diff
@@ -55,50 +54,62 @@ Factors that cannot be configured are hard coded for the current airframe.
...
+
+The default motor geometry is shown below. Motors have the same configuration fields as for the [multicopter geometry](#motor-geometry-multicopter).
 
 ![Geometry motor: tailsitter vtol](../../assets/config/actuators/qgc_geometry_tailsitter_motors.png)
 
 #### Motor Geometry: VTOL Tiltrotor
 
 The default VTOL tiltrotor motor geometry is shown below.
 
+![Tiltrotor VTOL Airframe](../../assets/airframes/types/VTOLTiltRotor.svg)
```
hamishwillee, line 108: That image is a hexarotor tiltrotor. 

It is just too big. Ditto with the quad tailstitter. 

FYI only. My point - worth rendering this to see what it looks like. To be clear, I think having an image is a good idea, but one that pushes the content a page down is annoying for readers.

I'll see if I can constrain these.

en/config/actuators.md
```diff
@@ -1,94 +1,255 @@
...
+
+First, the **Motors** drop-down setting lets you choose the number of motors, which is 4 for the example above.
+
+For each motor you can then set:
+- `Position X`: [X-position](#motor-position-coordinate-system), in metres.
+- `Position Y`: [Y-position](#motor-position-coordinate-system), in metres.
+- `Position Z`: [Z-position](#motor-position-coordinate-system), in metres.
+- (Advanced) `Direction CCW`: Checkbox to indicate motor spins counter-clockwise (uncheck for clockwise).
+- (Advanced) `Bidirectional`: Checkbox to indicate motor is [bidirectional](#bidirectional-motors) 
+- (Advanced) `Slew Rate`: ?
```
hamishwillee, line 88: What does this mean? Definition needed

en/config/actuators.md
```diff
@@ -1,94 +1,255 @@
...
-![Geometry MC (QGC)](../../assets/config/actuators/qgc_actuators_mc_geometry.png)
+### Motor Geometry
 
-A fixed wing airframe would instead display the parameters that define control surfaces, while a VTOL airframe could display both motors and control surfaces.
+The motor geometry section lets you set the number of motors, the relative position, and other properties of each motor. The motor properties are similar for all frames, but there are some exceptions.
 
+Multicopter airframe provides a diagram in QGC showing the X, Y positions, which is useful for checking whether the motors configuration is correct. To get a feeling for the motor layout for VTOL and other vehicles see the [Airframe Reference](../airframes/airframe_reference.md)
 
-### Conventions
+The motor geometry explanations for a number of different frames are described below.
```
hamishwillee, line 61: FYI This says "you are describing the explanations" not "describing the frames". I will fix.

en/config/actuators.md
```diff
@@ -1,94 +1,255 @@
...
 
-The following sections contain the conventions and explanations for configuring the geometry.
 
-#### Coordinate system
+#### Motor Position Coordinate System
```
hamishwillee, line 64: I wasn't entirely convinced this should go first. But let's try it here.

en/config/actuators.md
```diff
@@ -1,94 +1,255 @@
...
+- (Advanced) `Slew Rate`: ?
+
+The X, Y, Z positions are in [FRD coordinate frame, relative to the centre of gravity](#motor-position-coordinate-system).
+
+
+#### Motor Geometry: VTOL Quadrotor Tailsitter
+
+VTOL Quadrotor is a tailsitter with 4 motors like a quadcopter as shown below.
+
+![VTOL Quadrotor Tailsitter](../../assets/airframes/types/VTOLQuadRotorTailSitter.svg)
```
hamishwillee, line 97: These images are too big. Let's just link to other docs.

hamishwillee:
@junwoo091400 I've played with the geometry section a bit. I removed the images you added - they take up too much space and add nothing to the explanation. If you're working on a particular frame you already know the details of that frame. I have linked to the airframe referenced instead, which tells you want the outputs are. 

I tend to think that developers want to read as little as possible. That means you show them the things they need to know first and the detail last. You also consider that something is a reference - what you read many times is near the top, and what is read once is at the bottom. [all "if at all possible". Being succint is important.

I'm not going to change it now, but we might consider another approach for the geometry. Rather than splitting it by vehicle type we actually show all the fields and their explanations first for all frames. That's a good reference, though it doesn't have as much context as the way we have done it now. Something to think about.

Now I'll look at the control surface bit.

en/config/actuators.md
```diff
@@ -1,96 +1,250 @@
...
+
+![Control Surface Deflections](../../assets/config/actuators/plane_control_surface_convention.png)
+
+Control surfaces use the following deflection direction convention:
+
+#### Horizontal Control Surface (e.g. Aileron)
+
+> Upwards movement equals Positive deflection
+
+'Pitch UP' control input should command positive deflection to both ailerons
```
hamishwillee, line 171: All of this stuff is very interesting and has nothing to do with configuration of these surfaces. What we need here is an outline of how you set the scaling values. When are they 1? When are they -1. What does 0 mean. Why 0.5?

en/config/actuators.md
```diff
@@ -1,96 +1,235 @@
...
 
 If bidiectional motors are used, make sure to select the **Reversible** checkbox for those motors (the checkbox is displayed as an "advanced" option).
 
-![Reversible](../../assets/config/actuators/qgc_geometry_reversible.png)
+![Reversible](../../assets/config/actuators/qgc_geometry_reversible_marked.png)
 
 Note that you will need to also ensure that the ESC associated with bidirectional motors is configured appropriately (e.g. 3D mode enabled for DShot ESCs, which can be achieved via [DShot commands](../peripherals/dshot.md#commands)).
 
 
+### Control Surfaces Geometry
```
hamishwillee, line 149: @bkueng @junwoo091400 I have updated this control surfaces geometry section. It needs more input from @bkueng now.

Why? All the stuff about how control surfaces work doesn't tell people how to set the values here. 

@bkueng around about line 163 we have this:
```
- `Roll scale`: Roll scale (normalised: -1 to 1)
- `Pitch scale`: Pitch scale (normalised: -1 to 1)
- `Yaw scale`: Yaw scale (normalised: -1 to 1)
- `Trim`: ?
- `Skew`: ?
```
What should we put for each of these? I.e. what do values in the range correspond to? What is the input, and what is the output if (say) the value is 0.7?

Is the "Control Surface Deflection Convention" useful for explaining the above? Left in place in case it is!

en/config/actuators.md
```diff
@@ -1,96 +1,242 @@
...
+- `Angle at max tilt`: [Minimum tilt angle](#tilt-servo-coordinate-system), relative to the z-axis (degrees).
+- `Tilt direction`: `Towards front` (Positive X direction) or `Towards right` (Positive Y direction).
+- `Use for control`: [Tilt server used for yaw/pitch](#tilt-servos-for-yaw-pitch-control)
+  - `None`: Torque control is not used.
+  - `Yaw`: Tilt servos used to control yaw.
+  - `Pitch`: Tilt servos used to control pitch.
+  - `Both Yaw and Pitch`: Tilt servos are used to control both yaw and pitch.
+
+
+#### Tilt Servo Coordinate System
```
hamishwillee, line 210: @bkueng This is how I think tilt servo values work. This is linked to  from section above. The approach is "by example" - which I think removes the ambiguity.

en/config/actuators.md
bkueng: @sfuhrer I leave this up to you

en/config/actuators.md
junwoo091400: Aha gotcha 🙆‍♂️

en/config/actuators.md
junwoo091400: I agree. The Airfarme SVGs are quite confusing! Will create PR for that as well

junwoo091400: PR added here: https://github.com/PX4/PX4-Autopilot/pull/19686

en/config/actuators.md
```diff
@@ -1,96 +1,236 @@
...
+The fields are:
+
+- `Control Surfaces`: The number of control surfaces (set this first!)
+- `Type`: The type of each control surface: `LeftAileron`, `RightAileron`, `Elevator`, `Rudder`, `Left Elevon`, `Right Elevon`, `Left V-Tail`, `Right V-Tail`, `Left Flap`, `Right Flap`, `Airbrakes`, `Custom`.
+- `Roll scale`: Roll scale (normalised: -1 to 1)
+- `Pitch scale`: Pitch scale (normalised: -1 to 1)
+- `Yaw scale`: Yaw scale (normalised: -1 to 1)
+- `Trim`: ?
+- `Skew`: ?
+- `Lock control surfaces in hover`: `Enabled` or `Disabled`.
```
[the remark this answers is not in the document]
sfuhrer: Most VTOLs don't require the control surfaces in hover for control (exception: duo tailsitter that uses the elevons for pitch and yaw control). Sometimes you still want to enable them in hover already, as they can help the MC motors in stabilizing the vehicle in wind or when flying fast.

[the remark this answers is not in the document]
hamishwillee: Thanks @sfuhrer ! I guess if they aren't "required" then locking them prevents them moving around an potentially affecting vehicle control?

Is this OK?
```suggestion
- `Lock control surfaces in hover`:
  - `Enabled`: Most vehicles do not use control surfaces in hover. Use this setting to lock them so that they don't affect vehicle dynamics.
 - `Disabled`: Set this for vehicles that use control surfaces in hover, such as the duo tailsitter (which uses elevons for pitch and yaw control). It should also be set for vehicles that use control surfaces to provide additional stabilization in hover mode when moving at speed or in high winds.
```

en/config/actuators.md
hamishwillee: @sfuhrer Can you comment on this whole section on control surfaces?
In particular, as I asked in ^^^
- What should we put for each of the fields? I.e. what do values in the range correspond to? What is the input, and what is the output if (say) the value is 0.7?
  - In other words, consider someone with a VTOL or FW frame that has an elevon (say). How do they decide what number to put in here?
- How do we describe trim? 
  - I "think" it it is like an offset to the default expected position. So you have a servo where you'd liked the lowest input value to be at the default position. However the unpowered servo is actually a bit less than this. So you're saying that 0 is actually at 0.1 say and that code should automatically account for this when setting values.
  - But I could be completely wrong, and that still doesn't tell me what the scale is. So same question. I have a vehicle, what is this trim, and how do I set it.
- Ditto for skew?

en/config/actuators.md
```diff
@@ -1,96 +1,242 @@
...
+
+The scaling values are normalised (-1 to 1), decoupling the outputs of PX4 controllers from the physical servo control signals.
+
+
+#### Control Surface Deflection Convention
```
hamishwillee, line 173: @sfuhrer This section was originally added to help us describe the control surface values - as I ask here: https://github.com/PX4/PX4-user_guide/pull/1858/files#r881074960

I wasn't sure it was useful for that purpose so I left it in - i.e. to indicate the frame and direction of positive negative. 
Is it useful? If not we will remove it.

en/config/actuators.md
hamishwillee: I might revisit this a bit later, but for now adding the point.,

```suggestion
1. Setup the Motors Geometry correctly to include all the motors in your vehicle you want to configure
1. Select the Output tab where you want to assign the motors (if not assigned already)
```

en/config/actuators.md
hamishwillee: Beat means "simple". Even the most ignorant user will know that thrust is forwards on a fixed wing vehicle.

en/config/actuators.md
hamishwillee: Thanks @bkueng. I changed this to as shown below. If this is not OK let me know (since this text may be moved).

```suggestion
1. **Set the 'Disarmed' value for control surface outputs. The value depends on the vehicle.
```

Are there any particular "considerations" for this? I mean for a _failsafe_ value on an FW vehicle it is not uncommon to set the value such that the vehicle does a slow glide or circle to land. For an MC motor you want a _disarmed_ value to not power a motor. This is kind of FMI, but perhaps might be useful to provide additional hints above.

en/config/actuators.md
```diff
@@ -1,96 +1,250 @@
...
+- `Angle at max tilt`: [Minimum tilt angle](#tilt-servo-coordinate-system) in degrees, relative to the z-axis.
+- `Tilt direction`: `Towards front` (positive x direction) or `Towards right` (positive y direction).
+- `Use for control`: [Tilt servo used for yaw/pitch](#tilt-servos-for-yaw-pitch-control)
+  - `None`: Torque control is not used.
+  - `Yaw`: Tilt servos used to control yaw.
+  - `Pitch`: Tilt servos used to control pitch.
+  - `Both Yaw and Pitch`: Tilt servos are used to control both yaw and pitch.
+
+
+#### Tilt Servo Coordinate System
```
hamishwillee, line 211: Rewrote this again after improving the diagram. I understand this now.

en/config/actuators.md
```diff
@@ -1,96 +1,250 @@
...
+Because **θ<sub>1</sub>** would more negative (smaller) than **θ<sub>0</sub>**, it would be the `Angle at min tilt`.
+
+Similarly, a servo that moves:
+- between the upright and forward postions would have `min=0` and `max=90`. 
+- symmetrically 45 degrees around the upright position would have `min=-45` and `max=45`
+- between the upright and backward positions would have `min=-90` and `max=0`.
+:::
+
+The `Tilt direction` indicates whether the servo tilts in the plane towards the `Front` or `Right` of the vehicle.
+On the diagram this would be represented by **α** that can only take values of 0 (front) or 90 (right).
```
hamishwillee, line 237: The values might be wrong. My assumption is that tilt direction is 0 pointing forward, increasing positive in rotation towards the right.

en/config/actuators.md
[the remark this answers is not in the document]
sfuhrer: >Thanks @sfuhrer ! I guess if they aren't "required" then locking them prevents them moving around an potentially affecting vehicle control?

Yes correct, they usually do more harm then good. 

And yes, I like your proposal :+1:

en/config/actuators.md
sfuhrer: >What should we put for each of the fields? I.e. what do values in the range correspond to? What is the input, and what is the output if (say) the value is 0.7?

It's a normalized effectiveness of the actuator around the corresponding axis. The numbers that are entered there are directly put into the allocation matrix, that is then inverted to get from desired moments (normalized) to control signals. Increasing the scale will reduce the deflection of the control surfaces (as it gets inverted). 
This is very low level stuff that a normal user shouldn't care about, unless it is for the tuning of coupled control surfaces (like an elevon, that controls both pitch and roll). In that case we should mention what I wrote above (if you want more deflection of a certain actuator around a certain axis, you can reduce the scale). But really make sure that for most setups the default scales are all you need (they appear automatically when selecting the control surface type).

Trim: yes it's simply an offset that is added on this actuator, normally to have it centered without input.

>Ditto for skew?

I don't know, where is that used?

We can link the devsummit talk Beat and I will make on CA (once we have it), that will include a walk-through.

en/config/actuators.md
```diff
@@ -1,139 +1,395 @@
...
+![Control Surface Setup Example](../../assets/config/actuators/control_surfaces_geometry.png)
+
+The fields are:
+
+- `Control Surfaces`: The number of control surfaces (set this first!)
+- `Type`: The type of each control surface: `LeftAileron`, `RightAileron`, `Elevator`, `Rudder`, `Left Elevon`, `Right Elevon`, `Left V-Tail`, `Right V-Tail`, `Left Flap`, `Right Flap`, `Airbrakes`, `Custom`.
+- `Roll scale`: Roll scale (normalised: -1 to 1)
+- `Pitch scale`: Pitch scale (normalised: -1 to 1)
+- `Yaw scale`: Yaw scale (normalised: -1 to 1)
+- `Trim`: ?
```
hamishwillee, line 167: @sfuhrer Is this OK? Looking at it, my thought was "how do you work out what it is"? Is it effectively trial and error - i.e. in the actuator panel you set the actuator to the minimum value. If it isn't centred you change this value until it is?
```suggestion
- `Trim`: An offset added to the actuator so that it is centered without input (normally).
```

en/config/actuators.md
```diff
@@ -1,139 +1,395 @@
...
+- `Skew`: ?
```
hamishwillee, line 168: @bkueng Silvan says of skew "I don't know, where is that used?". It's an advanced option for actuators - but I haven't seen it set in any screens. Can you tell me the intent or some frame in which it is needed?

en/config/actuators.md
```diff
@@ -1,139 +1,413 @@
...
+- `Lock control surfaces in hover`:
+  - `Enabled`: Most vehicles do not use control surfaces in hover. Use this setting to lock them so that they don't affect vehicle dynamics.
+ - `Disabled`: Set this for vehicles that use control surfaces in hover, such as the duo tailsitter (which uses elevons for pitch and yaw control). It should also be set for vehicles that use control surfaces to provide additional stabilization in hover mode when moving at speed or in high winds.
+
+
+
+
+#### Actuator Roll, Pitch, and Yaw Scaling
```
hamishwillee, line 176: @sfuhrer FYI only I put the detail on scaling you provided here.

en/config/actuators.md
```diff
@@ -1,139 +1,413 @@
...
+
+![Control Surface Setup Example](../../assets/config/actuators/control_surfaces_geometry.png)
+
+The fields are:
+
+- `Control Surfaces`: The number of control surfaces (set this first!)
+- `Type`: The type of each control surface: `LeftAileron`, `RightAileron`, `Elevator`, `Rudder`, `Left Elevon`, `Right Elevon`, `Left V-Tail`, `Right V-Tail`, `Left Flap`, `Right Flap`, `Airbrakes`, `Custom`.
+- `Roll scale`: Roll scale (normalised: -1 to 1)
+- `Pitch scale`: Pitch scale (normalised: -1 to 1)
+- `Yaw scale`: Yaw scale (normalised: -1 to 1)
```
hamishwillee, line 166: @sfuhrer Thanks for your [response above](https://github.com/PX4/PX4-user_guide/pull/1858/files#r881534338) on these.  To improve this I added a section with your information about the scaling below - https://github.com/PX4/PX4-user_guide/pull/1858/files#r882271662

Then I link to this here. Is this OK?

```suggestion
- `Roll scale`: Effectiveness of actuator around roll axis (normalised: -1 to 1). [Generally you should use the default actuator value](#actuator-roll-pitch-and-yaw-scaling).
- `Pitch scale`: : Effectiveness of actuator around pitch axis (normalised: -1 to 1).  [Generally you should use the default actuator value](#actuator-roll-pitch-and-yaw-scaling).
- `Yaw scale`: Effectiveness of actuator around yaw axis (normalised: -1 to 1).  [Generally you should use the default actuator value](#actuator-roll-pitch-and-yaw-scaling).
```

en/config/actuators.md
sfuhrer: Yes I think it is useful. The convention is also written down in a comment in code: https://github.com/PX4/PX4-Autopilot/blob/96b03040491e727752751c0e0beed87f0966e6d4/src/modules/control_allocator/module.yaml#L492

en/config/actuators.md
sfuhrer: Yes that looks fine for now. And yes, that's what Trim is doing, yes I do it normally just by trial and error, but one could also check the convention for surface deflection. I don't think the "(normally)" is required.

en/config/actuators.md
hamishwillee: Thanks! Added back.

en/config/actuators.md
sfuhrer: Sounds correct to me.

en/config/actuators.md
```diff
@@ -1,139 +1,393 @@
...
+Motors and actuators (which are referred to as "functions") can be assigned to any physical output on any of the available busses.
+
+![Actuator Outputs - Multicopter diagram](../../assets/config/actuators/qgc_actuators_mc_outputs.png)
+
+
+:::note
+PWM AUX outputs are preferred over the PWM MAIN outputs for controlling motors (they have lower latency).
+:::
 
 PWM outputs are grouped according to the hardware groups of the autopilot.
```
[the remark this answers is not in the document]
sfuhrer: Yes I think that's right.

en/config/actuators.md
sfuhrer: I couldn't find any reference in code nor in QGC to "skew". Did you by chance misread and meant "Slew rate"?
image

That would be the same as we had on the legacy mixing system: https://docs.px4.io/v1.12/en/concept/mixing.html#summing-mixer
"A minimal actuator traversal time limit can also be specified in the output scaler (inverse of a slew rate)."

sfuhrer:
@hamishwillee @junwoo091400 can we do a combined effort to get this in still this week? This is turning into a very big (well mainly through the comment history) PR that is hard to keep the overview of.

en/config/actuators.md
```diff
@@ -1,139 +1,412 @@
...
 Each group allows to configure the PWM rate or DShot/Oneshot (if supported).
```
hamishwillee, line 282:
```suggestion
PWM outputs are grouped based on the protocols that they support.
In other words, the outputs that support DShot/Oneshot/PWM would be grouped together, separate from those that don't support Dshot (say).

```

hamishwillee:
> can we do a combined effort to get this in still this week? This is turning into a very big (well mainly through the comment history) PR that is hard to keep the overview of.

I can't. Off until Wednesday. Should be able to get as "good enough" next Wednesday.

What I have done is gone through and resolved "my" open issues with minor fixes. What is still needed:
- We need to know what "skew" means for a setting for MC and for FW.
- Review of the section on motor testing. 
- Move of the section on motor testing. I "think" it is actually mostly motor assignment if you aren't using the automated method. 
- Post process might look at the structure of the geometry section again, but a separate PR.

junwoo091400:
@hamishwillee I have addressed your comments with the latest commit like below:
1. Slew rate setting explanation has been added
2. Motor testing part has been fixed up (on Tilt Servo part)
3. Here I am not sure what you mean, as I think Actuator testing & configuring the diarmed/armed, etc values are tightly coupled, and I would like to know what you exactly mean by 'automated method'
4. Yes that's good for the separate PR

I think so far this is good for merge :+1:

junwoo091400:
@hamishwillee I have addressed your comments with the latest 2 commits like below:
1. Slew rate setting explanation has been added
2. Motor testing part has been fixed up (on Tilt Servo part)
3. Here I am not sure what you mean, as I think **Actuator testing & configuring the diarmed/armed, etc values are tightly coupled**, and so it won't be desirable to separate the testing / output setting. I would like to know what you exactly mean by 'automated method'
4. Yes that's good for the separate PR

I think so far this is good for merge :+1:

en/config/actuators.md
```diff
@@ -1,139 +1,413 @@
...
-The multicopter configuration screen looks like this.
+The view has three sections:
+- [Geometry](#geometry): Configure the geometry for the [selected airframe](../config/airframe.md).
+  This includes number, position and properties of [motors](#motor-geometry) and also the number and properties of [control surfaces](#control-surfaces-geometry) and [motor tilt servos](#motor-tilt-servo-geometry).
+- [Actuator Outputs](#actuator-outputs): Assign motors, control surfaces, and other actuators to specific output.
+- [Actuator Testing](#actuator-testing): Test that motors and actuators move in the direction / speed as expected.
+
+A quadcopter might have an setup screen similar to the one shown below.
+This defines a 4 motor copter with X-geometry, and maps each of the motors to DShot ESC.
+It also maps PWM400 AUX outputs for controlling a parachute and landing gear as well.
```
hamishwillee, line 33: Nope. You can do "It maps .... as well", or have it as it was. But the addition at the end is redundant.

```suggestion
It also maps PWM400 AUX outputs for controlling a parachute and landing gear.
```

en/config/actuators.md
```diff
@@ -45,8 +46,7 @@ This includes the number and position of [motors](#motor-geometry), and the numb
 For VTOL tiltrotor vehicles, it will also include the number and properties of [tilt servos](#motor-tilt-servo-geometry)
 
 :::note
-The UI is customised for the selected airframe.
-Only the motors and/or control surface options that are configurable for the airframe type are displayed.
+The UI shows the configured motor positions, customized to your vehicle (Currently only for Multicopter)
```
hamishwillee, line 49: This change is wrong. This is about the whole geometry section not the multicopter image. 

Or by example, if you have a vehicle with motors that you can't customise the geometry on, they won't be displayed."

I will revert.

en/config/actuators.md
```diff
@@ -1,139 +1,413 @@
...
-The UI displays a customised view for the selected type; if there are no configurable options this may display a static image of the frame geometry, or nothing at all.
+The geometry section is used to set any configurable geometry-related parameters for the selected [airframe](../config/airframe.md).
+This includes the number and position of [motors](#motor-geometry), and the number, function, and properties of [control surfaces](#control-surfaces-geometry).
+For VTOL tiltrotor vehicles, it will also include the number and properties of [tilt servos](#motor-tilt-servo-geometry)
 
-The screenshot below shows the geometry screen for a multicopter frame, which allows you to select the number of motors, their relative positions on the frame, and their expected spin directions (select "**Direction CCW**" for counter-clockwise motors).
-This particular frame also includes an image showing the motor positions, which dynamically updates as the motors settings are changed.
+:::note
+The UI shows the configured motor positions, customized to your vehicle (Currently only for Multicopter)
```
hamishwillee, line 49:
```suggestion
The UI is customised for the selected airframe:
- Only configurable fields for the selected airframe type are displayed (i.e. if motor geometry isn't configurable, the fields won't be visible).
- The motor position diagram is currently only displayed for multicopter frames.
```

en/config/actuators.md
hamishwillee: Actually, I'm going to merge the two points :-)

en/config/actuators.md
```diff
@@ -1,139 +1,413 @@
...
 
-![Geometry MC (QGC)](../../assets/config/actuators/qgc_actuators_mc_geometry.png)
+### Motor Geometry
 
-A fixed wing airframe would instead display the parameters that define control surfaces, while a VTOL airframe could display both motors and control surfaces.
+The motor geometry section lets you set the number of motors, the relative position, and other properties of each motor.
 
+Most motor properties apply to all frames.
+A few properties apply to specific frames.
+For example, `Tilted-by` and `axis` are only shown & relevant for [Tiltrotor VTOL](#motor-geometry-vtol-tiltrotor) and [Standard VTOL](#motor-geometry-standard-vtol) vehicles, respectively.
```
hamishwillee, line 58:
```suggestion
For example, `Tilted-by` and `axis` are only relevant for [Tiltrotor VTOL](#motor-geometry-vtol-tiltrotor) and [Standard VTOL](#motor-geometry-standard-vtol) vehicles, respectively.
```

en/config/actuators.md
```diff
@@ -1,139 +1,413 @@
...
+- `Roll scale`: Effectiveness of actuator around roll axis (normalised: -1 to 1). [Generally you should use the default actuator value](#actuator-roll-pitch-and-yaw-scaling).
+- `Pitch scale`: : Effectiveness of actuator around pitch axis (normalised: -1 to 1).  [Generally you should use the default actuator value](#actuator-roll-pitch-and-yaw-scaling).
+- `Yaw scale`: Effectiveness of actuator around yaw axis (normalised: -1 to 1).  [Generally you should use the default actuator value](#actuator-roll-pitch-and-yaw-scaling).
+- `Trim`: An offset added to the actuator so that it is centered without input. This might be determined by trial and error.
+- `Slew Rate`: Minimum time allowed for the motor/servo signal to pass through the full output range.
+   - It is intended for actuators that may be damaged if they move too fast — like the tilting actuators on a tiltrotor VTOL vehicle. It can be used to limit the rate of change of an actuator (if this is not specified then no rate limit is applied).
+   - For example, a setting of 2.0 means that the motor/servo signal can only go from 0 to 1 in minimum x seconds (in case of reversible motors, the range is -1 to 1).
```
hamishwillee, line 172:
```suggestion
- `Slew Rate`: Minimum time allowed for the motor/servo signal to pass through the full output range, in seconds.
   - It is intended for actuators that may be damaged if they move too fast — like the tilting actuators on a tiltrotor VTOL vehicle.
     The setting limits the rate of change of an actuator (if not specified then no rate limit is applied).
   - For example, a setting of 2.0 means that the motor/servo should not be commanded to move from 0 to 1 at a rate that completes the operation in less than 2 seconds (in case of reversible motors, the range is -1 to 1).
```

en/config/actuators.md
hamishwillee: I'm an idiot. Let us never speak of this again.

en/config/actuators.md
```diff
@@ -1,139 +1,483 @@
...
+
+The control surface configuaration sets output values such that the surfaces:
+
+1. have the correct idle position when disarmed.
+1. move in the directions as defined in the [Control Surface Convention](#control-surface-direction-conceptual).
+
+![Control Surface Disarmed 1500 Setting](../../assets/config/actuators/control_surface_disarmed_1500.png)
+
+For each control surface:
+1. Set the `Disarmed` value for Control Surface outputs to `1500` (note, this assumes we're using a PWM actuator where 1500 is the middle of the range)
```
hamishwillee, line 432: @junwoo091400 I gave it a big edit of the output section. Quite happy with the structure now - following your idea that they are entwined with testing. Also separated out an output assignment section that is "manual".

Just this section and the one after to go. My problem is that, while I get the point, I don't think it is generic.

For a start, are you saying that all control surfaces are controlled using PWM values of range 1000 to 2000. Reading this I think so. Further, I think that you set the value always to 1500. Is that correct, or would you adjust it if it was in the wrong position? Or would you adjust the trim value in the geometry?

en/config/actuators.md
```diff
@@ -1,139 +1,483 @@
...
+   - Unlike motors, when `disarmed` we usually control surfaces to be in the middle position of their range.
+   - Adjust the value until the control is positioned correctly.
+2. Set the Output mode to `PWM 50Hz`, instead of default PWM 400 Hz when using a PWM output for Control Surfaces
```
hamishwillee, line 435: What? Now you're saying I change my mode, and now I use PWM50Hz rather that PWM400Hz? This makes no sense.

What I "think" you mean is that you generally don't need as high speed control for actuators, so you're recommending a slow protocol to get more reliability. Is that right?

So I think you need to start by saying all of that BEFORE you have any of these steps. And then describe the rest of the setting relative to just one PWM setting.

en/config/actuators.md
```diff
@@ -1,139 +1,483 @@
...
+   - If your servo is creating weird noise, it is most likely because of the "PWM 400Hz" output, which is usually not accepted by commercial Servos.
+   - In case your Servo supports 100, 200 or 400 Hz command, you can of course set the output mode to the maximum PWM frequency it supports!
+3. Move Control Surface slider upwards (positive command) and verify that it moves in the direction defined in the [Control Surface Convention](#control-surface-direction-conceptual).
+   1. If the control surface moves in the opposite direction, click on the 'Rev Range' checkbox to reverse the range
+4. Move the slider to the middle and check if the Control Surfaces are aligned in the neutral position of the wing
+
+    ![Control Surface Trimming](../../assets/config/actuators/control_surface_trim.png)
+   1. If it is not aligned, you can set the **Trim** value for each control surface.
```
hamishwillee, line 443: Now we're talking about trimming. I'm even more confused

en/config/actuators.md
```diff
@@ -1,139 +1,483 @@
...
+   3. Note that as mentioned above as a safety feature of slider, unless you actually touch the slider, the actuator command doesn't get sent.
+      Therefore **you must manually click and put the slider in the middle position to test the trim**
+   4. Or, an easier way to test without using the slider would be to set the [`COM_PREARM_MODE`](../advanced_config/parameter_reference.md#COM_PREARM_MODE) parameter to `Always`.
+      1. Which will enable the control of Servos even when the vehicle is disarmed, and will constantly be applying the Trim setting to the Control Surfaces
+      2. Here you can try setting different values for the Trim and check the alignment, and then settle on the value you are happy with.
+
+
+### Tilt Servo Setup
+
+![Tilt Servo Setup](../../assets/config/actuators/tilt_servo_setup.png)
```
hamishwillee, line 454: Probably the same problems as above.

en/config/actuators.md
junwoo091400: I think the `will not be commanded` fits better than `should not be commanded`, since control allocation module is the one in charge, and hence `should` implies that user has to take that into account somehow :thinking:

en/config/actuators.md
junwoo091400: The 'Disarmed' value and 'Trim' values are decoupled and can be quite confusing.

'Disarmed' is which command the Servo will be commanded when the vehicle is disarmed. And it's absolute value / precision is not so relevant, as you just want to have the control surfaces in a reasonable position (and not in min/max position).

'Trim' is relevant for when the control surface is actively being controlled. As far as I know, if the trim = 0, command will be based on the average of the minimum / maximum PWM range. So the 'disarmed' isn't related at all.

en/config/actuators.md
junwoo091400: Gotcha, and yes it's to prevent the Servo from making weird noises when it's commanded with PWM at 400 Hz. And because the default configuration is 400 Hz, I added the command there :thinking: 

Since it applies to the whole commands, I agree it's better to put it in the beginning.

en/config/actuators.md
junwoo091400: Not sure why you are confused. This is a procedure of setting up a control surface that any user would go through :thinking:

junwoo091400:
I have applied your comments @hamishwillee. I have however did not move the `Trim` part out of the Actuator configuration section, as it is unavoidable that user sets the `Trim` during the Actuator Configuration step (as it is actually not Geometry specific: Which, you can set without the knowledge of the actual hardware).

Let me know if this is good :wink:

en/config/actuators.md
```diff
@@ -1,139 +1,484 @@
...
-  A trim value can be configured for control surfaces, which is also applied to the test slider.
-  :::
+![Actuator Testing Slider](../../assets/config/actuators/vtol_tiltrotor_sliders_example.png)
+
+The section has an **Enable Sliders** switch that must be toggled before sliders can be used.
+The sliders can power the motors/servos across their full range of motion, and "snap" to the disarmed and minimum positions.
+
+:::note
+After you toggle the **Enable sliders** switch, actuators/motors won't do anything until the corresponding slider is _moved_.
+This is a safety feature to prevent sudden motor movements after switch is enabled.
```
hamishwillee, line 315: I'm going to leave "to prevent sudden motor movements after switch is enabled" but I think it is "redundant". 

Why? Because the less people have to read the better. It is necessary to know that the actuators won't move unless the sliders move in this case. It is "good" to know that it is intentional behaviour. The extra detail of to prevent sudden motor movements is easily inferred.

It is "correct" though.

en/config/actuators.md
```diff
@@ -312,25 +312,26 @@ The sliders can power the motors/servos across their full range of motion, and "
...
 :::note
 After you toggle the **Enable sliders** switch, actuators/motors won't do anything until the corresponding slider is _moved_.
-This is a safety feature.
+This is a safety feature to prevent sudden motor movements after switch is enabled.
 :::
 
 Sliders can be used to verify the following:
 
-1. Motors and actuators are assigned to the expected output.
+1. Actuators (Motors, Control surfaces, etc.) are assigned to the expected output.
```
hamishwillee, line 320: Nice!

en/config/actuators.md
```diff
@@ -312,25 +312,26 @@ The sliders can power the motors/servos across their full range of motion, and "
...
 1. Motors don't spin when at the `disarmed` PWM output value
 1. Motors barely spin at the `minimum` PWM output value.
 1. Motors give **positive thrust** in the expected direction
-1. Control Surfaces are in the correct idle position when disarmed.
+1. Control Surfaces are in the correct idle position for `disarmed` output value
```
hamishwillee, line 324: Yes, good consistency change. This is what I was trying to capture in my changes - we're slotting to the point that feeds the specified field value :-).

en/config/actuators.md
```diff
@@ -1,139 +1,484 @@
...
+1. Motor Tilt Servos are in the correct idle position for `disarmed` output value
+1. Motor Tilt Servos move in the direction as defined in the [Tilt Servo Convention](#motor-tilt-servos-conceptual)
+
+
+## Output Assignment and Configuration
+
+In the _Actuator Outputs_ section on the right panel, you can assign and configure outputs, optionally using the [Actuator Testing](#actuator-testing) sliders:
+- MC vehicles that have connected motors to PWM outputs can use the [Identify & Assign Motors](#multicopter-pwm-identify-assign-motors) button to perform motor assignment "semi-automatically".
+- Motor output assignment can be done/checked using sliders (see [Output Assignment (Manual)](#output-assignment-manual)).
+- Actuator output assignment can be done/checked using sliders (see [Output Assignment (Manual)](#output-assignment-manual)).
```
hamishwillee, line 335: Merge these two:
```suggestion
- Output assignment (of motors and actuators) can be done/checked using sliders (see [Output Assignment (Manual)](#output-assignment-manual)).
```

en/config/actuators.md
```diff
@@ -1,139 +1,484 @@
...
+1. Motors give **positive thrust** in the expected direction
+1. Control Surfaces are in the correct idle position for `disarmed` output value
+1. Control Surfaces move in the direction as defined in the [Control Surface Convention](#control-surface-direction-conceptual)
+1. Motor Tilt Servos are in the correct idle position for `disarmed` output value
+1. Motor Tilt Servos move in the direction as defined in the [Tilt Servo Convention](#motor-tilt-servos-conceptual)
+
+
+## Output Assignment and Configuration
+
+In the _Actuator Outputs_ section on the right panel, you can assign and configure outputs, optionally using the [Actuator Testing](#actuator-testing) sliders:
```
hamishwillee, line 332:
```suggestion
Outputs are assigned to functions and configured in the [Actuator Outputs](#actuator-outputs) section, while the  [Actuator Testing](#actuator-testing) sliders are commonly used to determine appropriate configuration values to enter.
```

en/config/actuators.md
```diff
@@ -1,139 +1,484 @@
...
+To assign that motor to an output you simply select the corresponding motor displayed in the screen.
+QGC will then spin the next motor for you to assign, and so on.
+
+Instructions:
+1. Setup the motor geometry to match the motors on your frame.
+1. Select the PWM tab where you want to assign the motors.
+1. Click the **Identify & Assign Motors** button.
+1. One motor will start spinning (click **Spin Motor Again** if it stops spinning too quickly to note.)
+
+   Click on the corresponding motor in the geometry section.
```
hamishwillee, line 364: Changing this back. Click and select are synonyms for a button, but click does not work on a touch screen.

```suggestion
   Select the corresponding motor in the geometry section.
```

en/config/actuators.md
```diff
@@ -1,139 +1,483 @@
...
+   - Verify that the motor doesn't spin in this position.
+   - If the motor spins, reduce the corresponding PWM `Disarmed` value in the [Actuator Outputs](#actuator-outputs) section to below the level at which it still spins.
+2. Slowly move the slider up until it snaps to the _minimum_ position.
+   In this position the motor is set to the outputs `Minimum` value.  
+   - Verify that the motor is spinning very slowly in this position.
+   - If the motor is not spinning, or spinning too fast you will need to adjust the corresponding PWM `Disarmed` value in the [Actuator Outputs](#actuator-outputs) such that the motors barely spin.
+   
+     ![PWM Minimum Output](../../assets/config/actuators/pwm_minimum_output.png)
+     :::note
+     For DShot output, this is not required <!-- any, or just the minimum check? -->
```
hamishwillee, line 412: Is it just this step or any of this configuration not required for DShot?

en/config/actuators.md
```diff
@@ -1,139 +1,493 @@
...
+1. Set the `Disarmed` value so that the surfaces will stay at neutral position when disarmed.
+   This is usually around `1500` for PWM servos.
+2. Move the slider for the surface upwards (positive command) and verify that it moves in the direction defined in the [Control Surface Convention](#control-surface-direction-conceptual).
+   - If the control surface moves in the opposite direction, click on the `Rev Range` checkbox to reverse the range.
+3. Move the slider again to the middle and check if the Control Surfaces are aligned in the neutral position of the wing
+   - If it is not aligned, you can set the **Trim** value for the control surface.
+     :::note
+	 This is done in the `Trim` setting of the Geometry panel, usually by "trial and error".
+	 ![Control Surface Trimming](../../assets/config/actuators/control_surface_trim.png)
+	 :::
```
hamishwillee, line 449:
```suggestion
     :::
```

hamishwillee:
@junwoo091400 Done and merged! Thanks so much for all your hard work on this - it has a lot more information than the original version and should be a great help for PX4 users in v1.13 and beyond.

@bkueng I've merged, but would appreciate your review of the [Control Surface](https://docs.px4.io/master/en/config/actuators.html) and Tilt Rotor configuration sections. I made a few comments I'm not sure on - like recommending Dshot over PWM400 servo if you really need a high rate servo as "better value" (I know you probably don't need Dshot anyway for control surfaces).

en/config/actuators.md
junwoo091400: I agree! It's redundant :disappointed:, now that I reflect back on it! I just remembered Beat telling me the reason for this safety feature, and I thought it wouldn't be so easily inferrable, but I guess it is :wink:

en/config/actuators.md
junwoo091400: Actuator Testing also is used for determining control surface movement direction, trim and so on, so maybe this change makes it too specific to one aspect of Actuator Testing?

en/config/actuators.md
junwoo091400: Haven't thought about that!

en/config/actuators.md
hamishwillee: @junwoo091400 Good point, I patched this in https://github.com/PX4/PX4-user_guide/commit/20a4ae785c91976983a3d75d2692f410da67208d

PR #1863, opened 2022-04-28
changed -w as  NUM_WORKERS instead of warn

khanasif786:


en/simulation/ros_interface.md
```diff
@@ -41,7 +41,7 @@ roslaunch mavros px4.launch fcu_url:="udp://:14540@127.0.0.1:14557"
 :::note
 It can be useful to call *roslaunch* with the `-w` (warn) and/or `-v` (verbose) in order to get warnings about missing dependencies in your setup. For example:
 ```sh
-roslaunch mavros px4.launch fcu_url:="udp://:14540@127.0.0.1:14557"
+roslaunch -v mavros px4.launch fcu_url:="udp://:14540@127.0.0.1:14557"
```
hamishwillee, line 44: @Jaeyoung-Lim Is this something you could review?

Jaeyoung-Lim: I don't think this should be set as the default. Since it is already explained with respect to the `-w` and `-v` flags I would suggest to not add this as default.

khanasif786: @Jaeyoung-Lim. Actually the -w flag doesn't work, I checked that. Well, I will be happy to close.

Jaeyoung-Lim: @khanasif786 Could you maybe update the `-w` flag as the correct definition? 
```
  -w NUM_WORKERS, --numworkers=NUM_WORKERS
                        override number of worker threads. Only valid for core
                        services.
```

khanasif786: @Jaeyoung-Lim . Yeah, I was going wrong after seeing -w as warn. Here warn having a different work. I have updated according to NUM_WORKERS, but don't you think we should add these args to example command also.

khanasif786:
@Jaeyoung-Lim Any update ?

en/simulation/ros_interface.md
hamishwillee: @Jaeyoung-Lim Follow on question here ^^^^.

Jaeyoung-Lim (approved):

hamishwillee (approved):
Thanks very much @khanasif786 and @Jaeyoung-Lim - much appreciated.

PR #1867, opened 2022-04-30
Update multi vehicle simulation supported vehicles

elpiel:
The list was updated accordingly based on the script:

https://github.com/PX4/PX4-Autopilot/blob/6fc857772dc57d95ed1d8ecb351cea6aacea42aa/Tools/gazebo_sitl_multiple_run.sh#L23


hamishwillee (approved):
Awesome, thanks very much!

FYI @Jaeyoung-Lim What's the best way to avoid us getting out of sync here again?

PR #1872, opened 2022-05-02
Update beaglebone_blue.md

Superbear88:


hamishwillee (approved):
@Superbear88 Thanks! I am assuming that the root@beaglebone changed to debian@beaglebone  - but I have not tested this.

Generally when you make a change it really helps if you add a description and supporting information. That makes it easy for me to decide that a change is valid. So here you could state in the description that 

> "root@beaglebone changed to debian@beaglebone in [VERSION ?], as you can see in this [INSERT LINK TO INFO/PROOF]

PR #1879, opened 2022-05-07
Add note about QGC manual communication link for ethernet

danylevskyi:
I was confused by the phrase `QGroundControl should connect automatically.` while configuring an ethernet connection.
In my case, I had to manually add a communication link.

Screenshot 2022-04-28 at 15 33 51

Screenshot 2022-04-28 at 15 32 32

Slack thread: https://px4.slack.com/archives/C581ST17V/p1648008264584879

### DIFF
--- a/en/advanced_config/ethernet_setup.md
+++ b/en/advanced_config/ethernet_setup.md
@@ -158,6 +158,10 @@ To connect a GCS to PX4 over Ethernet:
 1. QGroundControl should connect automatically.
 
 :::note
+In case if QGC doesn't connect automatically try to add communication link manually. in QGC go to Application Settings → Comm Links and create UDP connection with port `14550` and server address 192.168.0.4 (Pixhawk IP address). 
+:::
+
+:::note
 [PX4 Ethernet Port Configuration](#px4-ethernet-port-configuration) is not required because it is setup by default for connecting to a GCS.
 :::
 


mrpollo:
I think the original claim should say the flight controller needs an IP assigned, which typically comes via DHCP, and if there's no such service on the network you need to configure it manually. That should clear things up

What do you think @hamishwillee ?

hamishwillee:
@mrpollo I don't think so, though it is worth saying. All that resolves is how PX4 might get its address, but having any old address does not allow QGC to connect to you. Even if it is assigned by DHCP. 

My understanding:
- QGC listens for MAVLink on port 14550 on its computer by default. That works for Sim, because PX4 and QGC run on the same computer. 
- It could work on a network with just one PX4 vehicle too, if it multicasts. This is what I thought was happening. In this case if QGC and PX4 are on the same subnet I think that QGC listens on its port 14550 and the connection works.
- What will always work is if QGC listens on the IP address and port of the PX4 node - that's what is being set up above.

I think we probably need to explicitly say that you will need to set up QGC as indicated here, because it is the most "general" solution that should always work. 

But I will look at this next week.

hamishwillee:
FYI, fixed in https://github.com/PX4/PX4-user_guide/pull/1884

Thanks or reporting @danylevskyi

PR #1880, opened 2022-05-07
Basic update of FW controllers documentation and TECS diagrams

kaklik:
Documentation facts have been updated to match the PX4 source code. 

- Explain [coordinated turn](https://discuss.px4.io/t/fixed-wing-attitude-controller-turn-coordination-example/26378/2)
- Add a note about the possibility of disabling the airspeed scaling. 
- Explain [Throttle pass through in attitude controller](https://discuss.px4.io/t/thrust-controller-using-throttle-set-point-from-tecs/16159/3). 

TODO: [TECS controllers diagrams need an update](https://discuss.px4.io/t/fixed-wing-tecs-controller-disconnect-with-documentation/19983). Unfortunately, a source of the diagrams seems to be unavailable.


hamishwillee:
Thanks! I'll look properly Wednesday, in the meantime, source for images is usually listed in the source code of the page in a comment. In this case I think the relevant image is  on draw.io: https://drive.google.com/file/d/1ibxekmtc6Ljq60DvNMplgnnU-JOvKYLQ/view?usp=sharing - Request access from dev team by trying to open it.

kaklik:
> Thanks! I'll look properly Wednesday, in the meantime, source for images is usually listed in the source code of the page in a comment. In this case I think the relevant image is on draw.io: https://drive.google.com/file/d/1ibxekmtc6Ljq60DvNMplgnnU-JOvKYLQ/view?usp=sharing - Request access from dev team by trying to open it.

Hi! Unfortunately, that is not the TECS controller diagram, but the FW attitude controller diagram. TECS is slightly above that on the documentation page. I noticed that some diagrams have a link to sources, but the TECS controller does not.

hamishwillee:
@sfuhrer @RomanBapst Can you guys please comment on the TEC docs changes and where the source of the controller diagram source might be?

This is the stuff that was added to the old developer guide in https://github.com/PX4/PX4-Devguide/pull/779

sfuhrer:
> @sfuhrer @RomanBapst Can you guys please comment on the TEC docs changes and where the source of the controller diagram source might be?
> 
> This is the stuff that was added to the old developer guide in [PX4/PX4-Devguide#779](https://github.com/PX4/PX4-Devguide/pull/779)

Yes the TECS structure indeed changed recently. @RomanBapst did the old TECS diagram when I remember correctly, maybe he still has the source. Would be nice to have it in an open and easy to edit way, as it's for likely gonna change again in the next months.

RomanBapst:
@kaklik I found one of the diagrams on the drive:
https://drive.google.com/file/d/1m73F6MKW_CYu2E89CrFhUCeEJ6tKJFkW/view?usp=sharing
Can you please check if you can access it?

hamishwillee:
@kaklik I have added an image link for the tecs throttle (from Roman) that can be opened in draw io below the diagram (as an html comment in the source at http://docs.px4.io/master/en/flight_stack/controller_diagrams.html#total-energy-control-loop )

The tecs pitch diagram he links above https://github.com/PX4/PX4-user_guide/pull/1880#issuecomment-1124578229 is an HTML file. To open that I had to download it and then import it to https://app.diagrams.net/ - not quite as convenient. But should give you something to start from. We should provide the same kind of URL as for the throttle case.

Ping Roman when you've made some changes. He's the expert in this, and best person to review.

kaklik:
Ok, Thanks @RomanBapst  for the finding source of the TECS Throttle diagram!  It seems that I am able to open it by the procedure described by @hamishwillee .  Is there some possibility that the TECS Pitch control diagram could be found also? 

The diagram update seems to take some more time. Therefore I suggest splitting the diagram update to another issue/PR. The current pull request is not directly related to the diagrams update, therefore I think it could be merged.

kaklik:
Here is [redrawn TECS pitch control diagram](https://drive.google.com/file/d/1bZtFULYmys-_EQNhC9MNcKLFauc7OYJZ/view?usp=sharing).

TECS_pitch drawio

In case both updated diagrams should be considered as correct -  @RomanBapst  ?  Then both should be added to this pull request. 

> 
> Would be nice to have it in an open and easy to edit way, as it's for likely gonna change again in the next months.

@sfuhrer do you mean some specific change in the TECS structure, which is planned to happen in the next months?

RomanBapst:
@kaklik The pitch loop looks ok. However, in the throttle loop the feedforward term contains feedback which does not make sense, right?
The feedforward part is only the predicted throttle (throttle_predicted). I think that should be split up to make it more clear.

RomanBapst:
@kaklik What do you think about adding a separate diagram for the specific energy rates / setpoints?
It would be nice to see how those are calculated as well.

RomanBapst:
@kaklik Thanks, that looks better!

hamishwillee (approved):
@RomanBapst I assume that your comment means this is ready to merge. Am doing so.

Thanks very much @kaklik

PR #1887, opened 2022-05-16
ULog File Format Document Improvement

junwoo091400:
## Background
While writing the Part 4 of the PX4 uORB Explained Series, I found that ULog File Format document is quite hard to read. Therefore here's the improvement PR!

## Changes from this PR
1. Change Message Type sections from lists into concrete header sections
2. Add Table of 'Messages' section for Definitions / Data section, to
give an overview of the types of messages logged
3. Rephrased and organized doc sentences to make it more easily
understandable

## Note
In case interested, here's the published Part 3 of the uORB Explained Series : https://px4.io/px4-uorb-explained-part-3-the-deep-stuff/

### DIFF
--- a/en/dev_log/ulog_file_format.md
+++ b/en/dev_log/ulog_file_format.md
@@ -5,7 +5,7 @@ The format is self-describing, i.e. it contains the format and [uORB](../middlew
 
 ::: note
 The [system logger](../modules/modules_system.md#logger) allows the default set of logged topics to be replaced with a custom set that is defined on an SD Card.
-See [Logging > SD Card Configuration](../dev_log/logging.html#sd-card-configuration) for more information.
+See [Logging > SD Card Configuration](../dev_log/logging#sd-card-configuration) for more information.
 :::
 
 ULog commonly logs uORB topics related to (but not limited to) the following sources:
@@ -32,7 +32,10 @@ The following binary types are used for logging. They all correspond to the type
 Additionally the types can be used as an array: e.g. `float[5]`.
 
 Strings (`char[length]`) do not usually contain the termination NULL character `'\0'` at the end.
-String comparisons are case sensitive.
+
+:::note
+String comparisons are case sensitive, which should be taken into account when comparing the uORB topic names when [adding subscriptions](#a-subscription-message).
+:::
 
 ## ULog File Structure
 
@@ -80,10 +83,10 @@ struct message_header_s {
 - `msg_size` is the size of the message in bytes without the header.
 - `msg_type` defines the content, and is a single alphabet character.
 
-In practice, we include the `msg_size` and `msg_type` in each message definitions as the first 2 members, so the header `message_header_s` never gets used directly in most cases.
+In practice, we include the `msg_size` and `msg_type` in each message definitions as the first 2 members of different message definitions.
 
 :::note
-Messages described below are prefixed with the single alphabet that corresponds to it's `msg_type` in the beginning.
+Messages below are prefixed with the single alphabet that corresponds to it's `msg_type`.
 :::
 
 
@@ -100,7 +103,7 @@ The message types in this section are:
 6. [Default Parameter](#q-default-parameter-message)
 
 
-#### 'B': Flag Bits Message
+#### 'B': Flag Bits Message (`msg_type = 'B'`)
 
 :::note
 This message must be the **first message** right after the header section, so that it has a fixed constant offset from the start of the file!
@@ -168,7 +171,8 @@ Some field names are special:
 - `timestamp`: every [Subscription Message](#a-subscription-message) must include a timestamp field
   - Its type can be: `uint64_t` (currently the only one used), `uint32_t`, `uint16_t` or `uint8_t`.
   - The unit is always microseconds, except for in `uint8_t` where the unit is in milliseconds.
-  - A log writer must make sure to log messages often enough to be able to detect wrap-arounds and a log reader must handle wrap-arounds (and take into account dropouts).
+  - A log writer must make sure to log messages often enough to be able to detect **wrap-arounds** (when the timestamp overflows the data type and goes back to 0)
+  - Also, the log reader must handle wrap-arounds (and take into account dropouts).
   - The timestamp must always be monotonic increasing for a message series with the same `msg_id`.
 - `_padding{}`: field names that start with `_padding` (e.g. `_padding[3]`) should not be displayed and their data must be ignored by a reader.
   - These fields can be inserted by a writer to ensure correct alignment.
--- a/en/dev_log/ulog_file_format.md
+++ b/en/dev_log/ulog_file_format.md
@@ -1,7 +1,8 @@
 # ULog File Format
 
 ULog (Universal-Log) is the file format used for logging messages. The format is self-describing, i.e. it contains the format and [uORB](../middleware/uorb.md) message types that are logged.
-This document is meant to be the ULog File Format Spec Documentation, especially for anyone who is interested in developing a ULog Parser / Application and needs to decode ULog file.
+This document is meant to be the ULog File Format Spec Documentation.
+It is intended especially for anyone who is interested in writing a ULog parser / serializer and needs to decode / encode files.
 
 PX4 uses ULog to log uORB topics as messages related to (but not limited to) the following sources:
 - **Device inputs:** Sensors, RC input, etc.
@@ -102,7 +103,7 @@ The message types in this section are:
 This message must be the **first message** right after the header section, so that it has a fixed constant offset from the start of the file!
 :::
 
-This message provides an information to the log parser whether the log is parsable or not.
+This message provides information to the log parser whether the log is parsable or not.
 
 ```c
 struct ulog_message_flag_bits_s {
@@ -114,11 +115,12 @@ struct ulog_message_flag_bits_s {
 ```
 
 - `compat_flags`: compatible flag bits
-  - These flags indicate whether any information that is compatible with the ULog file format is in the log.
+  - These flags indicate the presence of features in the log file that are compatible with any ULog parser.
   - `compat_flags[0]`: *DEFAULT_PARAMETERS* (Bit 0): if set, the log contains [default parameters message](#q-default-parameter-message)
 
   The rest of the bits are currently not defined and must be set to 0.
-  These bits can be used for future ULog changes that are compatible with existing parsers. For example, adding a new message types, which can be indicated by setting one of the Bits, and the existing parsers can just ignore the new message type in the log.
+  These bits can be used for future ULog changes that are compatible with existing parsers.
+  For example, adding a new message type can be indicated by defining a new bit in the standard, and existing parsers will ignore the new message type.
   It means parsers can just ignore the bits if one of the unknown bits is set.
 - `incompat_flags`: incompatible flag bits.
   - `incompat_flags[0]`: *DATA_APPENDED* (Bit 0): if set, the log contains appended data and at least one of the `appended_offsets` is non-zero.
@@ -128,11 +130,12 @@ struct ulog_message_flag_bits_s {
   If a parser finds any of these bits set that isn't specified, it must refuse to parse the log.
 - `appended_offsets`: File offset for appended data.
   If no data is appended, all offsets must be zero.
-  This can be used to reliably append data for logs that may stop in the middle of a message. For example, hardfault dumps in PX4.
+  This can be used to reliably append data for logs that may stop in the middle of a message.
+  For example, crash dumps.
 
   A process appending data should do:
   - set the relevant `incompat_flags` bit
-  - set the first `appended_offsets` as the length of the log file without the appended data, as that is where the data would start
+  - set the first `appended_offsets` that is currently 0 to the length of the log file without the appended data, as that is where the new data will start
   - append any type of messages that are valid for the Data section.
 
 It is possible that there are more fields appended at the end of this message in future ULog specifications.
@@ -141,7 +144,7 @@ If the `msg_size` is bigger than expected (currently 40), any additional bytes m
 
 #### 'F': Format Message
 
-Format message defines a singe message name and it's inner fields in a single string.
+Format message defines a single message name and it's inner fields in a single string.
 
 ```c
 struct message_format_s {
@@ -222,7 +225,7 @@ key | Description | Example for value
 `value_len` represents the data size of the `value`. This is described in the `key`.
 :::
 
-- For PX4, the format of `ver_sw_release` and `ver_os_release` is: 0xAABBCCTT, where AA is **major**, BB is **minor**, CC is patch and TT is the **type**.
+- The format of `ver_sw_release` and `ver_os_release` is: 0xAABBCCTT, where AA is **major**, BB is **minor**, CC is patch and TT is the **type**.
   - **Type** is defined as following: `>= 0`: development, `>= 64`: alpha version, `>= 128`: beta version, `>= 192`: RC version, `== 255`: release version.
   - For example, `0x010402FF` translates into the release version v1.4.2.
 
@@ -238,7 +241,7 @@ struct ulog_message_info_multiple_header_s {
   uint8_t is_continued; // can be used for arrays
   uint8_t key_len;
   char key[key_len];
-  char value[header.msg_size-key_len]
+  char value[header.msg_size-2-key_len]
 };
 ```
 
--- a/en/dev_log/ulog_file_format.md
+++ b/en/dev_log/ulog_file_format.md
@@ -62,26 +62,28 @@ The header is a fixed-size section and has the following format (16 bytes):
 ----------------------------------------------------------------------
 ```
 
-- **File Magic (7 Bytes):** File type indicator that reads "ULog"
+- **File Magic (7 Bytes):** File type indicator that reads "ULogXYZ where XYZ is the magic bytes sequence `0x01 0x12 0x35`"
 - **Version (1 Byte):** File format version (currently 1)
 - **Timestamp (8 Bytes):** `uint64_t` integer that denotes when the logging started in microseconds.
 
 ### Definition & Data Section Message Header
 
-The _Definitions and Data_ section consists of a stream of **messages** that always starts with this header: 
+The _Definitions and Data_ sections contain a number of **messages**. Each message is preceded by this header:
 
 ```c
 struct message_header_s {
-	uint16_t msg_size;
-	uint8_t msg_type
+  uint16_t msg_size;
+  uint8_t msg_type;
 };
 ```
 
 - `msg_size` is the size of the message in bytes without the header.
 - `msg_type` defines the content, and is a single alphabet character.
 
+In practice, we include the `msg_size` and `msg_type` in each message definitions as the first 2 members, so the header `message_header_s` never gets used directly in most cases.
+
 :::note
-Message types described below are noted with the single alphabet `msg_type` characteristic in the beginning.
+Messages described below are prefixed with the single alphabet that corresponds to it's `msg_type` in the beginning.
 :::
 
 
@@ -90,12 +92,12 @@ Message types described below are noted with the single alphabet `msg_type` char
 The definitions section contains basic information such as software version, uORB Topic format, initial parameter values, and so on.
 
 The message types in this section are:
-1. [Flag Bits](#b--flag-bits-message)
-2. [Format Definition](#f--format-definition-message)
-3. [Information](#i--information-message)
-4. [Multi Information](#m--multi-information-message)
-5. [Parameter](#p--parameter-message)
-6. [Default Parameter](#q--default-parameter-message)
+1. [Flag Bits](#b-flag-bits-message)
+2. [Format Definition](#f-format-message)
+3. [Information](#i-information-message)
+4. [Multi Information](#m-multi-information-message)
+5. [Parameter](#p-parameter-message)
+6. [Default Parameter](#q-default-parameter-message)
 
 
 #### 'B': Flag Bits Message
@@ -106,21 +108,24 @@ This message must be the **first message** right after the header section, so th
 
 ```c
 struct ulog_message_flag_bits_s {
-  struct message_header_s;
+  uint16_t msg_size;
+  uint8_t msg_type = 'B';
   uint8_t compat_flags[8];
   uint8_t incompat_flags[8];
   uint64_t appended_offsets[3]; ///< file offset(s) for appended data if appending bit is set
 };
 ```
 
-- `compat_flags`: compatible flag bits.
-  - `compat_flags[0]`, bit 0, *DEFAULT_PARAMETERS*: if set, the log contains parameter defaults (message 'Q').
+- `compat_flags`: compatible flag bits
+  - These flags indicate any information that is compatible with the existing ULog is in the log.
+  - `compat_flags[0]`: *DEFAULT_PARAMETERS* (Bit 0): if set, the log contains [default parameters message](#q-default-parameter-message)
 
-  The rest of the bits is currently not defined and all must be set to 0.
+  The rest of the bits are currently not defined and must be set to 0.
   These bits can be used for future ULog changes that are compatible with existing parsers.
   It means parsers can just ignore the bits if one of the unknown bits is set.
 - `incompat_flags`: incompatible flag bits.
-  The LSB bit of index 0 is set to one if the log contains appended data and at least one of the `appended_offsets` is non-zero.
+  - `incompat_flags[0]`: *DATA_APPENDED* (Bit 0): if set, the log contains appended data and at least one of the `appended_offsets` is non-zero.
+  
   All other bits are undefined and must be set to 0.
   If a parser finds one of these bits set, it must refuse to parse the log.
   This can be used to introduce breaking changes that existing parsers cannot handle.
@@ -137,9 +142,9 @@ It is possible that there are more fields appended at the end of this message in
 This means a parser must not assume a fixed length of this message.
 If the message is longer than expected (currently 40 bytes), the exceeding bytes must just be ignored.
 
-#### 'F': Format Definition Message
+#### 'F': Format Message
 
-Format definition that can be logged or used in another format definitions as a nested type.
+Format message defines a single uORB topic's name and it's inner fields in a single string.
 
 ```c
 struct message_format_s {
@@ -148,23 +153,24 @@ struct message_format_s {
 };
 ```
 
-* `format` is a plain-text string with the following format: `message_name:field0;field1;`
+- `format` is a plain-text string with the following format: `message_name:field0;field1;`
+  - There can be an arbitrary amount of fields (minimum 1), separated by `;`.
 
-There can be an arbitrary amount of fields (minimum 1), separated by `;`.
+A `field` has the format: `type field_name`, or for an array: `type[array_length] field_name` is used (only fixed size arrays are supported).
 
-A field has the format: `type field_name` or `type[array_length] field_name` for arrays (only fixed size arrays are supported).
-
-`type` is one of the basic binary types or a `message_name` of another format definition (nested usage).
-A type can be used before it's defined.
-There can be arbitrary nesting but no circular dependencies.
+A `type` is one of the [basic binary types](#data-types) or a `message_name` of another format definition (nested usage).
+- A type can be used before it's defined.
+  - e.g. The message `MessageA:MessageB[2] msg_b` can come before the `MessageB:uint_8[3] data`
+- There can be arbitrary nesting but **no circular dependencies**
+  - e.g. `MessageA:MessageB[2] msg_b` & `MessageB:MessageA[4] msg_a`
 
 Some field names are special:
-- `timestamp`: every [Subscription Message](#a--subscription-message) must include a timestamp field
+- `timestamp`: every [Subscription Message](#a-subscription-message) must include a timestamp field
   - Its type can be: `uint64_t` (currently the only one used), `uint32_t`, `uint16_t` or `uint8_t`.
-  - The unit is always microseconds, except for in `uint8_t` it's milliseconds.
+  - The unit is always microseconds, except for in `uint8_t` where the unit is in milliseconds.
   - A log writer must make sure to log messages often enough to be able to detect wrap-arounds and a log reader must handle wrap-arounds (and take into account dropouts).
   - The timestamp must always be monotonic increasing for a message series with the same `msg_id`.
-- Padding: field names that start with `_padding` should not be displayed and their data must be ignored by a reader.
+- `_padding{}`: field names that start with `_padding` (e.g. `_padding[3]`) should not be displayed and their data must be ignored by a reader.
   - These fields can be inserted by a writer to ensure correct alignment.
   - If the padding field is the last field, then this field will not be logged, to avoid writing unnecessary data.
   - This means the `message_data_s.data` will be shorter by the size of the padding.
@@ -172,16 +178,21 @@ Some field names are special:
 
 #### 'I': Information Message
 
+The Information message defines a dictionary type definition `key` : `value` pair for any information, including but not limited to Hardware version, Software version, Build toolchain for the software, etc.
+
 ```c
-struct message_info_s {
-  struct message_header_s header;
+struct ulog_message_info_header_s {
+  uint16_t msg_size;
+  uint8_t msg_type = 'I';
   uint8_t key_len;
-  char key[key_len];
-  char value[header.msg_size-1-key_len]
+  char key[255];
 };
 ```
-* `key` is a plain string, as in the format message (can also be a custom type), but consists of only a single field without ending `;`, eg. `float[3] myvalues`
-* `value` contains the data as described by `key`
+
+- `key_len`: Length of the key value
+- `key`: String with the `key` and `value` information combined: e.g. `char[5] sys_toolchain_ver9.4.0`
+  - The `key` part has the format identical as the field in the [Format message](#f-format-message) (can also be a custom type), but consists of only a single field without ending `;`, eg. `char[5] sys_toolchain_ver`
+  - The `value` contains the data corresponding to the `key`, and is written right after the `key`, e.g. `9.4.0`.
 
 :::note
 A key : value pair defined in the Information message should be unique. Meaning there shouldn't be more than one definition with the same key value!
@@ -203,8 +214,8 @@ key | Description | Example for value
 `char[7] sys_os_ve`r | OS version (git tag) |  "9f82919"
 `uint32_t ver_os_release` | OS version (see below) |  0x010401ff
 `char[7] sys_toolchain` | Toolchain Name |  "GNU GCC"
-`char[5] sys_toolchain_ver | Toolchain Version |  "6.2.1"
-char[16] sys_mcu` | Chip name and revision |  "STM32F42x, rev A"
+`char[5] sys_toolchain_ver` | Toolchain Version |  "6.2.1"
+`char[16] sys_mcu` | Chip name and revision |  "STM32F42x, rev A"
 `char[12] sys_uuid` | Unique identifier for vehicle (eg. MCU ID) |  "392a93e32fa3"...
 `char[7] log_type` | Type of the log (full log if not specified) | "mission"
 `char[10] replay` | File name of replayed log if in replay mode | "log001.ulg"
@@ -215,15 +226,15 @@ The `key_len` value in `char key[key_len]` has no fixed value.
 Therefore the key lengths in the example above (3, 9, 2, 7, ...) may not be exactly same with your ULog's information message.
 :::
 
-* The format of `ver_sw_release` and `ver_os_release` is: 0xAABBCCTT, where AA is **major**, BB is **minor**, CC is patch and TT is the **type**.
-  * **Type** is defined as following: `>= 0`: development, `>= 64`: alpha version, `>= 128`: beta version, `>= 192`: RC version, `== 255`: release version.
-  * For example, 0x010402ff translates into the release version v1.4.2.
+- The format of `ver_sw_release` and `ver_os_release` is: 0xAABBCCTT, where AA is **major**, BB is **minor**, CC is patch and TT is the **type**.
+  - **Type** is defined as following: `>= 0`: development, `>= 64`: alpha version, `>= 128`: beta version, `>= 192`: RC version, `== 255`: release version.
+  - For example, `0x010402FF` translates into the release version v1.4.2.
 
 This message can also be used in the Data section (this is however the preferred section).
 
 #### 'M': Multi Information Message
 
-The same as the information message, except that there can be multiple messages with the same key (parsers store them as a list).
+Multi information message serves the same purpose as the information message, but for cases where the `value` length exceeds  (parsers store them as a list).
 
 ```c
 struct ulog_message_info_multiple_header_s {
@@ -241,7 +252,7 @@ Parsers can store all information multi messages as a 2D list, using the same or
 
 #### 'P': Parameter Message
 
-Same format as the [Information Message](#i--information-message).
+Parameter message in the _Definitions_ section defines the parameter values of the vehicle when the logger started. It uses the same format as the [Information Message](#i-information-message).
 
 ```c
 struct message_info_s {
@@ -257,6 +268,8 @@ If a parameter dynamically changes during runtime, this message can also be [use
 The data type is restricted to `int32_t` and `float`.
 
 #### 'Q': Default Parameter Message
+
+The default parameter message defines the default value of a parameter for a given vehicle and setup.
 
 ```c
 struct ulog_message_parameter_default_header_s {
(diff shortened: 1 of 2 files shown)


en/dev_log/ulog_file_format.md
```diff
@@ -23,14 +29,12 @@ They all correspond to the types in C:
 |double             |  8            |
 |bool, char         |  1            |
 
-Additionally all can be used as an array, eg. `float[5]`.
-In general all strings (`char[length]`) do not contain a `'\0'` at the end.
+Additionally the types can be used as an array : eg. `float[5]`. In case of logging of strings : (`char[length]`), it usually does not contain the termination NULL character `'\0'` at the end.
```
hamishwillee, line 32: Note again, why have all that space around the colon? `an array : eg.`.
The first sentence is an improvement, the second is not. I will fix.

en/dev_log/ulog_file_format.md
```diff
@@ -23,14 +29,12 @@ They all correspond to the types in C:
...
 String comparisons are case sensitive.
 
+## ULog structure
 
-## File structure
-
-The file consists of three sections:
+ULog consists of the following three sections:
```
hamishwillee, line 37: You're talking about a file here, so you wouldn't normally rename to "ulog", and if you do, you'd say "the ULog file format".

SO this would reasonably change to
```
The ULog file format consists of the following three sections:
```
or 
```
The file consists of the following three sections:
```

en/dev_log/ulog_file_format.md
```diff
@@ -41,6 +45,9 @@ The file consists of three sections:
 ----------------------
 ```
 
+Description on each section is explained below.
```
hamishwillee, line 48: Note, this says in English "below I will explain the description" not "below I will explain the section".

en/dev_log/ulog_file_format.md
```diff
@@ -41,233 +48,318 @@ The file consists of three sections:
...
 ```
 ----------------------------------------------------------------------
 | 0x55 0x4c 0x6f 0x67 0x01 0x12 0x35 | 0x01         | uint64_t       |
 | File magic (7B)                    | Version (1B) | Timestamp (8B) |
 ----------------------------------------------------------------------
 ```
 
-Version is the file format version, currently 1.
-Timestamp is a `uint64_t` integer, denotes the start of the logging in microseconds.
+- **File Magic (7 Bytes):** File type indicator that reads "ULog"
```
hamishwillee, line 65: Note the two characters at the end. THese don't render in gitbook, and Notepad shows them as: 
image

Is there another way to describe this - e.g. "..that reads "ULogXY, where XY is [whatever it is]"

en/dev_log/ulog_file_format.md
```diff
@@ -41,233 +48,318 @@ The file consists of three sections:
...
+- **Version (1 Byte):** File format version (currently 1)
+- **Timestamp (8 Bytes):** `uint64_t` integer that denotes when the logging started in microseconds.
 
+### Definition & Data Section Message Header
 
-### Definitions Section
+The _Definitions and Data_ section consists of a stream of **messages** that always starts with this header:
```
hamishwillee, line 71: Are you saying that:
- every message in both sections starts with this header
- before all the stream of messages for the definitions and data this header is displayed
- before the stream of messages for the definitions section this header is display and also it is displayed again before the stream for data?

This sentence will need to change based on which one of the above is true.

en/dev_log/ulog_file_format.md
junwoo091400: I thought ULog itself is a file format, therefore there is no need to state it again (even the document's name is `ulog_file_format.md`)

junwoo091400:
Regarding the format side of things, I think  it will be great to update the Style guide in the doc: https://docs.px4.io/master/en/contribute/docs.html#style-guide, so that we can have concrete examples showing how to do bulleted points, capitalization and so on :wink:

en/dev_log/ulog_file_format.md
junwoo091400: Ohhh yes I am Explaining the Description :rofl:, my english is :camel:

junwoo091400: Thank you for the catch!

en/dev_log/ulog_file_format.md
```diff
@@ -1,17 +1,23 @@
...
+::: note
+The [system logger](../modules/modules_system.md#logger) allows the default set of logged topics to be replaced with a custom set that is defined on an SD Card.
+See [Logging > SD Card Configuration](../dev_log/logging.html#sd-card-configuration) for more information.
+:::
 
-It can be used for logging device inputs (sensors, etc.), internal states (cpu load, attitude, etc.) and `printf` log messages.
+ULog commonly logs uORB topics related to (but not limited to) the following sources:
+- **Device inputs:** Sensors, RC input, etc.
+- **Internal states:** CPU load, attitude, EKF state, etc.
+- **Messages:** MAVLink communications, `PX4_INFO()`, Custom `printf` statements, and so on.
```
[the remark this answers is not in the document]
junwoo091400: Gotcha, will do that

en/dev_log/ulog_file_format.md
junwoo091400: > e.g. "..that reads "ULogXY, where XY is [whatever it is]"

I can say "ULogXYZ, where XYZ are the magic 3 bytes: 0x01 0x12 0x35"

en/dev_log/ulog_file_format.md
junwoo091400: > * every message in both sections starts with this header

This is the correct case. I think I can change it to:

The _Definitions and Data_ section consists of **messages**, and each message always starts with this header:

en/dev_log/ulog_file_format.md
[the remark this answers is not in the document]
junwoo091400: @bkueng Is it correct to say that Messages part (MAVLink message, PX4_INFO, etc) are also a uORB topic? Or how should we phrase this part?:

>  ULog commonly logs uORB topics related to (but not limited to) the following sources:

en/dev_log/ulog_file_format.md
```diff
@@ -29,12 +29,15 @@ The following binary types are used for logging. They all correspond to the type
 |double             |  8            |
 |bool, char         |  1            |
 
-Additionally the types can be used as an array : eg. `float[5]`. In case of logging of strings : (`char[length]`), it usually does not contain the termination NULL character `'\0'` at the end.
+Additionally the types can be used as an array: e.g. `float[5]`.
+
+Strings (`char[length]`) do not usually contain the termination NULL character `'\0'` at the end.
 String comparisons are case sensitive.
```
junwoo091400, line 35: @bkueng I am still wondering if this part is still relevant / necessary for the User Doc. Could we bring a contextual example that shows when 'case sensitie-ness' matters?

Is it used in Flight Review? ulog python library? etc.

en/dev_log/ulog_file_format.md
```diff
@@ -73,8 +77,8 @@ struct message_header_s {
 };
 ```
 
-* `msg_size` is the size of the message in bytes without the header
-* `msg_type` defines the content, and is a single alphabet character.
+- `msg_size` is the size of the message in bytes without the header.
+- `msg_type` defines the content, and is a single alphabet character.
 
 :::note
 Message types described below are noted with the single alphabet `msg_type` characteristic in the beginning.
```
junwoo091400, line 84: Is this clear enough for the readers to recognize that the: `"'B': Flag Bits Message"` indicates that the `msg_type` of that message corresponds to the character `'B'`?

Because this is one of the first things I was so confused about the document, as it wasn't so clear what all those 'B', 'f', 'I', 'M', etc. meant :thinking:

en/dev_log/ulog_file_format.md
hamishwillee: YOu could be right - not sure it is defined as such though. But would you say "Zip consists of" or "Mission plan consists of"?

en/dev_log/ulog_file_format.md
```diff
@@ -23,14 +29,12 @@ They all correspond to the types in C:
 |double             |  8            |
 |bool, char         |  1            |
 
-Additionally all can be used as an array, eg. `float[5]`.
-In general all strings (`char[length]`) do not contain a `'\0'` at the end.
+Additionally the types can be used as an array : eg. `float[5]`. In case of logging of strings : (`char[length]`), it usually does not contain the termination NULL character `'\0'` at the end.
 String comparisons are case sensitive.
 
+## ULog structure
```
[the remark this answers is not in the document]
hamishwillee: Ignore that, it makes no difference here.

All I was suggesting (now fixed) is `## ULog Structure`

en/dev_log/ulog_file_format.md
hamishwillee: That works.

en/dev_log/ulog_file_format.md
hamishwillee: Do you mean "starts with" or is "preceded by":
```suggestion
The _Definitions and Data_ sections contain a number of **messages**. Each message is preceded by this header: 
```

en/dev_log/ulog_file_format.md
hamishwillee: No. We need to make that clear right up the top of the doc somewhere. I had same question and forgot to raise it.

en/dev_log/ulog_file_format.md
```diff
@@ -41,233 +48,318 @@ The file consists of three sections:
...
+```
+struct ulog_message_flag_bits_s {
+  struct message_header_s;
+  uint8_t compat_flags[8];
+  uint8_t incompat_flags[8];
+  uint64_t appended_offsets[3]; ///< file offset(s) for appended data if appending bit is set
+};
+```
+
+- `compat_flags`: compatible flag bits.
```
junwoo091400, line 116: @bkueng Is there a better way to phrase this? Since I am considering explaining this part like:
> Compatible flags indicate if the log includes any special message that is compatible with the existing parsers is in the uLog

And the `incompat_flags` part like:
> Incompatible flags indicate whether any special message that is incompatible with the existing parsers is in the uLog

But I am not sure if I understand this part correctly (Also out of 64 bits, only 1 bit is being used for now :thinking:), could you give your input on this?

en/dev_log/ulog_file_format.md
```diff
@@ -41,233 +48,318 @@ The file consists of three sections:
...
+  This can be used to reliably append data for logs that may stop in the middle of a message.
+
+  A process appending data should do:
+  - set the relevant `incompat_flags` bit,
+  - set the first `appended_offsets` that is 0 to the length of the log file,
+  - then append any type of messages that are valid for the Data section.
+
+It is possible that there are more fields appended at the end of this message in future ULog specifications.
+This means a parser must not assume a fixed length of this message.
+If the message is longer than expected (currently 40 bytes), the exceeding bytes must just be ignored.
```
junwoo091400, line 138: This part is still very vague / hard to understand

en/dev_log/ulog_file_format.md
```diff
@@ -41,233 +48,318 @@ The file consists of three sections:
...
+
+`type` is one of the basic binary types or a `message_name` of another format definition (nested usage).
+A type can be used before it's defined.
+There can be arbitrary nesting but no circular dependencies.
+
+Some field names are special:
+- `timestamp`: every [Subscription Message](#a--subscription-message) must include a timestamp field
+  - Its type can be: `uint64_t` (currently the only one used), `uint32_t`, `uint16_t` or `uint8_t`.
+  - The unit is always microseconds, except for in `uint8_t` it's milliseconds.
+  - A log writer must make sure to log messages often enough to be able to detect wrap-arounds and a log reader must handle wrap-arounds (and take into account dropouts).
```
junwoo091400, line 165: @bkueng I don't understand this part. Could you explain?

en/dev_log/ulog_file_format.md
```diff
@@ -137,51 +142,57 @@ It is possible that there are more fields appended at the end of this message in
...
+  char key[255];
 };
 ```
-* `key` is a plain string, as in the format message (can also be a custom type), but consists of only a single field without ending `;`, eg. `float[3] myvalues`
-* `value` contains the data as described by `key`
+
+- `key_len`: Length of the key value
+- `key`: String with the `key` and `value` information combined: e.g. `char[5] sys_toolchain_ver9.4.0`
+  - The `key` part has the format identical as the field in the [Format message](#f-format-message) (can also be a custom type), but consists of only a single field without ending `;`, eg. `char[5] sys_toolchain_ver`
+  - The `value` contains the data corresponding to the `key`, and is written right after the `key`, e.g. `9.4.0`.
```
junwoo091400, line 195: It was very confusing to explain how the key string of the struct differs from the concept of key : value of the information we are encompassing inside the key string.

Maybe worth changing the naming in the PX4's side to clarify what this field actually does? @bkueng

en/dev_log/ulog_file_format.md
```diff
@@ -137,51 +142,57 @@ It is possible that there are more fields appended at the end of this message in
...
 #### 'I': Information Message
 
+The Information message defines a dictionary type definition `key` : `value` pair for any information, including but not limited to Hardware version, Software version, Build toolchain for the software, etc.
+
 ```c
-struct message_info_s {
-  struct message_header_s header;
+struct ulog_message_info_header_s {
+  uint16_t msg_size;
+  uint8_t msg_type = 'I';
```
junwoo091400, line 186: So actually, it turns out that currently in PX4 the struct definitions don't use the `message_header_s` struct in the beginning of each message struct's definition. So the documentation was wrong (out of date).

So I propose putting in the actual implementation like so (only edited 2 messages to be in this format), what do you think? @hamishwillee

en/dev_log/ulog_file_format.md
junwoo091400: I added a [comment](https://github.com/PX4/PX4-user_guide/pull/1887#discussion_r876820169) about this, where the proposed change will expose the `msg_type` character clearly in the struct definition!

junwoo091400:
Fixed the links between headers & updated struct formats to be C, converted * bullet points to - based

en/dev_log/ulog_file_format.md
hamishwillee: What it means is that the size of the timestamp isn't infinite, and it will rollover back to zero after a while. You can detect this is you log with some frequency, but if your log length is too infrequent you won't know if your latest item is actually at timestamp X or timestamp X + the interval for the whole timer to cycle once.

en/dev_log/ulog_file_format.md
junwoo091400: PR here : https://github.com/PX4/PX4-Autopilot/pull/19675

en/dev_log/ulog_file_format.md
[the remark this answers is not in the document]
bkueng: Hamish is correct, it's either the tagged or non-tagged string messages.
```
**String messages:** `printf` statements, including `PX4_INFO()` and `PX4_ERR()`.
```

en/dev_log/ulog_file_format.md
bkueng: It's relevant, don't just remove things. This doc is meant to be a semi-formal file format spec, that can be used to implement a file reader or writer. Please keep that in mind.

en/dev_log/ulog_file_format.md
```diff
@@ -41,233 +48,331 @@ The file consists of three sections:
...
+#### 'B': Flag Bits Message
+
+:::note
+This message must be the **first message** right after the header section, so that it has a fixed constant offset from the start of the file!
+:::
 
-The Definitions and Data sections consist of a stream of messages.
-Each starts with this header:
 ```c
-struct message_header_s {
```
bkueng, line 65: This was actually better for understanding, it makes things more clear and explicit. The fact that PX4 does it differently, isn't really relevant here.

en/dev_log/ulog_file_format.md
bkueng: > So the documentation was wrong (out of date).

See my comment above - this isn't meant to document what PX4 exactly does.

en/dev_log/ulog_file_format.md
```diff
@@ -41,233 +48,331 @@ The file consists of three sections:
...
-  	uint8_t is_continued; ///< can be used for arrays
-  	uint8_t key_len;
-  	char key[key_len];
-  	char value[header.msg_size-2-key_len]
-  };
-  ```
-  The same as the information message, except that there can be multiple messages with the same key (parsers store them as a list).
-  The `is_continued` can be used for split-up messages: if set to 1, it is part of the previous message with the same key.
-  Parsers can store all information multi messages as a 2D list, using the same order as the messages occur in the log.
+`char[3] sys_name` | Name of the system |  "PX4"
```
bkueng, line 207: I find this more confusing than before. The key is generic, but now it's tied to the example

en/dev_log/ulog_file_format.md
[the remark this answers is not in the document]
junwoo091400: Thank you!

en/dev_log/ulog_file_format.md
junwoo091400: Hmm yea that wouldn't make much sense 🤔

en/dev_log/ulog_file_format.md
junwoo091400: Gotcha 👌. One motive behind changing to this format was to expose the `msg_type` byte, and make it clear that each messages had their own type assigned 🤔. Will add it as a separate text!

en/dev_log/ulog_file_format.md
junwoo091400: I found the `value_len` part confusing when I first encountered the doc. Is there a better way to phrase it?

Since currently with the example, it is very clear that this is how you can use the key : value pair, but it's also not generic anymore. Maybe add more text above to state that it can be generic?

en/dev_log/ulog_file_format.md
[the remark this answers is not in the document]
hamishwillee: @bkueng I've been playing with a new MAVLink battery message for PX4 dev summit. The updated uorb topic shows up in the log, but I'm not seeing any mavlink messages. Do these have to be explicitly enabled? If so, how?

en/dev_log/ulog_file_format.md
hamishwillee: Yes, that is odd wording. @bkueng Is this what is meant? 
```suggestion
It is possible that there are more fields appended at the end of this message in future ULog specifications.
This means a parser must not assume a fixed length of this message.
If the message is longer than expected (currently 40 bytes), any additional bytes must be ignored/discarded.
```

en/dev_log/ulog_file_format.md
hamishwillee: @bkueng As an ordinary reader I did look at the introduction and assume that this document reflects what PX4 exactly does. Mostly because I assume this is a custom **PX4** file format definition. Not something abstract or from some external source.

We need to capture this in the introduction - ideally with some positioning for what this doc is/is not useful for in a PX4 context.

If it doesn't reflect PX4, where did this come from? And what use is the doc - just to get a general feel for the fields you might see?

en/dev_log/ulog_file_format.md
hamishwillee: @bkueng I think you are saying that the values @junwoo091400 has suggested are length values specific to PX4 implementation, and that some other flight stack using this spec might use different numbers. Is that right?

There are two possible fixes. 
1. We use `value_len` but we have a note at the bottom stating (something like):
   > `value_len` represents the data size required for storing the key. This will be key and system specific.
2. Use @junwoo091400 values, but at the end have a note:
  > The values for the value data-size in the table above are those used by PX4. Other systems may use other values.
  
I tend to think option 1 is better for a "generic spec" - because with option 2 people are more likely just to see the values and assume they can be hard coded (i.e. not read the spec properly). So I'd revert and use option "1" (assuming Beat agrees)

en/dev_log/ulog_file_format.md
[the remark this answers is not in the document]
bkueng: What do you mean by mavlink messages do not show up? Where do you expect them?

en/dev_log/ulog_file_format.md
bkueng: Yes

en/dev_log/ulog_file_format.md
bkueng: Yes the file format is not tied to PX4. It could be used by any project with similar requirements.

PX4 follows the file format precisely, but it may use different struct namings, definitions, etc. Things that are not relevant for the on-disk file format.

en/dev_log/ulog_file_format.md
```diff
@@ -2,16 +2,22 @@
 
 ULog is the file format used for logging system data.
 
-The format is self-describing, i.e. it contains the format and message types that are logged (note that the [system logger](../dev_log/logging.md) allows the *default set* of logged topics to be replaced from an SD card).
+ULog is **self-describing** : It contains the format and message types that are logged.
 
-It can be used for logging device inputs (sensors, etc.), internal states (cpu load, attitude, etc.) and `printf` log messages.
+::: note
+The [logger](../dev_log/logging.md#sd-card-configuration) allows the specifying a custom *default set* of logged topics with the `logger_topics.txt` file
```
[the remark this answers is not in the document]
junwoo091400: Thanks! I checked the change

junwoo091400:
I want to bring it to attention that `compat_flags` and `incompat_flags` are still so vague. And it really doesn't make much sense to me (it's abstract). Could you explain this a bit more so that users can understand it contextually? @bkueng 

My previous comment here: https://github.com/PX4/PX4-user_guide/pull/1887#discussion_r876753074

This [fix](https://github.com/PX4/PX4-user_guide/pull/1887#discussion_r881087705) wasn't what I meant by 'confusing' fyi, I want to explain the whole `compat` / `incompat` more clearly in general. @hamishwillee

en/dev_log/ulog_file_format.md
junwoo091400: Added to the doc!

junwoo091400:
Changes were applied, and we still have the following questions remaining:
1. Do we show the `header` part or, the stripped down `msg_size and msg_header` part on the message struct definitions? -> https://github.com/PX4/PX4-user_guide/pull/1887#discussion_r876820169
2. How do we explain the `compat_flags` and `incompat_flags` more clearly? -> https://github.com/PX4/PX4-user_guide/pull/1887#issuecomment-1137202556
3. How do we expose the fact that `msg_type` corresponds to the characters prefixed on each message type definition sections? -> https://github.com/PX4/PX4-user_guide/pull/1887#discussion_r876727481

en/dev_log/ulog_file_format.md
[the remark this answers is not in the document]
hamishwillee: I was confused. I was assuming there might be magic that automatically logs all mavlink messages. But that is not what this says - it logs uorb topics.

en/dev_log/ulog_file_format.md
```diff
@@ -41,233 +51,332 @@ The file consists of three sections:
...
+  uint8_t msg_type = 'B';
+  uint8_t compat_flags[8];
+  uint8_t incompat_flags[8];
+  uint64_t appended_offsets[3]; ///< file offset(s) for appended data if appending bit is set
 };
 ```
 
-`msg_size` is the size of the message in bytes without the header (`hdr_size`= 3 bytes).
-`msg_type` defines the content and is one of the following:
+- `compat_flags`: compatible flag bits
```
hamishwillee, line 122: > How do we explain the compat_flags and incompat_flags more clearly? -> https://github.com/PX4/PX4-user_guide/pull/1887#issuecomment-1137202556

I don't find this confusing, perhaps because the concept is also used in mavlink.

Essentially there are two types of changes you might make to protocol or log or whatever - compatible or incompatible. 

An incompatible change is one that makes the content usuable - you simply have to throw away everything in the scope of the change.
- In mavlink this might be something "like" compression of a message - if a parser were to see a compression incompatibility bit set and it didn't know what that meant it would have to drop the message
- in the log this means you have to discard the whole log if you don't implement the feature indicated by an incompatible bit. This might mean the indicated extension is essential to using the log. I "guess" it could also be like the mavlink case, which means it makes the log unparsable. 

A compatible change is one that a parser can just ignore - it should continue to parse as normal - the output might not include some information that is in the log, but that isn't essential, and the parser can just go it's merry way.
For mavlink this is something like a bit that indicates a particular message is high priority. If a parser doesn't implement high priority parsing and the bit is appended to the end so that parsers that don't understand it won't even see it - then it is compatible. For the log this is just "default parameters message" - you can use a log even if you just ignore this message.

Anyway, leave to you to decide if  more needs to be said.



What this says is that if any of the `compat_flags` flags are set a parser that d

hamishwillee:
@junwoo091400 Leaving for you and Beat to discuss 1, 3 here: https://github.com/PX4/PX4-user_guide/pull/1887#issuecomment-1137211226

I have commented on 2 https://github.com/PX4/PX4-user_guide/pull/1887#discussion_r882256016

hamishwillee:
Thanks you. I am broadly happy that adding some structure/headings is a good thing, and this looks good to me. But I see the owner as Beat.

@bkueng Are you happy for this to be merged?

en/dev_log/ulog_file_format.md
```diff
@@ -41,233 +50,328 @@ The file consists of three sections:
...
+  struct message_header_s header; // msg_type = 'A'
+  uint8_t multi_id;
+  uint16_t msg_id;
+  char message_name[header.msg_size-3];
+};
+```
+
+- `multi_id`: the same message format can have multiple instances, for example if the system has two sensors of the same type. The default and first instance must be 0.
+- `msg_id`: unique id to match [Logged data Message](#d-logged-data-message) data. The first use must set this to 0, then increase it.
+  - The same `msg_id` must not be used twice for different subscriptions, not even after unsubscribing.
```
bkueng, line 330: I don't see why it should not be possible to re-use after unsubscribing.

en/dev_log/ulog_file_format.md
```diff
@@ -1,17 +1,22 @@
 # ULog File Format
 
-ULog is the file format used for logging system data.
+ULog is the file format used for logging system data. The format is self-describing, i.e. it contains the format and [uORB](../middleware/uorb.md) message types that are logged.
 
-The format is self-describing, i.e. it contains the format and message types that are logged (note that the [system logger](../dev_log/logging.md) allows the *default set* of logged topics to be replaced from an SD card).
+::: note
+The [system logger](../modules/modules_system.md#logger) allows the default set of logged topics to be replaced with a custom set that is defined on an SD Card.
+See [Logging > SD Card Configuration](../dev_log/logging#sd-card-configuration) for more information.
```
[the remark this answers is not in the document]
junwoo091400: Removed! And yes it's irrelevant for the ULog File Format Spec!

en/dev_log/ulog_file_format.md
```diff
@@ -1,17 +1,22 @@
...
+:::
 
-It can be used for logging device inputs (sensors, etc.), internal states (cpu load, attitude, etc.) and `printf` log messages.
+ULog commonly logs uORB topics related to (but not limited to) the following sources:
```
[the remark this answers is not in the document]
junwoo091400: Changes applied!

en/dev_log/ulog_file_format.md
```diff
@@ -41,233 +50,328 @@ The file consists of three sections:
...
+  uint8_t compat_flags[8];
+  uint8_t incompat_flags[8];
+  uint64_t appended_offsets[3]; // file offset(s) for appended data if appending bit is set
 };
 ```
 
-`msg_size` is the size of the message in bytes without the header (`hdr_size`= 3 bytes).
-`msg_type` defines the content and is one of the following:
+- `compat_flags`: compatible flag bits
+  - These flags indicate any information that is compatible with the existing ULog is in the log.
```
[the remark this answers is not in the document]
junwoo091400: It was explained a bit by Hamish here: https://github.com/PX4/PX4-user_guide/pull/1887#discussion_r882256016, but I am still confused as well :shrug:. Any contextual example would be much more helpful!

en/dev_log/ulog_file_format.md
```diff
@@ -41,233 +50,328 @@ The file consists of three sections:
...
+  - e.g. The message `MessageA:MessageB[2] msg_b` can come before the `MessageB:uint_8[3] data`
+- There can be arbitrary nesting but **no circular dependencies**
+  - e.g. `MessageA:MessageB[2] msg_b` & `MessageB:MessageA[4] msg_a`
+
+Some field names are special:
+- `timestamp`: every [Subscription Message](#a-subscription-message) must include a timestamp field
+  - Its type can be: `uint64_t` (currently the only one used), `uint32_t`, `uint16_t` or `uint8_t`.
+  - The unit is always microseconds, except for in `uint8_t` where the unit is in milliseconds.
+  - A log writer must make sure to log messages often enough to be able to detect **wrap-arounds** (when the timestamp overflows the data type and goes back to 0)
+  - Also, the log reader must handle wrap-arounds (and take into account dropouts).
```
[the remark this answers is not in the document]
junwoo091400: I added the 'optional' part, but the dropouts part is quite unclear to me still :shrug: (not sure how it works)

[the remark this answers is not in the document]
junwoo091400: I edited it to make it more contextual :thinking:

en/dev_log/ulog_file_format.md
```diff
@@ -41,233 +50,328 @@ The file consists of three sections:
...
+#### 'I': Information Message
+
+The Information message defines a dictionary type definition `key` : `value` pair for any information, including but not limited to Hardware version, Software version, Build toolchain for the software, etc.
+
+```c
+struct ulog_message_info_header_s {
+  struct message_header_s header; // msg_type = 'I'
+  uint8_t key_len;
+  char key[key_len];
+  char value[header.msg_size-2-key_len]
```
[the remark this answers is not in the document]
junwoo091400: It was an mistake!

en/dev_log/ulog_file_format.md
junwoo091400: Fixed!

junwoo091400:
Applied all the comments, the new commit is pushed :wink:

en/dev_log/ulog_file_format.md
```diff
@@ -118,34 +114,34 @@ struct ulog_message_flag_bits_s {
...
-  This can be used to reliably append data for logs that may stop in the middle of a message.
+  This can be used to reliably append data for logs that may stop in the middle of a message. For example, hardfault dumps in PX4.
 
   A process appending data should do:
-  - set the relevant `incompat_flags` bit,
-  - set the first `appended_offsets` that is 0 to the length of the log file,
-  - then append any type of messages that are valid for the Data section.
+  - set the relevant `incompat_flags` bit
+  - set the first `appended_offsets` as the length of the log file without the appended data, as that is where the data would start
+  - append any type of messages that are valid for the Data section.
```
junwoo091400, line 136: I am still unsure why it's only the 'first' element of the array `appended_offsets` that is used. Why do we have 3 of them as an array then?

Also, I would like to check if the phrase "length of the log file without the appended data, as that is where the data would start" is correct.

In PX4, it seems that the `incompat_flags` isn't set anywhere, except for having a check in reply module for replaying the log. So where is this `incompat_flags` actually set? In PX4? Or external source? @bkueng

en/dev_log/ulog_file_format.md
junwoo091400: Thank you for the explanation! It was applied in the latest commit

junwoo091400:
Rebased on `main` branch and the confusion in `appended_offsets` was discussed with @bkueng. Should be good now 🙌

hamishwillee:
@bkueng Can you please merge this if you are happy with it?

en/dev_log/ulog_file_format.md
```diff
@@ -118,34 +114,34 @@ struct ulog_message_flag_bits_s {
...
   If no data is appended, all offsets must be zero.
-  This can be used to reliably append data for logs that may stop in the middle of a message.
+  This can be used to reliably append data for logs that may stop in the middle of a message. For example, hardfault dumps in PX4.
 
   A process appending data should do:
-  - set the relevant `incompat_flags` bit,
-  - set the first `appended_offsets` that is 0 to the length of the log file,
-  - then append any type of messages that are valid for the Data section.
+  - set the relevant `incompat_flags` bit
+  - set the first `appended_offsets` as the length of the log file without the appended data, as that is where the data would start
```
bkueng, line 135: `set the first `appended_offsets` that is currently 0 to the length of the log file without the appended data, as that is where the new data will start`

en/dev_log/ulog_file_format.md
```diff
@@ -118,34 +114,34 @@ struct ulog_message_flag_bits_s {
...
 
 It is possible that there are more fields appended at the end of this message in future ULog specifications.
 This means a parser must not assume a fixed length of this message.
-If the message is longer than expected (currently 40 bytes), any additional bytes must be ignored/discarded.
+If the `msg_size` is bigger than expected (currently 40), any additional bytes must be ignored/discarded.
 
 #### 'F': Format Message
 
-Format message defines a single uORB topic's name and it's inner fields in a single string.
+Format message defines a singe message name and it's inner fields in a single string.
```
bkueng, line 144: `single`

en/dev_log/ulog_file_format.md
```diff
@@ -205,29 +201,28 @@ Predefined information messages are:
...
 `int32_t time_ref_utc` | UTC Time offset in seconds | -3600 |
 
 :::note
-The `key_len` value in `char key[key_len]` has no fixed value.
-Therefore the key lengths in the example above (3, 9, 2, 7, ...) may not be exactly same with your ULog's information message.
+`value_len` represents the data size of the `value`. This is described in the `key`.
 :::
 
-- The format of `ver_sw_release` and `ver_os_release` is: 0xAABBCCTT, where AA is **major**, BB is **minor**, CC is patch and TT is the **type**.
+- For PX4, the format of `ver_sw_release` and `ver_os_release` is: 0xAABBCCTT, where AA is **major**, BB is **minor**, CC is patch and TT is the **type**.
```
bkueng, line 225: This does not only apply to PX4, please keep as it was

en/dev_log/ulog_file_format.md
```diff
@@ -243,7 +238,7 @@ struct ulog_message_info_multiple_header_s {
   uint8_t is_continued; // can be used for arrays
   uint8_t key_len;
   char key[key_len];
-  char value[header.msg_size-2-key_len]
+  char value[header.msg_size-key_len]
```
bkueng, line 241: This was correct with -2

en/dev_log/ulog_file_format.md
```diff
@@ -41,233 +46,327 @@ The file consists of three sections:
...
+6. [Default Parameter](#q-default-parameter-message)
+
+
+#### 'B': Flag Bits Message
+
+:::note
+This message must be the **first message** right after the header section, so that it has a fixed constant offset from the start of the file!
+:::
+
+This message provides an information to the log parser whether the log is parsable or not.
```
bkueng, line 105:
```suggestion
This message provides information to the log parser whether the log is parsable or not.
```

en/dev_log/ulog_file_format.md
```diff
@@ -1,16 +1,12 @@
 # ULog File Format
 
-ULog is the file format used for logging system data. The format is self-describing, i.e. it contains the format and [uORB](../middleware/uorb.md) message types that are logged.
+ULog (Universal-Log) is the file format used for logging messages. The format is self-describing, i.e. it contains the format and [uORB](../middleware/uorb.md) message types that are logged.
+This document is meant to be the ULog File Format Spec Documentation, especially for anyone who is interested in developing a ULog Parser / Application and needs to decode ULog file.
```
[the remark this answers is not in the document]
hamishwillee: FWIW Just as easy to do these kinds of things as suggestions.
```suggestion
This document is meant to be the ULog File Format Spec Documentation.
It is intended especially for anyone who is interested in writing a ULog parser / serializer and needs to decode / encode files.
```

en/dev_log/ulog_file_format.md
```diff
@@ -118,34 +114,34 @@ struct ulog_message_flag_bits_s {
 ```
 
 - `compat_flags`: compatible flag bits
-  - These flags indicate any information that is compatible with the existing ULog is in the log.
+  - These flags indicate whether any information that is compatible with the ULog file format is in the log.
```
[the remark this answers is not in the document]
hamishwillee:
```suggestion
  - These flags indicate the presence of features in the log file that are compatible with any ULog parser.
```

en/dev_log/ulog_file_format.md
```diff
@@ -118,34 +114,34 @@ struct ulog_message_flag_bits_s {
...
   - `compat_flags[0]`: *DEFAULT_PARAMETERS* (Bit 0): if set, the log contains [default parameters message](#q-default-parameter-message)
 
   The rest of the bits are currently not defined and must be set to 0.
-  These bits can be used for future ULog changes that are compatible with existing parsers.
+  These bits can be used for future ULog changes that are compatible with existing parsers. For example, adding a new message types, which can be indicated by setting one of the Bits, and the existing parsers can just ignore the new message type in the log.
```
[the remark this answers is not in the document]
hamishwillee:
```suggestion
  These bits can be used for future ULog changes that are compatible with existing parsers.
  For example, adding a new message type can be indicated by defining a new bit in the standard, and existing parsers will ignore the new message type.
```

en/dev_log/ulog_file_format.md
```diff
@@ -118,34 +114,34 @@ struct ulog_message_flag_bits_s {
...
-  If a parser finds one of these bits set, it must refuse to parse the log.
-  This can be used to introduce breaking changes that existing parsers cannot handle.
-- `appended_offsets`: File offsets (0-based) for appended data.
+  The rest of the bits are currently not defined and must be set to 0.
+  This can be used to introduce breaking changes that existing parsers cannot handle. For example, when an old ULog Parser that didn't have the concept of *DATA_APPENDED* reads the newer ULog, it would stop parsing the log as the log will contain out-of-spec messages / concepts.
+  If a parser finds any of these bits set that isn't specified, it must refuse to parse the log.
+- `appended_offsets`: File offset for appended data.
   If no data is appended, all offsets must be zero.
-  This can be used to reliably append data for logs that may stop in the middle of a message.
+  This can be used to reliably append data for logs that may stop in the middle of a message. For example, hardfault dumps in PX4.
```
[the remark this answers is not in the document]
hamishwillee:
```suggestion
  This can be used to reliably append data for logs that may stop in the middle of a message.
  For example, crash dumps.
```

en/dev_log/ulog_file_format.md
hamishwillee:
```suggestion
Format message defines a single message name and it's inner fields in a single string.
```

en/dev_log/ulog_file_format.md
hamishwillee:
```suggestion
  - set the first `appended_offsets` that is currently 0 to the length of the log file without the appended data, as that is where the new data will start
```

en/dev_log/ulog_file_format.md
hamishwillee: Revert
```suggestion
- The format of `ver_sw_release` and `ver_os_release` is: 0xAABBCCTT, where AA is **major**, BB is **minor**, CC is patch and TT is the **type**.
```

en/dev_log/ulog_file_format.md
hamishwillee: Revert
```suggestion
  char value[header.msg_size-2-key_len]
```

hamishwillee:
@bkueng I have accepted all your suggestions and comments (those you had as comments, I integrated as suggestions). This includes the cases where you said "revert".

@junwoo091400 the only remaining case is https://github.com/PX4/PX4-user_guide/pull/1887#discussion_r912932165 - "Can you keep (0-based)?".

Please fix that @junwoo091400 ASAP and then let me know. IF that is fixed and unless @bkueng has objections I plan to merge this TOMORROW. @junwoo091400 that doesn't mean you can't continue work on this, but I would prefer that done as a separate PR. This will then go into the v1.13 release I hope to make tomorrow.

en/dev_log/ulog_file_format.md
```diff
@@ -1,17 +1,19 @@
 # ULog File Format
 
-ULog is the file format used for logging system data.
+ULog (Universal-Log) is the file format used for logging messages. The format is self-describing, i.e. it contains the format and [uORB](../middleware/uorb.md) message types that are logged.
```
bkueng, line 3: @junwoo091400 as I told you I'm not sure if that's the exact origin
```suggestion
ULog is the file format used for logging messages. The format is self-describing, i.e. it contains the format and [uORB](../middleware/uorb.md) message types that are logged.
```

en/dev_log/ulog_file_format.md
```diff
@@ -41,233 +47,329 @@ The file consists of three sections:
...
+  These bits can be used for future ULog changes that are compatible with existing parsers.
+  For example, adding a new message type can be indicated by defining a new bit in the standard, and existing parsers will ignore the new message type.
+  It means parsers can just ignore the bits if one of the unknown bits is set.
+- `incompat_flags`: incompatible flag bits.
+  - `incompat_flags[0]`: *DATA_APPENDED* (Bit 0): if set, the log contains appended data and at least one of the `appended_offsets` is non-zero.
+  
+  The rest of the bits are currently not defined and must be set to 0.
+  This can be used to introduce breaking changes that existing parsers cannot handle. For example, when an old ULog Parser that didn't have the concept of *DATA_APPENDED* reads the newer ULog, it would stop parsing the log as the log will contain out-of-spec messages / concepts.
+  If a parser finds any of these bits set that isn't specified, it must refuse to parse the log.
+- `appended_offsets`: File offset for appended data.
```
bkueng, line 131:
```suggestion
- `appended_offsets`: File offset (0-based) for appended data.
```

en/dev_log/ulog_file_format.md
```diff
@@ -41,233 +47,329 @@ The file consists of three sections:
...
+
+  The rest of the bits are currently not defined and must be set to 0.
+  These bits can be used for future ULog changes that are compatible with existing parsers.
+  For example, adding a new message type can be indicated by defining a new bit in the standard, and existing parsers will ignore the new message type.
+  It means parsers can just ignore the bits if one of the unknown bits is set.
+- `incompat_flags`: incompatible flag bits.
+  - `incompat_flags[0]`: *DATA_APPENDED* (Bit 0): if set, the log contains appended data and at least one of the `appended_offsets` is non-zero.
+  
+  The rest of the bits are currently not defined and must be set to 0.
+  This can be used to introduce breaking changes that existing parsers cannot handle. For example, when an old ULog Parser that didn't have the concept of *DATA_APPENDED* reads the newer ULog, it would stop parsing the log as the log will contain out-of-spec messages / concepts.
```
bkueng, line 129:
```suggestion
  This can be used to introduce breaking changes that existing parsers cannot handle. For example, when an old ULog parser that didn't have the concept of *DATA_APPENDED* reads the newer ULog, it would stop parsing the log as the log will contain out-of-spec messages / concepts.
```

en/dev_log/ulog_file_format.md
```diff
@@ -41,233 +47,329 @@ The file consists of three sections:
...
+
+- The format of `ver_sw_release` and `ver_os_release` is: 0xAABBCCTT, where AA is **major**, BB is **minor**, CC is patch and TT is the **type**.
+  - **Type** is defined as following: `>= 0`: development, `>= 64`: alpha version, `>= 128`: beta version, `>= 192`: RC version, `== 255`: release version.
+  - For example, `0x010402FF` translates into the release version v1.4.2.
+
+This message can also be used in the Data section (this is however the preferred section).
+
+#### 'M': Multi Information Message
+
+Multi information message serves the same purpose as the information message, but for cases where the `value` length exceeds  (parsers store them as a list).
```
bkueng, line 236:
```suggestion
Multi information message serves the same purpose as the information message, but for long messages or multiple messages with the same key.
```

en/dev_log/ulog_file_format.md
```diff
@@ -322,50 +426,55 @@ The following messages belong to this section:
...
+
+* `sync_magic`: [0x2F, 0x73, 0x13, 0x20, 0x25, 0x0C, 0xBB, 0x12]
 
-- 'M': information message multi.
-  See above.
+#### 'O': Dropout message
 
-- 'P': parameter message.
-  See above.
+Mark a dropout (lost logging messages) of a given duration in ms
```
bkueng, line 444:
```suggestion
Mark a dropout (lost logging messages) of a given duration in ms.
```

hamishwillee:
Thanks. Looks good!

PR #1906, opened 2022-06-13
Add note to enable GPS Blending on ARK RTK GPS

AlexKlimaj:
With moving baseline on two cannodes, the flight controller can't ensure the rover is always GPS1 in uorb based on when they boot. If the rover comes up as instance 2, the heading is not used unless GPS Blending is enabled. With GPS blending enabled, the heading is always published, and the best receiver is used for gps position. The moving base produces a more accurate position, while the rover produces the heading.

Rover came up as GPS1
https://review.px4.io/plot_app?log=4bc33dda-df8c-4b4d-9814-d92eb397a384

Rover came up as GPS2
https://review.px4.io/plot_app?log=0a591982-8ee9-4386-9c9f-121bad7f4b1d


hamishwillee:
The only thoughts that went through my mind when reading this was:
- This was related to moving base so should it go in the moving base section rather than PX4 configuration.
- Does setting all the bits have any side effects? i.e. is there any reason not to set them or concerns.

@AlexKlimaj But since I'm sure you know what you're doing so I'm going to merge this. Thank you.

PR #1923, opened 2022-07-05
Control allocator page

bkueng:
@hamishwillee this adds the diagrams from our dev-summit presentation to a new control allocator page. This will later replace the Mixing page.

I know it's minimal and there's more that can be said. Let me know where you'd like to see more.


en/concept/control_allocation.md
```diff
@@ -0,0 +1,54 @@
...
+:::note
+How to setup mixing and outputs is documented [here](../config/actuators.md).
+:::
+
+
+## Actuator Control Pipeline
+
+Overview of the mixing pipeline in terms of modules and uORB topics:
+<!-- https://drive.google.com/file/d/1L2IoxsyB4GAWE-s82R_x42mVXW_IDlHP/view?usp=sharing -->
+<img alt="Pipeline Overview" src="../../assets/concepts/control_allocation_pipeline.png" width="1200" style="text-align: center; margin-left: -230px; margin-right: -230px; max-width: 200%"></img>
```
hamishwillee, line 25: The other image doesn't scale properly because it is fixed width. If we do it this way then the user can click the image to get it full screen.
```suggestion
![Pipeline Overview](../../assets/concepts/control_allocation_pipeline.png)
```

en/concept/control_allocation.md
```diff
@@ -0,0 +1,54 @@
...
+![Pipeline Overview](../../assets/concepts/control_allocation_pipeline.png)
+
+Notes:
+- The rate controller outputs torque and thrust setpoints
+- the `control_allocator` module:
+  - handles different geometries based on configuration
+  - does the mixing
+  - handles motor failures
+  - publishes the motor and servo control signals
+  - publishes the servo trims separately so they can be added to the actuator testing
```
hamishwillee, line 34: "so they can be added to the actuator testing" doesn't mean anything to me.

en/concept/control_allocation.md
```diff
@@ -0,0 +1,70 @@
+# Control Allocation (Mixing)
+
+The PX4 architecture ensures that the airframe layout does not require special case handling in the core controllers.
+
+Mixing means to take force commands (e.g. `turn right`) and translate them to actuator commands which control motors or servos.
+For a plane with one servo per aileron this means to command one of them high and the other low.
+The same applies for multicopters: pitching forward requires changing the speed of all motors.
+
+Separating the mixer logic from the actual attitude/rate controller greatly improves reusability.
```
hamishwillee, line 9: @bkueng This is a little confusing to me (I understand it, but find it hard to read). I have attempted to re-write and further capture that we also abstract the output function mapping.

```suggestion
PX4 takes desired torque and thrust commands from the core controllers and translates them to actuator commands which control motors or servos.

The translation depends on the physical geometry of the airframe.
For example, given a torque command to "turn right" (say):
- A plane with one servo per aileron will command one of servo high and the other low.
- A multicopter will yaw right by changing the speed of all motors.

PX4 separates this translation logic, which is referred to as "mixing" from the attitude/rate controller.
This ensures that the core controllers do not require special handling for each airframe geometry, and greatly improves reusability.

In addition, PX4 abstracts the mapping of output functions to specific hardware outputs.
This means that any motor or servo can be assigned to almost any controller output.  
```

en/concept/control_allocation.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Adding a new Geometry or Output Function
+
+See [this commit](https://github.com/PX4/PX4-Autopilot/commit/5cdb6fbd8e1352dcb94bd58918da405f8ff930d7) for how to add a new geometry.
+The QGC UI will then automatically show the right configuration UI when `CA_AIRFRAME` is set to the new geometry.
+
+[This commit](https://github.com/PX4/PX4-Autopilot/commit/a65533b46986e32254b64b7c92469afb8178e370) shows how to add a new output function.
+Any uORB topic can be subscribed and assigned to a function.
+
+## Setting the Default Airframe Geometry
```
hamishwillee, line 45: So I wanted to capture how this is enabled and how you set it for an airframe.

en/concept/geometry_files.md
```diff
@@ -1,5 +1,10 @@
 # Multicopter Geometry Files
 
+:::note
```
hamishwillee, line 3: Just a note/reminder that this is only relevant for v1.13 and earlier.

en/dev_airframes/adding_a_new_frame.md
```diff
@@ -2,7 +2,6 @@
 
 PX4 uses canned airframe configurations as starting point for airframes.
 The configurations are defined in [config files](#config-file) that are stored in the [ROMFS/px4fmu_common/init.d](https://github.com/PX4/PX4-Autopilot/tree/main/ROMFS/px4fmu_common/init.d) folder.
-The config files reference [mixer files](#mixer-file) that describe the physical configuration of the system, and which are stored in the [ROMFS/px4fmu_common/mixers](https://github.com/PX4/PX4-Autopilot/tree/main/ROMFS/px4fmu_common/mixers) folder.
```
hamishwillee, line 5: @bkueng This is minimal changes to adding a new frame to explain what you do. 

I'd have gone further except the wing-wing airframe does not use control allocation. To make an example of control allocation in place I'd have to have changed to another airframe, which would have made the rest of the mixer docs "off".

Anyway, if wing wing gets CA params then I would update this. Otherwise when we move to v1.14 this can all be deleted and we'll move to a different frame.

hamishwillee: Right now it just points to the control allocation doc for info about how things might work. I "think" that is enough.

en/concept/control_allocation.md
```diff
@@ -0,0 +1,70 @@
...
+- the output drivers:
+  - handle the hardware initialization and update
+  - use a shared library [src/libs/mixer_module](https://github.com/PX4/PX4-Autopilot/blob/main/src/lib/mixer_module/).
+    The driver defines a parameter prefix, e.g. `PWM_MAIN` that the library then uses for configuration.
+    Its main task is to select from the input topics and assign the right data to the outputs based on the user set `<param_prefix>_FUNCx` parameter values.
+    For example if `PWM_MAIN_FUNC3` is set to **Motor 2**, the 3rd output is set to the 2nd motor from `actuator_motors`.
+  - output functions are defined under [src/lib/mixer_module/output_functions.yaml](https://github.com/PX4/PX4-Autopilot/blob/main/src/lib/mixer_module/output_functions.yaml).
+- if you want to control an output from MAVLink, set the relevant output function to **Offboard Actuator Set x**, and then send the [MAV_CMD_DO_SET_ACTUATOR](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_ACTUATOR) MAVLink command.
+
+## Adding a new Geometry or Output Function
```
hamishwillee, line 37: @bkueng Assuming you are happy with MY changes I think we should merge this and iterate.

The specific part which needs more work (IMO) is this section - as I'm sure you are aware.

I am not sure this is sufficient for someone else to understand what they do for their new frame from this - say a "chinook". So I'd pull out some of the things from that doc which show the "have to do" vs the "specific case".

I really don't know what bits are most relevant, but perhaps an explanation of what you need in the YAML - e.g. https://github.com/PX4/PX4-Autopilot/commit/5cdb6fbd8e1352dcb94bd58918da405f8ff930d7#diff-c89677865693eb56c0125d5b99b9cac8b7c8bf774f51260fac6c4ce22f7a2137R382

Or we can get someone else to upgrade one of the frames and ask them to work out what they think is relevant?

bkueng: Yes there's a lot that could be said, but I'm not sure what people actually need. We can also see when someone goes through this the first time.
Adding output functions was already done by several people successfully.

This probably helps as a reference for the config: https://github.com/PX4/PX4-Autopilot/blob/main/validation/module_schema.yaml#L440=

en/concept/control_allocation.md
bkueng: Makes sense

en/concept/control_allocation.md
bkueng: This is copied from the existing page. Yes yours is better.

`can be assigned to almost any controller output` -> `can be assigned to almost any physical output`?
`The whole process is known as "control allocation"` -> I guess it can be interpreted in different ways, we typically use it as synonym for "mixing" (so w/o the physical output assignment). But then again we also use it to refer to the whole new architecture, which technically isn't accurate.

hamishwillee: OK, I have removed that last point "The whole process is known as "control allocation". It just seems odd to have two terms - then define mixing but not "control allocation".

Also took your first suggestion.

en/concept/control_allocation.md
hamishwillee: Let's say that then.
```suggestion
  - publishes the servo trims separately so they can be added as an offset when [testing actuators](../config/actuators.md#actuator-testing) (using the test sliders).
```

en/concept/control_allocation.md
```diff
@@ -0,0 +1,75 @@
...
+
+See [this commit](https://github.com/PX4/PX4-Autopilot/commit/5cdb6fbd8e1352dcb94bd58918da405f8ff930d7) for how to add a new geometry.
+The QGC UI will then automatically show the right configuration UI when `CA_AIRFRAME` is set to the new geometry.
+
+[This commit](https://github.com/PX4/PX4-Autopilot/commit/a65533b46986e32254b64b7c92469afb8178e370) shows how to add a new output function.
+Any uORB topic can be subscribed and assigned to a function.
+
```
hamishwillee, line 49: How about this reference for the yaml validation?

```suggestion

Note that parameters for control allocation are defined in [src/modules/control_allocator/module.yaml](https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/control_allocator/module.yaml)
The schema for this file is [here](https://github.com/PX4/PX4-Autopilot/blob/main/validation/module_schema.yaml#L440=) (in particular, search for the key `mixer:`

```

en/concept/control_allocation.md
hamishwillee: Fair enough. I can't judge. I added https://github.com/PX4/PX4-user_guide/pull/1923/files#r914562653 to include this reference - if you're happy with that then lets merge.

bkueng:
Thanks @hamishwillee. Good to merge from my side.

hamishwillee:
@bkueng Thanks very much. BTW, can you explain what needs to happen for http://docs.px4.io/master/en/payloads/ ?

I THINK that we need to map (up to ?) 3 outputs using "Offboard Actuator Set x" and then everything else should work the same as currently documented.
1. Are we still limited to 3?
2. Are any mapped by default? I.e. most frames prior to control allocation did map the first three AUX.

bkueng:
> I THINK that we need to map (up to ?) 3 outputs using "Offboard Actuator Set x" and then everything else should work the same as currently documented.

Yes, and you can use up to 6.

RC aux can be mapped as well, there's no default.

hamishwillee:
Thanks! I'll look at this next week.

PR #1937, opened 2022-07-08
Update on Multicopter combined position velocity diagram + minimal description 

heappy:
1. Update diagram in assets with following points
- renamed to stay consistent with prefix `mc*` instead of `px4_mc*`
- removed mode
- added feedforward for velocity and acceleration which are mode dependent
- changed `r - position` to `X - position`, adaption also for velocity and acceleration
- consistent diagram with velocity controller - see differential velocity path 
- removed projection and yaw since this is not part of position and velocity controller

2. Added minimal description
3. Removed old diagram

Please review @bresch @hamishwillee


bresch:
The diagram is better now, however, could you use the symbols we have defined in the documentation to be more "correct". I know the other ones also don't follow the conventions exactly, but we should start to be a bit more strict with that.
https://docs.px4.io/main/en/contribute/notation.html#symbols

hamishwillee:
@heappy Aside from comment ^^^, can you use the same name as the existing file (preferably) or delete the old file. If the location of the source for the image has changed, can you update that too (if you just updated the existing stored image then no change needed)

heappy:
Thanks for the feedback
@hamishwillee I removed the old file in assets/diagram - this resulted in the failed check. I'll rename it to the old file name with the changes from @bresch

bresch (approved):
Thanks!

hamishwillee:
All good. I'll test the build. Thanks so much for this!

en/flight_stack/controller_diagrams.md
```diff
@@ -63,11 +63,13 @@ The diagrams use the standard [PX4 notation](../contribute/notation.md) (and eac
 
 #### Combined Position and Velocity Controller Diagram
 
-![MC Position Controller Diagram](../../assets/diagrams/px4_mc_position_controller_diagram.png)
+![MC Position Controller Diagram](../../assets/diagrams/mc_combined_position_velocity_controller_diagram.png)
```
hamishwillee, line 66: Ah, name reverted but not in the docs.

heappy: Oh my bad sry

hamishwillee: No worries! Thanks for fixing up the diagram!

PR #1943, opened 2022-07-13
PX4 v1.13 MAVLink parachute support 

hamishwillee:
This updates parachute support docs for MAVLink and (partially) for actuators.

It follows on from the work in https://github.com/PX4/PX4-Autopilot/pull/18589


en/peripherals/parachute.md
```diff
@@ -27,50 +28,85 @@ Below are a few considerations when using parachutes:
...
 :::
 
+<!-- 
+The parachute can also be configured using the [Actuators](../config/actuators.md) screen.
+In this case ....
+-->
+
+
+### MAVLink Parachute Setup
+
```
hamishwillee, line 83: @ThomasDebrunner @MaEtUgR Can you please check this section.

I've looked at the code in https://github.com/PX4/PX4-Autopilot/pull/18589 and I want to confirm my understanding

The big picture is clear - the `MAV_CMD_DO_PARACHUTE` is emitted if the parachute is detected and healthy on failsafe. The `COM_PARACHUTE` parameter is checked an will warn if a parachute is expected but not present/healthy.

The small picture is less clear:
-  PX4 appears to base its health checks on getting the heartbeat - that then sets `SYS_STATUS.onboard_control_sensors_present_extended` and `.onboard_control_sensors_health_extended` for the recovery system bit. Is that right?
  -  What about the enabled flag? (`onboard_control_sensors_enabled_extended`) - doesn't seem to be set. I would kind have expected enabled to be based on the `COM_PARACHUTE` parameter.
  - I guess we're assuming that a parachute is always "in a good state". Otherwise we need a mechanism for it to actually tell the flight stack it is OK.
  - What is `COM_PARACHUTE` supposed to do other than be a trigger for sending events when the heartbeat is lost?
- Is the `MAV_CMD_DO_PARACHUTE` sent to the specific component that emits the MAV_TYPE_PARACHUTE? What if two are connected? Or is it just sent to current system ID and component 0 (broadcast).
- Is the `MAV_CMD_DO_PARACHUTE`  command forwarded? i.e. can a GCS trigger the parachute? Should it?

en/peripherals/parachute.md
```diff
@@ -27,50 +28,85 @@ Below are a few considerations when using parachutes:
...
+To enable flight termination:
+- Set [Safety](../config/safety.md) action to *Flight termination* for checks where you want the parachute to trigger.
+- Set [Failure Detector](../config/safety.md#failure-detector) pitch angles, roll angles and time triggers for crash/flip detection, and disable the failure/IMU timeout circuit breaker (i.e. set [CBRK_FLIGHTTERM=0](../advanced_config/parameter_reference.md#CBRK_FLIGHTTERM)).
+  
+:::note
+You can also configure an [external Automatic Trigger System (ATS)](../config/safety.md#external-automatic-trigger-system-ats) for failure detection.
 :::
+  
+### Motor Failsafe Settings
+
```
hamishwillee, line 50: @bkueng The original docs here were very much based around a particular example using the PWM MAIN output 7 on a quad with motors on MAIN1-4. You essentially had to manually set the FAIL values for each of the outputs.

I'm "part" of the way to restructuring this to work for actuators too.
1. Do we need to think about motor "FAIL" settings? My guess is that these are automatically set to disarmed values as part of normal actuator setup (?).
2. The parachute can go on any unused output, and with actuators I can set where it should be:
   image
   
   But how do I set the failsafe value?
3. What happens if the system is using UAVCAN? ie. for ESCs and to trigger the parachute. What do I do in this case?

hamishwillee: @bkueng ping ^^^

en/peripherals/parachute.md
```diff
@@ -27,50 +28,85 @@ Below are a few considerations when using parachutes:
...
 
 Flight termination (and hence parachute deployment) may be triggered by safety checks such as RC Loss, geofence violation, and so on, from attitude triggers and other failure detector checks, or by a command from a ground station.
-During flight termination PX4 sets PWM outputs to their "failsafe" values (failsafe values turn off motors, but may be used to turn on/trigger the parachute.
+During flight termination PX4 sets PWM outputs to their "failsafe" values (failsafe values turn off motors, but may be used to turn on/trigger the parachute).
+If a MAVLink parachute is connected and healthy, a command will be sent to activate it.
 
 Parachute setup therefore involves:
-- Connecting the parachute to unused outputs and setting their failsafe values to levels that will trigger the parachute (using appropriate parameters).
 - Configuring [flight termination](../advanced_config/flight_termination.md) as the appropriate action for those safety and failure cases where the parachute should be deployed.
+- Configure PX4 output levels to disable motors on failsafe.
```
bkueng, line 36: I'd add that this is already the default. For servos it's the center value.

en/peripherals/parachute.md
bkueng: > Do we need to think about motor "FAIL" settings? My guess is that these are automatically set to disarmed values as part of normal actuator setup (?).

Yes it's automatic, you typically don't need to care.

> But how do I set the failsafe value? Or does PX4 know that for a parachute maximum means "trigger" and minimum/disarmed means "wait"? So values get set appropriately for the function?

It goes to the maximum value. Yes it's per function. See the pararm description for https://docs.px4.io/main/en/advanced_config/parameter_reference.html#PWM_AUX_FUNC1.

> What happens if the system is using UAVCAN? ie. for ESCs and to trigger the parachute. What do I do in this case?

It works the same way. What do you mean by ESC?

hamishwillee: Thanks. Ignore the "for escs" part. That was when I was thinking that motors in failsafe have another state than disarmed . Ignore me.

en/peripherals/parachute.md
hamishwillee: Merging as this is, I think correct. I still need the detail in ^^^ and that request is tracked in #1977

en/peripherals/parachute.md
```diff
@@ -27,50 +29,81 @@ Below are a few considerations when using parachutes:
...
+
+
+### MAVLink Parachute Setup
+
+PX4 will trigger a connected and healthy parachute on failsafe by sending the command [MAV_CMD_DO_PARACHUTE](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_PARACHUTE) with the [PARACHUTE_RELEASE](https://mavlink.io/en/messages/common.html#PARACHUTE_ACTION) action.
+
+MAVLink parachute support is enabled by setting the parameter [COM_PARACHUTE=1](../advanced_config/parameter_reference.md#COM_PARACHUTE).
+PX4 will then indicate parachute status using the [MAV_SYS_STATUS_RECOVERY_SYSTEM](https://mavlink.io/en/messages/common.html#MAV_SYS_STATUS_RECOVERY_SYSTEM) bit in the [SYS_STATUS](https://mavlink.io/en/messages/common.html#SYS_STATUS) extended onboard control sensors fields:
+- `SYS_STATUS.onboard_control_sensors_present_extended`: MAVLink parachute present (based on heartbeat detection).
+- `SYS_STATUS.onboard_control_sensors_enabled_extended`: ?
```
hamishwillee, line 87: Used?

PR #1944, opened 2022-07-13
Payload control allocation docs

hamishwillee:
@bkueng This attempts first pass as payload docs in the light of control allocation. Its a little messy because we still need to cover mixers. 

I have also "touched" the camera docs because I am not sure what changes for these. See notes below. We can merge this and work on those separately if you like

### DIFF
--- a/en/payloads/README.md
+++ b/en/payloads/README.md
@@ -24,55 +24,65 @@ PX4 supports the following set of MAVLink commands/mission items for cameras tha
 * [MAV_CMD_DO_TRIGGER_CONTROL](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_TRIGGER_CONTROL) - start/stop capturing (using distance or time, as defined using above messages).
 
 The following topics show how to *connect* your camera configure PX4:
-* [Camera triggering](../peripherals/camera.md) from flight controller PWM or GPIO outputs, or via MAVLink.
+
+* [Camera triggering](../peripherals/camera.md) from flight controller PWM or GPIO outputs, or via MAVLink. 
 * [Camera timing feedback](../peripherals/camera.md#camera-capture) via hotshoe input.
 
 
 ## Cargo Drones ("Actuator" Payloads)
 
 Cargo drones commonly use servos/actuators to trigger cargo release, control winches, etc.
-PX4 supports servo and GPIO triggering via both RC and MAVLink commands.
+PX4 supports actuator triggering via both RC and MAVLink commands.
 
-### RC Triggering
+### Payload Outputs: Control Allocation
 
-You can map up to three RC channels to control servos/actuators attached to the flight controller using the parameters [RC_MAP_AUX1](../advanced_config/parameter_reference.md#RC_MAP_AUX1) to [RC_MAP_AUX3](../advanced_config/parameter_reference.md#RC_MAP_AUX3).
+If control allocation is enabled ([SYS_CTRL_ALLOC=1](../advanced_config/parameter_reference.md#SYS_CTRL_ALLOC)) you can specify up to 6 outputs to be controlled from either RC channels or MAVLink.
 
-The RC channels are *usually* mapped to the `AUX1`, `AUX2`, `AUX3` outputs of your flight controller (using a [mixer file](../concept/mixing.md) defined in your airfame).
+:::note
+Control allocation is supported from PX4 v1.13 but disabled by default.
+:::
+
+This is done in the [Actuators](../config/actuators.md#actuator-outputs) configuration screen by assigning the functions `Offboard Actuator Set 1` to `Offboard Actuator Set 6` functions to any desired payload outputs (in the [actuator outputs](../config/actuators.md#actuator-outputs) section).
+
+### Payload Outputs: Mixer Allocation
+
+The [mixer file](../concept/mixing.md) for the current airframe is used to specify the payload outputs for RC and MAVLink passthrough when control allocation is disabled.
+
+Usually just three outputs are mapped to the `AUX1`, `AUX2`, `AUX3` outputs of the flight controller.
 You can confirm which outputs are used for RC AUX passthrough on your vehicle in the [Airframe Reference](../airframes/airframe_reference.md).
-For example, [Quadrotor-X](../airframes/airframe_reference.md#quadrotor-x) has the normal mapping: "**AUX1:** feed-through of RC AUX1 channel", "**AUX2:** feed-through of RC AUX2 channel", "**AUX3:** feed-through of RC AUX3 channel".
+For example, [Quadrotor-X](../airframes/airframe_reference.md#quadrotor-x) has the normal mapping: 
+- "**AUX1:** feed-through of RC AUX1 channel"
+- "**AUX2:** feed-through of RC AUX2 channel"
+- "**AUX3:** feed-through of RC AUX3 channel"
 
 If your vehicle doesn't specify RC AUX feed-through outputs, then you can add them using using a custom [Mixer File](../concept/mixing.md) that maps [Control group 3](../concept/mixing.md#control-group-3-manual-passthrough) outputs 5-7 to your desired port(s).
 An example of such a mixer is the default passthrough mixer: [pass.aux.mix](https://github.com/PX4/PX4-Autopilot/blob/main/ROMFS/px4fmu_common/mixers/pass.aux.mix).
 
-:::note
-The same outputs used for "feed-through of RC AUX" may also be set using a MAVLink command (see [below](#mission-triggering)).
-PX4 will use the last value set through either mechanism.
-:::
 
+### RC Triggering
+
+The available passthrough (payload) outputs are mapped to RC channels using the [RC_MAP_AUXn](../advanced_config/parameter_reference.md#RC_MAP_AUX1) parameters.
+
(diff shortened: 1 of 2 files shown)


en/peripherals/camera.md
```diff
@@ -48,12 +48,14 @@ On a controller that only has an FMU, the pins map to `MAIN5` and `MAIN6` (e.g.
 At time of writing triggering only works on FMU pins - you can't trigger a camera using pins on the I/O board.
 :::
 
+
 :::warning
```
hamishwillee, line 52: So there is all this old stuff which said you had to use certain outputs for triggering, and that some combinations disabled other pins.
1. Is that all still true when using mixers?>
2. With control allocation, how does this work? I.e. is it still 2 pins you set? 
3. What do we say for camera capture?

en/payloads/README.md
```diff
@@ -24,55 +24,65 @@ PX4 supports the following set of MAVLink commands/mission items for cameras tha
...
+
+If control allocation is enabled ([SYS_CTRL_ALLOC=1](../advanced_config/parameter_reference.md#SYS_CTRL_ALLOC)) you can specify up to 6 outputs to be controlled from either RC channels or MAVLink.
 
-You can map up to three RC channels to control servos/actuators attached to the flight controller using the parameters [RC_MAP_AUX1](../advanced_config/parameter_reference.md#RC_MAP_AUX1) to [RC_MAP_AUX3](../advanced_config/parameter_reference.md#RC_MAP_AUX3).
+:::note
+Control allocation is supported from PX4 v1.13 but disabled by default.
+:::
 
-The RC channels are *usually* mapped to the `AUX1`, `AUX2`, `AUX3` outputs of your flight controller (using a [mixer file](../concept/mixing.md) defined in your airfame).
+This is done in the [Actuators](../config/actuators.md#actuator-outputs) configuration screen by assigning the functions `Offboard Actuator Set 1` to `Offboard Actuator Set 6` functions to any desired payload outputs (in the [actuator outputs](../config/actuators.md#actuator-outputs) section).
```
bkueng, line 45: Note that the ones for RC are the `RC_AUX x` functions

en/payloads/README.md
```diff
@@ -24,55 +24,65 @@ PX4 supports the following set of MAVLink commands/mission items for cameras tha
...
-PX4 will use the last value set through either mechanism.
-:::
 
+### RC Triggering
 
-### Mission Triggering
+The available passthrough (payload) outputs are mapped to RC channels using the [RC_MAP_AUXn](../advanced_config/parameter_reference.md#RC_MAP_AUX1) parameters.
 
-You can use the [MAV_CMD_DO_SET_ACTUATOR](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_ACTUATOR) MAVLink command to set (up to) three actuators values at a time, either in a mission or as a command.
+If control allocation is used then `RC_MAP_AUXn` maps to the output that has an offboard acutator function with the same `n` index (e.g. `RC_MAP_AUX2` maps specified channel to the output with function `Offboard Actuator Set 2`).
```
bkueng, line 66: Same as above

en/peripherals/camera.md
bkueng: 1. yes for mixers
2. yes for the ones that require 2 pins, you need to assign camera triggering to 2 outputs. Then for the ones that use PWM (e.g. Seagull MAP2), the whole PWM group cannot be used for anything else, you can only have camera triggering in that group.
3. you can also assign any of the outputs

bkueng (commented):
Thanks Hamish. Looks good, apart from the RC AUX having separate output functions.

en/payloads/README.md
hamishwillee: Not sure how this applies to this statement. Will discuss below.

en/payloads/README.md
hamishwillee: > Note that the ones for RC are the RC_AUX x functions

What do you mean?

What I've said is:
- You can map any output (AUX/MAIN/UAVCAN) to use for passthrough using actuator assignment of `Offboard Actuator Set x` if control allocation is enabled.
- You can similarly specify that some AUX outputs (only) are passthrough channels - e.g. the ones marked as "feed-through of RC AUX1 channel"
- You can map these to RC channels using - `RC_MAP_AUXn` where `n` and `x` (for the output) map 1:1.

I THINK perhaps you are saying that while you can map any output as passthrough, you can only use `RC_MAP_AUX` to map those outputs to RC channels if they are AUX outputs. There are no  `RC_MAP_MAIN` mappers, so you couldn't use RC passthrough for a non-AXU channel. Is that right?

I assume you can still use any outputs for mapping to the mavlink command to set an actuator?

en/peripherals/camera.md
hamishwillee: > Then for the ones that use PWM (e.g. Seagull MAP2), 

> you can also assign any of the outputs

Don't they all have to use PWM? 

I base this on the yaml which says FMU only https://github.com/PX4/PX4-Autopilot/blob/main/src/lib/mixer_module/output_functions.yaml#L40-L52

image

en/peripherals/camera.md
```diff
@@ -1,18 +1,19 @@
...
-Cameras can also (optionally) use the flight controller [camera capture pin](#camera-capture) to signal the exact moment that a photo/frame is taken.
-This allows more precise mapping of images to GPS position for geotagging, or the right IMU sample for VIO synchronization, etc.
 
 ## Trigger Configuration
 
-Camera triggering is usually configured from the *QGroundControl* [Vehicle Setup > Camera](https://docs.qgroundcontrol.com/en/SetupView/Camera.html#px4-camera-setup) section.
+Camera triggering is usually configured from the *QGroundControl* [Vehicle Setup > Camera](https://docs.qgroundcontrol.com/en/SetupView/Camera.html#px4-camera-setup) section. 
 
-![Trigger pins](../../assets/camera/trigger_pins.png)
+![Trigger pins](../../assets/camera/trigger_pins.png)
```
hamishwillee, line 16: FMI did you tidy up the UI here for when control allocation is used? Specifically, I would expect the hardware settings to disappear (I can't test)

image

hamishwillee:
Thanks @bkueng - updated with separate output functions for RC and MAVLink - did not read your response properly at first!

This is a bit complicated because there are two ways to do things. When control allocation is the default we can simplify this again.

There are a couple open questions re camera. Might be OK in which case we can merge.

en/peripherals/camera.md
bkueng: They use the PWM Pins, but not necessarily PWM signal. For some it's just a GPIO signal that's toggled.

en/peripherals/camera.md
bkueng: Good point, no I did not look at that.

hamishwillee: OK, leaving that with you. If you fix create a PR or issue for this can you ping me so I can add it as a comment?

en/peripherals/camera.md
```diff
@@ -54,9 +57,25 @@ With `TRIG_PINS=78`, you can use the AUX pins 1-6 as actuator outputs.
 Any other combination of pins can be selected, but this will disable use of the other FMU pins as outputs.
 :::
 
+### Trigger Setup using Control Allocation
+
+When _dynamic control allocation_ is enabled ([SYS_CTRL_ALLOC=1](../advanced_config/parameter_reference.md#SYS_CTRL_ALLOC)) camera trigger pins can be set on any FMU outputs.
+This is done on the *QGroundControl* [Actuators](../config/actuators.md) configuration screen by assigning the `Camera_Trigger` function on any FMU output.
+
+If using a PWM output that requires two pins (e.g. Seagull MAP2) you can assign to any two outputs.
```
hamishwillee, line 65:
```suggestion
If using trigger setup that requires two pins (e.g. Seagull MAP2) you can assign to any two outputs.
```

en/peripherals/camera.md
```diff
@@ -100,9 +114,24 @@ PX4 emits the MAVLink [CAMERA_TRIGGER](https://mavlink.io/en/messages/common.htm
 If camera capture is configured, the timestamp from the camera capture driver is used, otherwise the triggering timestamp.
 :::
 
+Camera capture/feedback is enabled in PX4 by setting [CAM_CAP_FBACK = 1](../advanced_config/parameter_reference.md#CAM_CAP_FBACK).
+
+If mixers are used, then the capture pin used depends on the hardware:
+- Pixhawk FMUv5x boards use the board-specific camera capture pin (PI0).
+- Other board use FMU PWM pin 6 (hardcoded) for camera capture.
+
+If _dynamic control allocation_ is enabled ([SYS_CTRL_ALLOC=1](../advanced_config/parameter_reference.md#SYS_CTRL_ALLOC)) the camera capture pin can be set on any FMU output.
```
hamishwillee, line 123:
```suggestion
If _dynamic control allocation_ is enabled ([SYS_CTRL_ALLOC=1](../advanced_config/parameter_reference.md#SYS_CTRL_ALLOC)) the camera capture pin can be set on any FMU output pin.
```

en/peripherals/camera.md
hamishwillee: @bkueng Thanks. I think I have cleaned this up now. You OK for me to merge?

hamishwillee:
Merging. I think this is good. Thanks for all the help @bkueng

PR #1957, opened 2022-07-14
Update collision_prevention.md

daxatino:
By default MPC.POS_MODE is set to 4, which disables Collision Prevention. Current documentation does not make evident that this parameter needs to be set correctly. I propose to include the paramater in the documentation to avoid further confusion.

### DIFF
--- a/en/computer_vision/collision_prevention.md
+++ b/en/computer_vision/collision_prevention.md
@@ -47,6 +47,7 @@ Parameter | Description
 <span id="CP_DELAY"></span>[CP_DELAY](../advanced_config/parameter_reference.md#CP_DELAY) | Set the sensor and velocity setpoint tracking delay. See [Delay Tuning](#delay_tuning) below.
 <span id="CP_GUIDE_ANG"></span>[CP_GUIDE_ANG](../advanced_config/parameter_reference.md#CP_GUIDE_ANG) | Set the angle (to both sides of the commanded direction) within which the vehicle may deviate if it finds fewer obstacles in that direction. See [Guidance Tuning](#angle_change_tuning) below.
 <span id="CP_GO_NO_DATA"></span>[CP_GO_NO_DATA](../advanced_config/parameter_reference.md#CP_GO_NO_DATA) | Set to 1 to allow the vehicle to move in directions where there is no sensor coverage (default is 0/`False`).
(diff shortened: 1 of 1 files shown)


hamishwillee:
@MaEtUgR Can you confirm if Collision Prevention is only enabled when 0: Simple position control or 3: Smooth position control (Jerk optimized) are enabled - i.e. it doesn't work in position mode when  acceleration based input is enabled - i.e. https://docs.px4.io/main/en/advanced_config/parameter_reference.html#MPC_POS_MODE = 4

daxatino:
In MPC_POS_MODE=4, it activates the Manual Acceleration flight mode, here:
[](https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/flight_mode_manager/tasks/ManualAcceleration/FlightTaskManualAcceleration.hpp
).
Its class derives from "**FlightTaskManualAltitudeSmoothVel**", which does not activate Collision Prevention behaviours. 
It should instead derive from "**FlightTaskManualPositionSmoothVel**".

hamishwillee:
@daxatino Collision prevention should be enabled in all position control modes. I'd be happy for you to create a PX4-Autopilot bug - but lets wait on @MaEtUgR.

hamishwillee:
@MaEtUgR Is on holiday. Should be back soon.

hamishwillee:
@MaEtUgR Ping. I realize your backlog is large. This is just putting back on your radar.

hamishwillee:
@MaEtUgR Ping. I have a nasty feeling you're on holiday again, but just in case you're back.

MaEtUgR:
@hamishwillee Sorry for the delay, that would have been easy to answer. Yes collision prevention does not work with MPC_POS_MODE 4 because
1. Setpoints are different
2. No one maintained or ported collision prevention
3. I have no easy way to test even just the old version e.g. in SITL so I didn't attempt myself

I had high hopes in a new version of collision prevention that came to my attention but it's neither ready nor compatible to the old one yet. I'm well aware of the problem and at some point will have to solve if no one else maintains collision prevention. In fact we plan to remove MPC_POS_MODE 3 once collision prevention is ported.

I hope this helps.

hamishwillee (approved):
Thanks @MaEtUgR - that means this is correct. Thanks for the fix @daxatino, merging now.

PR #1962, opened 2022-07-20
Actuators - add output functions listing

hamishwillee:
We haven't methodically documented the output functions, which meant that we missed a bunch of possible behaviours we need to document (although we are on the way).

This creates a listing of the functions that might be offered for an output. 

I might do further cross linking. I just want to make sure we can get some information here before I tidy.


en/config/actuators.md
```diff
@@ -313,6 +313,51 @@ Configuring the Camera Capture / Trigger input requires a reboot to take effect
...
+
+Output functions are used to map the "logical functions" of an airframe, such as `Motor 1` or `Landing gear`, to physical outputs like FMU output pin 2.
+This makes it easy to use a particular output pin for almost any purpose.
+
+Some functions are only relevant to particular frames or output types, and will not be offered on others.
+
+Functions include: 
+
+- `Disabled`: Output has no assigned function.
+- `Constant_Min`: Output set to constant minimum value.
```
hamishwillee, line 326: My assumption is that this will be set to whatever the minimum value field is set in the output right? While the constant_max would be set to the max field?
When would you want to set constant values?

en/config/actuators.md
```diff
@@ -313,6 +313,51 @@ Configuring the Camera Capture / Trigger input requires a reboot to take effect
...
+- `Constant_Max`: Output is set to constant maximum value.
+- `Motor 1` to `Motor 12`: Output is indicated motor.
+  Only motors allowed for airframe are displayed.
+- `Servo 1` to `Servo 8`: Servo output.
+   These are further assigned a specific meaning based on airframe, such as "tilt servo", "left aileron".
```
hamishwillee, line 331: So servo is never offered as assignment, so this is an assumption.

en/config/actuators.md
```diff
@@ -313,6 +313,51 @@ Configuring the Camera Capture / Trigger input requires a reboot to take effect
...
+- `Offboard Acutator Set 1` to `Offboard Acutator Set 6`: MAVLink Payload output.
```
hamishwillee, line 332: I will cross link this to the payload section later.

en/config/actuators.md
```diff
@@ -313,6 +313,51 @@ Configuring the Camera Capture / Trigger input requires a reboot to take effect
...
+- `Landing Gear`: Output is landing gear.
```
hamishwillee, line 333: Is there anything special you have to do for landing gear? (other than set this and wire PWM output?). We have no other instructions for using landing gear - would be good to create some - who's the expert?

en/config/actuators.md
```diff
@@ -313,6 +313,51 @@ Configuring the Camera Capture / Trigger input requires a reboot to take effect
...
+- `Parachute`: Output is parachute.
```
hamishwillee, line 334: What else should we say here. I haven't checked other PR on Parachute yet, but we need to understand what value is set for this normally and what is used for the "trigger" value.

en/config/actuators.md
```diff
@@ -313,6 +313,51 @@ Configuring the Camera Capture / Trigger input requires a reboot to take effect
...
+- `RC Roll`: Output is roll from RC.
+- `RC Pitch`: Output is pitch from RC.
+- `RC Throttle`: Output is throttle from RC.
+- `RC Yaw`: Output is yaw from RC
+- `RC Flaps`: Output is flaps from RC
```
hamishwillee, line 339: These are just RC direct passthrough right? So these can be assigned via other params to particular channels?

en/config/actuators.md
```diff
@@ -313,6 +313,51 @@ Configuring the Camera Capture / Trigger input requires a reboot to take effect
...
+
+The following functions can only be applied to FMU outputs:
+
+- `Camera_Trigger`: Output to trigger camera.
+  Enabled when [`TRIG_MODE==0`](../advanced_config/parameter_reference.md#TRIG_MODE).
+  Configured via `TRIG_*` parameters.
+- `Camera_Capture`: Input to get image capture notification.
+  Enabled when [CAM_CAP_FBACK==0](../advanced_config/parameter_reference.md#CAM_CAP_FBACK).
+  Configured via `CAM_CAP_*` parameters.
+- `PPS_Input`: PPS Capture.
```
hamishwillee, line 353: What is PPS?

en/config/actuators.md
```diff
@@ -313,6 +313,51 @@ Configuring the Camera Capture / Trigger input requires a reboot to take effect
 You should assign functions to the outputs that match your physical wiring of motors and servos, and use the [Actuator Testing](#actuator-testing) section described below to determine appropriate output parameter values.
 These steps are covered in [Output Assignment and Configuration](#output-assignment-and-configuration).
 
+### Output Functions
+
+Output functions are used to map the "logical functions" of an airframe, such as `Motor 1` or `Landing gear`, to physical outputs like FMU output pin 2.
+This makes it easy to use a particular output pin for almost any purpose.
+
+Some functions are only relevant to particular frames or output types, and will not be offered on others.
```
junwoo091400, line 321: This is specified in the Mixer of the control_allocator's [parameter definition yaml file](https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/control_allocator/module.yaml#L452-L868). Maybe worth mentioning? Or is it too much information for the users?

en/config/actuators.md
junwoo091400: That is correct. You don't directly assign the Servos as outputs, but rather as Tilt or Control Surfaces
image

en/config/actuators.md
junwoo091400: No, it just has 3 [action options](https://github.com/PX4/PX4-Autopilot/blob/main/msg/landing_gear.msg) and isn't that different from simple actuator. I think for 'instruction', it would just be:

1. Push the slider to maximum and verify that gear is up.
2. Push the slider to minimum and verify that gear is down
3. Adjust the 'reverse range' option appropriately, if the landing gear moves in the wrong direction

en/config/actuators.md
junwoo091400: It currently stays as Minimum (-1.0f) value, until a Failsafe is triggered, then the parachute gets deployed by setting the Maximum (1.0f) value.

So:
1. Move the slider to minimum and verify that parachute is in idle / stored state
2. Move the slider to maximum and verify that parachute is in deploy / open state
3. The parachute will get triggered automatically when the vehicle is in Failsafe mode

https://github.com/PX4/PX4-Autopilot/blob/55563eba49f87624d7b73356ea3d421fee7ec081/src/lib/mixer_module/functions.hpp#L244-L256

en/config/actuators.md
junwoo091400: Yes this is the direct passthrouh. Which 'other params' are you referring to?

en/config/actuators.md
junwoo091400: It's about capturing a pulse coming from a GPS module for time synchronization with the satellites (and the internal clock) I believe. 

https://github.com/PX4/PX4-Autopilot/blob/main/src/drivers/pps_capture/PPSCapture.cpp#L119-L154
https://old.fmad.io/blog-pps-time-synchronization.html

en/config/actuators.md
hamishwillee: Thanks

en/config/actuators.md
hamishwillee: https://docs.px4.io/main/en/advanced_config/parameter_reference.html#RC_MAP_PITCH and friends. These say what channel on the is passed through the output identified by `RC Pitch`and friends. I'll update that.

en/config/actuators.md
hamishwillee: I mentioned the specific cases that are constrained alongside the outputs, and link to the file at the end. Generally I don't think people need to know too much here other than that they might not see an output all the time.

en/config/actuators.md
hamishwillee: Thanks. I'll note the value set. But not the reason - since it is a bit speculative.

en/config/actuators.md
hamishwillee: I'm going to leave this part for now. If you know of an issue # for the problem though we could link it?

en/config/actuators.md
hamishwillee: Thanks. For this PR I will just update the minimum and maximum values.

en/config/actuators.md
hamishwillee: Thanks!

hamishwillee:
@bkueng This look OK to you? 

I probably need to do a separate topic on Landing gear and cross link, but should not block this. Ditto parachute case.

en/config/actuators.md
bkueng: > Oh actually it seems like it isn't even implemented properly

How so?

hamishwillee: FYI The main reason I want to leave this is that I need a lot more info than this about landing gear. But no more than this for this document.

hamishwillee: Changed my mind. Added landing gear/parachute sections at end. They will be OK, even if not exhaustive.

PR #1964, opened 2022-07-20
A whole lot of spell fixes

hamishwillee:
Found using https://github.com/codespell-project/codespell

Run interactively using:

```
codespell -i 3 -w -S ./advanced_config/parameter_reference.md -L ned
```


bkueng:
Add it to CI? Or are there too many exceptions?

hamishwillee:
@bkueng Its a good idea but I'd like to get the numbers down really low and think about it. 

In general if doing this I'd want to filter on the docs that the user had actually touched. That would then exclude things like params that are outside of the poster's control. 

Note, I  posted a PR on PX4-Autopilot to fix many of those cases today.

PR #1965, opened 2022-07-20
Improve Rover Documentation

junwoo091400:
## Motivation
A recent [post in the Discuss forum](https://discuss.px4.io/t/diy-rover-some-suggest-and-problem/28346#h-1the-px4-document-has-very-little-introduction-to-rover-5) about the poor documentation of using Rovers in PX4 has sparked my interest in improving the general documentation on Rovers.

Currently there are a lot of problems in terms of making a simple Rover work in PX4 since it is mostly focused on flight control and multicopter / fixed-wing vehicles. However, I think having a better documentation will at least attract more users to try out rover and lead to contributions / improvements in the end.

## Content
This is a first cut at improving the documentation and I would like to add the build documentation for a racing boat I am currently configuring as a follow up in a separate PR.

Let's make Rovers work well with PX4 :mechanical_arm: !

### DIFF
--- a/en/frames_rover/README.md
+++ b/en/frames_rover/README.md
@@ -1,24 +1,49 @@
-# Rover (UGVs)
+# Rovers (UGVs)
 
-![Traxxas Rover Picture](../../assets/airframes/rover/traxxas_stampede_vxl/final_side.jpg)
-
-PX4 provides support for Unmanned Ground Vehicles (UGVs). The set of configurations can be seen in [Airframes Reference > Rover](../airframes/airframe_reference.md#rover).
+PX4 supports rovers (Unmanned Ground Vehicles - UGVs) with [ackermann and differential](#rover-types) steering.
 
 This section contains build logs/instructions for assembling as well as configuring a number of UGV frames.
 
-## How to configure a Rover
+![Traxxas Rover Picture](../../assets/airframes/rover/traxxas_stampede_vxl/final_side.jpg)
 
-Using the [control allocation](../config/actuators.md), it is straight forward to setup the rover.
+## Rover Types
 
-First, determine your rover type as there are two types of Rover:
-1. Rover (Differential) : Rover that steers itself using different speeds of the wheels left and right. For example, the [r1 rover](../simulation/gazebo.md#ugv_differential) is differential type.
-2. Rover (Ackermann) : Rover with a steering mechanism equivalent to a regular car using the [ackermann geometry](https://en.wikipedia.org/wiki/Ackermann_steering_geometry). For example, the [gazebo rover](../simulation/gazebo.md#ugv_ackermann) is ackermann type.
+PX4 supports rovers with:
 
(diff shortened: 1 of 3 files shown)


hamishwillee:
There is some work to be done in control allocation: https://github.com/PX4/PX4-Autopilot/issues/19937

hamishwillee:
A worthy goal. I have fixed up the structure to be more extensible.

en/frames_rover/README.md
```diff
@@ -1,11 +1,49 @@
...
+
+For vehicles with Ackermann steering:
+
+1. In the [Airframe](../config/airframe.md) configuration, select the *Generic Ground Vehicle*.
+1. Set the [CA_AIRFRAME](../advanced_config/parameter_reference.md#CA_AIRFRAME) parameter to **Rover (Ackermann)** (5).
+1. Follow the [Actuators document](../config/actuators.md) to map the steering and throttle outputs as displayed.
+
+For vehicles with Differential steering:
+
+1. In the [Airframe](../config/airframe.md) configuration, select either the _Aion Robotics R1 UGV_ or _NXP Cup car: DF Robot GPX_
```
hamishwillee, line 33: @bkueng I created https://github.com/PX4/PX4-Autopilot/issues/19937 because (among other things) there is no generic differential frame. 
- Is the intent to create one?
- Is the approach here OK for now - ie copying these two frames. My assumption is that they will be able to add any additional throttle or whatever as needed.


Further, the assumption of our generic ackerman is one steering motor/servo and one for thrust. However the simulated rovers have multiple motors for steering and so on defined. 
So my question is, how configurable is this? If I add 2 left motors and 2 right motors will all of them be used. What about a 6 motor vehicle? 

It would be good to state our limits here :-)

en/frames_rover/README.md
```diff
@@ -1,11 +1,49 @@
...
+1. Set the [CA_AIRFRAME](../advanced_config/parameter_reference.md#CA_AIRFRAME) parameter to **Rover (Differential)** (6).
+1. Follow the [Actuators document](../config/actuators.md) to map the left and right motor and throttle outputs.
```
hamishwillee, line 35:
```suggestion
1. Follow the [Actuators document](../config/actuators.md) to map the left and right motors and throttle outputs.
```

en/frames_rover/README.md
hamishwillee: Thanks @bkueng - then I think this is probably OK for now. I'll merge. If you hate it let me know and we can fix as a post process.

hamishwillee:
Thanks @junwoo091400 

[Traxis](https://docs.px4.io/main/en/frames_rover/traxxas_stampede.html#traxxas-stampede-vxl) appears to be gone from the airframe already - its an ackermann so I presume we can just suggest the generic rover airframe ? Do you want to update that ?

Other things of interest are how it behaves in missions. Whether it can reverse, etc etc.

PR #1974, opened 2022-07-23
Add RC Link lost - Failsafe Action delay information

junwoo091400:
## About
As mentioned in [this PR comment](https://github.com/PX4/PX4-Autopilot/pull/16406#issuecomment-824620724), the behavior of `COM_RCL_ACT_T` was not documented in Safety documentation. And I noticed this discrepancy as reported in [this issue](https://github.com/PX4/PX4-Autopilot/issues/19952).

## Changes
1. Updated QGC screenshot to up to date UI (Icon layouts changed)
2. Added explanation on the `COM_RCL_ACT_T` parameter behavior


en/config/safety.md
```diff
@@ -69,6 +68,19 @@ The RC Loss failsafe is triggered if the RC transmitter link is lost in manual m
 
 ![Safety - RC Loss (QGC)](../../assets/qgc/setup/safety/safety_rc_loss.png)
 
+Note that PX4 has two delay/timeout settings which comes into play like this:
+
+1. RC link is lost (Transmitter turned off / vehicle goes out of Radio range)
+2. Timeout of [COM_RC_LOSS_T](../advanced_config/parameter_reference.md#COM_RC_LOSS_T)
+3. System considers RC link as lost, and RC control is invalid
+4. Vehicle enters auto loiter (hold) mode if in manual mode, or continues mission if in mission mode
```
hamishwillee, line 76: not true. Depends on COM_RCL_EXCEPT

hamishwillee:
Thanks @junwoo091400 

This has useful information, so I am going to merge it. I have subedited it a bit. The important there here is actually to make it clear that you should not touch the RC loss value in the editor and you may need to change the act timeout.

BUT this is not the way to fix this problem. The right way is to update QGroundControl to "do the right thing". 
- Minimal fix would be to replace COM_RC_LOSS_T  param (and name) with COM_RCL_ACT_T . YOu might have both but if you can only have one it must be COM_RCL_ACT_T 
- Nice to have would be adding checkboxes for COM_RCL_EXCEPT.

Then we can more or less leave things alone, because you should only be touching COM_RCL_ACT_T 

Can you make the necessary changes in QGC?

PR #1988, opened 2022-07-29
Fix mixer file syntax

mortenfyhn:
I think this mixer file syntax example is missing a colon, based on the actual mixer files in the PX4-Autopilot repo.


hamishwillee:
Absolutely. Thank you.

PR #1992, opened 2022-07-29
Add nokov supoort

duguguang:
nokov optical motion capture system can use the vrpn or ros package(https://wiki.ros.org/mocap_nokov) to Provide location support


hamishwillee:
Thanks. I did this in the commit https://github.com/PX4/PX4-user_guide/commit/1e61449631d3fb68d09ab0c91cfbab1d3f28cef1 

Note, only files in the /en path ever need to be updated. Everything else gets auto-updated via the translation system crowdin.

PR #2002, opened 2022-08-03
Change to host networking in container example

magicbycalvin:
When running the container example in bridged mode, QGroundControl reports that the UDP port is already in use when attempting to connect to the SITL simulation through the comms ports. This is rectified by including a note and making a slight modification to the example code.

### DIFF
--- a/en/test_and_ci/docker.md
+++ b/en/test_and_ci/docker.md
@@ -129,15 +129,19 @@ xhost +
 # Run docker and open bash shell
 docker run -it --privileged \
 --env=LOCAL_USER_ID="$(id -u)" \
--v ~/Projects/PX4-Autopilot:/src/PX4-Autopilot/:rw \
+-v ~/src/PX4-Autopilot:/src/PX4-Autopilot/:rw \
 -v /tmp/.X11-unix:/tmp/.X11-unix:ro \
--e DISPLAY=$DISPLAY \
+-e DISPLAY=:0 \
 --network host \
 --name=px4-ros px4io/px4-dev-ros2-foxy:2022-07-31 bash
 ```
 
 :::note
 We use the host network mode to avoid conflicts between the UDP port access control when using QGroundControl on the same system as the docker container.
+:::
+
+:::note
+In some situations, `DISPLAY` may need to be set to a different value such as `:1`. It may be possible to determine what that value is by using `echo $DISPLAY` or simply by changing `-e DISPLAY=:0` to `-e DISPLAY=$DISPLAY`. Note that making that change assumes that the environment variable `DISPLAY` has been set.
 :::
 
 If everything went well you should be in a new bash shell now. Verify if everything works by running, for example, SITL:
--- a/en/test_and_ci/docker.md
+++ b/en/test_and_ci/docker.md
@@ -141,7 +141,9 @@ We use the host network mode to avoid conflicts between the UDP port access cont
 :::
 
 :::note
-In some situations, `DISPLAY` may need to be set to a different value such as `:1`. It may be possible to determine what that value is by using `echo $DISPLAY` or simply by changing `-e DISPLAY=:0` to `-e DISPLAY=$DISPLAY`. Note that making that change assumes that the environment variable `DISPLAY` has been set.
+If you encounter the error "Can't open display: :0", `DISPLAY` may need to be set to a different value.
+On Linux (XWindow) hosts you can change `-e DISPLAY=:0` to `-e DISPLAY=$DISPLAY`.
+On other hosts you might iterate the value of `0` in  `-e DISPLAY=:0` until the "Can't open display: :0" error goes away.
 :::
 
 If everything went well you should be in a new bash shell now. Verify if everything works by running, for example, SITL:
--- a/en/test_and_ci/docker.md
+++ b/en/test_and_ci/docker.md
@@ -126,12 +126,16 @@ xhost +
 # Run docker and open bash shell
 docker run -it --privileged \
 --env=LOCAL_USER_ID="$(id -u)" \
--v ~/src/PX4-Autopilot:/src/PX4-Autopilot/:rw \
+-v ~/Projects/PX4-Autopilot:/src/PX4-Autopilot/:rw \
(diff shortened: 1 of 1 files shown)


en/test_and_ci/docker.md
```diff
@@ -126,13 +126,17 @@ xhost +
 # Run docker and open bash shell
 docker run -it --privileged \
 --env=LOCAL_USER_ID="$(id -u)" \
--v ~/src/PX4-Autopilot:/src/PX4-Autopilot/:rw \
+-v ~/Projects/PX4-Autopilot:/src/PX4-Autopilot/:rw \
```
hamishwillee, line 129: This bit is wrong. The bit you changed is the `<host_src>` and the example above explicitly states "share the directory **~/src/PX4-Autopilot** on the host computer"

en/test_and_ci/docker.md
```diff
@@ -126,13 +126,17 @@ xhost +
...
 -v /tmp/.X11-unix:/tmp/.X11-unix:ro \
--e DISPLAY=:0 \
--p 14570:14570/udp \
---name=mycontainer px4io/px4-dev-ros:2017-10-23 bash
+-e DISPLAY=$DISPLAY \
```
hamishwillee, line 131: This would assume you are pulling the value of $DISPLAY from somewhere.  Why, and where from?

en/test_and_ci/docker.md
```diff
@@ -126,13 +129,17 @@ xhost +
...
+--network host \
+--name=px4-ros px4io/px4-dev-ros2-foxy:2022-07-31 bash
 ```
 
+:::note
+We use the host network mode to avoid conflicts between the UDP port access control when using QGroundControl on the same system as the docker container.
+:::
```
hamishwillee, line 141: THanks @magicbycalvin 

@julianoes @JonasVautherin Do either of you know enough to evaluate this use of `--network host` change? Or can you recommend someone?

en/test_and_ci/docker.md
magicbycalvin: I forgot to change that back, I'll go ahead and fix that.

en/test_and_ci/docker.md
magicbycalvin: From my limited testing on Ubuntu machines (20.04 desktop on a Laptop and 22.04 desktop on an Intel NUC) I found that `$DISPLAY` had already been set by something else within the system. The reason I chose to use this instead of `:0` was because on my laptop, the proper display to use was actually `:1` which ended up causing me a half hour or so of grief trying to figure it out. By using `$DISPLAY` I am assuming that other machines also have had this value automatically set and will hopefully mitigate potential issues for future users. I can go ahead and add a note mentioning this though so that if `$DISPLAY` is not set, users will at least have some understanding of what is going on (or I can change it back to `:0` and still include a note about possibly needing to change that value).

en/test_and_ci/docker.md
julianoes: I'm never using this script, so I can't really comment on it. I also don't know what port forwarding 14570 was for.

hamishwillee: @julianoes Can you recommend who might know/who else I could ask?

JonasVautherin: I personally don't use my containers with `--network host`, but I don't really see a big problem in recommending it here. It's a pretty basic Docker feature, so people who know Docker will be fine removing it if they want to.

Overall if this helps beginners getting started, why not :+1:.

en/test_and_ci/docker.md
```diff
@@ -129,10 +132,18 @@ docker run -it --privileged \
...
+
+:::note
+In some situations, `DISPLAY` may need to be set to a different value such as `:1`. It may be possible to determine what that value is by using `echo $DISPLAY` or simply by changing `-e DISPLAY=:0` to `-e DISPLAY=$DISPLAY`. Note that making that change assumes that the environment variable `DISPLAY` has been set.
```
hamishwillee, line 144: Hi @magicbycalvin,

What the `-e` flag does (I think) is set environment flags in the container. Since we're passing in a value for DISPLAY from the host, I assume that what we're doing here is telling the container what display to use in the host.

Now the  $DISPLAY` environment variable is a magic number set by the XWindow system (according to the internet). So I would expect this to work any XWindow system like Ubuntu hosts, but not WIndows.

Further, what are the symptoms of getting the wrong display setting? Is it that gazebo might not display in the host machine, or something else?

Depending on the symptom I think we should reword like this:

```suggestion
If [symptoms of wrong display], `DISPLAY` may need to be set to a different value.
On Linux (XWindow) hosts you can change `-e DISPLAY=:0` to `-e DISPLAY=$DISPLAY`.
On other hosts you might iterate the value of `0` in  `-e DISPLAY=:0` until [symptom is fixed]
```

en/test_and_ci/docker.md
hamishwillee: As long as it works. Thanks!

en/test_and_ci/docker.md
hamishwillee: I have proposed a new wording (with questions to be answered) here: https://github.com/PX4/PX4-user_guide/pull/2002/files#r947330948

Otherwise I think this PR is good to go. THanks!

en/test_and_ci/docker.md
magicbycalvin: Good point about Windows environments (I didn't even know you could get PX4 working on Windows!)
I agree with your suggested change and went ahead and copied that over. The symptom of a wrong display is that you get the error `Can't open display: :0`. I included that message in the [symptom of wrong display] and [symptom is fixed] comments above.

hamishwillee:
Thanks @magicbycalvin - your fix looks good. Appreciate your patience getting this through.

FYI only, yes Windows is supported. The official support via Cygwin is broken in the main branch due to lack of knowledge about getting the packaging tools working. We're very likely to move to WSL2 soon - it's great - Linux on Windows: http://docs.px4.io/main/en/dev_setup/dev_env_windows_wsl.html#windows-wsl2-based-development-environment

PR #2003, opened 2022-08-05
Fixing the reference for mavlink start changes

gubertoli:
The file that must be changed to enable broadcasting and specific address streaming is the one that call `mavlink start`. This file is the `px4-rc.mavlink` instead of the `rcS`.


hamishwillee:
Thanks @gubertoli !

PR #2011, opened 2022-08-13
add parameters section for airspeed configuration

avcuenes:


en/sensor/airspeed.md
```diff
@@ -43,6 +43,14 @@ Enable each type using its [corresponding parameter](../advanced_config/paramete
 
 Airspeed calibration information can be found in: [Basic Configuration > Airspeed](../config/airspeed.md).
 
+### Parameters
+
```
hamishwillee, line 47: Thanks @avcuenes 

I don't _think_ so. Listing them in this doc means that they should apply to all airspeed sensors. But if you look at the definitions they appear to apply to a particular model the SDP3x, though CAL_AIR_TUBED_MM might be used elsewhere:

- CAL_AIR_CMODEL (INT32)	Airspeed sensor compensation model for the SDP3x
- CAL_AIR_TUBELEN (FLOAT) | Airspeed sensor tube lengthComment: See the CAL_AIR_CMODEL explanation on how 
- CAL_AIR_TUBED_MM (FLOAT) | Airspeed sensor tube diameter. Only used for the Tube Pressure Drop Compensation

So my take is that we should include the ones above that must be set for _all_ airspeed sensors in this page. We might also say that there are other params specific to airspeed setup. 

If any of these must be set for a specific airspeed sensor then they should be mentioned in that sensor.

@sfuhrer Can you please comment on this?

sfuhrer: I'm pretty sure all 3 are only used for SDP3x sensors.

hamishwillee:
@avcuenes , @sfuhrer 

I have subedited this to:
- Make it clear that airspeed sensors are not unlike other sensors in that they need to be started by default. This is a general rule in PX4 from v1.13 and not an exception. I need to work out what is started by default and mark everything else as "probably not started by default". This is a good start.
- Sensor-specific configuration of parameters should be done in the appropriate page if one exists. There is no such page for this sensor, which is a pity. So I have created a section for this case and added the params there.

Thanks both of you. Docs are better because of this.

PR #2005, opened 2022-08-05
add a few information about airspeed sensor and usage of it

avcuenes:

PR #2014, opened 2022-08-18
airspeed - set the primary source

hamishwillee:
@sfuhrer @avcuenes This PR replaces https://github.com/PX4/PX4-user_guide/pull/2005 because I can't easily remove the changes to translations.

@sfuhrer It is worth mentioning [ASPD_PRIMARY](https://docs.px4.io/main/en/advanced_config/parameter_reference.html#ASPD_PRIMARY) here.

Two questions:

1. ASPD_PRIMARY sets an index. How can I tell/find out which index matches to which sensor I have enabled?
2. What does it mean to be the primary? Is this just weighted in some way in EKF2 or is it the ONLY thing used by EKF2 and then dropped if EKF2 somehow decides the first value is invalid. Are there docs I can point to?

### DIFF
--- a/en/sensor/airspeed.md
+++ b/en/sensor/airspeed.md
@@ -26,7 +26,6 @@ All the above sensors are connected via the I2C bus/port.
 Additionally, the [Avionics Anonymous Air Data Computer](https://www.tindie.com/products/avionicsanonymous/uavcan-air-data-computer-airspeed-sensor/) can be connected to the UAVCAN bus to determine not only high-accuracy airspeed, but also true static pressure and air temperature via onboard barometer and an OAT probe.
 :::
 
-
 ## Configuration
 
 ### Enable Airspeed Sensors
@@ -39,7 +38,11 @@ Enable each type using its [corresponding parameter](../advanced_config/paramete
 - **TE MS5525:** [SENS_EN_MS5525DS](../advanced_config/parameter_reference.md#SENS_EN_MS5525DS)
 - **Eagle Tree airspeed sensor:** [SENS_EN_ETSASPD](../advanced_config/parameter_reference.md#SENS_EN_ETSASPD)
 
-### Sensor-specific configuration
+### Set Primary Source
+
+If you have enabled multiple airspeed sensors enabled, you can select which one is the primary source for measurements using [ASPD_PRIMARY](../advanced_config/parameter_reference.md#ASPD_PRIMARY).
+
+### Sensor-specific Configuration
 
 Other than enabling the sensor, sensor-specific configuration is often not required.
 If it is needed, it should be covered in the appropriate sensor page (for example [TFSLOT > Configuration](./airspeed_tfslot.md\#configuration)).
--- a/en/sensor/airspeed.md
+++ b/en/sensor/airspeed.md
@@ -38,7 +38,7 @@ Enable each type using its [corresponding parameter](../advanced_config/paramete
 - **TE MS5525:** [SENS_EN_MS5525DS](../advanced_config/parameter_reference.md#SENS_EN_MS5525DS)
 - **Eagle Tree airspeed sensor:** [SENS_EN_ETSASPD](../advanced_config/parameter_reference.md#SENS_EN_ETSASPD)
 
-You should also check [ASPD_PRIMARY](../advanced_config/parameter_reference.md#ASPD_PRIMARY) is `1` (the default).
+You should also check [ASPD_PRIMARY](../advanced_config/parameter_reference.md#ASPD_PRIMARY) is `1` (see next section - this is the default).
 
 ### Multiple Airspeed Sensors
 
--- a/en/sensor/airspeed.md
+++ b/en/sensor/airspeed.md
@@ -38,9 +38,19 @@ Enable each type using its [corresponding parameter](../advanced_config/paramete
 - **TE MS5525:** [SENS_EN_MS5525DS](../advanced_config/parameter_reference.md#SENS_EN_MS5525DS)
 - **Eagle Tree airspeed sensor:** [SENS_EN_ETSASPD](../advanced_config/parameter_reference.md#SENS_EN_ETSASPD)
 
-### Set Primary Source
+You should also check [ASPD_PRIMARY](../advanced_config/parameter_reference.md#ASPD_PRIMARY) is `1` (the default).
 
-If you have enabled multiple airspeed sensors enabled, you can select which one is the primary source for measurements using [ASPD_PRIMARY](../advanced_config/parameter_reference.md#ASPD_PRIMARY).
+### Multiple Airspeed Sensors
+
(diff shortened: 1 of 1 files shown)


sfuhrer:
>ASPD_PRIMARY sets an index. How can I tell/find out which index matches to which sensor I have enabled?

There is currently sadly no option way to link it to a specific "sensor ID". Index 1 is the first one started, index 2 the second etc. 
And the value of -1 indicates an "airspeed sensor less" setup and 0 is "using synthetic airspeed estimation".

>What does it mean to be the primary? Is this just weighted in some way in EKF2 or is it the ONLY thing used by EKF2 and then dropped if EKF2 somehow decides the first value is invalid. 

EKF2 only gets the validated airspeed source. The validation happens outside in the airspeed selector. Primary means that this source is used as long as it's valid, and otherwise there is a fallback to a secondary instance if available.

>Are there docs I can point to?

Not atm, can you propose a good spot to add it to?

hamishwillee:
@sfuhrer Thanks!

> Not atm, can you propose a good spot to add it to?

I'll try add it here. 
1. Is there anything to say about the airspeed selector? i.e. what does it see at high level as things that would make one selected over another?
2. The user now has to start these themselves. So they will either have to start them in the right order "manually" and note that or have a config that starts them for their airfame.

en/sensor/airspeed.md
```diff
@@ -39,7 +38,21 @@ Enable each type using its [corresponding parameter](../advanced_config/paramete
 - **TE MS5525:** [SENS_EN_MS5525DS](../advanced_config/parameter_reference.md#SENS_EN_MS5525DS)
 - **Eagle Tree airspeed sensor:** [SENS_EN_ETSASPD](../advanced_config/parameter_reference.md#SENS_EN_ETSASPD)
 
-### Sensor-specific configuration
+You should also check [ASPD_PRIMARY](../advanced_config/parameter_reference.md#ASPD_PRIMARY) is `1` (the default).
+
+### Multiple Airspeed Sensors
```
hamishwillee, line 43: @sfuhrer I'm going to merge this, but would be good if you could check it, and we might want to adjust.

I assume:
1. The airspeed selection is continuous - i.e. it is run to check the airspeed sensor is valid all the time, and triggers the fallback automatically, independent of EFK.
2. Can EKF decide it does not trust the incoming data and force a new selection? Or does it just reset on some check?
3. What checks does the airspeed selector do? If we know what they are I can perhaps find a place here for that.
4. If the EKF does further checks to validate, what are they?

en/sensor/airspeed.md
```diff
@@ -39,7 +38,21 @@ Enable each type using its [corresponding parameter](../advanced_config/paramete
 - **TE MS5525:** [SENS_EN_MS5525DS](../advanced_config/parameter_reference.md#SENS_EN_MS5525DS)
 - **Eagle Tree airspeed sensor:** [SENS_EN_ETSASPD](../advanced_config/parameter_reference.md#SENS_EN_ETSASPD)
 
-### Sensor-specific configuration
+You should also check [ASPD_PRIMARY](../advanced_config/parameter_reference.md#ASPD_PRIMARY) is `1` (the default).
```
hamishwillee, line 41:
```suggestion
You should also check [ASPD_PRIMARY](../advanced_config/parameter_reference.md#ASPD_PRIMARY) is `1` (see next section - this is the default).
```

en/sensor/airspeed.md
```diff
@@ -39,7 +38,21 @@ Enable each type using its [corresponding parameter](../advanced_config/paramete
...
 - **Eagle Tree airspeed sensor:** [SENS_EN_ETSASPD](../advanced_config/parameter_reference.md#SENS_EN_ETSASPD)
 
-### Sensor-specific configuration
+You should also check [ASPD_PRIMARY](../advanced_config/parameter_reference.md#ASPD_PRIMARY) is `1` (see next section - this is the default).
+
+### Multiple Airspeed Sensors
+
+If you have multiple airspeed sensors then you can select which sensor is _preferred_ as the primary source using [ASPD_PRIMARY](../advanced_config/parameter_reference.md#ASPD_PRIMARY), where `1`, `2` and `3` reflect the order in which the airspeed sensors were started:
+- `-1`: Disabled (no airspeed information used). 
+- `0`: Synthetic airspeed estimation (groundspeed minus windspeed)
```
hamishwillee, line 47: The [EKF2 docs](http://docs.px4.io/main/en/advanced_config/tuning_the_ecl_ekf.html#synthetic-sideslip) on synthetic sideslip say "is enabled by setting the [EKF2_FUSE_BETA](http://docs.px4.io/main/en/advanced_config/parameter_reference.html#EKF2_FUSE_BETA) parameter to `1`"

So what's happening here? Do we also need to set ASPD_PRIMARY?

hamishwillee:
Thanks @avcuenes for kicking this off, and @sfuhrer for your advice. Silvan, couple more questions above for you as follow up. The docs are correct as far as they go (I think) but we can say a little more maybe to provide a more complete story.

en/sensor/airspeed.md
sfuhrer: 1. correct.
2. no there is no EKF->airspeed selector communication. The EKF gets from the selector what the selector thinks is best, and then can decide if it wants to consume the airspeed data or not.
3. You can see a rough explanation on the checks within the parameters. See ASPD_DO_CHECKS. If you need more info let me know, I also would have some example plots to add (motivation for the checks).
4. EKF fuses airspeed if it's above EKF2_ARSP_THR and has a reasonably low innovation compared to groundspeed minus windspeed.

en/sensor/airspeed.md
sfuhrer: No it's two different things. ASPD_PRIMARY (one could challenge the name...) set to 0 will make the selector output a "approximated" airspeed value, while setting the FUSE_BETA to 1 will just also EKF to estimate the wind without an airspeed sensor present. 
@bresch @RomanBapst would you know of a case where FUSE_BETA actually shouldn't be turned on a FW or VTOL in FW flight? Is it even necessary to keep this parameter? I haven't set up any vehicle where it wasn't enabled.

hamishwillee: OK. I will ignore this case.

en/sensor/airspeed.md
hamishwillee: Thank you! Captured in https://github.com/PX4/PX4-user_guide/pull/2049

en/sensor/airspeed.md
bresch: I would say that almost all planes have enough positive yaw stiffness (i.e.: directional stability) to estimate wind using sideslip fusion. The only case where it should be disabled is for a fighter aircraft with an unstable yaw axis or a skid-to-turn drone where the zero sideslip assumption is too far from the truth during turns.

PR #2016, opened 2022-08-18
Redirect root page to main

hamishwillee:
Vuepress had a root page "all about PX4" which needs to exist because we use a multi-lingual setup. This has the wrong page title generated and I can't seem to fix that.

However this page does not need to exist - what "PX4" is should be on the PX4 site. This is a docs site.

So what I have done is redirected the page to the english top page, always.

FYI @mrpollo


mrpollo:
finally... thank you

PR #2022, opened 2022-08-29
Remove erroneous space in path

noahbliss:
There is a space in a file path in one of the example commands. This PR removes it.


hamishwillee (approved):
Thanks @noahbliss !

PR #2023, opened 2022-08-29
Update micrortps.md

maxpolzin:
On Ubuntu 22.04 with px4_msgs@7f89976091235579633935b7ccaab68b2debbe19 and PX4_Autopilot@e780a583cd2f3c0c42978ff0e5c010f104443267, micrortps_client is not found. The micrortps_agent seems to be working though.

### DIFF
--- a/en/middleware/micrortps.md
+++ b/en/middleware/micrortps.md
@@ -145,7 +145,7 @@ The PX4 Autopilot firmware initialization code may in future automatically start
 For example, in order to run the *Client* daemon with SITL connecting to the Agent via UDP:
 
 ```sh
-micrortps_client start -t UDP
+micrortps_agent start -t UDP
 ```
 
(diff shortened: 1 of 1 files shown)


en/middleware/micrortps.md
```diff
@@ -145,7 +145,7 @@ The PX4 Autopilot firmware initialization code may in future automatically start
 For example, in order to run the *Client* daemon with SITL connecting to the Agent via UDP:
 
 ```sh
-micrortps_client start -t UDP
+micrortps_agent start -t UDP
```
hamishwillee, line 148: Hi @maxpolzin 

This isn't correct. The section is definitely about building the client, and the client is what is on PX4. The agent should be built on your companion computer or whatever.

If it is not present then there is some other problem though. Can you raise as a discussion board question. My assumption would be that you haven't built with rtps included.

PR #2034, opened 2022-09-07
Add THEPeach K1/R1

ThePeach-kr:
Hi,

We developed a Flight Controller and even completed a flight test.
I would like to add a description of our K1/R1 Flight Controller to the official PX4 documentation.

Thanks


hamishwillee:
That's very cool.

1. What versions of PX4 can it run? Specifically, you state that it is built via `make thepeach_r1_default` but I don't think that is even in the mainline branch right?
2. What level of testing has been done? Has this been given to the PX4 flight team to try out?
3. Are there any flight logs etc?
4. There is some boilerplate text we need on all of flight stacks that states how they are supported. From http://docs.px4.io/main/en/flight_controller/mindpx.html you can see these notes:

image

image

ThePeach-kr:
many thanks to applying

Here's answers your questions.
1. Currently, I requested PX4-Autopilot. 
- PX4-Autopilot: https://github.com/PX4/PX4-Autopilot/pull/20176
- PX4-Bootloader: https://github.com/PX4/PX4-Bootloader/pull/219
2. Our team tested it on its own. Should I provide PX4 team to proceed with the test?
3. Here`s flight log.
- K1: https://review.px4.io/plot_app?log=525fdf48-a445-4986-ae7e-566ff69a41d8
- R1: https://review.px4.io/plot_app?log=924d4fd3-f51d-4c5e-b4f5-aef5f562b501
4. Yes, I`ll do that.

Thank you for your advice.

hamishwillee:
Thanks! @bkueng @davids5 Please see this PR. Who's the best person to get the logs looked at. 

My thinking is that if it gets accepted into PX4 mainline that would be the point that we think about incorporating these documents.

PR #2053, opened 2022-09-14
Mixers removed - update camera trigger docs

hamishwillee:
This updates camera trigger/capture docs to remove mixer specific instructions.

Part of docs cleanup following https://github.com/PX4/PX4-Autopilot/pull/20108


en/peripherals/camera.md
```diff
@@ -114,25 +104,12 @@ PX4 emits the MAVLink [CAMERA_TRIGGER](https://mavlink.io/en/messages/common.htm
 If camera capture is configured, the timestamp from the camera capture driver is used, otherwise the triggering timestamp.
 :::
 
-Camera capture/feedback is enabled in PX4 by setting [CAM_CAP_FBACK = 1](../advanced_config/parameter_reference.md#CAM_CAP_FBACK).
-
-If mixers are used, then the capture pin used depends on the hardware:
-- Pixhawk FMUv5x boards use the board-specific camera capture pin (PI0).
-- Other board use FMU PWM pin 6 (hardcoded) for camera capture.
+## MAVLink Command Interface
```
hamishwillee, line 107: The driver accepts these MAVLink commands:
- [MAV_CMD_DO_TRIGGER_CONTROL](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_TRIGGER_CONTROL)
- [MAV_CMD_DO_DIGICAM_CONTROL](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_DIGICAM_CONTROL)
- [MAV_CMD_DO_SET_CAM_TRIGG_DIST](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_CAM_TRIGG_DIST)

My understanding is that it ignores the other commands, such as those for the MAVLink Camera Protocol but simply forwards those through to a connected MAVLink camera right?

en/peripherals/camera.md
```diff
@@ -11,25 +11,25 @@ This timestamp can be used for several applications, including: timestamping pho
 
 ## Trigger Configuration
 
-Camera triggering is usually configured from the *QGroundControl* [Vehicle Setup > Camera](https://docs.qgroundcontrol.com/master/en/SetupView/Camera.html#px4-camera-setup) section. 
+Camera triggering is usually configured from the *QGroundControl* [Vehicle Setup > Camera](https://docs.qgroundcontrol.com/master/en/SetupView/Camera.html#px4-camera-setup) section.
 
-![Trigger pins](../../assets/camera/trigger_pins.png) 
+![Trigger pins](../../assets/camera/trigger_pins.png)
```
hamishwillee, line 16: This image shows the AUX pins being set in the QGC interface. Did that UI get fixed now that the config can only be done through Actuators?

en/peripherals/camera.md
```diff
@@ -99,6 +80,15 @@ The full list of parameters pertaining to the camera trigger module can be found
 Cameras can also (optionally) use a camera capture pin to signal the exact moment when a photo/frame is taken.
 This allows more precise mapping of images to GPS position for geotagging, or the right IMU sample for VIO synchronization, etc.
 
+Camera capture/feedback is enabled in PX4 by setting [CAM_CAP_FBACK = 1](../advanced_config/parameter_reference.md#CAM_CAP_FBACK).
```
hamishwillee, line 83: Is CAM_CAP_FBACK needed now you know this by the output pin being set?

bkueng: Yes the driver still needs to be enabled so that it's started.

en/peripherals/camera.md
bkueng: No, but thanks for the reminder. I'll look into that.

en/peripherals/camera.md
bkueng: I don't know the details here.

en/peripherals/camera.md
bkueng: I checked and the `Hardware Settings` are only shown if the param exists. So it should be fine already.

hamishwillee: Excellent. I just cropped the image.

en/peripherals/camera.md
hamishwillee: Thanks. I've reviewed the spec and updated this section. I've also recommended using MAVLink cameras (i.e. highlighted the differences).

It might not be 100% accurate as it is untested but it does correctly identify the limits of each approach.

hamishwillee:
@bkueng Thanks for you help.

I have updated this to make it more specifically "Camera Configuration". We really need a lot more on cameras, in particular related to using MAVLink cameras - including listing cameras that support the camera protocol.

Unfortunately I can't do that - and no one who knows the options seems interested in going over this.

Upshot. This fixes the mixer updates and is a bit better, but more work could be done.

PR #2071, opened 2022-09-28
Purge mixer from system startup

hamishwillee:
This removes mixers from the startup doc.

@bkueng Can you please review. Can you also confirm that you can't stick a full airframe config file on the SD card - it only extends existing configurations?

### DIFF
--- a/en/concept/system_startup.md
+++ b/en/concept/system_startup.md
@@ -74,23 +74,29 @@ The resulting boot log has detailed information about the boot sequence and shou
 ### Replacing the System Startup
 
 In most cases customizing the default boot is the better approach, which is documented below.
-If the complete boot should be replaced, create a file `/fs/microsd/etc/rc.txt`, which is located in the `etc` folder on the microSD card. If this file is present nothing in the system will be auto-started.
+If the complete boot should be replaced, create a file `/fs/microsd/etc/rc.txt`, which is located in the `etc` folder on the microSD card.
+If this file is present nothing in the system will be auto-started.
 
 ### Customizing the System Startup
 
-The best way to customize the system startup is to introduce a [new airframe configuration](../dev_airframes/adding_a_new_frame.md).
-If only tweaks are wanted (like starting one more application) special hooks in the startup can be used.
+The best way to customize the system startup is to introduce a [new frame configuration](../dev_airframes/adding_a_new_frame.md).
+The frame configuration file can be included in the firmware or on an SD Card.
+
+If you only need to "tweak" the existing configuration, such as starting one more application or setting the value of a few parameters, you can specify these by creating two files in the `/etc/` directory of the SD Card:
+
+- [/etc/config.txt](#customizing-the-configuration-config-txt): modify parameter values
+- [/etc/extras.txt](#starting-additional-applications-extras-txt): start applications
+
+The files are described below.
 
 :::warning
 The system boot files are UNIX FILES which require UNIX LINE ENDINGS.
 If editing on Windows use a suitable editor.
 :::
 
-There are two main hooks.
-Note that the root folder of the microsd card is identified by the path `/fs/microsd`.
-
-- /fs/microsd/etc/config.txt
-- /fs/microsd/etc/extras.txt
+:::note
+These files are referenced in PX4 code as `/fs/microsd/etc/config.txt` and `/fs/microsd/etc/extras.txt`, where the root folder of the microsd card is identified by the path `/fs/microsd`.
+:::
 
 #### Customizing the Configuration (config.txt)
 
@@ -104,7 +110,7 @@ param set-default PWM_MAIN_DIS3 1000
 param set-default PWM_MAIN_MIN3 1120
 ```
 
-#### Starting additional applications
+#### Starting Additional Applications (extras.txt)
 
 The `extras.txt` can be used to start additional applications after the main system boot.
 Typically these would be payload controllers or similar optional custom components.
--- a/en/concept/system_startup.md
+++ b/en/concept/system_startup.md
@@ -73,9 +73,10 @@ The resulting boot log has detailed information about the boot sequence and shou
 
 ### Replacing the System Startup
 
-In most cases customizing the default boot is the better approach, which is documented below.
-If the complete boot should be replaced, create a file `/fs/microsd/etc/rc.txt`, which is located in the `etc` folder on the microSD card.
-If this file is present nothing in the system will be auto-started.
+The whole boot can be replaced by creating a file `/etc/rc.txt` on the microSD card with a new configuration (nothing in the old configuration will be auto-started, and if the file is empty, nothing at all will be started).
+
+Customizing the default boot is almost always a better approach.
+This is documented below.
 
 ### Customizing the System Startup
 
--- a/en/concept/system_startup.md
+++ b/en/concept/system_startup.md
@@ -94,17 +94,14 @@ Note that the root folder of the microsd card is identified by the path `/fs/mic
 
 #### Customizing the Configuration (config.txt)
 
-The `config.txt` file can be used to modify shell variables (parameters).
+The `config.txt` file can be used to modify parameters.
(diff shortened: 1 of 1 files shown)


en/concept/system_startup.md
```diff
@@ -68,33 +68,45 @@ The resulting boot log has detailed information about the boot sequence and shou
...
-* /fs/microsd/etc/config.txt
-* /fs/microsd/etc/extras.txt
-* /fs/microsd/etc/mixers/NAME_OF_MIXER
+- /fs/microsd/etc/config.txt
+- /fs/microsd/etc/extras.txt
 
 #### Customizing the Configuration (config.txt)
 
-The `config.txt` file can be used to modify shell variables.
+The `config.txt` file can be used to modify shell variables (parameters).
```
dagar, line 97: Maybe shift this to modifying parameters? There are a few things left that could be set with shell variables, but for the most part it's all moving to parameters.

en/concept/system_startup.md
```diff
@@ -68,33 +68,45 @@ The resulting boot log has detailed information about the boot sequence and shou
...
 It is loaded after the main system has been configured and *before* it is booted.
 
+For example, you could create a file on the SD card, `etc/config.txt` with this content:
+
+```
+set PWM_AUX_OUT 1234
```
dagar, line 103: All of this is done with parameters now.

en/concept/system_startup.md
hamishwillee: @dagar Absolutely. I was reading this as "using the shell to modify parameters".

```suggestion
The `config.txt` file can be used to modify parameters.
```

en/concept/system_startup.md
```diff
@@ -68,33 +68,45 @@ The resulting boot log has detailed information about the boot sequence and shou
...
+- /fs/microsd/etc/config.txt
+- /fs/microsd/etc/extras.txt
 
 #### Customizing the Configuration (config.txt)
 
-The `config.txt` file can be used to modify shell variables.
+The `config.txt` file can be used to modify shell variables (parameters).
 It is loaded after the main system has been configured and *before* it is booted.
 
+For example, you could create a file on the SD card, `etc/config.txt` with this content:
```
hamishwillee, line 100:
```suggestion
For example, you could create a file on the SD card, `etc/config.txt` with that sets parameter values as shown:
```

en/concept/system_startup.md
```diff
@@ -68,33 +68,45 @@ The resulting boot log has detailed information about the boot sequence and shou
...
+
+```
+set PWM_AUX_OUT 1234
+set PWM_AUX_DISARMED 1500
+set PWM_AUX_MIN 1000
+set PWM_AUX_MAX 2000
+set PWM_AUX_RATE 50
```
hamishwillee, line 107: Show params instead
```suggestion
param set-default PWM_MAIN_DIS3 1000
param set-default PWM_MAIN_MIN3 1120
```

hamishwillee:
Thanks @dagar - updated as you suggested. Good to merge otherwise?

I'm still hoping to confirm you can only extend an existing airframe config on the SD card -  not define a whole one that might be selected by the user in QGC.

bkueng (approved):
There's support for external airframes: https://github.com/PX4/PX4-Autopilot/blob/main/ROMFS/px4fmu_common/init.d/rcS#L207.
Basically if you use a `SYS_AUTOSTART > 1000000`, it will execute `ext_autostart/rc.autostart` (https://github.com/PX4/PX4-Autopilot/blob/main/ROMFS/px4fmu_common/init.d/rc.autostart_ext#L6).

hamishwillee:
@bkueng Excellent, but I'm a bit confused about where you put the file and what you do / do not have to do.

So I have my file, let's call it `1000002_hamish_frame`. 
- I started from the generic quad file [4001_quad_x](https://github.com/PX4/PX4-Autopilot/blob/main/ROMFS/px4fmu_common/init.d/airframes/4001_quad_x), 
- set up my geometry, outputs, and tuning params, then got the different params.
- used `param show-for-airframe` to work out the different params.
- copied `4001_quad_x`, appended the new params, renamed it to  `1000002_hamish_frame`

Looking at your comment I think then that 
- I need to set `SYS_AUTOSTART` to `1000002`.
  - This will then start `etc/init.d/rc.autostart_ext` in the firmware
  - Which will try to start `. ${SDCARD_EXT_PATH}/rc.autostart` on the SDK card, which I think is "/fs/microsd/etc/rc.autostart"
- Do I need to create "/fs/microsd/etc/rc.autostart"? If so, what does it look like?
- Do I copy `1000002_hamish_frame` into `/fs/microsd/etc/` 
- My  `1000002_hamish_frame` depends on the same things as `4001_quad_x` - e.g. `. ${R}etc/init.d/rc.mc_defaults`. Will it find this or do I need to copy that into SD card as well? (or is it optional - it will look for files on SD card first, then firmware)?

I'm hoping the answer is just set the  `SYS_AUTOSTART` appropriately and copy your config file to the location!

bkueng:
>      Do I need to create "/fs/microsd/etc/rc.autostart"? If so, what does it look like?

Your file `1000002_hamish_frame` should be `$SD_CARD/ext_autostart/rc.autostart`. This is the simple version. If you have multiple external airframes (as we at Auterion do), `rc.autostart` would check SYS_AUTOSTART and call the relevant script. I don't think we need this here though.

> My 1000002_hamish_frame depends on the same things as 4001_quad_x - e.g. . ${R}etc/init.d/rc.mc_defaults. Will it find this or do I need to copy that into SD card as well? (or is it optional - it will look for files on SD card first, then firmware)?

It will find it.

en/concept/system_startup.md
```diff
@@ -68,34 +68,49 @@ The resulting boot log has detailed information about the boot sequence and shou
...
-  * For custom applications: The system was out of RAM. Run the `free` command to see the amount of free RAM.
-  * A software fault or assertion resulting in a stack trace
+- For custom applications: The system was out of RAM. Run the `free` command to see the amount of free RAM.
+- A software fault or assertion resulting in a stack trace
 
 ### Replacing the System Startup
 
-In most cases customizing the default boot is the better approach, which is documented below. If the complete boot should be replaced, create a file `/fs/microsd/etc/rc.txt`, which is located in the `etc` folder on the microSD card. If this file is present nothing in the system will be auto-started.
+In most cases customizing the default boot is the better approach, which is documented below.
+If the complete boot should be replaced, create a file `/fs/microsd/etc/rc.txt`, which is located in the `etc` folder on the microSD card.
```
hamishwillee, line 77: Should we say `/etc/rc.txt` on the SD card. The `/fs/microsd/` is just how PX4 sees it - and irrelevant to the user at this point.

en/concept/system_startup.md
```diff
@@ -68,34 +68,49 @@ The resulting boot log has detailed information about the boot sequence and shou
...
+If this file is present nothing in the system will be auto-started.
```
hamishwillee, line 78: How then do you start stuff? Is the idea that you put all your config in this file?

hamishwillee:
OK, a couple more questions added, but I think this is ready to merge and maybe do fixes, if needed, as a post process. Re review appreciated.

I added info on adding config file to SD card in #2078

en/concept/system_startup.md
bkueng: Yes

en/concept/system_startup.md
hamishwillee, line 78: Improve the wording around replacing the whole boot.

```suggestion
The whole boot can be replaced by creating a file `/etc/rc.txt` on the microSD card with a new configuration (nothing in the old configuration will be auto-started, and if the file is empty, nothing at all will be started).

Customizing the default boot is almost always a better approach.
This is documented below.
```

hamishwillee: Thanks. All fixed in https://github.com/PX4/PX4-user_guide/pull/2071/files#r988606065

PR #2072, opened 2022-09-30
Package Delivery Feature Documentation

junwoo091400:
# About
This is a documentation for the recently merged PR: https://github.com/PX4/PX4-Autopilot/pull/19963

## About Package Delivery
This feature allows users to plan a mission where a Gripper can be configured to drop the package and accomplish a package delivery :gift: 

## TODOs
- [x] Add missing images
- [x] Add architecture documentation
- [x] Add tables (formatted)


junwoo091400:
@hamishwillee How do I format Tables in the User doc?

hamishwillee:
Standard markdown tables work - of which there are a few variants. Personally I do what best suits the information I have. For wide tables I tend to use the format with no start or end markers. I will sometimes align the first few column's | markers

```
col1H | another col | yetAnother col
--- | --- | --- 
Some content | yet more thingy | Why is Hamish so good looking? Many have asked this question :-)
Beer                | yet more thingy | Why is Hamish so good looking? Many have asked this question :-)
Sausages        | something          | Is there anything they can't do?
```

col1H | another col | yetAnother col
--- | --- | --- 
Some content | yet more thingy | Why is Hamish so good looking? Many have asked this question :-)
Beer                | yet more thingy | Why is Hamish so good looking? Many have asked this question :-)
Sausages        | something          | Is there anything they can't do?

hamishwillee:
Change out of draft when you're ready for review. 

My two bits for now.
1. Gripper Integration, should be under peripherals http://docs.px4.io/main/en/peripherals/ not advanced configuration. Copy at least the part of the structure for the page where you indicate a couple of purchase options.
2. Package delivery should probably be associated with payloads and cameras, so perhaps under there - http://docs.px4.io/main/en/payloads/. YOu might also argue its something to do with missions, so we could cross link from the mission page.
3. Package delivery - who is this doc for? I would split it into developers, integrators and users. Developers want to extend the feature and understand touch points. Integrators want to know how to add a gripper and any settings they have to touch. Users want to know how to define a mission with a gripper. If they must be combined I would structure doc for developers last.

junwoo091400:
> My two bits for now.

Thanks for the reply, and sorry for getting back on this so late :skull: 

Regarding your points, I would then structure the doc in following way:

1. For users: In the payload (http://docs.px4.io/main/en/payloads/#cargo-drones-actuator-payloads) doc, I would add a section for "Delivery using a Gripper" section, that points to the mission planning documentation (as the doc is already quite long, I think it makes sense to separate out mission planning). 
2. For users: Under the missions (http://docs.px4.io/main/en/flying/missions.html) doc, I would add a doc detailing how the Package delivery mission can be configured. Like done here: https://docs.auterion.com/flight-operations/auterion-mission-control/plan/mission-plan/package-delivery-mission
3. For integrator: Under the peripherals (http://docs.px4.io/main/en/peripherals/) doc, I would add a 'Gripper' section, where I would also add documentation about Servo Gripper that is currently supported, to explain how the Gripper should be set up. Like done here: https://docs.auterion.com/avionics/skynode/airframe-integration/gripper-integration
4. For developer: I would add a sub-section under advanced features (http://docs.px4.io/main/en/advanced_features/), to explain how a Gripper based package delivery mission works internally, as done briefly here: https://docs.auterion.com/avionics/skynode/airframe-integration/gripper-integration/setting-up-payload-delivery-with-gripper#payload-delivery-architecture-diagram

I will update the PR, and let me know if I missed any points! :+1:

junwoo091400:
Added the changes mentioned in the last comment

junwoo091400:
@hamishwillee Ready for review :wink:. The image sizes are still quite big (aware of that), but I would first want to get sanity checks from you regarding the structure from you! :+1:

hamishwillee:
Re your suggestions above that sounds pretty logical - I'll look in a minute.

The only caveat is "advanced features" - I don't think I'd put anything there. It is my intent to throw away that section altogether, merging relevant information where it needs to be and making a single features page in the intro somewhere.

Why? None of it is particularly advanced or stays advanced. All this means is that the features are seen as "too hard".

hamishwillee:
Unrelated comment:
image

1. Ideally this sentence should only be displayed if someone _tries_ to add the same button action twice.
2. The sentence is not great English: 
   > Assigning the same action to multiple buttons requires the press of all those buttons for the action to be taken. This is useful to prevent accidental button presses for critical actions like Arm or Emergency stop.

   I suggest that you don't need to explain WHY this rule exists here. It's a rule. The less text you make people read the better. 
   
   I would suggest:
   
   > Multiple buttons that have the same action must be pressed simultaneously to invoke the action.

en/flying/package_delivery_mission.md
```diff
@@ -0,0 +1,84 @@
+# Package delivery mission planning
```
hamishwillee, line 1: All headings in all docs - make sure are "First letter capitalization".

```suggestion
# Package Delivery Mission Planning
```

en/flying/package_delivery_mission.md
```diff
@@ -0,0 +1,84 @@
...
+
+A package delivery mission allows user to plan & execute a package delivery. Note that this is only supported for vehicles that has a [delivery mechanism hardware](../peripherals/gripper.md) configured & installed.
+
+
+## Operator control configuration
```
hamishwillee, line 6:
```suggestion
## Operator Control Configuration
```

en/flying/package_delivery_mission.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Delivery Mission Planning and Configuration (Gripper)
+
+Currently the package delivery feature only supports a [Gripper mechanism](../peripherals/gripper.md). Therefore the following documentation focuses on configuring a delivery mission using a Gripper.
+
+The package delivery feature using a gripper is embedded in the action waypoint "Gripper Mechanism". To enable different cases for different vehicle types and delivery behaviors, it has been decoupled as a separate waypoint. This is illustrated in the diagram below, where a clear **pre-waypoint** and **action-waypoint** cascade structure allows enabling different delivery use cases.
+
+![Package delivery action waypoint item methodology](../../assets/flying/package_delivery_action_waypoint_item.jpg)
+
+### Create a standard mission waypoint
```
hamishwillee, line 19:
```suggestion
### Create a Standard Mission Waypoint
```

en/flying/package_delivery_mission.md
```diff
@@ -0,0 +1,84 @@
...
+Before using the package delivery feature, if feasible please setup the [gripper action button](../peripherals/gripper.md#griper-action-joystick-button-mapping-in-qgc) to have a manual control over the Gripper's state.
+
+
+## Delivery Mission Planning and Configuration (Gripper)
+
+Currently the package delivery feature only supports a [Gripper mechanism](../peripherals/gripper.md). Therefore the following documentation focuses on configuring a delivery mission using a Gripper.
+
+The package delivery feature using a gripper is embedded in the action waypoint "Gripper Mechanism". To enable different cases for different vehicle types and delivery behaviors, it has been decoupled as a separate waypoint. This is illustrated in the diagram below, where a clear **pre-waypoint** and **action-waypoint** cascade structure allows enabling different delivery use cases.
+
+![Package delivery action waypoint item methodology](../../assets/flying/package_delivery_action_waypoint_item.jpg)
```
hamishwillee, line 17: I understand the point of this diagram - the actual release is preceded by some other operation. The diagram shows an MC doing land, and a VTOL doing land or drop. But an MC is a "VTOL" so an MC can do land and drop too.

en/flying/package_delivery_mission.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Create a standard mission waypoint
+
+Follow the [mission documentation](missions.md) on how to setup a Mission Start, Takeoff waypoints. Then, plan whatever waypoints you would like to execute as well.
+
+### Create a package delivery waypoint
```
hamishwillee, line 23:
```suggestion
### Create a Package Delivery Waypoint
```

en/flying/package_delivery_mission.md
```diff
@@ -0,0 +1,84 @@
...
+![Action waypoint](../../assets/flying/qgc_package_delivery_action_waypoint.png)
+
+* Then configure the action for the Release mechanism in the waypoint configuration. Set it to "Release" in order to release the package.
+* The gripper ID does not need to be set for now.
+
+![Gripper action setting](../../assets/flying/qgc_mission_plan_gripper_action_setting.png)
+
+At this point, the Release Mechanism waypoint would have disappeared from the map since it's an Action with no positional information.
+
+### Examples of Package delivery mission plan
```
hamishwillee, line 52:
```suggestion
### Examples Plans
```

en/flying/package_delivery_mission.md
```diff
@@ -0,0 +1,84 @@
...
+* **Pre-waypoint**: Land mission item
+* **Action**: Gripper Release (Not shown on the 2D map, only in the right panel)
+
+![Land and Release example](../../assets/flying/land_and_release_package_delivery_mission_example.png)
+
+Notice how the altitude graph shows the 'Land' item, also on the right panel.
+
+### Special Remarks
+
+#### RTL waypoint for a landed package delivery missions
```
hamishwillee, line 78:
```suggestion
#### RTL Waypoint for Package Delivery with Landing
```

en/flying/package_delivery_mission.md
```diff
@@ -0,0 +1,84 @@
+# Package Delivery Mission Planning
```
hamishwillee, line 1: 1. So a land waypoint involves landing, doing gripper, taking off again.
   - How do you stop the system from auto disarming? Or is the theory that the gripper will unlatch and the next waypoint be executed so quickly that the disarm won't happen? Do you need to tune anything to make this happen?
   - Did you give up on the idea of having something wait in place? i.e. a loiter waypoint on the ground, kicked off through the UI to go to next waypoint?
   - I assume that PX4 will just take off again to the new waypoint? i.e. no need for an explicit second takeoff waypoint?
   - Is that still true on a VTOL frame? They seem to like TAKEOFF waypoints.
2. An MC can hover, so it can drop a package too. See my note on the image below. 
3. If you're on a VTOL do you need to force a transition to MC first? Should we say that? 

I will need to subedit this a bit more for English, but I like it.

en/SUMMARY.md
```diff
@@ -88,6 +88,7 @@
     * [Landing \(Fixed Wing\)](flying/fixed_wing_landing.md)
     * [Takeoff \(Fixed Wing\)](flying/fixed_wing_takeoff.md)
   * [Missions](flying/missions.md)
+    * [Package Delivery Mission Planning](flying/package_delivery_mission.md)
```
hamishwillee, line 91: Removing "Planning" - it is inferred - and in sidebar we benefit from shorter titles.
```suggestion
    * [Package Delivery Mission](flying/package_delivery_mission.md)
```

en/flying/package_delivery_mission.md
```diff
@@ -0,0 +1,84 @@
...
+
+A package delivery mission allows user to plan & execute a package delivery. Note that this is only supported for vehicles that has a [delivery mechanism hardware](../peripherals/gripper.md) configured & installed.
```
hamishwillee, line 3:
```suggestion
A package delivery mission allows user to plan and execute a package delivery. Note that this is only supported for vehicles that has a [delivery mechanism hardware](../peripherals/gripper.md) configured & installed.
```

hamishwillee: Use & when you have to. Screen readers hate them.

en/flying/package_delivery_mission.md
hamishwillee: For the images, for example, something like this is possible for the waypoint editor - highlighting the buttons you indicate should be selected
image

en/peripherals/gripper.md
```diff
@@ -0,0 +1,106 @@
+# Gripper Peripherals
+
+![High-load gripper example](../../assets/hardware/grippers/highload_gripper_example.jpg)
+
+Grippers are a commonly used mechanism for UAV delivery applications. There are multiple types of interfaces (e.g. PWM, CAN) for different products in the market.
+
+## Supported Gripper types
```
hamishwillee, line 7:
```suggestion
## Supported Gripper Types
```

en/peripherals/gripper.md
```diff
@@ -0,0 +1,106 @@
...
+
+PX4 supports following Gripper types, check the individual page to learn how to integrate them.
+
+- [Servo Gripper with PWM](gripper_servo.md)
+
+:::note
+The [Roboclaw driver](../modules/modules_driver.md#roboclaw) isn't supported as a Gripper yet. However in the future it may get supported as a Gripper.
+:::
+
+## Gripper actuator mapping
```
hamishwillee, line 17:
```suggestion
## Gripper Actuator Mapping
```

en/peripherals/gripper.md
```diff
@@ -0,0 +1,106 @@
...
+
+To enable the output of the Gripper, set the Function of the output port the gripper is connected to as `Gripper`.
+
+:::note
+Make sure to set the correct PWM frequency (usually commercial servo / grippers only support 50Hz)! If you mis-configure the frequency you risk damaging the gripper.
+:::
+
+![Gripper output mapping](../../assets/config/gripper/qgc_gripper_output_setup.png)
+
+## Griper action joystick button mapping in QGC
```
hamishwillee, line 27:
```suggestion
## Gripper Action Joystick Button Mapping in QGC
```

en/peripherals/gripper.md
```diff
@@ -0,0 +1,106 @@
...
+
+Therefore unless you [configure the gripper for package delivery](#setting-up-gripper-for-package-delivery), gripper will not be functioning!
+:::
+
+
+## Setting up Gripper for Package Delivery
+
+Gripper is commonly used for features like [Package Delivery](../advanced_features/package_delivery.md). This section describes how to setup the Payload Delivery for Gripper.
+
+### Note on using Joystick button action during package delivery mission
```
hamishwillee, line 52:
```suggestion
### Note on Using Joystick Button Action During Package Delivery Mission
```

en/peripherals/gripper.md
```diff
@@ -0,0 +1,106 @@
...
+
+If you forcefully command the gripper to 'Close' position while the package delivery is happening (gripper open action), gripper won't be able to finish the open action, and the mission will come to a halt for a payload delivery mission item timeout, then resume the mission.
+
+### Enable payload delivery feature (Gripper)
```
hamishwillee, line 56:
```suggestion
### Enable Payload Delivery Feature (Gripper)
```

en/peripherals/gripper.md
```diff
@@ -0,0 +1,106 @@
...
+
+Set [`PD_GRIPPER_EN`](../advanced_config/parameter_reference.md#PD_GRIPPER_EN) parameter to 1 (reboot required after change).
+
+### Enabling Pre-arm mode
```
hamishwillee, line 60:
```suggestion
### Enabling Pre-arm Mode
```

en/peripherals/gripper.md
```diff
@@ -0,0 +1,106 @@
...
+
+As an alternative test method, you can utilize the [Joystick capability](#griper-action-joystick-button-mapping-in-qgc) to trigger gripper open and close actions as well.
+
+:::note
+If you get an error message such as "[payload_deliverer] not running", make sure that you have gone through the setup procedures above and either run `payload_deliverer start` command in the Nuttx shell
+:::
+
+After measuring the opening and closing actuation time, set [PD_GRIPPER_TO](../advanced_config/parameter_reference.md#PD_GRIPPER_TO) to the longer timeout of the two.
+
+### Mission timeout
```
hamishwillee, line 94:
```suggestion
### Mission Timeout
```

en/peripherals/gripper_servo.md
```diff
@@ -0,0 +1,34 @@
...
+
+Wire color | Purpose
+---     | ---
+Brown   | Ground
+Red     | Power
+Yellow  | PWM Signal
+
+You would need to connect them into a PWM input of the Flight controller appropriately.
+
+### Compatibility checks
```
hamishwillee, line 25:
```suggestion
### Compatibility Checks
```

en/peripherals/gripper.md
```diff
@@ -0,0 +1,106 @@
+# Gripper Peripherals
+
+![High-load gripper example](../../assets/hardware/grippers/highload_gripper_example.jpg)
+
+Grippers are a commonly used mechanism for UAV delivery applications. There are multiple types of interfaces (e.g. PWM, CAN) for different products in the market.
+
+## Supported Gripper Types
+
+PX4 supports following Gripper types, check the individual page to learn how to integrate them.
```
hamishwillee, line 9:
```suggestion
PX4 supports the following Gripper types, check the individual page to learn how to integrate them.
```

en/peripherals/gripper.md
```diff
@@ -0,0 +1,106 @@
+# Gripper Peripherals
+
+![High-load gripper example](../../assets/hardware/grippers/highload_gripper_example.jpg)
+
+Grippers are a commonly used mechanism for UAV delivery applications. There are multiple types of interfaces (e.g. PWM, CAN) for different products in the market.
+
+## Supported Gripper Types
+
+PX4 supports the following Gripper types, check the individual page to learn how to integrate them.
```
hamishwillee, line 9: What does it mean to "support" a gripper type? I think essentially it means that for a hardware-integrated gripper you can assign the gripper output. More generally it means that you can configure it to be triggered when the gripper commmand is encountered in a mission or via a command. Right? 

So the roboclaw is not supported because you can't set it up to close when a gripper command is sent. Right?

en/peripherals/gripper.md
```diff
@@ -0,0 +1,106 @@
...
+
+- [Servo Gripper with PWM](gripper_servo.md)
+
+:::note
+The [Roboclaw driver](../modules/modules_driver.md#roboclaw) isn't supported as a Gripper yet. However in the future it may get supported as a Gripper.
+:::
+
+## Gripper Actuator Mapping
```
hamishwillee, line 17: Should this go into the servo docs? I appreciate that you might later have a UAVCAN servo that is almost the same, but I think in that case it might still be good to have a separate image. Thoughts?

en/peripherals/gripper.md
```diff
@@ -0,0 +1,106 @@
...
+
+:::note
+Make sure to set the correct PWM frequency (usually commercial servo / grippers only support 50Hz)! If you mis-configure the frequency you risk damaging the gripper.
+:::
+
+![Gripper output mapping](../../assets/config/gripper/qgc_gripper_output_setup.png)
+
+## Gripper Action Joystick Button Mapping in QGC
+
+QGC provides the Button mapping interface, which allows to map delivery mechanism actions to specific buttons of the [Joystick](../config/joystick.md) connected to QGC.
```
hamishwillee, line 29: Can an RC controller switch be mapped in http://docs.px4.io/main/en/config/flight_mode.html#flight-mode-selection

I'm guessing not, but we should make that clear somewhere.

en/peripherals/gripper.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Gripper Actuation Time
+
+You need to specify an actuation time, which specifies the time it takes to open or close the gripper. As most grippers don't have a sensor to detect successful actuation (close/opening), the payload delivery feature will rely on this value to estimate gripper's position.
+
+To measure gripper actuation time, go to the MAVLink Shell in AMC and execute the following code while the drone is on a bench and the propellers are removed:
+
+`> payload_deliverer gripper_test`
+
+Then observe how long it takes for the gripper to go from open position to closed position, and vise versa for opening. Note down the actuation time for both cases and choose the maximum measurement for the actuation time.
```
hamishwillee, line 84:
```suggestion
Then observe how long it takes for the gripper to go from open position to closed position, and vice versa for opening. Note down the actuation time for both cases and choose the maximum measurement for the actuation time.
```

en/peripherals/gripper.md
```diff
@@ -0,0 +1,106 @@
...
+
+Set the [`PD_GRIPPER_TYPE`](../advanced_config/parameter_reference.md#PD_GRIPPER_TYPE) to the type of Gripper you have.
+
+Note that for now we only support a [Servo Gripper](gripper_servo.md), so you will be setting the value to `Servo`.
+
+### Gripper Actuation Time
+
+You need to specify an actuation time, which specifies the time it takes to open or close the gripper. As most grippers don't have a sensor to detect successful actuation (close/opening), the payload delivery feature will rely on this value to estimate gripper's position.
+
+To measure gripper actuation time, go to the MAVLink Shell in AMC and execute the following code while the drone is on a bench and the propellers are removed:
```
hamishwillee, line 80:
```suggestion
To measure gripper actuation time, go to the [MAVLink Shell](../debug/mavlink_shell.md) in QGC  and execute the following code while the drone is on a bench and the propellers are removed:
```

en/peripherals/gripper.md
```diff
@@ -0,0 +1,106 @@
...
+![Gripper action mapping](../../assets/config/gripper/qgc_gripper_actions_joystick.png)
+
+You can test the actions by clicking on the buttons each and checking Gripper's movements. If the gripper doesn't move as expected, please make sure that the Gripper integration was set up properly.
+
+If you press "Grab" button while the gripper is opening, it will automatically abort releasing behavior and go to the Close position, effectively cancelling the release command.
+
+:::note
+Note that currently in PX4 the Gripper isn't supported as a standalone driver, but is instead tied to the [payload delivery feature](../advanced_features/package_delivery.md).
+
+Therefore unless you [configure the gripper for package delivery](#setting-up-gripper-for-package-delivery), gripper will not be functioning!
```
hamishwillee, line 44: To confirm, the package delivery setup is what links the mavlink gripper command and the gripper actuator output right?

So if you don't enable the payload delivery feature the system might get the mavlink command to grip/ungrip but won't pass that through to the servo?

en/peripherals/gripper.md
```diff
@@ -0,0 +1,106 @@
...
+
+![High-load gripper example](../../assets/hardware/grippers/highload_gripper_example.jpg)
+
+Grippers are a commonly used mechanism for UAV delivery applications. There are multiple types of interfaces (e.g. PWM, CAN) for different products in the market.
+
+## Supported Gripper Types
+
+PX4 supports the following Gripper types, check the individual page to learn how to integrate them.
+
+- [Servo Gripper with PWM](gripper_servo.md)
```
hamishwillee, line 11: What about mavlink gripper support? Is it just that you don't have one to test with? What I'd expect here is that if you get the gimbal command and it isn't addressed to the FC it should be forwarded to the component that can handle it.

en/peripherals/gripper_servo.md
```diff
@@ -0,0 +1,34 @@
...
+Yellow  | PWM Signal
+
+You would need to connect them into a PWM input of the Flight controller appropriately.
+
+### Compatibility Checks
+
+Before connecting the cable, double-check the following requirements:
+
+* **Signal line voltage level**: Check the data-sheet of your gripper mechanism to find the voltage level of the signal line. And make sure this is compatible with the voltage level of the pins of your Flight Controller.
+* **Power requirements of gripper**: Check the mechanism's data-sheet to find out the power line voltage level requirements. Depending on that, the gripper can be either connected directly to the [power module](../power_module/README.md) or connected to a 5V line.Alternatively a custom voltage regulator can be used to output any other voltage required.
```
hamishwillee, line 30:
```suggestion
* **Power requirements of gripper**: Check the mechanism's data-sheet to find out the power line voltage level requirements. Depending on that, the gripper can be either connected directly to the [power module](../power_module/README.md) or connected to a 5V line.
  Alternatively a custom voltage regulator can be used to output any other voltage required.
```

en/peripherals/gripper.md
hamishwillee: Actually I think "yes", along with the specific configuration for the servo. There might be duplication, but the servo setup will be easier to follow.

en/peripherals/gripper_servo.md
```diff
@@ -0,0 +1,35 @@
+# Servo Gripper with PWM
+
+This section describes how to connect and configure a Servo type Gripper with PWM input.
+
+![High-load gripper example](../../assets/hardware/grippers/highload_gripper_example.jpg)
+
```
hamishwillee, line 6: Do we know the model of this gripper - could we have a list here of supported grippers, with this as the first one?

en/flying/package_delivery_mission.md
```diff
@@ -0,0 +1,84 @@
+# Package Delivery Mission Planning
+
+A package delivery mission allows user to plan and execute a package delivery. Note that this is only supported for vehicles that has a [delivery mechanism hardware](../peripherals/gripper.md) configured & installed.
```
hamishwillee, line 3: All of this only talks about grippers. In the advanced features, you talk about winches. To what level are winches supported?

en/advanced_features/package_delivery.md
```diff
@@ -0,0 +1,49 @@
+# Package Delivery in Mission
```
hamishwillee, line 1: This should all move to the development section somewhere - it is mostly interesting if you want understand how the current things are integrated, and how you might extend it. 

When yuou think it is done, let's get Beat to review this part.

junwoo091400:
> MOre generally, I don't want anything in "advanced features". I plan on removing that because most of what is in there is just a mixed back of features. I plan to move the info at some point and have a single features page in the intro somewhere.

I agree! Having all these *specializations are not so ideal 👍

en/flying/package_delivery_mission.md
junwoo091400: Valid point, it can also drop from the sky. I will update the diagram!

hamishwillee: Cool. Do we also want to mention winches here?

en/peripherals/gripper.md
junwoo091400: Right, ones that we can configure to be a gripper is what we support. And since there are numerous other gripper types (e.g. EPM Gripper: https://ardupilot.org/copter/docs/common-electro-permanent-magnet-V3.html), I narrowed down the scope to a 'Servo Gripper'.

And this shouldn't depend on whether it can be used in mission (right now, gripper driver is tied to `payload_deliverer` module, but it should be a separate entity actually), and ideally every gripper should be controllable via the gripper MAVLink command.

And since the gripper command doesn't work, roboclaw isn't the supported gripper as you said!

en/peripherals/gripper.md
junwoo091400: I moved most of this to the Gripper Servo (along with the image, as it is PWM specific), and left generic comments only in the gripper doc

en/peripherals/gripper.md
junwoo091400: Related to the [comment above](https://github.com/PX4/PX4-user_guide/pull/2072#discussion_r1020336768), gripper's functionality shouldn't be dependent on the package delivery feature ultimately.

And so for now, package delivery setup is a necessary evil for getting gripper to work, but that will be resolved in the future.

en/peripherals/gripper.md
junwoo091400: Good point! And that's totally possible with current setup (haven't tested yet) theoretically. I'm not aware of a real MAVLink Gripper I can get off the shelf & test with. I will ask around and let you know if I get to test it :thinking: 

Or if you are aware of it, let me know as well :+1:

en/peripherals/gripper_servo.md
junwoo091400: Ah sure, I will ask!

junwoo091400: Hmm, it seems like it is just a generic gripper from a company named 'Southco', and it's not really a well known product. Would it be beneficial to include it regardless? (Concern is that it doesn't represent the off the shelf gripper that well)

en/flying/package_delivery_mission.md
junwoo091400: Oh so for now winch is not supported, so technically I should remove that mention in advanced feature, but it is coming in the future.

Shall I remove the mentino?

en/advanced_features/package_delivery.md
junwoo091400: Ah yes, I think it would be nice to have a development section for control allocation, ekf, etc :thinking: 

Definitely not in the advanced_features though :thinking:

junwoo091400:
Regarding your comment on mission planning: https://github.com/PX4/PX4-user_guide/pull/2072#discussion_r1017426960 @hamishwillee 

- Auto disarming is disabled by default when in mission, so when we are landed to deliver package, we don't disarm
- Waiting waypoint: That hasn't been considered. Since the timeout is implied inside the Gripper command itself (as it has some acknowledgement mechanism)
- Take-off after package delivery mission item: Yes, this is the way it should be, and having a blunt 'waypoint' right after the delivery mission item is wrong. But since in the GCS having to place an extra 'take-off' waypoint is a bit of a hassle, that wasn't being done :cry: 
- Yes, for VTOL we should handle MC to FW transition as well! Currently I haven't tested it with VTOL

en/flying/package_delivery_mission.md
hamishwillee: @junwoo091400 Do we also want to mention winches here?

en/flying/package_delivery_mission.md
hamishwillee: No, we should mention that we only support grippers right now.

en/peripherals/gripper.md
hamishwillee: I don't know of one. I'll think on how we might integrate this.

en/peripherals/gripper.md
hamishwillee: Yes, I get that. I was hoping you would confirm the "detail". But I think I'm correct.

en/peripherals/gripper_servo.md
hamishwillee: Yes I think so, though perhaps with a disclaimer that this is a tested servo, not necessarily a recommendation.

en/payloads/README.md
```diff
@@ -28,12 +28,15 @@ The following topics show how to *connect* and configure a camera:
...
+## Cargo Drones (Package delivery)
 
-Cargo drones commonly use servos/actuators to trigger cargo release, control winches, and so on.
-PX4 supports actuator triggering using both RC and MAVLink commands.
+Cargo drones commonly use servos/actuators to trigger cargo release, control winches, and so on. PX4 supports actuator triggering using both RC and MAVLink commands, as well as providing a native [package delivery in mission feature](../flying/package_delivery_mission.md).
 
-### RC Payload Control
+### Delivery using Package delivery feature
+
+Read more in [package delivery in mission feature](../flying/package_delivery_mission.md) documentation.
```
hamishwillee, line 37: Note, will expand this a bit.

en/advanced/package_delivery.md
```diff
@@ -0,0 +1,49 @@
+# Package Delivery in Mission
```
hamishwillee, line 1: FYI I moved this here into development section. It will not longer be a target for other docs other than developers. I will pull out some of it for the package delivery intro in cargo.

en/flying/package_delivery_mission.md
```diff
@@ -0,0 +1,91 @@
...
+:::warning
+DO NOT place a "Return to Launch" item right after the Package delivery waypoint (Pre-waypoint + Action waypoint). This will result in vehicle idling at the landed coordinate.
+:::
+
+This will allow the vehicle to take-off from the package delivery waypoint, since that isn't possible with a RTL waypoint. Due to safety reasons([related issue](https://github.com/PX4/PX4-Autopilot/pull/20044)), "Return To Launch" is disabled when vehicle is landed.
+
+
+## Manual Control of Gripper in Missions
+
+If joystick buttons are [configured for the gripper open/close actions]() then these can be triggered manually in a mission or any other mode.
```
hamishwillee, line 88: Note, link broken. I pulled this section out of the gripper doc. I'll add link when I'm happy with the section and have uploaded changes.

en/peripherals/gripper.md
hamishwillee: @junwoo091400 Thanks. So what "interface" is roboclaw? I assume it is not PWM, or it is more complicated than PWM servo? I..e I really don't want to talk about the driver per-se, more about the protocol.

en/peripherals/gripper.md
hamishwillee: Initially I'll just add a note that it is theoretically supported. Still working on this doc.

en/peripherals/gripper.md
```diff
@@ -0,0 +1,107 @@
...
+## Supported Grippers
+
+There are many different gripper mechanisms ("jaws", "fingers", "electromagnets") and interfaces (PWM, CAN, MAVLink, and so on).
+
+PX4 supports grippers that have simple triggers to hold and release, and that use the following interfaces (see linked documents for details):
+
+- [PWM Servo Gripper](gripper_servo.md) - Grippers connected to autopilot PWM outputs
+- **MAVLink Gripper** (Untested) - Grippers that support the [MAV_CMD_DO_GRIPPER](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_GRIPPER) MAVLink command.
+
+:::note
```
hamishwillee, line 23: How does Roboclaw interface work?

en/peripherals/gripper.md
```diff
@@ -0,0 +1,107 @@
...
+
+1. Set [`PD_GRIPPER_EN`](../advanced_config/parameter_reference.md#PD_GRIPPER_EN) parameter to 1 (reboot required after change).
+1. Set [`PD_GRIPPER_TYPE`](../advanced_config/parameter_reference.md#PD_GRIPPER_TYPE) to match your Gripper.
+   For example, set to `Servo` for a [Servo Gripper](gripper_servo.md).
+
+### Gripper Actuator Mapping
+
+To enable the output of the Gripper, set the Function of the output port where the gripper is connected to as `Gripper` in the [Actuators](../config/actuators.md#actuator-outputs) tab.
+
+<!-- Replace this with actuator diagram -->
```
hamishwillee, line 53: @junwoo091400 Have we got a screenshot of a gripper setup in control allocation? Would be good here.

en/peripherals/gripper.md
```diff
@@ -0,0 +1,107 @@
...
+
+- [PWM Servo Gripper](gripper_servo.md) - Grippers connected to autopilot PWM outputs
+- **MAVLink Gripper** (Untested) - Grippers that support the [MAV_CMD_DO_GRIPPER](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_GRIPPER) MAVLink command.
+
+:::note
+The [Roboclaw driver](../modules/modules_driver.md#roboclaw) isn't supported as a Gripper yet.
+However in the future it may get supported as a Gripper.
+:::
+
+## Using a Gripper
```
hamishwillee, line 28: Note, new section - contains all the "using" information.

en/peripherals/gripper.md
```diff
@@ -0,0 +1,107 @@
+# Gripper Peripherals
+
+Grippers are mechanical devices that can be integrated with an unmanned vehicle to hold (grip) and release payloads.
+
+PX4 allows grippers to triggered automatically in [Payload Delivery Missions](../flying/package_delivery_mission.md) or manually using a Joystick.
+
+![High-load gripper example](../../assets/hardware/grippers/highload_gripper_example.jpg)
+
+:::note
```
hamishwillee, line 9: FYI new note. I realised that if used as a generic actuator a gripper is not a gripper. So this links to the actuator docs - if someoen wants to do that they know enough now to find it. The rest of this doc is specific to your solution.

en/peripherals/gripper.md
```diff
@@ -0,0 +1,107 @@
...
+:::
+
+## Supported Grippers
+
+There are many different gripper mechanisms ("jaws", "fingers", "electromagnets") and interfaces (PWM, CAN, MAVLink, and so on).
+
+PX4 supports grippers that have simple triggers to hold and release, and that use the following interfaces (see linked documents for details):
+
+- [PWM Servo Gripper](gripper_servo.md) - Grippers connected to autopilot PWM outputs
+- **MAVLink Gripper** (Untested) - Grippers that support the [MAV_CMD_DO_GRIPPER](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_GRIPPER) MAVLink command.
```
hamishwillee, line 21: Note, so we need to talk about setup and limitations, but I think this is good for now. Perhaps a topic as a follow up when we have one we can test :-)

en/peripherals/gripper.md
```diff
@@ -0,0 +1,107 @@
...
+
+You can test the actions by clicking on the mapped buttons and checking for gripper movement.
+If the gripper doesn't move as expected check the package delivery configuration and actuator mapping are set up properly.
+
+### Enable Pre-ARM Mode
+
+We need to set a parameter to enable [pre-arming](../advanced_config/prearm_arm_disarm.md).
+This keeps the motors disabled but allows the gripper to be opened and closed for attaching the payload (avoiding potential danger from spinning propellers).
+
+Set [`COM_PREARM_MODE`](../advanced_config/parameter_reference.md#COM_PREARM_MODE) to `Always`.
```
hamishwillee, line 72: Do we need a caveat or warning, and if so what? I.e. if it is completely safe to always set prearm mode, then why isn't this always enabled or would default-on. My thinking is that there are case where prearm might be dangerous.

en/peripherals/gripper.md
```diff
@@ -0,0 +1,107 @@
...
+
+### Configure Gripper Actuation Time
```
hamishwillee, line 74: I merged the profiling topic into this, but I have not re-written it yet. So consider subedit "up to here".

en/peripherals/gripper_servo.md
junwoo091400: Finally found it! https://southco.com/en_any_int/r4-em-r22-161

en/peripherals/gripper.md
junwoo091400: For now it just takes in `actuator_controls_0` topic, and directly maps the throttle and yaw output of the flight controller into the drive & turn commands of the Roboclaw. So I am not sure if it would work so well now, and it probably needs to be refactored to use Control allocation Actuators architecture, right @bkueng ?

en/peripherals/gripper.md
junwoo091400: Right, I could add a section on using Gripper as a 'generic actuator', as that is probably the most common use case!

en/peripherals/gripper.md
junwoo091400: Added!

en/peripherals/gripper.md
junwoo091400: Added in https://github.com/PX4/PX4-user_guide/pull/2072/commits/8cb6995af87a95eeadbb32f22a5dad2be1c2b1a2

en/peripherals/gripper.md
junwoo091400: I have asked @MaEtUgR about it!

en/advanced/package_delivery.md
junwoo091400: > I will pull out some of it for the package delivery intro in cargo.

What do you mean by that?

en/flying/package_delivery_mission.md
junwoo091400: I added the url now!

en/flying/package_delivery_mission.md
junwoo091400: Definitely true, thanks for the tip!

en/peripherals/gripper.md
junwoo091400: Yep especially this hasn't been tested :thinking:

en/peripherals/gripper.md
hamishwillee: What I mean is "does it have a PWM interface"? How is it connected? I'd like to say something like "Roboclaw is not yet supported because XXX bus is not integrated with YYYY"

en/peripherals/gripper.md
hamishwillee: A gripper is a gripper (as you have documented). A generic actuator that happens to be mapped to a gripper is not a gripper :-). It's enough to have this note. I have tidied it up.

en/peripherals/gripper_servo.md
hamishwillee: Thanks !!! I moved this note into a "Supported Grippers" section in the servo. This allows us to easily separate the mavlink and other grippers into particular types, depending on what people want to do.

en/flying/package_delivery_mission.md
hamishwillee: Impossible after the fact. But you could halve the height (as shown) and then zoom the mission. Obviously I can't fake zooming the mission, but you can see how you might do this.

image

en/peripherals/gripper.md
hamishwillee: @junwoo091400 I have updated the whole of the gripper and gripper servo docs. For those docs only question is this one ^^^

hamishwillee:
@junwoo091400 Nearly there. Just got the mission document itself to go (payloads, gripper, gripper servo, and development docs done), Will try to finish that tomorrow.

hamishwillee:
Sorry, did not get to this today. Tomorrow!!!

en/advanced/package_delivery.md
hamishwillee: What I mean is that if some of this is relevant for "end users" rather than "developers", I pull put it in the end user docs.

en/peripherals/gripper.md
hamishwillee: Thanks @junwoo091400 - What did he say?

hamishwillee:
@junwoo091400 I've subedited this all now.

The mission planning is much simpler - we don't need that image or any of that stuff about "pre waypoints" as it is too many words for a simple concept. Essentially you have a mission item to say where you make the release and another to make the release. If you want the release on the ground then you turn it into a land waypoint. I kept your examples though and added some of the other information you provided.

Also all the stuff about VTOL vs MC vs FW is not really necessary to mention - its just whether the vehicle can land, and whether the package can safely be dropped.

I think this is good. It would be nice to iterate with answers to https://github.com/PX4/PX4-user_guide/pull/2072#discussion_r1023674051 and how roboclaw works, but we can wait on that.

I will merge so this doesn't have to wait for additional feedback. BUT I would like you to review again please!!!!!

en/peripherals/gripper.md
MaEtUgR: I said I so far always enabled prearm mode especially for preflight surface checks. I'm guessing it was previously disabled by default such that if you accidentally plug a motor on a servo pin the chances are lower that it spins up. But that assumption is flawed since servos usually have 1500 as disarmed values. Also assignment of actuators is now much better with control allocation. So the only remaining thing I see is to save the servos from extra work while the vehicle is powered on the bench 🤷‍♂️ But a fixed-wing pro like @sfuhrer might have more insights.

hamishwillee: @MaEtUgR Thanks. IMO we don't particularly need a warning. As you say, prearm is only unsafe if somehow motors or "dangerous actuators" can move in this state. So possibly risky on first setup, but not for use once properly configured.

MaEtUgR: Yes, exactly 👍

sfuhrer: Context: https://github.com/PX4/PX4-Autopilot/pull/17309#issuecomment-992205444
TL;DR: I think we shouldn't just change the default to "Always", but stay on the safe side.
A user then should make the deliberate decision (after setting up the vehicle the first time and testing the wiring) whether one wants to have the servos already enabled by without arming or in prearm mode. But in general "arming" should be active for all actuators, not just the motors.

hamishwillee: Fair enough. I think at this point you've probably already verified your motors and servos are set up properly and prearm is safe. So no need for a warning. No harm on being on the safe side for the setting either.

PR #2074, opened 2022-10-03
failsafe: add simulation page & update existing page

bkueng:
Updates from https://github.com/PX4/PX4-Autopilot/pull/20172.

I'll add the CI logic later on to automatically update the simulation from PX4.


en/config/safety.md
```diff
@@ -20,18 +20,20 @@ Action | Description
 <a id="action_hold"></a>[Hold mode](../flight_modes/hold.md) | The vehicle will enter *Hold mode*. For multicopters this means the vehicle will hover, while for fixed/wing the vehicle will circle.
 <a id="action_return"></a>[Return mode](../flight_modes/return.md) | The vehicle will enter *Return mode*. Return behaviour can be set in the [Return Home Settings](#return-mode-settings) (below).
 <a id="action_land"></a>[Land mode](../flight_modes/land.md) | The vehicle will enter *Land mode*, and lands immediately.
+<a id="action_disarm"></a>Disarm | Stops the motors immediately.
```
hamishwillee, line 23: How does this differ from "Lockdown" option below?

hamishwillee: More precisely, I see these options in the simulation:
image 
image

So would I be correct in saying:
- There is no "lockdown action".
- Terminate is flight termination, disarm is disarm. They both disarm the motors, but flight termination also sets failsafe values. FMI, why would you need both?

en/config/safety_simulation.md
```diff
@@ -0,0 +1,29 @@
...
+
+The simulation can also be executed locally:
+
+```sh
+make run_failsafe_web_server
+```
+
+ <iframe src="failsafe/index.html?no-title=1" frameborder="0" height="1400px" style="text-align: center; margin-left: -230px; margin-right: -230px;" width="1200"></iframe>
+
+
```
hamishwillee, line 29:
```suggestion

```

en/config/safety_simulation.md
```diff
@@ -0,0 +1,27 @@
...
+     The failsafe state machine can override this in case of a failsafe.
+4. Check the action under **Output**
+5. Check what happens when changing mode or **Move the RC sticks**
+6. Play with different settings and conditions!
+
+
+The simulation can also be executed locally:
+
+```sh
+make run_failsafe_web_server
```
hamishwillee, line 24: This didn't work for me on today's main and was not shown in `make list_config_targets`. 
- This is coming?
- I presume it runs a server and the command prompt shows you what URL to look at?
- This is useful if you want to try it with your own failsafe changes? i.e. why would you want to do this?

en/config/safety_simulation.md
```diff
@@ -0,0 +1,27 @@
+# Failsafe State Machine Simulation
+
+This page can be used to simulate the actions of the PX4 failsafe state machine under all possible configurations and conditions.
+
+The simulation runs the same code in the browser as is executed on the vehicle, in real-time (the simulation is automatically kept in sync with the latest version of the code).
+Note that any delayed action (`COM_FAIL_ACT_T`) will also be delayed in the simulation.
+
+To use it:
+
+1. First configure the parameters on the left.
```
hamishwillee, line 10: FYI I modified the list a bit, but not sure if it is better or not. I wanted to make clear that you must do the parameters and vehicle first, and that all other settings may impact each other - i.e. you might want to change conditions before or after you change things in the state.

en/config/safety_simulation.md
```diff
@@ -0,0 +1,27 @@
...
+6. Play with different settings and conditions!
+
+
+The simulation can also be executed locally:
+
+```sh
+make run_failsafe_web_server
+```
+
+ <iframe src="failsafe/index.html?no-title=1" frameborder="0" height="1400px" style="text-align: center; margin-left: -230px; margin-right: -230px;" width="1200"></iframe>
```
hamishwillee, line 27: FYI only, there is a lot of whitespace displayed after the embedded content.

en/config/safety.md
hamishwillee: PS IFF this is the same result/danger as termination is it protected by the circuit breaker? (CBRK_FLIGHTTERM)

hamishwillee:
This looks good. Happy to merge - just wanted to give you a chance to see/respond to the comments and subedits.

I played with the simulator a bit. It's great, but would be harder to use if you didn't already know what the various params are. Are the parameters manually added?

Looking at the params you show, these are new or not documented in safety. Are they important? Can you tell me more about them? I would like to add specific sections on what causes the failsafe and what settings are appropriate.
- COM_ACT_FAIL_ACT
- COM_IMB_PROP_ACT
- COM_LKDOWN_TKO
- COM_QC_ACT
- COM_SPOOLUP_TIME

There are some we document that don't seem to be in the simulation - are they relevant? E.g.
- COM_POS_FS_DELAY
- CBRK_VELPOSERR - and any other CBRK_
- NAV_GPSF_LT
- etc. Just wondering where the cutoff is.

en/config/safety_simulation.md
bkueng: Yes that's on purpose as it will grow over time. I did not see a way to make it fit to the actual content.

en/config/safety_simulation.md
bkueng: The PR isn't merged yet, but you can test it there. It will tell you to first install the SDK then point you to the url.

> This is useful if you want to try it with your own failsafe changes? i.e. why would you want to do this?

Yes for development. Or if you use a specific version where you want to verify the exact behavior.

en/config/safety.md
bkueng: I'm removing the lockdown action and replace it with disarming, as this should be easier to understand.
Lockdown is currently still used for the kill switch, so it's not removed completely.

bkueng:
The parameters are automatically added - it's all the parameters the failsafe state machine accesses. So they all have an effect, one way or another.
Would it help to also show the full param description? It currently only shows the short description on the '?'.

The other parameters are indirectly relevant: they define when/if the corresponding conditions get triggered. That logic is outside of the failsafe state machine (and thus the simulation).

en/config/safety_simulation.md
hamishwillee: It does no harm. I'm fine with this. But thanks for letting me know.

en/config/safety.md
hamishwillee: @bkueng So to be clear:
1. Lockdown is not a failsafe action - it exists for the kill switch but is not something you can select as an action for any failsafe. I will therefore remove it.
2. The term "disarm" is certainly more clear than "lockdown"
3. Disarm turns off motors, while flight termination sets stuff to it's failsafe values (i.e. turns off motors, but might deploy stuff)
   - Is that right?
   - IF so, why is this considered less severe than flight termination? Is it because disarming is revertable? 
   - This is just as dangerous as flight termination right - the vehicle drops out of the air. Is it interlocked by CBRK_FLIGHTTERM or some other circuit breaker?
   - Can flight termination do other stuff, like quadchute? My assumption is NO.

en/config/safety_simulation.md
```diff
@@ -0,0 +1,27 @@
...
+2. Set the vehicle type
+3. Set the other values in the **State** or any of the flags under **Conditions**
+   - The **Intended Mode** corresponds to the commanded mode via RC or GCS (or external script).
+     The failsafe state machine can override this in case of a failsafe.
+4. Check the action under **Output**
+5. Check what happens when changing mode or **Move the RC sticks**
+6. Play with different settings and conditions!
+
+
+The simulation can also be executed locally:
```
hamishwillee, line 21:
```suggestion
The simulation can also be executed locally in order to test a specific version or set of changes:
```

en/config/safety_simulation.md
hamishwillee: OK. This will all work when you merge then. Sounds good.

hamishwillee:
> The parameters are automatically added - it's all the parameters the failsafe state machine accesses. So they all have an effect, one way or another.
>Would it help to also show the full param description? It currently only shows the short description on the '?'.

I don't think so. I just think this is developer stuff, and is limited by the fact that you have to know what fits with what. Also some of the headings dont seem to match content - i.e. this isn't all battery stuff.

image

I'm not sure what would make this easier. 
- Perhaps a ? next to each condition indicating which modes it might affect - i.e. I can guess "Global position estimate invalid" affects any transition into an auto mode, but a newbie might not. Personally I don't know what "Angular velocity invalid" might affect.
- Perhaps some kind of cross linking between parameters and conditions - For example, if I select a condition that affects battery, colour-code highlighting the battery params - and so on.
- Filter the whole lot by one failsafe - e.g. battery, and only display those parts.

But even it you do nothing this is still good :-). It is possible for people to dig.

en/config/safety.md
bkueng: Correct.

> Is it interlocked by CBRK_FLIGHTTERM or some other circuit breaker?

No, you need to actively select it in the first place.

> Can flight termination do other stuff, like quadchute? My assumption is NO.

No

bkueng:
> Also some of the headings dont seem to match content - i.e. this isn't all battery stuff.

There's a new paragraph, just without title - we could add more spacing there.
Feel free to play around with the layouting, or let me know if I should change it.

> Perhaps a ? next to each condition indicating which modes it might affect - i.e. I can guess "Global position estimate invalid" affects any transition into an auto mode, but a newbie might not. Personally I don't know what "Angular velocity invalid" might affect.

This could be done, is a bit of work though.

The other suggestions make sense as well - but it's all manual work that need to be maintained (e.g. when parameters change).

hamishwillee:
Thanks @bkueng , I'm going to merge this. Will add a comment on the rest after.

hamishwillee:
W.r.t. the battery thingy, the paragraph break isn't sufficient. You've otherwise done a good job of using grouping to make it "somewhat clear" what condition might apply to what param. 

I would put headings for the other two groupings instead of paragraphs - if there are any that are sensible. If not, then you might minimally put a heading "Other" after the battery conditions to cover both groups.

I'd also consider having text after the **Mode** heading having text something like "Conditions affecting switch to Intended mode" so the relationship to the field in "State" is spelled out.

I don't see what else you can do without creating some kind of manual link between conditions an parameters. I think it's a legitimate concern for ongoing maintenance. If it were me I'd wait and see how much problem people really have with this before investing more time.

hamishwillee:
@bkueng Further, I am a muppet, having just merged this. Do you think that the PR will go in within the next week or so? If so I'd rather leave this in. IF it is going to be a while I'll revert and recreate the PR.

PR #2080, opened 2022-10-05
removed unnecessary '$' from commands

khanasif786:


hamishwillee (approved):
Thanks, I agree. Generally you only need a $ to differentiate between commands you enter and responses from the system, but that does not apply here.

PR #2083, opened 2022-10-06
Remove MOT_ORDERING from Kakute

hamishwillee:
The `MOT_ORDERING` is no longer in PX4 but was marked as an important setting.
I have deleted it, but is there something else we should be saying here?

@bkueng No urgency, but I'm sure you'll know.

### DIFF
--- a/en/flight_controller/kakutef7.md
+++ b/en/flight_controller/kakutef7.md
@@ -9,7 +9,6 @@ The *Kakute F7* from Holybro is a flight controller board designed for racers.
 
 <img src="../../assets/flight_controller/kakutef7/board.jpg" width="400px" title="Kakute F7" />
 
-
 :::note
 This flight controller is [manufacturer supported](../flight_controller/autopilot_manufacturer_supported.md).
 :::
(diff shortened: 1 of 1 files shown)


bkueng (approved):
We could say that it comes with the betaflight motor ordering, and if you use the motor ordering/assignment UI it will take care of this.

en/flight_controller/kakutef7.md
```diff
@@ -97,12 +96,13 @@ The firmware can be installed in any of the normal ways:
 
 ## Configuration
 
+If you use a 4-in-1 ESC with Betaflight/Cleanflight motor assignment you can use the [Actuator](../config/actuators.md) configuration UI to set the motor ordering appropriately.
```
hamishwillee, line 99: Thanks @bkueng . I don't understand what the Betaflight assignment means but I assume it is some kind of standard output wiring to motor assignment, and that this doesn't matter to us since we can set any function to any output in software.

If so, there is nothing much other than this to say, so I am merging.
Thanks!

bkueng: It's the motor ordering: https://learnassets.getfpv.com/learn/wp-content/uploads/2019/06/26225548/Motor-Order.jpg. But yes, it doesn't matter, it's just different from the PX4 default.

PR #2085, opened 2022-10-06
Gimbal - replace mixer with control allocation

hamishwillee:
This attempts to update the Gimbal docs to replace old mixer information with control allocation.

I've moved it from development into Hardware/Flight Controller Peripherals since it is not "developer only" (to me that means "I am writing PX4 code", not "I am using existing code" - this is for integrators).

@julianoes Can you sanity check the changes from a gimbal point of view. It might be worth us saying a bit about PX4 being a gimbal manager? 

@bkueng Can you sanity check this from a control allocation point of view. 
- I assume that we are no longer tied to AUX ports for the allocation.

Other comments inline.


en/SUMMARY.md
```diff
@@ -290,6 +290,7 @@
         * [Zubax Orel](dronecan/zubax_orel.md)
       * [VESC](peripherals/vesc.md)
     * [Camera](peripherals/camera.md)
+    * [Gimbal \(Mount\) Configuration](advanced/gimbal_control.md)
```
hamishwillee, line 293: Note, moved to "integrator space".

en/advanced/gimbal_control.md
```diff
@@ -1,82 +1,64 @@
-# Gimbal Control Setup
+# Gimbal Configuration
```
hamishwillee, line 1: @julianoes Normally we'd have a parent gimbal page, that might include some overview and purchase links. This is more like the camera page adjacent - it is all about config.

Are there any gimbals we know work?

en/advanced/gimbal_control.md
```diff
@@ -1,82 +1,64 @@
...
+After selecting the input mode, reboot the vehicle to start the mount driver.
+
+You should set `MNT_MODE_IN` to one of: `RC (1)`, `MAVlink gimbal protocol v2 (4)` or `Auto (0)` (the other options are deprecated).
+If you select `Auto (0)`, the gimbal will automatically select either RC or or MAVLink input based on the latest input (note that to switch from MAVLink to RC requires a large stick motion).
+
+The output is set using the [MNT_MODE_OUT](../advanced_config/parameter_reference.md#MNT_MODE_OUT) parameter.
+This is set to output on a PXM port by default (`AUX (0)`).
+If the [MAVLink Gimbal Protocol v2](https://mavlink.io/en/services/gimbal_v2.html) is supported by your gimbal, you should instead use/select `MAVLink gimbal protocol v2 (2)`.
+
+The full list of parameters for setting up the mount driver can be found in [Parameter Reference > Mount](../advanced_config/parameter_reference.md#mount) (the most relevant values are described below).
```
hamishwillee, line 28: There are a whole bunch of offset parameters etc - I assume these are all still "valid" in the age of control allocation?

en/advanced/gimbal_control.md
```diff
@@ -1,82 +1,64 @@
...
 
 ## MAVLink Gimbal (MNT_MODE_OUT=MAVLINK)
 
-To enable a MAVLink gimbal, first set parameter [MNT_MODE_IN](../advanced_config/parameter_reference.md#MNT_MODE_IN) to `MAVLINK_DO_MOUNT` and [MNT_MODE_OUT](../advanced_config/parameter_reference.md#MNT_MODE_OUT) to `MAVLINK`.
```
hamishwillee, line 24: Note, updated to indicate use the current API. Reasonable?

en/advanced/gimbal_control.md
```diff
@@ -1,82 +1,64 @@
...
+A common configuration is to have a serial connection to the gimbal from the Flight Controller `TELEM2` port (assuming `TELEM2` is free).
 For this configuration you would set:
 - [MAV_1_CONFIG](../advanced_config/parameter_reference.md#MAV_1_CONFIG) to **TELEM2** (if `MAV_1_CONFIG` is already used for a companion computer (say), use `MAV_2_CONFIG`).
 - [MAV_1_MODE](../advanced_config/parameter_reference.md#MAV_1_MODE) to **NORMAL**
 - [SER_TEL2_BAUD](../advanced_config/parameter_reference.md#SER_TEL2_BAUD) to manufacturer recommended baud rate.
 
-This will enable the user to command the gimbal using [MAV_CMD_DO_MOUNT_CONTROL](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_MOUNT_CONTROL) and [MAV_CMD_DO_MOUNT_CONFIGURE](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_MOUNT_CONFIGURE).
-
+This will enable MAVLink to command the gimbal using [MAV_CMD_DO_MOUNT_CONTROL](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_MOUNT_CONTROL) and [MAV_CMD_DO_MOUNT_CONFIGURE](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_MOUNT_CONFIGURE).
```
hamishwillee, line 42: So this subset of messages are the original ones. @julianoes I think these are part of MAVLink Gimbal protocol v2 as well right? Are there any other new commands etc we should mention here?

en/advanced/gimbal_control.md
```diff
@@ -1,82 +1,64 @@
...
+## Gimbal on Flight Controller PWM Output (MNT_MODE_OUT=AUX)
 
-A mixer file is required to define the mapping for the output pins and the [mount mixer](https://github.com/PX4/PX4-Autopilot/blob/main/ROMFS/px4fmu_common/mixers/mount.aux.mix) is automatically selected (this overrides any AUX mixer provided by the airframe configuration).
+The gimbal can be connected to the Flight controller PWM ports by setting the output mode to `MNT_MODE_OUT=AUX`.
 
-The output assignment is as following:
-- **AUX1**: Pitch
-- **AUX2**: Roll
-- **AUX3**: Yaw
-- **AUX4**: Shutter/retract
```
hamishwillee, line 47: @bkueng There does not appear to be a gimbal shutter/retract CA option. Does there need to be/do I need to mention anything here? 

I suspect that's a camera thing and should not be mentioned.

en/advanced/gimbal_control.md
```diff
@@ -1,82 +1,64 @@
...
+The output pins that are used to control the gimbal are set in the [Acuator Configuration > Outputs](/en/config/actuators.md#actuator-outputs) by selecting any three unused Actuator Outputs and assigning the following output functions:
+- `Gimbal Roll`: Output controls gimbal roll.
+- `Gimbal Pitch`: Output controls Gimbal pitch.
+- `Gimbal Yaw`: Output controls Gimbal pitch.
 
-### Customizing the mixer configuration
+For example, you might have the following settings to assign the gimbal roll, pitch and yaw to AUX1-3 ouptuts.
```
hamishwillee, line 54:
```suggestion
For example, you might have the following settings to assign the gimbal roll, pitch and yaw to AUX1-3 outputs.
```

en/advanced/gimbal_control.md
```diff
@@ -1,82 +1,64 @@
...
-M: 1
-O:      10000  10000      0 -10000  10000
-S: 2 1  10000  10000      0 -10000  10000
-
-# yaw
-M: 1
-O:      10000  10000      0 -10000  10000
-S: 2 2  10000  10000      0 -10000  10000
-```
+![Gimbal Actuator config](../../assets/config/actuators/qgc_actuators_gimbal.png)
```
hamishwillee, line 56: The image is just a control actuator image at top of this page. 

How do we choose the ranges to set here for the min, max, disarmed? 

I.e. is the same as in http://docs.px4.io/main/en/config/actuators.html#output-assignment-manual - you basically wiggle the actuator test bar to confirm the gimbal moves in the right axis, then set the range such that middle is disarmed position, and there is an up range and down range around the central point. So you increase to top value that the gimbal can physically move and that is the PWM value to set, ditto for the bottom of the range?

en/advanced/gimbal_control.md
```diff
@@ -85,21 +67,23 @@ make px4_sitl gazebo_typhoon_h480
...
 Start the simulation with (no parameter needs to be changed for that):
+
 ```
 make px4_sitl gazebo_typhoon_h480
 ```
+
 Make sure it's armed, eg. with `commander takeoff`, then use the following command to control the gimbal (for example):
+
 ```
 vmount test yaw 30
```
hamishwillee, line 84: Does this still work.

en/advanced/gimbal_control.md
julianoes: We should no longer mention these old commands and use the v2 ones instead.

en/advanced/gimbal_control.md
julianoes: 1. This should still work.
2. That would be nice, if you feel like trying it out.

en/advanced/gimbal_control.md
bkueng: Yes.

en/advanced/gimbal_control.md
bkueng: Yes that was removed.

en/advanced/gimbal_control.md
bkueng: Yes either by testing or the gimbal docs might specify the range.

hamishwillee: THanks. I added a note to this effect.

en/advanced/gimbal_control.md
hamishwillee: Not really, at this time :-).

en/advanced/gimbal_control.md
```diff
@@ -1,82 +1,72 @@
...
-For this configuration you would set:
+For example, if the `TELEM2` port on the flight controller is unused you can connect it to the gimbal and set the following PX4 parameters:
 - [MAV_1_CONFIG](../advanced_config/parameter_reference.md#MAV_1_CONFIG) to **TELEM2** (if `MAV_1_CONFIG` is already used for a companion computer (say), use `MAV_2_CONFIG`).
 - [MAV_1_MODE](../advanced_config/parameter_reference.md#MAV_1_MODE) to **NORMAL**
 - [SER_TEL2_BAUD](../advanced_config/parameter_reference.md#SER_TEL2_BAUD) to manufacturer recommended baud rate.
 
-This will enable the user to command the gimbal using [MAV_CMD_DO_MOUNT_CONTROL](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_MOUNT_CONTROL) and [MAV_CMD_DO_MOUNT_CONFIGURE](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_MOUNT_CONFIGURE).
+With this setup, ground stations can send [MAVLink Gimbal Commands](https://mavlink.io/en/services/gimbal_v2.html#gimbal-manager-messages) to PX4, such as `MAV_CMD_DO_SET_ROI_LOCATION` to track region of interest and `MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW` to control the gimbal attitude, and so on.
 
+:::note
```
hamishwillee, line 46: @julianoes I'm going to merge this as "better than it was" and we can iterate. I think this is probably close to correct, but perhaps not "complete"

However can you re-review this for sanity?

For example, this note states my assumption that this setup makes PX4 a gimbal manager. That might not be so.

A few more questions:

1. Are there any mavlink gimbal protocol v2 we know might work?
2. If they use an old protocol gimbal, can you still drive it using the new protocol messages - e.g. such as setting the ROI. Similarly, can I drive a PWM gimbal with all the instructions in the gimbal manager docs. The docs say that the gimbal manager will translate the input to the format expected but I suspect it might be more nuanced than that? Or to put it another way, does the PX4 gimbal manager do everything you might expect to convert all gimbal manager commands to roll, pitch, yaw ?
3. If not, what isn't supported?
4. Given the gimbal manager does all those things (say) why would we recommend a mavlink gimbal over using PX4 with a PWM connected gimbal?
5. What should someone do if there is a mavlink gimbal on the system that supports gimbal manager? Do we need to document that?

en/advanced/gimbal_control.md
hamishwillee: Done, I pointed to the gimbal manager commands with a few examples

en/advanced/gimbal_control.md
julianoes: 1. Gremsy
2. Yes, PX4 translates between v2 and v1, or v2 and PWM. So it should do it all. If you find something that doesn't work right, it's a bug.
3. Don't know.
4. If you want to fiddle with PWM, feel free. If something should just work then v2 would be recommended
5. You mean a gimbal that implements the manager? In that case don't enable gimbal control in PX4. In that case commands and mission items would be sent to the gimbal manager.
6. QGC is not really maintained or updated so :shrug:.

PR #2090, opened 2022-10-08
Replace referrals to master by main

michael-aviate-it:
As suggested by https://github.com/github/renaming and obviously already conducted in the PX4 repo, I would like to contribute by updating the documentation.
(Only stumbled over it, since I got a "fatal: Couldn't find remote ref master" error when following it through ;) )


hamishwillee (approved):
Thanks very much!

PR #2094, opened 2022-10-10
New PX4-Bridge

akshata-01:
- Intro
- Why this approach?
- Architectural Overview
- A note in the RTPS has been added.

### DIFF
--- a/en/middleware/microdds.md
+++ b/en/middleware/microdds.md
@@ -1,0 +1,37 @@
+# MicroDDS
+
+The *PX4-Fast RTPS(DDS) Bridge*, which was also referred to as as the *microRTPS Bridge*, has now been replaced with MicroDDS bridge.
+
+The new approach provides a faster and simpler method of integrating applictaions running and linked in DDS domains (including ROS nodes), making it easy to share sensor data, commands, and other vehicle information.
+
+The following guide describes the new PX4 bridge architecture, and shows the reader how to write a simple *Micro DDS* application to subscribe * publish telemetry updates from the PX4 Autopilot.
+
+:::note
+Micro XRCE-DDS is a software solution that enables communictaion with an existing DDS network. 
+:::
+
+## Why this approach is of interest?
+
+Initially, micro-RTPS served as the middleware and its consistent improvement evolved into what's known as the microDDS that provides the user a native interface between the flight controller and the mission computer. 
+The notable difference between the two is what kind of transport protocol each supports. The newest version supports UDP/TCP/Serial/Custom transport protocol.  
+
+The Micro-XRCE-DDS-Gen has 2 major advantages, i.e. it can be used to generate client code and build px4_msgs along with microdds_client together creating a single library.  
+
+:::tip
+
+:::
+
+## Architectural overview
+
+### microDDS Bridge
+
+The [*microDDS*](https://micro-xrce-dds.docs.eprosima.com/en/stable/introduction.html) provides DDS publisher subcription approach wherein the client is on a low-resource consumption device and the agent is connected to with the DDS gobal data space.
+
+#### The microdds Client
+The *Client* is the PX4 Autopilot middleware daemon process that runs on the flight controller. The client can either publish or subcribe directly to data topics because the agent itself acts on behalf of the [client](https://micro-xrce-dds.docs.eprosima.com/en/stable/client.html)
+
+#### The microdds Agent
+The *Agent* runs as a daemon process that's defined by what the clients defines it to be and hence it's dynamic. It's a server that acts an an intermediary between the client and the DDS Global Data Space. 
+
+#### microdds Communication
+microDDS has APIs that allow writing time-critical applications and can be used over many transport protocols supporting TCP,UDP over Ethernet Wi-Fi & 6LoWPAN and Bluetooth. The user can also customise this making microDDS transport-agnostic.
--- a/en/middleware/micrortps_copy.md
+++ b/en/middleware/micrortps_copy.md
@@ -1,0 +1,388 @@
+# RTPS/DDS Interface: PX4-Fast RTPS(DDS) Bridge
+
+:::note
+This approach now has an advanced version evolved into what's called the Micro-DDS PX4 Bridge which provides seamless communication between the autopilot and mission computer.
+:::
+
+The *PX4-Fast RTPS(DDS) Bridge*, which is also referred to as as the *microRTPS Bridge*, adds a Real Time Publish Subscribe (RTPS) interface to the PX4 Autopilot, enabling the exchange of [uORB messages](../middleware/uorb.md) between the various PX4 Autopilot internal components and (offboard) *Fast DDS* applications in realtime.
+
+This allows us to better integrate with applications running and linked in DDS domains (including ROS nodes), making it easy to share sensor data, commands, and other vehicle information.
+
+The following guide describes the RTPS/DDS bridge architecture, and shows the reader how to write a simple *Fast DDS* application to subscribe to telemetry updates from the PX4 Autopilot.
+
+:::note
+RTPS is the underlying protocol of the Object Management Group's (OMG) Data Distribution Service (DDS) standard.
+It aims to enable scalable, real-time, dependable, high-performance and inter-operable data communication using the publish/subscribe pattern.
+
+*Fast DDS* is a very lightweight cross-platform implementation of the latest version of the RTPS protocol / DDS middleware.  It was previously named *Fast RTPS*.
+:::
+
+## When should RTPS be used?
+
+RTPS should be used when you need to reliably share time-critical/real-time information between the flight controller and offboard components. It is instrumental in cases where offboard software needs to become a *peer* of software components running in PX4 (sending and receiving uORB topics).
+
+Possible use cases include communicating with robotics libraries for computer vision and other use cases where real-time data to/from actuators and sensors is essential for vehicle control.
+
+*Fast DDS* is not intended as a replacement for MAVLink. [MAVLink](../middleware/mavlink.md) remains the most appropriate protocol for communicating with ground stations, gimbals, cameras, and other offboard components (although *Fast DDS* may open other opportunities for working with some peripherals at higher frequencies).
+
+:::tip
+It is possible to use Fast RTPS(DDS) over slower links (e.g., radio telemetry) by being mindful of your link's extra constraints. Keep in mind you can easily overload your telemetry channel.
+:::
+
+## Architectural overview
+
+### microRTPS Bridge
+
+The *microRTPS* bridge exchanges messages between PX4 and DDS-participant applications, seamlessly converting between the [uORB](../middleware/uorb.md) and RTPS/DDS messages used by each system.
+
+![basic example flow](../../assets/middleware/micrortps/architecture.png)
+
+The main elements of the architecture are the client and agent processes shown in the diagram above.
+
+#### The microRTPS Client
+The *Client* is the PX4 Autopilot middleware daemon process that runs on the flight controller. This client subscribes to uORB topics published by other PX4 Autopilot components and sends any updates to the *Agent* (via a UART or UDP port), and also receives messages from the *Agent* and publishes them as uORB messages to the PX4 Autopilot.
+
+#### The microRTPS Agent
+The *Agent* runs as a daemon process on an offboard computer (outside the flight controller). This agent watches for uORB update messages from the *Client* and (re)publishes them over RTPS, and also subscribes to "uORB" RTPS/DDS messages from other DDS-participant applications and forwards them to the *Client*.
+
+#### microRTPS Agent/Client Communication
+The *Agent* and *Client* are connected via a serial link (UART) or UDP network, and the uORB information is [CDR serialized](https://en.wikipedia.org/wiki/Common_Data_Representation) before being sent (*CDR serialization* provides a common format for exchanging serial data between different platforms).
+
+The *Agent* and any *Fast DDS* applications are connected via UDP and may be on the same or another device. In a typical configuration, they will be on the same system (e.g., a development computer, Linux companion computer, or compute board), connected to the *Client*. This can be through a Wifi link or USB.
+
+## Code generation
+
+### Dependencies
+Fast DDS 2.0.0 or later and Fast-RTPS-Gen 1.0.4 (not later!) must be installed in order to generate the required code, and continue to the next steps. [Follow the installation guide.](../dev_setup/fast-dds-installation.md)
+
+:::note
+RTPS has been adopted as the middleware for the ROS 2 (Robot Operating System).
+
+For information about how to use this interface within the ROS 2 applications and development workflows, see [PX4-ROS 2 bridge](../ros/ros2_comm.md).
+:::
+
+### Fast RTPS(DDS) Applications (ROS Independent)
+
+All the code needed to create, build and use the bridge is automatically generated when PX4-Autopilot is compiled.
+
+The *Client* application is also compiled and built into the firmware as part of the normal build process.
+The *Agent* must be separately/manually compiled for the target computer.
+
+:::tip
+Most users will not need to do so. Still, the bridge can be [manually generated](micrortps_manual_code_generation.md), providing a more detailed overview of the build process and useful for troubleshooting.
+:::
+
+## Supported uORB messages
+
+The generated bridge code will enable a specified subset of uORB topics to be published/subscribed via RTPS, regardless if you are deploying a ROS application or not.
+
+For *automatic code generation* there's a *yaml* definition file in the PX4 **PX4-Autopilot/msg/tools/** directory called **uorb_rtps_message_ids.yaml**. This file defines the set of uORB messages to be used with RTPS, whether the messages are to be sent, received or both, and the RTPS ID for the message to be used in DDS/RTPS middleware.
+
+:::note
+It's essential to note that every RTPS message **needs** an ID to be set in this file.
+:::
+
+```yaml
+rtps:
+  - msg: actuator_armed
+    id: 0
+  - msg: actuator_control
+    id: 1
+  - ...
+  - msg: airspeed
+    id: 5
+    send: true
+  - msg: battery_status
+    id: 6
+    send: true
+  - msg: camera_capture
+    id: 7
+  - msg: camera_trigger
+    id: 8
+    receive: true
+  - ...
+  - msg: sensor_baro
+    id: 63
+    receive: true
+    send: true
+```
+
+<a id="client_firmware"></a>
+## Client (PX4/PX4-Autopilot)
+
+The *Client* source code is generated, compiled and built into the PX4 Autopilot firmware as part of the normal build process.
+
+To build the PX4 Autopilot firmware for NuttX/Pixhawk flight controllers use the `_rtps` feature in the configuration target.
+
+For example, to build RTPS for px4_fmu-v4:
+```sh
+make px4_fmu-v4_rtps
+```
+
+To build the firmware for a SITL target:
+```sh
+make px4_sitl_rtps
+```
+
+The *Client* application can be launched from [NuttShell/System Console](../debug/system_console.md).
+The command syntax is shown below (you can specify a variable number of arguments):
+
+```sh
+> micrortps_client start|stop|status [options]
+  -t <transport>          [UART|UDP] Default UART
+  -d <device>             UART device. Default /dev/ttyACM0
+  -l <loops>              How many iterations will this program have. -1 for infinite. Default -1.
+  -w <sleep_time_ms>      Time in ms for which each iteration sleep. Default 1ms
+  -b <baudrate>           UART device baudrate. Default 460800
+  -p <poll_ms>            Time in ms to poll over UART. Default 1ms
+  -r <reception port>     UDP port for receiving. Default 2019
+  -s <sending port>       UDP port for sending. Default 2020
+  -i <ip_address>         Select IP address (remote) values: <x.x.x.x>. Default: 127.0.0.1
+```
+
+:::note
+By default the *Client* runs as a daemon, but it does not automatically start, and must be manually launched.
+
+The PX4 Autopilot firmware initialization code may in future automatically start the *Client* as a permanent daemon process.
+:::
+
+For example, in order to run the *Client* daemon with SITL connecting to the Agent via UDP:
+
+```sh
+micrortps_client start -t UDP
+```
+
+## Agent in an Offboard Fast DDS interface (ROS-independent)
+
+The *Agent* code is automatically *generated* when you build the associated PX4 Autopilot firmware, and you can find the source here: `build/<target-platform>/src/modules/micrortps_bridge/micrortps_client/micrortps_agent/`.
+
+To build the *Agent* application, compile the code:
+
+```sh
+cd build/<target-platform>/src/modules/micrortps_bridge/micrortps_client/micrortps_agent
+mkdir build && cd build
+cmake ..
+make
+```
+
+The command syntax for the *Agent* is listed below:
+
+```sh
+$ ./micrortps_agent [options]
+  -t <transport>          [UART|UDP] Default UART.
+  -d <device>             UART device. Default /dev/ttyACM0.
+  -w <sleep_time_us>      Time in us for which each iteration sleep. Default 1ms.
+  -b <baudrate>           UART device baudrate. Default 460800.
+  -p <poll_ms>            Time in ms to poll over UART. Default 1ms.
+  -r <reception port>     UDP port for receiving. Default 2019.
+  -s <sending port>       UDP port for sending. Default 2020.
+  -n <set namespace>      Set a namespace for the micrortps_agent.
+```
+
+To launch the *Agent*, run `micrortps_agent` with appropriate options for specifying the connection to the *Client* (the default options connect from a Linux device to the *Client* over a UART port).
+
+As an example, to start the *micrortps_agent* with connection through UDP, issue:
+
+```sh
+./micrortps_agent -t UDP
+```
+
+## Creating a Fast DDS Listener application
+
+Once the *Client* (on the flight controller) and the *Agent* (on an offboard computer) are running and connected, *Fast DDS* applications can publish and subscribe to uORB topics on the PX4 Autopilot using RTPS.
+
+This example shows how to create a *Fast DDS* "listener" application that subscribes to the `sensor_combined` topic and prints out updates published from the PX4 Autopilot. A connected RTPS application can run on any computer on the same network as the *Agent*. For this example the *Agent* and *Listener application* will be on the same computer.
+
+The *fastrtpsgen* script can be used to generate a simple RTPS application from an IDL message file.
+
+:::note
+RTPS messages are defined in IDL files and compiled to C++ using *fastrtpsgen*.
+
+When building the bridge code, IDL files are generated for the uORB messages that may be sent/received, these IDL files are needed when you create a Fast DDS application to communicate with the PX4 Autopilot.
+
+You can find them in following path per build target:
+**build/BUILDPLATFORM/src/modules/micrortps_bridge/micrortps_agent/idl/*.idl**.
+:::
+
+Enter the following commands to create the application:
+
+```sh
+cd /path/to/PX4/PX4-Autopilot
+cd build/px4_sitl_rtps/src/modules/micrortps_bridge
+mkdir micrortps_listener
+cd micrortps_listener
+fastrtpsgen -example x64Linux2.6gcc ../micrortps_agent/idl/sensor_combined.idl
+```
+
+This creates a basic subscriber and publisher, and a main-application that you can run. 
+
+In order to print the data from the sensor combined topic, modify the following methods in **sensor_combined_Subscriber.cxx**:
+- `init()`: To change the subscription topic name (by default, the micrortps agent publishes the data on the named topic: `fmu/sensor_combined/out`),
+- `onNewDataMessage()`: To print the received sensor combined data.
+
+```cpp
+bool sensor_combinedSubscriber::init(Subscriber* sub)
+{
+    // Create RTPSParticipant
+
+    ParticipantAttributes PParam;
+    PParam.rtps.setName("Participant_subscriber"); //You can put the name you want
+    mp_participant = Domain::createParticipant(PParam);
+    if(mp_participant == nullptr)
+    {
+        return false;
+    }
+
+    //Register the type
+
+    Domain::registerType(mp_participant, static_cast<TopicDataType*>(&myType));
+
+    // Create Subscriber
+
+    SubscriberAttributes Rparam;
+    Rparam.topic.topicKind = NO_KEY;
+    Rparam.topic.topicDataType = myType.getName(); //Must be registered before the creation of the subscriber
+    Rparam.topic.topicName = "fmu/sensor_combined/out";
+    mp_subscriber = Domain::createSubscriber(mp_participant,Rparam, static_cast<SubscriberListener*>(&m_listener));
+    if(mp_subscriber == nullptr)
+    {
+        return false;
+    }
+    return true;
+}
+```
+
+```cpp
+void sensor_combinedSubscriber::SubListener::onNewDataMessage(Subscriber* sub)
+{
+    // Take data
+    sensor_combined_ st;
+
+    if(sub->takeNextData(&st, &m_info))
+    {
+        if(m_info.sampleKind == ALIVE)
+        {
+            // Print your structure data here.
+            ++n_msg;
+            std::cout << "\n\n\n\n\n\n\n\n\n\n";
+            std::cout << "Sample received, count=" << n_msg << std::endl;
+            std::cout << "=============================" << std::endl;
+            std::cout << "gyro_rad: " << st.gyro_rad().at(0);
+            std::cout << ", " << st.gyro_rad().at(1);
+            std::cout << ", " << st.gyro_rad().at(2) << std::endl;
+            std::cout << "gyro_integral_dt: " << st.gyro_integral_dt() << std::endl;
+            std::cout << "accelerometer_timestamp_relative: " << st.accelerometer_timestamp_relative() << std::endl;
+            std::cout << "accelerometer_m_s2: " << st.accelerometer_m_s2().at(0);
+            std::cout << ", " << st.accelerometer_m_s2().at(1);
+            std::cout << ", " << st.accelerometer_m_s2().at(2) << std::endl;
+            std::cout << "accelerometer_integral_dt: " << st.accelerometer_integral_dt() << std::endl;
+            std::cout << "magnetometer_timestamp_relative: " << st.magnetometer_timestamp_relative() << std::endl;
+            std::cout << "magnetometer_ga: " << st.magnetometer_ga().at(0);
+            std::cout << ", " << st.magnetometer_ga().at(1);
+            std::cout << ", " << st.magnetometer_ga().at(2) << std::endl;
+            std::cout << "baro_timestamp_relative: " << st.baro_timestamp_relative() << std::endl;
+            std::cout << "baro_alt_meter: " << st.baro_alt_meter() << std::endl;
+            std::cout << "baro_temp_celcius: " << st.baro_temp_celcius() << std::endl;
+
+        }
+    }
+}
+```
+
+To build and run the application on Linux:
+
+```sh
+make -f makefile_x64Linux2.6gcc
+bin/*/sensor_combined_PublisherSubscriber subscriber
+```
+
+Now you should see the sensor information being printed out:
+
+```sh
+Sample received, count=10119
+Received sensor_combined data
+=============================
+gyro_rad: -0.0103228, 0.0140477, 0.000319406
+gyro_integral_dt: 0.004
+accelerometer_timestamp_relative: 0
+accelerometer_m_s2: -2.82708, -6.34799, -7.41101
+accelerometer_integral_dt: 0.004
+magnetometer_timestamp_relative: -10210
+magnetometer_ga: 0.60171, 0.0405879, -0.040995
+baro_timestamp_relative: -17469
+baro_alt_meter: 368.647
+baro_temp_celcius: 43.93
+```
+
+:::note
+Make sure the *Client* is running, if the *Listener application* does not print anything.
+:::
+
+## Setting up the bridge with real hardware
+
+This section is work-in-progress.
+
+## Troubleshooting
+
+### Client reports that selected UART port is busy
+
+If the selected UART port is busy, it's possible that the MAVLink application is already being used.
+If both MAVLink and RTPS connections are required you will have to either move the connection to use another port or using the available protocol splitter for PX4 and companion computers.
+
+:::tip
+A quick/temporary fix to allow bridge testing during development is to stop MAVLink from *NuttShell*:
+```sh
+mavlink stop-all
+```
+:::
+
+### Agent not built/fastrtpsgen is not found
+
+The *Agent* code is generated using a *Fast DDS* tool called *fastrtpsgen*.
+
+If you haven't installed Fast DDS in the default path then you must specify its installation directory by setting the `FASTRTPSGEN_DIR` environment variable before executing *make*.
+
+On Linux/Mac this is done as shown below:
+
+```sh
+export FASTRTPSGEN_DIR=/path/to/fastrtps/install/folder/bin
+```
+
+:::note
+This should not be a problem if [Fast DDS is installed in the default location](../dev_setup/fast-dds-installation.md).
+:::
+
+### Enable UART on a companion computer
+
+For UART transport on a Raspberry Pi or any other companion computer you will have to enable the serial port:
+
+1. Make sure the `userid` (default is pi on a Raspberry Pi) is a member of the `dialout` group:
+
+   ```sh
+   groups pi
+   sudo usermod -a -G dialout pi
+   ```
+1. For the Raspberry Pi in particular, you need to stop the GPIO serial console that is using the port:
+
+   ```sh
+   sudo raspi-config
+   ```
+
+   In the menu showed go to **Interfacing options > Serial**.
+   Select **NO** for *Would you like a login shell to be accessible over serial?*. Valid and reboot.
+1. Check UART in kernel:
+
+   ```sh
+   sudo vi /boot/config.txt
+   ```
+
+   And make sure that the `enable_uart` value is set to 1:
+   ```
+    enable_uart=1
+   ```
+
+
+## Helpful Resources
+
+* [Fast DDS Installation](../dev_setup/fast-dds-installation.md)
+* [Manually Generate Client and Agent Code](micrortps_manual_code_generation.md)


en/middleware/micro_dds_edits.md
```diff
@@ -0,0 +1,377 @@
+# MicroDDS
```
hamishwillee, line 1: You've named this `micro_dds_edits.md` and the other one `micrortps_copy.md`. These names are reflected in the URLs, so the second name breaks the links. The first case is just a silly name.

hamishwillee:
Have the changes that remove micro rtps and add micro_dds been merged into PX4? What is the PR for that?

en/middleware/micro_dds_edits.md
```diff
@@ -0,0 +1,377 @@
...
+
+## Why this approach is of interest?
+
+Initially, micro-RTPS served as the middleware and its consistent improvement evolved into what's known as the microDDS that provides the user a native interface between the flight controller and the mission computer. 
+The notable difference between the two is what kind of transport protocol each supports. The newest version supports UDP/TCP/Serial/Custom transport protocol.  
+
+The Micro-XRCE-DDS-Gen has 2 major advantages, i.e. it can be used to generate client code and build px4_msgs along with microdds_client together creating a single library.  
+
+:::tip
+
```
hamishwillee, line 22: What's in here?

en/middleware/micro_dds_edits.md
hamishwillee, line 1: I'm guessing you have micro_dds_edits and micro_dds so you can track changes. That's what github does for you. From a reviewer point of view, having both is not using the tools. I'll delete.

en/middleware/micro_dds_edits.md
akshata-01: This was initially for my reference and I didn't add it in the SUMMARY thinking it'll be harmless. So it's nothing, thanks for removing it

en/middleware/micro_dds_edits.md
akshata-01: Yes, sure

akshata-01:
> Have the changes that remove micro rtps and add micro_dds been merged into PX4? What is the PR for that?
> 
> NOte, I have converted into a "draft" as we can't merge it until a) it is finished and reviewed b) the associated PR is merged.

Sure, [Here](https://github.com/PX4/PX4-Autopilot/pull/20227)

en/middleware/micro_dds_edits.md
```diff
@@ -0,0 +1,377 @@
...
+The *PX4-Fast RTPS(DDS) Bridge*, which was also referred to as as the *microRTPS Bridge*, has now been replaced with MicroDDS bridge.
+
+The new approach provides a faster and simpler method of integrating applictaions running and linked in DDS domains (including ROS nodes), making it easy to share sensor data, commands, and other vehicle information.
+
+The following guide describes the new PX4 bridge architecture, and shows the reader how to write a simple *Micro DDS* application to subscribe * publish telemetry updates from the PX4 Autopilot.
+
+:::note
+Micro XRCE-DDS is a software solution that enables communictaion with an existing DDS network. 
+
+:::
```
hamishwillee, line 12: YOu might want to completely invert this, starting from the sentence you currently have in the note:

> Micro XRCE-DDS is a software solution that enables communication with an existing DDS network. 

 and ending with a sentence like the one you currently have first:
 
 > The  MicroDDS bridge replaces and earlier solution, the *PX4-Fast RTPS(DDS) Bridge*.
> The new approach provides a faster and simpler method of integrating applictaions running and linked in DDS domains (including ROS nodes), making it easy to share sensor data, commands, and other vehicle information.
  blah blah


Why? It is best to say what this document is about. Right now the fact it is replacing fastRTPS is interesting, but we won't care after the next release. We just say what it is, and add the fact it is better as a comment or a section further down.

akshata-01: Noted. 
Btw I understand I caused a confusion, 
en/middleware/microdds.md is the most updated doc. 

I'll be sure to not create copies next time.

en/middleware/micro_dds_edits.md
hamishwillee: Ah, so I'll look again tomorrow when you've deleted this doc.

The problem with this stuff is that if you commit it, and it does not get deleted it then becomes part of the mainline. The next person to come along then sees two document with similar names - which is the right one, and so on.

hamishwillee:
These are some architecture overview summaries I want checked with @TSC21 . This is what provides me the delta to understand what is the difference in the microRTPS vs microDDS options.

## XRCE Bridge

image

Pretty much the same as the old model in that you have a fastCDR channel and the client/agent bidirectionally translate uorb to DDS

Presumably client source code is generated based on uorb and included in PX4.
- Previously for a non-ROS2 app the agent code was created at same time for same messages defined in yaml. YOu could build whenever. However for ROS 2 you needed different IDL, so instead the process generated `px4_msgs` and `px4_ros_com`, from which you could build agent code that would be compatible with the client on PX4.
- This was needed because DDS and ROS2 had a different IDL/message structure. SO a ROS2 app was different. It also needed a consistent API to match against.

So what is changed here is that you only need one form of IDL for DDS and no different format for ROS2. This is because the differences in forms are handed by the integration service to ROS2. 

What is cool about this is that in the new version you can build the agent off a release, since all the messages to share etc are known. There is no need to maintain  `px4_ros_com`.

1. That all about right?
2. Is the set of messages that are supported still built based on yaml as per http://docs.px4.io/main/en/middleware/micrortps.html#supported-uorb-messages ?

## microROS

image

So microROS is a similar client/agent architecture except that it translates from ROS2 to uorb, and presumably abstracts all the IDL stuff. 
My guess is that it uses the same uorb, but it does away with templates etc because you know the target IDL. 

Again, you don't need to do anything to maintain compatibility because you build off a release. About right?

Same kinds of questions thought - confirm build messages based on same YAML source.

## General questions

1. Are microROS and microDDS built using the same yaml definition of the uorb topics?
2. I presume that we would recommend using DDS for a DDS network that might also connect to ROS, but if we know we're connecting to a pure ROS network we use microROS?
3. Can both be present and enabled on PX4 at the same time? If not, how enable/disable?
4. Any other big picture things to mention?

en/SUMMARY.md
```diff
@@ -313,7 +313,6 @@
       * [Ubuntu Setup](dev_setup/dev_env_linux_ubuntu.md)
       * [Windows Setup](dev_setup/dev_env_windows_wsl.md)
       * [Visual Studio Code IDE](dev_setup/vscode.md)
-      * [Fast DDS installation](dev_setup/fast-dds-installation.md)
```
hamishwillee, line 316: Note, this was how to install the old toolchain. We will need to replaces this with a new doc. I have deleted the doc. I have not redirected - just created one redirection for the main document.

en/SUMMARY.md
```diff
@@ -611,7 +609,7 @@
   * [ROS](ros/README.md)
     * [ROS 2](ros/ros2.md)
       * [ROS 2 User Guide](ros/ros2_comm.md)
-      * [ROS 2 microRTPS Offboard Control Example](ros/ros2_offboard_control.md)
+      * [ROS 2 microRTPS Offboard Control Example](ros/ros2_offboard_control.md)
```
hamishwillee, line 612: Note the title. Touched this because we will need to rewrite this example for the new architecture.

en/SUMMARY.md
```diff
@@ -538,8 +537,7 @@
       * [wind](msg_docs/wind.md)
       * [yaw_estimator_status](msg_docs/yaw_estimator_status.md)
     * [MAVLink Messaging](middleware/mavlink.md)
-    * [RTPS/DDS Interface: PX4-Fast RTPS(DDS) Bridge](middleware/micrortps.md)
```
hamishwillee, line 541: Deleted link. Doc still exits, but just redirects to the DDS doc. That has a note explaining where FastRTPS went.

en/SUMMARY.md
```diff
@@ -538,8 +537,7 @@
...
-      * [Manually Generate Client/Agent](middleware/micrortps_manual_code_generation.md)
```
hamishwillee, line 542: We may need to replace this link with a new doc. Could well have the same name. But will reflect microDDS.

en/advanced_config/ethernet_setup.md
```diff
@@ -203,10 +203,10 @@ However this is not recommended because the default configuration is optimised f
 
 
 ## ROS2 Setup Example
-
-Prerequisites:
-
-- You have a supported autopilot hardware with RTPS feature enabled firmware on it by using [this guide](../middleware/micrortps.md#client-px4-px4-autopilot).
```
hamishwillee, line 209: Note, touched this because the document refers to RTPS. Need to verify status for XRCE microDDS and microROS.

en/dev_setup/building_px4.md
```diff
@@ -256,7 +256,7 @@ make [VENDOR_][MODEL][_VARIANT] [VIEWER_MODEL_DEBUGGER_WORLD]
 - **VENDOR:** The manufacturer of the board: `px4`, `aerotenna`, `airmind`, `atlflight`, `auav`, `beaglebone`, `intel`, `nxp`, etc.
     The vendor name for Pixhawk series boards is `px4`.
 - **MODEL:** The *board model* "model": `sitl`, `fmu-v2`, `fmu-v3`, `fmu-v4`, `fmu-v5`, `navio2`, etc.
-- **VARIANT:** Indicates particular configurations: e.g. `rtps`, `lpe`, which contain components that are not present in the `default` configuration. Most commonly this is `default`, and may be omitted.
+- **VARIANT:** Indicates particular configurations: e.g. `rtps`, `lpe`, which contain components that are not present in the `default` configuration. Most commonly this is `default`, and may be omitted.
```
hamishwillee, line 259: Note `rtps` config. Is there a new config? What are the options after the change?

en/dev_setup/dev_env_linux_arch.md
```diff
@@ -7,7 +7,7 @@ See [Toolchain Installation](../dev_setup/dev_env.md) for information about the
 
 The PX4-Autopilot repository provides a convenient script to set your Arch installation up for PX4 development: [Tools/setup/arch.sh](https://github.com/PX4/PX4-Autopilot/blob/main/Tools/setup/arch.sh). <!-- NEED px4_version -->
 
-The script installs (by default) all tools to build PX4 (without RTPS) for NuttX targets and run simulation with *jMAVsim*.
+The script installs (by default) all tools to build PX4 for NuttX targets and run simulation with *jMAVsim*.
```
hamishwillee, line 10: We might change this back to re-add "without DDS"

en/dev_setup/dev_env_linux_ubuntu.md
```diff
@@ -7,7 +7,6 @@ This environment can be used to build PX4 for [most PX4 targets](../dev_setup/de
 * [Gazebo Simulation](../simulation/gazebo.md)
 * [Raspberry Pi](#raspberry-pi)
 * [ROS (1)](#ros-gazebo) (Robotics Operating System)
-* [Fast DDS](../dev_setup/fast-dds-installation.md) - Required for ROS2
```
hamishwillee, line 10: IF we have an installation topic(s), this would be added back as a link

en/dev_setup/dev_env_linux_ubuntu.md
```diff
@@ -207,15 +206,6 @@ To install the development toolchain:
 :::
 
 
-<a id="fast_dds"></a>
```
hamishwillee, line 210: IF we have an installation topic(s), this would be added back or replaced with link to new information.

en/hardware/porting_guide_config.md
```diff
@@ -39,7 +39,7 @@ Builds will silently ignore any missing or miss-spelled modules in the  `*.px4bo
 Each PX4 board must have a `default.px4board` configuration and can have an optional `bootloader.px4board configuration`.
 However you can add also separate configurations under a different label e.g. `rtps.px4board`.
 Note that by default the configuration of `rtps.px4board` inherits all settings set in `default.px4board`.
-When changing the `rtps.px4board` it only stores the delta of the Kconfig keys that are different compared to `default.px4board`, this is useful to simplify configurations management
+When changing the `rtps.px4board` it only stores the delta of the Kconfig keys that are different compared to `default.px4board`, this is useful to simplify configurations management.
```
hamishwillee, line 42: Note, rtps mentioned

en/middleware/microdds.md
```diff
@@ -0,0 +1,74 @@
+# MicroDDS
+
+Micro XRCE-DDS is a software solution that enables communication with an existing DDS network. 
+
+The new approach provides a faster and simpler method of integrating applictaions running and linked in DDS domains (including ROS nodes), making it easy to share sensor data, commands, and other vehicle information.
+
+The following guide describes the new PX4 bridge architecture, and shows the reader how to write a simple *Micro DDS* application to subscribe * publish telemetry updates from the PX4 Autopilot.
+
+:::note
```
hamishwillee, line 9: @akshata-01 FYI I switched this around so this note about fast RTPS is at the bottom. The top should all be interesting info about microDDS. I haven't rewritten that.

en/middleware/micrortps.md
```diff
@@ -1,384 +1,3 @@
-# RTPS/DDS Interface: PX4-Fast RTPS(DDS) Bridge
+<Redirect to="microdds" />
```
hamishwillee, line 1: Note, deleted content. Not relevant once PR goes in. The note on the microDDS page explains what happened to it and where you go now if you're in old versions

en/middleware/micrortps_manual_code_generation.md
```diff
@@ -1,187 +0,0 @@
-# Manually Generate Client and Agent Code
```
hamishwillee, line 1: So as noted in SUMMARY.md where I removed the link - we might need a topic like this for microDDS and microROS. IN which case a new one would be created explaining how.

en/ros/README.md
```diff
@@ -26,7 +26,7 @@ PX4 supports both the "original" ROS and ROS 2, with the following configuration
 
 Unveiled at the [PX4 Developer Summit 2020](https://www.youtube.com/watch?v=lZ8crGI16qA) (and [ROS World 2020](https://www.youtube.com/watch?v=8XRkzHqQSf0)), the PX4 Development team, announced the plans to support microROS.
 
-* microRTPS:  microRTPS bridge with Fast DDS (The current stable ROS 2 interface)
-* micro XRCE-DDS: DDS on PX4 (Next step!)
-* micro ROS: ROS 2 on the PX4 (Our Target!)
+* microRTPS:  microRTPS bridge with Fast DDS (The current stable ROS 2 interface) 
+* micro XRCE-DDS: DDS on PX4 (Next step!) 
+* micro ROS: ROS 2 on the PX4 (Our Target!)
```
hamishwillee, line 31: We need to update this to explain the current state - as I understand it, this is all done.

en/ros/ros1_via_ros2.md
```diff
@@ -20,14 +20,14 @@ The PX4 development team recommend that all users [upgrade to ROS 2](../ros/ros2
 
 ## Overview
 
-The application pipeline for ROS (1) bridged over ROS 2 is shown below.
+The application pipeline for ROS (1) bridged over ROS 2 is shown below.
```
hamishwillee, line 23: I've "touched" this doc just to remind up it will need to be updated to explain how to do ROS 1 over ROS 2 with new architecture(s)

en/ros/ros2_comm.md
```diff
@@ -11,7 +11,7 @@ For information about using the *microRTPS bridge* **without ROS 2**, see the [R
 :::
 
 :::note
-For a more detailed and visual explanation on how to use PX4 with ROS 2 see these presentations from the PX4 maintainers:
+For a more detailed and visual explanation on how to use PX4 with ROS 2 see these presentations from the PX4 maintainers:
```
hamishwillee, line 14: We will need to review this whole doc in the light of microROS and microDDS

en/ros/ros2_offboard_control.md
```diff
@@ -6,7 +6,7 @@ If you are operating on a real vehicle be sure to have a way of gaining back man
 :::
 
 :::warning
```
hamishwillee, line 8: Ditto review this whole doc in light of microROS and microDDS

en/simulation/README.md
```diff
@@ -100,7 +100,7 @@ The different parts of the system connect via UDP, and can be run on either the
   Simulators then exchange information with PX4 using the [Simulator MAVLink API](#simulator-mavlink-api) described above.
   PX4 on SITL and the simulator can run on either the same computer or different computers on the same network.
   :::note
-  Simulators can also use the *microRTPS bridge* ([PX4-FastRTPS Bridge](../middleware/micrortps.md)) to directly interact with PX4 (i.e. via [UORB topics](../middleware/uorb.md) rather than MAVLink).
+  Simulators can also use the *microRTPS bridge* ([PX4-FastRTPS Bridge](../middleware/micrortps.md)) to directly interact with PX4 (i.e. via [UORB topics](../middleware/uorb.md) rather than MAVLink).
```
hamishwillee, line 103: Update section in line with microRTPS/DDS

en/middleware/micro_dds_edits.md
hamishwillee: Deleted now.

en/middleware/micro_dds_edits.md
hamishwillee: No worries. I've moved things around to good order. COntent now needs to be written following that pattern.

hamishwillee:
@akshata-01 OK, so I've been through and deleted the main RTPS legacy docs, redirecting one of them to the DDS doc. That DDS doc now has the note explaining where RTPS went. 

The docs I deleted are now a "hint" - that was something about RTPS that we might need to now do with microDDS and/or microROS.

I've also "touched" a whole lot of places that have the letters rtps - so those are all hints of things to do as well. We should create the main docs then go back and search for the term again, tidying things up as needed.

There is also a big question post above just clarifying "for me" that this is as I understand it. Doesn't change very much other than we need a separate page for microROS and microDDS and to understand where each are used.

You might already understand all these things, in which case, get writing :-)

en/dev_setup/building_px4.md
beniaminopozzan: All `rtps` config have been removed, the `microdds_client` is included by default in most of the boards but the user has to manually start it acting on the appropriate parameters.
In simulation, the module is automatically started using udp protocol and port 8888

hamishwillee:
Hi @Jaeyoung-Lim Can we go through this (or can you respond to it): https://github.com/PX4/PX4-user_guide/pull/2094#issuecomment-1284879342 

That's kind of the starting point before I know the right questions to ask about this. Yes I could dig more in code and original PRs, but I only work 2 days per week, and if we do it that way, I should get to this sometime next decade.

en/middleware/microdds.md
```diff
@@ -0,0 +1,74 @@
...
+```sh
+microdds_client status #Checks client status 
+```
+![Simulation](../../assets/middleware/microdds/simulation.png)
+
+TERMINAL 2
+
+```sh
+cd ~/px4_ros2_ws #Intro your ROS2 directory 
+source /opt/ros/humble/setup.bash
```
Jaeyoung-Lim, line 69: It seems that this document is assuming that you are running ros2 humble. The current default supported Ubuntu distribution of PX4 is Ubuntu 20.04, and this supports ROS Foxy. 

ROS Humble only supports Ubuntu 22.04, which PX4 doesn't fully support yet. therefore, I think we need to set the current ros2 support to ROS foxy at least for now

Jaeyoung-Lim (changes requested):

en/middleware/microdds.md
```diff
@@ -0,0 +1,74 @@
...
+#### The microdds Agent
+The *Agent* runs as a daemon process that's defined by what the clients defines it to be and hence it's dynamic. It's a server that acts an an intermediary between the client and the DDS Global Data Space. 
+
+#### microdds Communication
+microDDS has APIs that allow writing time-critical applications and can be used over many transport protocols supporting TCP,UDP over Ethernet Wi-Fi & 6LoWPAN and Bluetooth. The user can also customise this making microDDS transport-agnostic.
+
+## Test
+
+Let’s test! 
+You can check the microdds running using a grep command (ps aux | grep micro)
```
Jaeyoung-Lim, line 48: This test is specifically for a SITL instance. Are there plans to show how this can be tested on real hardware?

en/middleware/microdds.md
```diff
@@ -0,0 +1,74 @@
...
+
+TERMINAL 1
```
Jaeyoung-Lim, line 50: Not really a fan of terminal 1 / terminal 2 -> Could we change this to more informative descriptions on what we are running exactly?

en/middleware/microdds.md
```diff
@@ -0,0 +1,74 @@
...
+![Simulation](../../assets/middleware/microdds/simulation.png)
+
+TERMINAL 2
+
+```sh
+cd ~/px4_ros2_ws #Intro your ROS2 directory 
+source /opt/ros/humble/setup.bash
+source install/setup.bash
+ros2 topic list -v #Displays available topic 
+ros2 topic echo fmu/out/vehicle_status #You can check any available topic
```
Jaeyoung-Lim, line 72: This would require the user installing and setting up ros2. Shouldn't this be explained more on how ros2 could be setup, or at least link the docs?

en/middleware/microdds.md
```diff
@@ -0,0 +1,74 @@
...
+microdds_client status #Checks client status 
+```
+![Simulation](../../assets/middleware/microdds/simulation.png)
+
+TERMINAL 2
+
+```sh
+cd ~/px4_ros2_ws #Intro your ROS2 directory 
+source /opt/ros/humble/setup.bash
+source install/setup.bash
```
Jaeyoung-Lim, line 70:
```suggestion
source install/local_setup.bash
```

en/middleware/microdds.md
```diff
@@ -0,0 +1,74 @@
...
+
+```sh
+microdds_client status #Checks client status 
+```
+![Simulation](../../assets/middleware/microdds/simulation.png)
+
+TERMINAL 2
+
+```sh
+cd ~/px4_ros2_ws #Intro your ROS2 directory
```
Jaeyoung-Lim, line 68: The example is just doing a `ros2 topic echo`. Therefore a ros2 workspace is not necessary

en/middleware/microdds.md
dirksavage88: Usually local setup is sourced if you have just the ros2 installed without an underlay. However, I'm imaging users running an overlay with their own custom/modified packages but also needing to source the underlay where the RO2 base install is.

source: [Configuring environment](https://docs.ros.org/en/foxy/Tutorials/Beginner-CLI-Tools/Configuring-ROS2-Environment.html)

hamishwillee:
@Jaeyoung-Lim Thanks for that. Can you answer MY questions that is - everything in https://github.com/PX4/PX4-user_guide/pull/2094#issuecomment-1284879342

What you're reviewed is great, but it could be missing heaps of nuance if the "architecture" level stuff is wrong.

beniaminopozzan:
@hamishwillee , I can't answer all your questions, but for sure

> Is the set of messages that are supported still built based on yaml as per http://docs.px4.io/main/en/middleware/micrortps.html#supported-uorb-messages ?

Yes, it is. The yaml file is
https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/microdds_client/dds_topics.yaml

en/ros/ros1_via_ros2.md
```diff
@@ -20,14 +20,14 @@ The PX4 development team recommend that all users [upgrade to ROS 2](../ros/ros2
 
 ## Overview
```
TSC21, line 21: This all needs to be reviewed and retested to see if it still works. I would even replace this with eProsima's [Integration Service](https://integration-service.docs.eprosima.com/en/latest/) (proposed on one of my talks)

TSC21 (changes requested):

en/middleware/microdds.md
```diff
@@ -0,0 +1,74 @@
+# MicroDDS
+
+Micro XRCE-DDS is a software solution that enables communication with an existing DDS network. 
+
+The new approach provides a faster and simpler method of integrating applictaions running and linked in DDS domains (including ROS nodes), making it easy to share sensor data, commands, and other vehicle information.
```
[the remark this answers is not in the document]
mrpollo: I agree, @akshata-01 please rewrite this sentence "This interface allows for a fast and simple method of..." no need to compare against what is no longer available

en/middleware/microdds.md
mrpollo: It would be great to have those docs

Can you help me create one? any direction you can provide will help me create something

en/middleware/microdds.md
mrpollo: I agree, TERMINAL 1 vs 2 is the wrong way of doing this.

I'll review and update

en/middleware/microdds.md
mrpollo: The plan is to upgrade PX4 to Ubuntu 22.04 as well. (correct me if I'm wrong @dagar)

mrpollo (commented):
I'll spend some time reviewing your comments @Jaeyoung-Lim & @TSC21, I agree we make a lot of assumptions that need to be explained or called out, and we also need to make sure we use the proper name/terminology throughout eg, MicroDDS vs XRCE-DDS, and many more thanks to @TSC21 

@dagar can you also please review?

en/dev_setup/dev_env_linux_arch.md
dirksavage88: @hamishwillee there are multiple ways to get micro-ROS and XRCE-dds:

1) Microxrcedds client + agent **built from source**: [xrcedds install](http://micro-xrce-dds.readthedocs.io/en/latest/installation.html) *you can install client and/or agent, as all of these packages are part of a superbuild cmake list, and will install dependencies for you!
2) Standalone microxrce-dds agent via snap install: [xrce agent standalone snap](https://snapcraft.io/micro-xrce-dds-agent)
3) Micro-ROS via ros2 colcon build [uros ros2 package:](https://github.com/micro-ROS/micro_ros_setup) * requires running a setup + build script using 'ros2 run' as well as the colcon build tools. 
4) micro-ROS via snap package: [uros snap](https://snapcraft.io/micro-ros-agent)

As TSC21 stated, fastdds is still a dependency, but the microxrcedds install from source also call out fastcdr among other packages (foonathon_memory, googletest, spdlog...). However, the **SuperBuild.cmake will attempt to find the package, if not found it will clone and build as a dependency**. The installation process is very simplified compared to the old docs.

beniaminopozzan:
Any update on this?

hamishwillee:
Hi @beniaminopozzan,

No. **I'm not going to start this until @TSC21 or @Jaeyoung-Lim answer [my basic questions](https://github.com/PX4/PX4-user_guide/pull/2094#issuecomment-1284879342).**

hamishwillee:
PS It's an 80/20 rule thing - with the answers to those questions I can start this and likely get it done in a couple of days. Without that I am likely to have to write this full of caveats. Not just frustrating, but likely to take 5-10 days, and unlikely to be correct in all respects. I have far more useful things I could be doing with my 2 days per week.

TSC21:
I think we should hold while we start getting into a much for tighter and advanced disteibuted architecture: https://github.com/PX4/PX4-Autopilot/pull/20707. Unless @bkueng and @dagar disagree.

beniaminopozzan:
I agree with @hamishwillee on the importance of this documentation as at the moment there is none about the new microdds bridge.

Jaeyoung-Lim:
@hamishwillee Sorry that this thread has been blocked by my response

Unfortunately the questions you have asked are a little outside the boundary of what I can comment on. I am using the interface, but I lack the understanding of the inner workings of the interface.

While I will be happy to comment on instructions on how to use the interface and how to set it up, I am also in the process of learning about the interface 😓

bkueng:
Yes this is independent from my PR. @TSC21 could you please help Hamish?

hendjoshsr71:
Might I suggest marking a few of the ROS2 install docs as "Outdated" until this larger effort is completed?

I just went through installing ROS2 on 22.04 and got there almost on my own. But then got stuck for a bit until I saw Julian's message on Discord with some instructions he created here,

https://gist.github.com/julianoes/adbf76408663829cd9aed8d14c88fa29

hamishwillee:
@hendjoshsr71 Yes. I will do that today.

en/ros/README.md
hamishwillee: @TSC21 I'm ramping up on this again. So to be clear, we have XRCE-DDS client library and microROS client library on PX4 right. What we don't yet do is have any code running on the FC that directly calls the microROS library - i.e. no "ROS executors".
1. Correct?
2. But in theory we could start writing ROS nodes on PX4 now?

en/middleware/microdds.md
```diff
@@ -0,0 +1,74 @@
...
+
+### microDDS Bridge
+
+The [*microDDS*](https://micro-xrce-dds.docs.eprosima.com/en/stable/introduction.html) provides DDS publisher subcription approach wherein the client is on a low-resource consumption device and the agent is connected to with the DDS gobal data space.
+
+#### The microdds Client
+The *Client* is the PX4 Autopilot middleware daemon process that runs on the flight controller. The client can either publish or subcribe directly to data topics because the agent itself acts on behalf of the [client](https://micro-xrce-dds.docs.eprosima.com/en/stable/client.html)
+
+#### The microdds Agent
+The *Agent* runs as a daemon process that's defined by what the clients defines it to be and hence it's dynamic. It's a server that acts an an intermediary between the client and the DDS Global Data Space.
```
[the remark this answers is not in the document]
hamishwillee: I could, but I will probably re-write most of this based on [your presentation](https://www.youtube.com/watch?v=F5oelooT67E) once I understand how it works.

en/dev_setup/building_px4.md
hamishwillee: @beniaminopozzan Thanks. I presume by appropriate parameters etc you mean the command line here: http://docs.px4.io/main/en/modules/modules_system.html#examples-3 ?

@Jaeyoung-Lim I guess this is the same question as https://github.com/PX4/PX4-user_guide/pull/2094/files#r1008915158 - how do we get it working on real hardware? Do you know of any example/tutorial for real hardware yet?

Wouldn't it be better if all these options were presented as  parameters so that you could configure these in the airframe file? 

If not, is the suggestion that you do so in an SD card startup file?

en/dev_setup/dev_env_linux_arch.md
hamishwillee: @TSC21"Are we targeting to use this without or with ROS2 only?"

Good question - isn't that up to you to answer? For RTPS docs we did include both cases.

1. My understanding is that both options are supported - right? 
2. My further assumption is that the vast majority of users will be accessing via ROS2. If so I suggest that we target these two cases completely separately, and start with instructions that assume ROS2 consumers. Does that make sense?
3. IN that case, what are the simplest instructions we can use to set up fastDDR, microROS and XRCE-dds _agent_ code? My assumption is that it is the [stuff in this gist](https://gist.github.com/julianoes/adbf76408663829cd9aed8d14c88fa29#micro-ros) down to the microROS topic - Can you confirm, or is there a better way? 
   > @dirksavage88 I think this is your option "3" above. 
4. My understanding is that you don't need to build the microROS or XRCE-DDS modules yourself directly, this just happens as part of the build. 
   - is that right? 
   - If you don't have FastDDS / ROS installed, does that just mean the module isn't built and included in firmware, but everything else will work?

hamishwillee: @dirksavage88 For the "lower priority"/non-ROS2 case what would you recommend for installation - i.e. what do you need? Do you know of any meaningful tutorials? For RTPS we had this listener: https://docs.px4.io/main/en/middleware/micrortps.html#creating-a-fast-dds-listener-application

en/middleware/microdds.md
```diff
@@ -0,0 +1,74 @@
...
+```
+![Simulation](../../assets/middleware/microdds/simulation.png)
+
+TERMINAL 2
+
+```sh
+cd ~/px4_ros2_ws #Intro your ROS2 directory 
+source /opt/ros/humble/setup.bash
+source install/setup.bash
+ros2 topic list -v #Displays available topic
```
hamishwillee, line 71: Which uorb topics are available as ROS topics? This used to be done in Yaml. Is it all of them? https://github.com/PX4/PX4-Autopilot/blob/main/msg/CMakeLists.txt#L288

en/middleware/microdds.md
hamishwillee: @TSC21 @mrpollo So we're now in Jan 2022 - my assumption is that ROS2 humble is the target, and that this is on Ubuntu 22.04. Or at least that is what I will try out.

- Has PX4 been validated/updated to Ubuntu 22.04? Is there an ETA or a plan for this?

en/ros/README.md
beniaminopozzan: @hamishwillee I'll try to answer:
1. No, there is the XRCE-DDS client library but no microROS client library yet
2. For the above reason, I guess not yet

en/dev_setup/building_px4.md
beniaminopozzan: @hamishwillee precisely. But on real hardware there is also the `serial_config` parameter like the old micrortps. However, I never tried it yet and it does not allow for specifying custom namespaces.

en/dev_setup/dev_env_linux_arch.md
beniaminopozzan: @hamishwillee not sure about 1. but I definitively agree on 2.
About 3. Assuming to use Foxy, Humble of Rolling as ROS2 distro (not Galactic) I would say that the simplest instructions are just cloning in the workspace the bare Agent library https://github.com/eProsima/Micro-XRCE-DDS-Agent/ and build it together with the other ros packages (px4_msgs in particular). The library will automatically fetch all its dependencies and in the end the agent will be started with just `MicroXRCEAgent ....`

en/middleware/microdds.md
```diff
@@ -0,0 +1,74 @@
...
+TERMINAL 2
+
+```sh
+cd ~/px4_ros2_ws #Intro your ROS2 directory 
+source /opt/ros/humble/setup.bash
+source install/setup.bash
+ros2 topic list -v #Displays available topic 
+ros2 topic echo fmu/out/vehicle_status #You can check any available topic
+```
+![ROS2 Topic](../../assets/middleware/microdds/topic.png)
```
hamishwillee, line 74: This tutorial stops early. Looking at this more complete one: https://gist.github.com/julianoes/adbf76408663829cd9aed8d14c88fa29#offboard-example-and-px4_msgs

This imports the following:
```
git clone https://github.com/PX4/px4_ros_com.git src/px4_ros_com
git clone https://github.com/Jaeyoung-Lim/px4-offboard.git src/px4-offboard
````

@Jaeyoung-Lim 
1. I understand that ROS needs to know what ROS nodes are present, and that these are provided by `PX4/px4_msgs` - is that right? 
2. Presumably the definitions change depending on the build. Does that mean we'll take release snapshots of these.
3. If I build locally, where do the px4_msgs get generated? Or to put it another way, if I create a custom build of PX4, how do I expose messages to my ROS setup?
4. These look just like normal uorb topic definitions - how do they get turned into something your script/ROS can use?
5. I thought that `px4_ros_com` was no longer needed? Why does that get imported?

File: en/dev_setup/dev_env_linux_arch.md (hunk 10 above)
Comment (hamishwillee) [reply en/dev_setup/dev_env_linux_arch.md:10]: @beniaminopozzan Thanks! The reason I kind of thought using microROS installations as per tutorial might be good was future proofing. I.e. if we know that micro ROS contains XRCE-DDS agent and we expect micro ROS to be supported, then there is no particular cost in just following the micro ROS setup. Or am I being naïve? 

IN addition, I already have instructions to follow for micro ROS. That reduces my testing.

Of course if someone was to provide the step by step instructions for setup, things might be different :-)

File: en/middleware/microdds.md (hunk 32 above)
Comment (hamishwillee) [reply en/middleware/microdds.md:74]: Actually, thinking about this, my guess is that  you need `PX4/px4_msgs` if working with XRCE-DDS "alone" in order to tell the toolchain how to build the ROS topics. I assume that the colcon build builds everything, and somewhere in the definitions there is info to tell it to build the ROS2 topics from the messages.

I don't think you need the px4_ros_com at all, ever. If I'm wrong, when is it useful?

So how is microROS different? My guess based on the presentation is that you don't need px4_msgs either. I think in this case the client sends ROS messages by default so there is no need to do any conversion later on. Is that correct? IF not, can you clarify the "usage" difference?

File: en/middleware/microdds.md (hunk 31 above)
Comment (beniaminopozzan) [reply en/middleware/microdds.md:71]: @hamishwillee , the uorb topics to expose to ROS2 can be selected editing this https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/microdds_client/dds_topics.yaml

File: en/middleware/microdds.md (hunk 32 above)
Comment (beniaminopozzan) [reply en/middleware/microdds.md:74]: `PX4/px4_msgs` is still needed regardless which "agent", bare XRCE-DDS or microROS, you use. Currently it provides the message definitions to ROS2 (you can try running either agents without this package sourced, no error will appear until you actually try to echo a topic, then ROS2 will complain about the missing message definition). I don't know if in the future this will change, like it will be possible to compile the messages directly from the PX4 main repo.

The best part about this new system is that the message definitions on the PX4 side and the ROS2 side are the same. Therefore you only need a copy-paste of all the `.msg` files from PX4 to ROS2 to keep the two in synchronized.

px4_ros_com as you said is no more needed. It contains a few examples of nodes and a library to transform frames between ROS2 and PX4 conventions, but it is no more a dependency.

According to the official repo https://github.com/micro-ROS/micro-ROS-Agent, microROS agent is just a ROS2 node that wraps the micro XRCE-DDS Agent. So the communication is always between micro XRCE-DDS Agent and micro XRCE-DDS Client. The agent then exposes the topics to ROS2 .

File: en/dev_setup/dev_env_linux_arch.md (hunk 10 above)
Comment (beniaminopozzan) [reply en/dev_setup/dev_env_linux_arch.md:10]: @hamishwillee 

> Note, @beniaminopozzan says that microROS client is not installed. So presumably this install approach works because microROS dependencies include the XRCE-DDS agent.

Micro ros client is "just" microdds client (the one that PX4 uses) for the low level communication (with the microdds agent) plus the ROS API to have ROS in the device https://micro.ros.org/docs/concepts/client_library/introduction/
the pipeline is something like
low-resource-device  -- microRosClient -- micro XRCE-DDS client ----- micro XRCE-DDS Agent  -- microRosAgent -- ROS2
Right now, PX4 directly connects to micro XRCE-DDS client.

> The reason I kind of thought using microROS installations as per tutorial might be good was future proofing. I.e. if we know that micro ROS contains XRCE-DDS agent and we expect micro ROS to be supported, then there is no particular cost in just following the micro ROS setup. Or am I being naïve?

No, it make sense!

Regrading the step by step instructions, if you are ok with a really bare-minimum version, they are like this:


Provided that you have ROS2 foxy, humble or rolling installed, let's assume humble (like here https://gist.github.com/julianoes/adbf76408663829cd9aed8d14c88fa29#ros2)

1. create a ros2 workspace
```sh
mkdir -p ~/px4_ros_ws/src
cd ~/px4_ros_ws/src
```
2. clone the dependencies `px4_msgs` and `micro XRCE-DDS Agent`
```sh
git clone https://github.com/eProsima/Micro-XRCE-DDS-Agent.git
```
3. Source the main environment and compile the workspace
```sh
cd ~/px4_ros_ws/
colcon build
```
4. Source the compiled workspace and run the Agent (assuming a the default sitl scenario where the client tries to connect using udp4 on port 8888 like here https://gist.github.com/julianoes/adbf76408663829cd9aed8d14c88fa29#px4-sitl-with-gazebo-classic)
```sh
source install/local_setup.bash
MicroXRCEAgent udp4 -p 8888
```
at this point the tutorial can resume from here

> Leave it running, and open a new terminal for the next step.

https://gist.github.com/julianoes/adbf76408663829cd9aed8d14c88fa29#check-px4-topics

hamishwillee: Thanks very much. I think that's more than enough for me to start on this next week. Much appreciated.

hamishwillee: @beniaminopozzan Re your comment:

> Micro ros client is "just" microdds client (the one that PX4 uses) for the low level communication (with the microdds agent) plus the ROS API to have ROS in the device https://micro.ros.org/docs/concepts/client_library/introduction/
the pipeline is something like: 
> `low-resource-device -- microRosClient -- micro XRCE-DDS client ----- micro XRCE-DDS Agent -- microRosAgent -- ROS2`


A few follow on questions. 
1. I do understand that microROS allows ROS2 on the embedded controller. So it is a ROS2 API layer that connects to the micro XRCE-DDS client.
  - Why do you need a microROS agent? When we query our messages from ROS2 in the examples they recognise the ROS2 topics from uORB via XRCE-DDS, so what is the agent doing for us?

> Right now, PX4 directly connects to micro XRCE-DDS client.

Yes. But to be precise:
2. Does PX4 support/build the microROSClient in source?
3. Does PX4 include the client in firmware?

Basically I'm trying to understand "what happens next" - my assumption is that if 2 and 3 are "yes" then next step would be for someone to write a ros2 app on PX4, and that they could do that now. 

4. Is there a compelling use case for microROS. We generally have been pushing people to companion for years because the FC is resource constrained and harder to write for.

beniaminopozzan: > I do understand that microROS allows ROS2 on the embedded controller. So it is a ROS2 API layer that connects to the micro XRCE-DDS client.

Yes, I have the same understanding of microRO.S

> Why do you need a microROS agent?

I really don't know. Having all components under the "ROS2" hat is nice (I mean starting the Agent with `ros2 run micro_ros_agent micro_ros_agent`) but I guess there is something more, not strictly necessary, some other convenient feature maybe?

> Does PX4 support/build the microROSClient in source?
> Does PX4 include the client in firmware?

Not right now. Right now only micro XRCE-DDS Client is included in the firmware (in as a submodule of the [microdds](https://github.com/PX4/PX4-Autopilot/tree/main/src/modules/microdds_client) module).

> Basically I'm trying to understand "what happens next"

I can't help you here, let me ping @dagar for these aspects.

hamishwillee: What does it mean if `ros2 topic list` works, but when you try echo one of the displayed topics you get an error?

```
ubuntu@ubuntu-virtual-machine:~$ ros2 topic echo /fmu/out/vehicle_status
The message type 'px4_msgs/msg/VehicleStatus' is invalid
```

beniaminopozzan: I don't know if this applies anytime that error message appears, but in this case it means that ROS2 doesn't have the message definition for `px4_msgs/msg/VehicleStatus` (or that the message definition that it has is different from the one used to build that message, I have to check that). Did you not build the package `px4_msgs` o did you not source the associated environment?

hamishwillee: Thanks @beniaminopozzan - I think I'm a slow learner. Got it.

ViniciusAbrao:
I have learned a lot in this pull request. While we do not have a official documentation, the @Jaeyoung-Lim  ' s offboard package has given I huge documentation. I have recorded a video trying the package in the Foxy version of ROS2. All is working as expected. I would like to share with you in case it can help somehow: https://youtu.be/f5qEXDTZOug

hamishwillee:
@ViniciusAbrao Thanks for taking the time to post that - I've just skipped through it but looks cool - certainly useful for anyone coming to this PR.

This PR is superseded by another one https://github.com/PX4/PX4-user_guide/pull/2259 . This uses XRCE-DDS agent rather than microROS as for Jae's tutorial. 

There are a couple of things that are not obvious from this current PR and from Jae's examples:
1. You can just use XRCE-DDS - you don't need microROS. That's only needed if you have ROS nodes written in PX4 itself (not yet used). The microROS client provides libraries to allow that. Not clear what the agent does "over and above" XRCE-DDS Agent - no one seems to know.
2. The px4_ros_com  package isn't needed (at all)
3. Mismatched qos settings in a subscriber can stop you subscribing. [still getting my head around this].

When the docs are done, would love another video from you working through the XRCE-DDS tutorial.

ViniciusAbrao:
@hamishwillee thank you for your feedback. I got all your points but I have thinking about some points in the future that might have relation with what you said. For example:

1- I could see that the XRCE-DDS implemented in the modules is currently responsible for the ROS topics, but not yet for the ROS services. 

2- Looking into the eProsima tutorials: https://micro-xrce-dds.docs.eprosima.com/en/latest/client.html
I did not find yet how to implement ROS services.

3- Looking into the microROS tutorials: https://micro.ros.org/docs/tutorials/programming_rcl_rclc/service/
There is some documentation about how to start doing it. 

So at this point, I think that microROS also must be in the PX4 side, in order to simplify the implementations from the client side. If not, going directly through XRCE-DDS would be harder. Does it make sense?

hamishwillee:
> So at this point, I think that microROS also must be in the PX4 side, in order to simplify the implementations from the client side. If not, going directly through XRCE-DDS would be harder. Does it make sense?

Does not make sense sorry unless the summary is "until microROS is being used in PX4 we can't really talk about microROS yet".

The thing to note is that XRCE-DDS is used by microROS. microROS is a layer of libraries that let you do "native" ROS nodes on PX4 rather than uORB, and translating to DDS. What actually gets sent should be pretty much the same, which is why you can use either the XRCE-DDS or microROS agent on the companion to read the topics from PX4.

I don't see to much benefit in talking aobut microROS right now since PX4 doesn't implement/include the microROS client yet. YOu can use the microROS agent or the XRCE-DDS agent on the companion, but I'd rather talk about the XRCE-DDS agent for now - at least until someone explains what value it adds.

ViniciusAbrao:
Yes, I agree with all your comment. The only thing I am highlighting is the need of the microROS in the PX4. Otherwise I can not see how to implement Services like were done in the past by MAVROS (arm, set_mode, take off, land ... ). I might be confused at this point, but I did not find yet how to do it directly in the XRCE-DDS, without the layer of libraries provided by microROS.

If I am wrong, I suggest that the documentation presents how to do it in general:
- How to write topics, services, custom msgs, and so on, directly in the XRCE-DDS client.

mrpollo:
Hey @hamishwillee, hopefully this video by @DanMesh helps with some of the concepts https://youtu.be/CiajtefNQvE

hamishwillee:
>> Yes, I agree with all your comment. The only thing I am highlighting is the need of the microROS in the PX4. Otherwise I can not see how to implement Services like were done in the past by MAVROS (arm, set_mode, take off, land ... ). I might be confused at this point, but I did not find yet how to do it directly in the XRCE-DDS, without the layer of libraries provided by microROS.

> If I am wrong, I suggest that the documentation presents how to do it in general:

> How to write topics, services, custom msgs, and so on, directly in the XRCE-DDS client.

So MAVROS wraps MAVLink - i.e. it turns a ROS command into a MAVLink command. That MAVLink command is turned into NAV commands and uorb topics and PX4-side stuff. 
If you have XRCE-DDS you can publish to uorb topics from ROS. So if everything you need to do in PX4 to takeoff (say) can be represented by uorb topics then you can do that using ROS on the companion side. However if you need access to something on PX4 that is not represented in uORB (or just uORB) then yes, a PX4-side ROS node be needed.

So 2 points:
- I don't know how _all_ those operations are implemented in PX4 or if there are companion side nodes already written to do this without mavros. I suspect that most of them can be done with uorb -e.g. by publishing to http://docs.px4.io/main/en/msg_docs/action_request.html and http://docs.px4.io/main/en/msg_docs/vehicle_command.html
- Docs will cover using ROS to interface in general, but I think you make a very good point that there are standard things you might want to do such as takeoff, and these should be documented.

@beniaminopozzan what's the story with standard libraries in ROS2 for doing stuff like takeoff, arm, etc?

hamishwillee:
@ViniciusAbrao 

> Is the documentation only for users in the agent side?

YES. Right now.

Until PX4 starts using it client-side it is too early for docs. 

> How the PX4's user is going to implement ROS services (for example) without microROS in the client?

See my last answer ^^^. Most of them are likely to be implementable by posting to existing uorb topics. I actually don't know of a use case that needs microROS yet.

hamishwillee:
> Hey @hamishwillee, hopefully this video by @DanMesh helps with some of the concepts https://youtu.be/CiajtefNQvE

Thanks @mrpollo. Nothing I didn't know about ROS, but it does explain some of the work Beat is doing around flight modes in MAVLink and PX4.

ViniciusAbrao:
> @ViniciusAbrao
> 
> > Is the documentation only for users in the agent side?
> 
> YES. Right now.
> 
> Until PX4 starts using it client-side it is too early for docs.
> 
> > How the PX4's user is going to implement ROS services (for example) without microROS in the client?
> 
> See my last answer ^^^. Most of them are likely to be implementable by posting to existing uorb topics. I actually don't know of a use case that needs microROS yet.

With "ros2 topic list" we can see the topics implemented by the client, and we only have access to them. In the XRCE_DDS client module of PX4 we can see that implementation.

With "ros2 service list" there is not any of that services required to Mission PULL/PUSH, arm, change ....

In the @Jaeyoung-Lim tutorial, He did all that directly in the QGroundControl, and after the drone is on the flight, turned to offboard (also in QGroundControl) and so the ROS node is able to act. 

I am trying to see how to implement all that stuffs. If directly in the Agent side, perfect, I do not need to change the firmware (as expected). However, even in this case, the implementation in the agent side probably will be done with microROS (looks like I am a supporter of it ), because, otherwise, I think posting to existing uorb topics without microROS should be harder.

Sorry for extend my comment. It has been done only to clarify.

beniaminopozzan:
@hamishwillee 

> what's the story with standard libraries in ROS2 for doing stuff like takeoff, arm, etc?

I'm not really sure what you mean with that. At the best of my knowledge there are no such libraries. You have to manually create a publisher to `vehicle_command` and send the right message to takeoff, arm, ecc . I just checked https://github.com/PX4/px4_ros_com/pull/179 and it has everything to get a good idea of the required steps to go into offboard, arm, takeoff, land, all without even opening QGC. @ViniciusAbrao this should be useful for you too.
OH, NO microROS at all in this setup.

> I think posting to existing uorb topics without microROS should be harder.

It is not trivial but neither impossible IMO.

ViniciusAbrao:
> @hamishwillee
> 
> > what's the story with standard libraries in ROS2 for doing stuff like takeoff, arm, etc?
> 
> I'm not really sure what you mean with that. At the best of my knowledge there are no such libraries. You have to manually create a publisher to `vehicle_command` and send the right message to takeoff, arm, ecc . I just checked [PX4/px4_ros_com#179](https://github.com/PX4/px4_ros_com/pull/179) and it has everything to get a good idea of the required steps to go into offboard, arm, takeoff, land, all without even opening QGC. @ViniciusAbrao this should be useful for you too. OH, NO microROS at all in this setup.
> 
> > I think posting to existing uorb topics without microROS should be harder.
> 
> It is not trivial but neither impossible IMO.

@beniaminopozzan I could not find a mention to the Mission PULL/PUSH, which were ROS Services in the past with MAVROS. Do you have some suggestion?

beniaminopozzan:
@ViniciusAbrao oh there is no support for those features yet.

ViniciusAbrao:
> @ViniciusAbrao oh there is no support for those features yet.

 @beniaminopozzan 

I see ... that is the point. Should not be necessary the microROS in the client side in order to implement ROS Services? Otherwise, looks like it is not possible to do it directly with eProsima Micro XRCE-DDS Client.

hamishwillee:
> > @ViniciusAbrao oh there is no support for those features yet.
> 
> @beniaminopozzan
> 
> I see ... that is the point. Should not be necessary the microROS in the client side in order to implement ROS Services? Otherwise, looks like it is not possible to do it directly with eProsima Micro XRCE-DDS Client.

I don't know what these push/pull things do, but I'd argue that you do not require the microROS client for this. It might make thing _easier_ to write as a ROS node on the device, or it might not. This could be done as a PX4 module that is triggered from an external ROS node via uORB.

ViniciusAbrao:
> > > @ViniciusAbrao oh there is no support for those features yet.
> > 
> > 
> > @beniaminopozzan
> > I see ... that is the point. Should not be necessary the microROS in the client side in order to implement ROS Services? Otherwise, looks like it is not possible to do it directly with eProsima Micro XRCE-DDS Client.
> 
> I don't know what these push/pull things do, but I'd argue that you do not require the microROS client for this. It might make thing _easier_ to write as a ROS node on the device, or it might not. This could be done as a PX4 module that is triggered from an external ROS node via uORB.
> 
> Before you started you'd want to look at the requirement: it might make sense to do it somehow by file sharing.

@hamishwillee The push/pull are responsible for include/exclude points in the mission during the flight. This is very common when whe are working in the Mission Mode, when we previously the flight define the points where the drone need to pass.

@beniaminopozzan If both the ROS service server and ROS service client are in the agent side of the bridge, that is ok.... But I can imagine that for some security reason, or maybe in a scenario with more than one agent, the ROS service server  (this is unique) should be in the Client side of the bridge, and the ROS service client would be in the Agents of the bridge (more than one if necessary in case of, for example, both companion computer and ground station computer).

ViniciusAbrao:
> @hamishwillee
> 
> > what's the story with standard libraries in ROS2 for doing stuff like takeoff, arm, etc?
> 
> I'm not really sure what you mean with that. At the best of my knowledge there are no such libraries. You have to manually create a publisher to `vehicle_command` and send the right message to takeoff, arm, ecc . I just checked [PX4/px4_ros_com#179](https://github.com/PX4/px4_ros_com/pull/179) and it has everything to get a good idea of the required steps to go into offboard, arm, takeoff, land, all without even opening QGC. @ViniciusAbrao this should be useful for you too. OH, NO microROS at all in this setup.
> 
> > I think posting to existing uorb topics without microROS should be harder.
> 
> It is not trivial but neither impossible IMO.

I have translated the code to Python including: arm, take off, offboard flight and land:

https://github.com/ViniciusAbrao/px4_ros2_xrcedds/blob/master/offboard_control.py

ViniciusAbrao:
Working with real hardware (pixhawk 6C), looks like the client is not automatically started... Does anybody know how start the microdds_client ? is it via mavlink? I am trying to do it but if I work with mavlink I can not simultaneously work with the xrce-dds birdge !!

I tried to open the mavlink console in the QGroundControl, start the client, so I close the QGround and in the sequence I start the Agent. Following it I just have the following in the agent terminal:

[1676258724.774920] info | TermiosAgentLinux.cpp | init | running... | fd: 3
[1676258724.775022] info | Root.cpp | set_verbose_level | logger setup | verbose_level: 6

Looks like there is not any ROS topic being published. How do you manage it to see the topics?

mrpollo:
Hey everyone, I appreciate the help testing the PR. However, let's keep this comment thread about the changeset and start filing support issues here https://github.com/PX4/PX4-Autopilot/issues/21120 

This will help us focus on getting this PR through the finish line.

hamishwillee:
@ViniciusAbrao Re "Working with real hardware (pixhawk 6C), looks like the client is not automatically started... " - correct, that is covered in the PR that replaces this. You would normally have a line that starts it in system startup for the particular channel you're using. That will be different than the one used for mavlink unless you invent a protocol splitter (Auterion have one, but not sure it is open sourced)

hamishwillee:
I am closing this as most of the work has now been done elsewhere, and this has been data mined. Thank you @akshata-01 for taking the lead on this to get it kicked off.

mrpollo:
Thank you @akshata-01 🎉

Meowchine1:
Hello,
Maybe this is not the most appropriate theme, but I really need help — sorry for the interruption.

My goal is to send [roll, pitch, yaw, thrust] commands from ROS 2 to PX4. I was following this guide:
[ROS 2 User Guide (PX4-ROS 2 Bridge) | PX4 User Guide (v1.12)](https://docs.px4.io/v1.12/en/ros/ros2_comm.html)

However, I quickly realized that the documentation is outdated and that px4_ros_com is now deprecated ([reference](https://github.com/PX4/PX4-user_guide/pull/2094)).

I’m not targeting `make px4_sitl_rtps` at the moment, and now I’m really confused about the current supported way to send actuator commands (like `actuator_controls_0`) from ROS 2 to PX4.

Could someone please clarify the modern and recommended approach for this use case?

hamishwillee:
What version of PX4 are you using @Meowchine1 ? The docs in the library match the version. If you're using a recent version of PX4 you would use https://docs.px4.io/main/en/ros2/user_guide.html and check out https://docs.px4.io/main/en/ros2/px4_ros2_interface_lib.html

But the right place to ask these questions is the discussion forums. This area is for bug reports

PR #2099, opened 2022-10-13
Mixer purge fixes

hamishwillee:
@bkueng If you run a search on "mixer" in our docs source you get about 85 hits. Most are good or at least OK.

This is a pure "highlighting PR" with questions inline - ie. it is not to be merged. In some cases you will note questions like "can you create a PR to fix this page". Those ones are beyond me.

FYI
- Created https://github.com/PX4/PX4-Autopilot/pull/20447 to fix some of these hits.

### DIFF
--- a/en/concept/architecture.md
+++ b/en/concept/architecture.md
@@ -75,7 +75,7 @@ Its goal is to adjust the value of the process variable such that it matches the
 The output is a correction to eventually reach that setpoint.
 For example the position controller takes position setpoints as inputs, the process variable is the currently estimated position, and the output is an attitude and thrust setpoint that move the vehicle towards the desired position.
 
-A **mixer** takes force commands (such as "turn right") and translates them into individual motor commands, while ensuring that some limits are not exceeded.
+A **mixer** takes force commands (such as "turn right") and translates them into individual motor commands, while ensuring that some limits are not exceeded. 
 This translation is specific for a vehicle type and depends on various factors, such as the motor arrangements with respect to the center of gravity, or the vehicle's rotational inertia.
 
 
--- a/en/flight_controller/raspberry_pi_pilotpi_rpios.md
+++ b/en/flight_controller/raspberry_pi_pilotpi_rpios.md
@@ -227,18 +227,13 @@ It is also possible to just compile the code with command:
 
 You need to check these extra items to get your vehicle work properly.
 
-#### Mixer file 
+#### Actuator Configuration
 
-Mixer file is defined in `pilotpi_xx.conf`: 
+First set the [CA_AIRFRAME](../advanced_config/parameter_reference.md#CA_AIRFRAME) parameter for your vehicle.
 
-```sh
-mixer load /dev/pwm_output0 etc/mixers/quad_x.main.mix 
-```
+You will then be able to assign outputs using the normal [Actuator Configuration](../config/actuators.md) configuration screen (an output tab will appear for the RasPi PWM output driver).
 
-All available mixers are stored in `etc/mixers`. 
-You can create one by yourself as well.
-
-#### External compass
+#### External Compass
 
 In the startup script(`*.config`), you will find
 
--- a/en/flight_controller/raspberry_pi_pilotpi_rpios.md
+++ b/en/flight_controller/raspberry_pi_pilotpi_rpios.md
@@ -227,15 +227,16 @@ It is also possible to just compile the code with command:
 
 You need to check these extra items to get your vehicle work properly.
 
-#### Mixer file
+#### Mixer file 
 
-Mixer file is defined in `pilotpi_xx.conf`:
+Mixer file is defined in `pilotpi_xx.conf`: 
 
 ```sh
-mixer load /dev/pwm_output0 etc/mixers/quad_x.main.mix
+mixer load /dev/pwm_output0 etc/mixers/quad_x.main.mix 
 ```
 
-All available mixers are stored in `etc/mixers`. You can create one by yourself as well.
+All available mixers are stored in `etc/mixers`. 
+You can create one by yourself as well.
 
 #### External compass
 
--- a/en/flight_modes/README.md
+++ b/en/flight_modes/README.md
@@ -416,7 +416,7 @@ Key for understanding the table is as follows:
 
 Symbol | Description
 --- | ---
-M | Manual control via RC sticks. RC input is sent directly to the output mixer.
+M | Manual control via RC sticks. RC input is sent directly to the output mixer. 
 S | Assistance from autopilot to stabilize the attitude. RC input is required. Position of RC stick maps to the orientation of vehicle.
 S<sub>rate</sub> |  Assistance from autopilot to stabilize the attitude rate. RC input is required. Position of RC stick maps to the rate of rotation of vehicle in that orientation.
 S<sup>+</sup> | Assistance from autopilot to hold position or altitude against wind. RC input is required.
--- a/en/flight_modes/README.md
+++ b/en/flight_modes/README.md
@@ -153,7 +153,7 @@ th {
  <td>M</td>
  <td></td>
  <td><p>RC mode for performing acrobatic maneuvers e.g. rolls, flips, stalls and acrobatic figures.</p>
-<p>RPY stick inputs are translated to angular rate commands that are stabilized by autopilot. Throttle is passed directly to the output mixer.</p></td>
+<p>RPY stick inputs are translated to angular rate commands that are stabilized by autopilot. Throttle is passed directly to control allocation.</p></td>
 </tr>
 
 
@@ -165,7 +165,7 @@ th {
  <td>M</td>
  <td>M</td>
  <td></td>
- <td><p>RC mode where stick input is sent directly to the output mixer (for "fully" manual control).</p>
+ <td><p>RC mode where stick input is sent directly to control allocation (for "fully" manual control).</p>
    <p>This is the only mode that overrides the FMU (commands are sent via the safety coprocessor). It provides a safety mechanism that allows full control of throttle, elevator, ailerons and rudder via RC in the event of an FMU firmware malfunction.
    </p>
   </td>
@@ -332,7 +332,7 @@ th {
  <td>M</td>
  <td></td>
  <td><p>RC mode for performing acrobatic maneuvers e.g. flips, rolls and loops.</p> 
-  <p>RC RPY stick inputs control the rate of angular rotation around the respective axes. Throttle is passed directly to the output mixer.  When sticks are centered the vehicle will stop rotating, but remain in its current orientation (e.g. possibly inverted) and moving according to its current momentum.</p>
+  <p>RC RPY stick inputs control the rate of angular rotation around the respective axes. Throttle is passed directly to control allocation.  When sticks are centered the vehicle will stop rotating, but remain in its current orientation (e.g. possibly inverted) and moving according to its current momentum.</p>
  </td>
 </tr>
 
@@ -416,7 +416,7 @@ Key for understanding the table is as follows:
 
 Symbol | Description
 --- | ---
-M | Manual control via RC sticks. RC input is sent directly to the output mixer. 
+M | Manual control via RC sticks. RC input is sent directly to control allocation.
 S | Assistance from autopilot to stabilize the attitude. RC input is required. Position of RC stick maps to the orientation of vehicle.
 S<sub>rate</sub> |  Assistance from autopilot to stabilize the attitude rate. RC input is required. Position of RC stick maps to the rate of rotation of vehicle in that orientation.
 S<sup>+</sup> | Assistance from autopilot to hold position or altitude against wind. RC input is required.
--- a/en/flight_stack/controller_diagrams.md
+++ b/en/flight_stack/controller_diagrams.md
@@ -25,7 +25,7 @@ The diagrams use the standard [PX4 notation](../contribute/notation.md) (and eac
 
 * K-PID controller. See [Rate Controller](../config_mc/pid_tuning_guide_multicopter.md#rate-controller) for more information.
 * The integral authority is limited to prevent wind up.
-* The outputs are limited (in the mixer), usually at -1 and 1.
+* The outputs are limited (in the control allocation module), usually at -1 and 1.
 * A Low Pass Filter (LPF) is used on the derivative path to reduce noise (the gyro driver provides a filtered derivative to the controller).
 
   :::note
--- a/en/flight_stack/controller_diagrams.md
+++ b/en/flight_stack/controller_diagrams.md
@@ -211,8 +211,8 @@ The inputs into this block are called "virtual" as, depending on the current VTO
 For a standard and tilt-rotor VTOL, during transition the fixed-wing attitude controller produces the rate setpoints, which are then fed into the separate rate controllers, resulting in torque commands for the multicopter and fixed-wing actuators.
 For tailsitters, during transition the multicopter attitude controller is running.
 
-The outputs of the VTOL attitude block are separate torque and force commands for the multicopter (typically `actuator_controls_0`) and fixed-wing (typically `actuator_controls_1`) actuators. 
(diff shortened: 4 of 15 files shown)


en/flight_modes/README.md
```diff
@@ -416,7 +416,7 @@ Key for understanding the table is as follows:
 
 Symbol | Description
 --- | ---
-M | Manual control via RC sticks. RC input is sent directly to the output mixer.
+M | Manual control via RC sticks. RC input is sent directly to the output mixer.
```
hamishwillee, line 419: Many places say something like "is sent directly to the output mixer.". My understanding is that this is correct - you might still refer to it as an output mixer.

Would be be OK/better off writing this like "is sent directly the the control allocator"?

en/flight_stack/controller_diagrams.md
```diff
@@ -25,7 +25,7 @@ The diagrams use the standard [PX4 notation](../contribute/notation.md) (and eac
 
 * K-PID controller. See [Rate Controller](../config_mc/pid_tuning_guide_multicopter.md#rate-controller) for more information.
 * The integral authority is limited to prevent wind up.
-* The outputs are limited (in the mixer), usually at -1 and 1. 
+* The outputs are limited (in the mixer), usually at -1 and 1.
```
hamishwillee, line 28: Again, can I use the term "in the mixer"?

en/flight_stack/controller_diagrams.md
```diff
@@ -211,8 +211,8 @@ The inputs into this block are called "virtual" as, depending on the current VTO
 For a standard and tilt-rotor VTOL, during transition the fixed-wing attitude controller produces the rate setpoints, which are then fed into the separate rate controllers, resulting in torque commands for the multicopter and fixed-wing actuators.
 For tailsitters, during transition the multicopter attitude controller is running.
 
-The outputs of the VTOL attitude block are separate torque and force commands for the multicopter (typically `actuator_controls_0`) and fixed-wing (typically `actuator_controls_1`) actuators.
-These are handled in an airframe-specific mixer file (see [Mixing](../concept/mixing.md)).
+The outputs of the VTOL attitude block are separate torque and force commands for the multicopter (typically `actuator_controls_0`) and fixed-wing (typically `actuator_controls_1`) actuators. 
+These are handled in an airframe-specific mixer file (see [Mixing](../concept/mixing.md)).
```
hamishwillee, line 215: So obviously this is no longer correct - at least "handled in the mixer file" and suggestion to see mixing". Can you subedit this page in its entirety so I don't screw it up?

en/flight_controller/raspberry_pi_pilotpi_rpios.md
```diff
@@ -227,15 +227,16 @@ It is also possible to just compile the code with command:
 
 You need to check these extra items to get your vehicle work properly.
 
-#### Mixer file
```
hamishwillee, line 230: This RaPi setup suggests you use the pilotpi_xx.conf that loads the quad main mixer and maps it to /dev/pwm_output0 as shown below.

How does this get fixed - i.e. I know that you'd assign a geometry - just not sure how it gets mapped to outputs and what you do with the configuration file.

en/hardware/reference_design.md
```diff
@@ -31,10 +31,10 @@ The diagram below shows the division of bus and functional responsibilities betw
 Some Pixhawk-series controllers are built without the I/O board in order to reduce space or complexity, or to better address certain board use-cases.
 
 The I/O board is disabled by setting parameter [SYS_USE_IO=0](../advanced_config/parameter_reference.md#SYS_USE_IO).
-When the I/O board is disabled:
-- The MAIN mixer file is loaded into the FMU (so the "MAIN" outputs listed in the [Airframe Reference](../airframes/airframe_reference.md) appear on the port labeled AUX).
+When the I/O board is disabled:
```
hamishwillee, line 34: @bkueng Can I get a PR from you to update this doc?

en/simulation/flightgear.md
```diff
@@ -124,7 +124,7 @@ INFO  [simulator] Waiting for simulator to accept connection on TCP port 4560
 INFO  [simulator] Simulator connected on TCP port 4560.
 INFO  [commander] LED: open /dev/led0 failed (22)
 INFO  [commander] Mission #3 loaded, 9 WPs, curr: 8
-INFO  [init] Mixer: etc/mixers-sitl/plane_sitl.main.mix on /dev/pwm_output0
+INFO  [init] Mixer: etc/mixers-sitl/plane_sitl.main.mix on /dev/pwm_output0
```
hamishwillee, line 127: I don't want to run this simulator again just to test this. My thinking is that if I delete this line, very little would be lost right?

en/simulation/flightgear_vehicles.md
```diff
@@ -121,9 +121,9 @@ The file content meaning is as follows:
 * `FgModel` - a precise case sensitive name of the FlightGear model corresponding to "XXXX-set.xml" in the model directory (where XXXX is the model name).
 * `Url` is optional and it is not currently used. It is intended for future use to auto-download the models from web
 * `Controls` - the most important part of the process of adding a vehicle. This section contains the mapping between the PX4 mixer file and [FlightGear property tree](http://wiki.flightgear.org/Property_tree).
-  * The first number in a list selects a PX4 mixer output.
+  * The first number in a list selects a PX4 mixer output.
```
hamishwillee, line 124: Advice on what to do with this? It might be completely orthogonal to our control allocation stuff.

en/concept/architecture.md
```diff
@@ -75,7 +75,7 @@ Its goal is to adjust the value of the process variable such that it matches the
 The output is a correction to eventually reach that setpoint.
 For example the position controller takes position setpoints as inputs, the process variable is the currently estimated position, and the output is an attitude and thrust setpoint that move the vehicle towards the desired position.
 
-A **mixer** takes force commands (such as "turn right") and translates them into individual motor commands, while ensuring that some limits are not exceeded.
+A **mixer** takes force commands (such as "turn right") and translates them into individual motor commands, while ensuring that some limits are not exceeded.
```
bkueng, line 78: That can stay - the concept is the same (we're generally using mixer and allocation interchangeably)

en/flight_controller/raspberry_pi_pilotpi_rpios.md
bkueng: This is now the same as for pixhawks: you will have an output page for the rpi's pwm output driver.
You might have to set `CA_AIRFRAME` though.

en/flight_modes/README.md
bkueng: You can still use mixer here. Or `output allocation`.

en/flight_stack/controller_diagrams.md
bkueng: Here I'd use control allocation module, as it refers to the implementation.

en/flight_stack/controller_diagrams.md
bkueng:
```
The outputs of the VTOL attitude block are separate torque and force commands for the multicopter and fixed-wing actuators (two instances for `vehicle_torque_setpoint` and `vehicle_thrust_setpoint`). 
These are handled in an airframe-specific control allocation class.
```

en/hardware/reference_design.md
bkueng: This isn't relevant anymore and can be removed.
The mixing behavior won't change if you change `SYS_USE_IO`, except that you won't have the UI to configure the MAIN outputs anymore.
The behavior of the parameter changed now to not start the io driver if disabled. This means you won't have the MAIN outputs anymore, and also no RC if RC goes through the IO. I'm not sure if we even have to mention this here.

en/simulation/flightgear.md
bkueng: Yes you can just remove it.

en/simulation/flightgear_vehicles.md
bkueng: Yes it's the counterpart to the allocation on the simulator side. Essentially it just means they need to match.

en/flight_controller/raspberry_pi_pilotpi_rpios.md
bkueng: Yes you have to set `CA_AIRFRAME` as well. We'll need to improve that. It should be using the existing startup scripts and airframes.

en/flight_modes/README.md
hamishwillee: Thanks. I prefer "output allocation". Mostly because the more we are consistent with terminology, the less likely we are to have questions about "where did the mixer go" etc.
I've changed it globally-ish via search replace in this Pr.

en/flight_stack/controller_diagrams.md
hamishwillee: Thank you. I'll take that suggestion

```suggestion
The outputs of the VTOL attitude block are separate torque and force commands for the multicopter and fixed-wing actuators (two instances for `vehicle_torque_setpoint` and `vehicle_thrust_setpoint`). 
These are handled in an airframe-specific control allocation class.
```

en/simulation/flightgear_vehicles.md
hamishwillee: OK. I think this can be just left alone. Thanks.

hamishwillee, line 124:
```suggestion
  * The first number in a list selects a PX4 mixer output.
```

en/hardware/reference_design.md
hamishwillee: @bkueng It is not quite as easy as that. The problem is that FCs seem to always label the PWM used by default by motor outputs as "MAIN" irrespective of whether it is actually the FMU/AUX. Agree the RC stuff is not so relevant.

I've corrected the text as I understood it. Please recheck.

en/flight_controller/raspberry_pi_pilotpi_rpios.md
hamishwillee: Thanks. Updated this bit. @bkueng Can you please check the modification. We will have to fix this up when you sort the startup scripts.

hamishwillee:
@bkueng I think this is ready for review.

Note, there are still a bunch of fixes to do, but these are mostly in the build logs - which all suggest you just take the existing airframe config. But since all those are gone now we're screwed.
Silvan suggested we purge the very old ones - point to control allocation for the rest. We'll seek new ones to provide extras or airframe config for the SD card. Separate PR, maybe tomorrow.

en/hardware/reference_design.md
```diff
@@ -30,18 +30,13 @@ The diagram below shows the division of bus and functional responsibilities betw
...
-The I/O board is disabled by setting parameter [SYS_USE_IO=0](../advanced_config/parameter_reference.md#SYS_USE_IO).
-When the I/O board is disabled:
-- The MAIN mixer file is loaded into the FMU (so the "MAIN" outputs listed in the [Airframe Reference](../airframes/airframe_reference.md) appear on the port labeled AUX).
-  The AUX mixer file isn't loaded, so outputs defined in this file are not used.
-- RC input goes directly to the FMU rather than via the IO board.
-
-Flight controllers without an I/O board have `MAIN` ports, but they *do not* have `AUX` ports.
-Consequently they can only be used in [airframes](../airframes/airframe_reference.md) that do not use `AUX` ports, or that only use them for non-essential purposes (e.g. RC passthrough).
-They can be used for most multicopters and *fully* autonomous vehicles (without a safety pilot using RC control), as these typically only use `MAIN` ports for motors/essential controls.
+In this case [SYS_USE_IO](../advanced_config/parameter_reference.md#SYS_USE_IO) is set to `0` so that the I/O driver is not started.
```
bkueng, line 33: I would add something like '(If the board has an I/O, the driver can manually be disabled by setting SYS_USE_IO to 0)`.

en/hardware/reference_design.md
```diff
@@ -30,18 +30,13 @@ The diagram below shows the division of bus and functional responsibilities betw
...
+Because there are no I/O board outputs, the [Actuator Configuration](../config/actuators.md) will not display the MAIN tab.
```
bkueng, line 34: It will show the correct tab: e.g. on pixracer it will only show MAIN (no AUX).

The only case where it's not correct, is for boards with the same target, like the Pixhawk 4 vs Pixhawk 4 mini, which does not have an IO. In that case the mini only shows the AUX tab.

en/hardware/reference_design.md
```diff
@@ -30,18 +30,13 @@ The diagram below shows the division of bus and functional responsibilities betw
...
+Airframe actuator configurations should typically assign essential outputs to the AUX outputs (FMU) so that the configuration can still be used on controllers that do not have MAIN outputs (I/O board).
```
bkueng, line 35: Is this actually relevant?
Either way it's the opposite (yes it's confusing): it's the MAIN params/outputs that are used on both.

en/hardware/reference_design.md
hamishwillee: Sure, but why would you ever want to if you had a board with I/O? I

en/hardware/reference_design.md
hamishwillee: @bkueng Arrrgggg.  I am now possibly confused. To confirm:
- The bus used for MAIN and AUX is set by the target like `px4_fmu-v5_default` or `px4_fmu-v4_default`
- So for `px4_fmu-v5_default` FMU is set to AUX and IO is set to MAIN .
  - If you run this on Pixhawk 4 then screen-printed labels and actuators tabs are as you would expect.
  - If you run on Pixhawk 4 mini there is no IO so Actuator UI would say AUX and printing would say main.
- For `px4_fmu-v4_default` (Pixracer) there is only and FMU and it is mapped to MAIN - so there is both MAIN tab and main screenwriting.
  - Presumably this works because only Pixracer runs on this target. However you'd have similar effects on Pixhawk mini.

IN summary, the screen printing and displayed bus mostly match.

en/hardware/reference_design.md
hamishwillee: I know why I wrote this - I thought with control allocation FMU was always fixed to being AUX in all cases. This may clear up depending on your answer ^^^^

Historically advice was to use MAIN (IO) for essential outputs (i.e. motor 1 to X) and AUX (FMU) was used for less essential (gripper servo say). On board without IO MAIN was mapped to FMU so this advice still held - MAIN was always present.

From what I can tell with control allocation, MAIN will be present if you have an IO board or for targets that only have an FMU (so they map it to MAIN). MAIN will not be present on cases like the PX4 mini which only have AUX. Upshot, the advice is wrong in one of those two cases. 

> it's the MAIN params/outputs that are used on both.

The actuator docs says "PWM AUX outputs are preferred over the PWM MAIN outputs for controlling motors (they have lower latency)."; 

What should I say? :-)

bkueng: Correct.

> "PWM AUX outputs are preferred over the PWM MAIN outputs for controlling motors (they have lower latency)."

This is also still correct. You could add "If the board has an I/O chip, PWM AUX outputs are preferred..."

en/hardware/reference_design.md
bkueng: Yes exactly.

en/hardware/reference_design.md
bkueng: There's few reasons now. Reducing the cpu load a bit if you don't need it. You can also leave it away completely.

en/hardware/reference_design.md
hamishwillee: I think the current way is fine/more useful - or at least adding the detailed explanation won't help anyone.

I've removed the suggestion that main should always be used since you can no longer reliably assign to main and know the configuration will work on all airframes.

hamishwillee:
@bkueng I'm pretty sure this is good now. I'm going to merge because it is a pain to keep this rebased. It would be good if you could recheck the reference design changes though.

bkueng:
Double-checked, looks good

hamishwillee:
Thanks for checking!

PR #2100, opened 2022-10-13
Minor fixes to look at airframe reference

hamishwillee:
I'm trying to move away from refering to airframe unless I actually mean "air" frame. This fixes up some cases.

### DIFF
--- a/en/frames_plane/wing_wing_z84.md
+++ b/en/frames_plane/wing_wing_z84.md
@@ -73,6 +73,6 @@ The images below give a rough idea about the assembly process, which is simple a
 
 ## Airframe Configuration
 
-Select the Z-84 in the flying wing section of the QGC airframe config:
+Select the Z-84 in the flying wing section of the QGC airframe config: 
 
-![QGC - select firmware for West Wing](../../assets/airframes/fw/wing_wing/qgc_firmware_flying_wing_west_wing.png)
+![QGC - select firmware for West Wing](../../assets/airframes/fw/wing_wing/qgc_firmware_flying_wing_west_wing.png) 
--- a/en/advanced/parameters_and_configurations.md
+++ b/en/advanced/parameters_and_configurations.md
@@ -1,12 +1,12 @@
 # Parameters & Configurations
 
-PX4 uses the *param subsystem* (a flat table of `float` and `int32_t` values) and text files (for mixers and startup scripts) to store its configuration.
+PX4 uses the *param subsystem* (a flat table of `float` and `int32_t` values) and text files (for startup scripts) to store its configuration.
 
 This section discusses the *param* subsystem in detail.
 It covers how to list, save and load parameters, and how to define them and make them available to ground stations.
 
 :::tip
-[System startup](../concept/system_startup.md) and the way that [airframe configurations](../dev_airframes/adding_a_new_frame.md) work are detailed on other pages. 
+[System startup](../concept/system_startup.md) and the way that [frame configuration](../dev_airframes/adding_a_new_frame.md) startup scripts work are detailed on other pages.
 :::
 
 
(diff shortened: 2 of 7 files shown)


en/frames_plane/wing_wing_z84.md
```diff
@@ -73,6 +73,6 @@ The images below give a rough idea about the assembly process, which is simple a
 
 ## Airframe Configuration
 
-Select the Z-84 in the flying wing section of the QGC airframe config:
+Select the Z-84 in the flying wing section of the QGC airframe config: 
 
-![QGC - select firmware for West Wing](../../assets/airframes/fw/wing_wing/qgc_firmware_flying_wing_west_wing.png)
+![QGC - select firmware for West Wing](../../assets/airframes/fw/wing_wing/qgc_firmware_flying_wing_west_wing.png)
```
hamishwillee, line 78: @sfuhrer  This PR is mostly about only using aiframe when I mean a flying vehicle, otherwise frame. It can be ignored.

But this provides me an opportunity to ask about what to do for this frame: https://docs.px4.io/main/en/frames_plane/wing_wing_z84.html#wing-wing-z-84-pixracer-build

This refers to selecting the Wing Wing config, which no longer exists. I'd normally say - lets delete it because it is incredibly old.

But we have no other PX4 frame builds.

Is there any chance that you or someone like you can arrange the creation of a full plane build that uses bits that people can get hold of now, and which shows how to set up the control allocation?

Otherwise, welcome your suggestions.

sfuhrer: We now have "generic" airframes for the most general "types" of VTOLs and FWs:
FW:
- 2100_standard_plane
- 3000_generic_wind
VTOL:
- 13000_generic_standard_vtol
- 13100_generic_vtoltiltrotor
- 13200_generic_tailsitter_vtol

Additionally there are a couple of "off-the-shelf" integrated vehicles that can serve as examples:
- 2106_albatross
- 13013_deltaquad
- 13014_babyshark

The idea is to select the generic frame that fits your plane, and then complete the setup yourself (starting with the Allocation, then Output mapping, sensor calibration, RC config and then ofc tuning). I hope that the guide we have for the Allocation is good enough for even fairly beginners, same for the tuning guide for MC, but we should overhaul the FW tuning. Not to speak of adding a VTOL transition tuning guide. 
My idea was further to add a page for "existing builds" where developers could share their setup (build instructions, px4 parameter file to load automatically, flight logs etc). I can start by adding it for https://www.foxtechfpv.com/zmo-120-vtol-fpv.html, we've been flying it now for a while.

hamishwillee: Thanks @sfuhrer . I get all that but it does not solve the problem. 

We already have the "existing builds" for VTOL - peer docs of 
- http://docs.px4.io/main/en/frames_vtol/vtol_quadplane_fun_cub_vtol_pixhawk.html
- http://docs.px4.io/main/en/frames_vtol/vtol_tailsitter_caipiroshka_pixracer.html
- http://docs.px4.io/main/en/frames_vtol/vtol_tiltrotor_eflite_convergence_pixfalcon.html

And just one for FW
- http://docs.px4.io/main/en/frames_plane/wing_wing_z84.html

Telling me that we have generic ones now does not fix them. Providing instructions that people could use instead would fix the problem. But that's hard - I don't think we can automate this from the old mixer config right?

The babyshark etc are cool for config files, but not for the rest of this stuff about assembly. The build logs are good because they show a bit about real world assembly. We could do a lot more here. For example, we say "I got these motors and servos" but not much about the selection criteria.

Yes, going on, people will do this from scratch and it will be better . WOuld love your https://www.foxtechfpv.com/zmo-120-vtol-fpv.html to be added. We also need to keep encouraging people to contribute these.

But what do I do about all the existing build logs in the meantime?

The reason I raised this for Wing Wing is that it is the only FW frame we have. So if I just delete that, we have nothing.

PR #2103, opened 2022-10-14
add new manufacturer supported autopilot: VOXL 2

modaltb:
Add new manufacturer supported flight controller, VOXL 2.
Where applicable, we are linking out to already public facing documents.


hamishwillee:
@modaltb Thanks for this. Github would not let me subedit the PR, so I have recreated in PR #2106 and will close this.

Mostly just minor modifications to shrink the images.
It isn't quite "standard" format, but I appreciate that you probably don't want to maintain two sets of documentation.

PR #2105, opened 2022-10-18
Update ignition_gazebo.md

jmackay2:


Jaeyoung-Lim (approved):
Thanks!

en/simulation/ignition_gazebo.md
```diff
@@ -18,44 +18,48 @@ See [Simulation](../simulation/README.md) for general information about simulato
 ## Installation (Ubuntu Linux)
 
 :::note
-These instructions were tested on Ubuntu 18.04 and Ubuntu 20.04
+These instructions were tested on Ubuntu 18.04, Ubuntu 20.04, and Ubuntu 22.04
 :::
 
 1. Install the usual [Development Environment on Ubuntu LTS / Debian Linux](../dev_setup/dev_env_linux_ubuntu.md).
-1. Install Ignition Gazebo following the [installation instructions](https://github.com/Auterion/px4-simulation-ignition#readme) (`sudo` may be required):
```
hamishwillee, line 25: FYI @Jaeyoung-Lim FYI I removed this link as the Auterion docs did not match the updates.

More generally, where possible all docs should go in the guide. Sometimes they might go in PX4 tree. But linking to docs in Auterion tree is annoying.

But these

en/simulation/ignition_gazebo.md
```diff
@@ -18,44 +18,48 @@ See [Simulation](../simulation/README.md) for general information about simulato
...
    ```
 
 ## Running the Simulation
 
 Ignition Gazebo SITL simulation can be conveniently run through a `make` command as shown below:
+
 ```bash
 cd /path/to/PX4-Autopilot
-make px4_sitl ignition
+make px4_sitl gz_x500
```
hamishwillee, line 40: Hmmm, @Jaeyoung-Lim

FMI Why this naming convention? Does this reflect that you'll use `gz_` for ignition frames in future?

hamishwillee:
Thanks very much for this - the instructions "work", though on my Ubuntu 20.05 the gazebo simulation windows started but was black. That would be some other problem though :-)

hamishwillee:
PS @Jaeyoung-Lim What happens if you have both normal gazebo and ignition fortress installed? Do you think I could add this to ubuntu.sh setup script as well as normal gazebo?

Jaeyoung-Lim:
@hamishwillee This would not introduce any problems. However, installing two simulations as a dependency might be a bit too much IMHO

I think the problem now is that PX4 does not guide you how to install gzsim when somebody tries to run it without installing. I was trying to look into this : 😅

hamishwillee:
> However, installing two simulations as a dependency might be a bit too much IMHO

I strongly disagree because:
- we need it in CI to make sure it is being hammered/tested. That means in ubuntu.sh.
- you can turn off SIM if you like, and there is nothing to stop us adding an additional condition on this.

hamishwillee:
@Jaeyoung-Lim Awesome - are you doing that or me?

PR #2112, opened 2022-10-20
More clear about gimbal support

hamishwillee:
This is a follow up to #2085 to better explain gimbal support on PX4. Still draft, as I'm after confirmations.

The update is to better capture the story for multiple gimbals and PX4. It is draft because I don't know the answers yet.

So the way that the gimbal protocol works, 
- High level gimbal manager API and low level device API.
- there is a 1:1 relationship between gimbals manager and gimbal device.
- Ground station talks to gimbal manager, which then sends commands to the connected low level device.
- a gimbal might implement either or both sets of messages. 
  - If it implemented just the high level API you would have to connect to the gimbal using that API.
  - If it implements the low level API then you could control it with "some" gimbal manager - i.e. PX4 could be the thing that sends it messages.
- The ground station can discover any gimbal manager on the connected system using provided messages.

The gimbal docs on MAVLink assume that a mavlink node like PX4 will autodiscover all gimbal devices and roll out gimbal manager instances to manage them. Is that what it does?

What I think has been done (minimally) is to implement the gimbal manager protocol in PX4. This allows PX4 to publish itself as a gimbal manager and control either a PWM gimbal or a gimbal that implements the gimbal device API connected via mavlink.
I **think** then that the assumption is that if you have other gimbal devices on the device they will have to have their own managers and that you will have to configure PX4 to forward any traffic from GCS to them.

Is that right? Or does PX4 also creates a manager for every other device on the system?

Is the first gimbal component ID "reserved" for a hardware connected PWM gimbal? i.e. how is the ID allocated in this case? My assumption, since the manager uses the gimbal id to connect to, is that these must be individually made to be different.

### DIFF
--- a/en/advanced/gimbal_control.md
+++ b/en/advanced/gimbal_control.md
@@ -57,8 +57,6 @@ You can support additional gimbals provided that they:
 - Each gimbal must have a unique component id.
   For a PWM connected gimbal this will be the component ID of the autopilot
 
-Note that MAVLink does not support control of individual cameras in missions.
-Therefore any command sent in a mission will be broadcast to all connected cameras.
 
 ## Gimbal on FC PWM Output (MNT_MODE_OUT=AUX)
 
--- a/en/advanced/gimbal_control.md
+++ b/en/advanced/gimbal_control.md
@@ -27,26 +27,24 @@ The output is set using the [MNT_MODE_OUT](../advanced_config/parameter_referenc
 By default the output is set to a PXM port (`AUX (0)`).
 If the [MAVLink Gimbal Protocol v2](https://mavlink.io/en/services/gimbal_v2.html) is supported by your gimbal, you should instead select `MAVLink gimbal protocol v2 (2)`.
 
-
 The full list of parameters for setting up the mount driver can be found in [Parameter Reference > Mount](../advanced_config/parameter_reference.md#mount).
 The relevant settings for a number of common gimbal configurations are described below.
 
 ## MAVLink Gimbal (MNT_MODE_OUT=MAVLINK)
 
-To enable a MAVLink gimbal, first set parameter [MNT_MODE_IN](../advanced_config/parameter_reference.md#MNT_MODE_IN) to `MAVlink gimbal protocol v2` and [MNT_MODE_OUT](../advanced_config/parameter_reference.md#MNT_MODE_OUT) to `MAVLink gimbal protocol v2`. 
+Each physical gimbal device on the system must have its own high level gimbal manager, which is discoverable by a ground station using the MAVLink gimbal protocol.
+The ground station sends high level [MAVLink Gimbal Manager](https://mavlink.io/en/services/gimbal_v2.html#gimbal-manager-messages) commands to the manager, and it will in turn send appropriate lower level "gimbal device" commands to control the gimbal.
+
+PX4 can be configured as the gimbal manager to control a single gimbal device (which can either be physically connected or be a MAVLink gimbal).
+Additional MAVLink gimbals may be controlled by their own gimbal manager components (PX4 will forward the MAVLink traffic).
+
+To enable a MAVLink gimbal, first set parameter [MNT_MODE_IN](../advanced_config/parameter_reference.md#MNT_MODE_IN) to `MAVlink gimbal protocol v2` and [MNT_MODE_OUT](../advanced_config/parameter_reference.md#MNT_MODE_OUT) to `MAVLink gimbal protocol v2`.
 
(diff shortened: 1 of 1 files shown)


en/advanced/gimbal_control.md
```diff
@@ -27,26 +27,38 @@ The output is set using the [MNT_MODE_OUT](../advanced_config/parameter_referenc
...
 For example, if the `TELEM2` port on the flight controller is unused you can connect it to the gimbal and set the following PX4 parameters:
 - [MAV_1_CONFIG](../advanced_config/parameter_reference.md#MAV_1_CONFIG) to **TELEM2** (if `MAV_1_CONFIG` is already used for a companion computer (say), use `MAV_2_CONFIG`).
 - [MAV_1_MODE](../advanced_config/parameter_reference.md#MAV_1_MODE) to **NORMAL**
 - [SER_TEL2_BAUD](../advanced_config/parameter_reference.md#SER_TEL2_BAUD) to manufacturer recommended baud rate.
 
-With this setup, ground stations can send [MAVLink Gimbal Commands](https://mavlink.io/en/services/gimbal_v2.html#gimbal-manager-messages) to PX4, such as `MAV_CMD_DO_SET_ROI_LOCATION` to track region of interest and `MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW` to control the gimbal attitude, and so on.
+### Multiple Gimbal Support
+
+PX4 does not automatically create gimbal managers for other gimbal devices on the system.
```
julianoes, line 50:
```suggestion
PX4 does not automatically create gimbal managers for more than one gimbal devices on the system.
```

en/advanced/gimbal_control.md
```diff
@@ -27,26 +27,38 @@ The output is set using the [MNT_MODE_OUT](../advanced_config/parameter_referenc
...
+  This may require that traffic forwarding be configured between PX4, the GCS, and the gimbal.
+- Each gimbal must have a unique component id.
+  For a PWM connected gimbal this will be the component ID of the autopilot
 
-:::note
-In Gimbal Protocol v2 terms, this setup configures PX4 as a _Gimbal Manager_.
-The flight stack handles high level "gimbal manager" commands and mission items, sending appropriate lower level "gimbal device" commands to control the gimbal.
-:::
+Note that MAVLink does not support control of individual cameras in missions.
+Therefore any command sent in a mission will be broadcast to all connected cameras.
```
julianoes, line 61: I don't think that's true. You can specify the gimbal device in the commands. E.g. in [MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW) you can set the "Gimbal device ID" in `param7`.

en/advanced/gimbal_control.md
hamishwillee: @julianoes Hmmmm How about this:

```suggestion
PX4 automatically creates a gimbal manager for a connected PWM gimbal or the first MAVLink gimbal device it detects on any interface.
It does not automatically create gimbal manager for any other MAVLink gimbal devices it detects. 
```

en/advanced/gimbal_control.md
hamishwillee, line 61:
```suggestion
```

julianoes: Right, so you're saying you would want some sort of gimbal ID, rather than a component ID. And the reasoning is that you don't know the gimbals connected when you plan a mission offline. That's true but I would argue that you also don't know the gimbal IDs essentially, you just don't really know without connecting.

en/advanced/gimbal_control.md
julianoes: I think that's ok. However, both what I suggested and what you wrote is assuming PX4 doesn't do anything stupid when it detects two gimbals like toggling back and forth between them. Well, it might do just that as it's untested. :smile:

en/advanced/gimbal_control.md
hamishwillee: Good point.

A gimbal id would be slightly better than a component ID for the limited case where you can assume all gimbal payloads are the same. IN this case you could set the ID for each gimbal without having to know what cameras are attached, and create a generic mission that would work on any vehicle that had three cameras.

All a bit abstract though, since something still needs to map ids to gimbals etc.

If gimbal payloads might be different then you might be able to specify by requirement on the camera - i.e. ROI command to go to gimbals that support IR cameras, or whatever. That seems like something  you might want, and we could think about in the far future.

Upshot. I have convinced myself that IDs make enough sense for now, and we have to live with non-generic missions until we have a stakeholder who sees value in hammering this out.

en/advanced/gimbal_control.md
hamishwillee: I need to make that assumption :-). I was hoping you'd tell me that it is correct. I'll merge this, but we might get a bug report if it proves not to be true.

hamishwillee:
OK. Merging this. I'll add info about supported mavlink gimbals as a separate PR.

PR #2127, opened 2022-11-02
Minor Turbo Timber Evolution fixes post merge

chris1seto:
Just a couple of tweaks to edits that went in immediately before the merge. 
* Fix empty param file
* Fix blurb about FPV integration 
* Add new information about a *very* recent test flight to confirm the suggested power system is working correctly


en/frames_plane/turbo_timber_evolution.md
```diff
@@ -64,8 +64,10 @@ For this portion of the assembly, simply follow the manual.
...
 
-As this build will be especially heavy, and therefore likely demand higher average power from the ESC, the default 60A ESC was replaced with an 80A Turnigy PLUSH-32 during testing.
-The stock motor will also be replaced with a [higher power motor](https://hobbyking.com/en_us/turnigy-aerodrive-sk3-3548-840kv-brushless-outrunner-motor.html) (this has not yet been flight-tested).
+As this build will be especially heavy, and therefore likely demand higher average power from the ESC, the stock 60A ESC was replaced with an 80A Turnigy PLUSH-32 during testing.
+The stock motor was also replaced with a [higher power motor](https://hobbyking.com/en_us/turnigy-aerodrive-sk3-3548-840kv-brushless-outrunner-motor.html.
+The stock prop was replaced with an [APC 13x4](https://www.apcprop.com/product/13x4/) for better efficiency from the stock tri-blade prop.
+In testing, I found this new esc, motor, and prop combination to perform well.
```
hamishwillee, line 70:
```suggestion
As this build will be especially heavy, and therefore likely demand higher average power from the ESC, the stock 60A ESC was replaced with an 80A Turnigy PLUSH-32 during testing.
The stock motor was also replaced with a [higher power motor](https://hobbyking.com/en_us/turnigy-aerodrive-sk3-3548-840kv-brushless-outrunner-motor.html).
The stock propeller was replaced with an [APC 13x4](https://www.apcprop.com/product/13x4/) for better efficiency than the stock tri-blade prop.
This new ESC, motor, and propeller combination performs well in testing.
```

en/frames_plane/turbo_timber_evolution.md
```diff
@@ -178,10 +174,10 @@ Heatshrink was used to secure the dupont connector of the cable such that it can
 
 #### FPV Camera (Caddx Vista)
 
-
-Another custom cable was made to connect the Caddx Vista to the FCU UART and power.
+Another custom cable was made to connect the Caddx Vista FPV camera to the FCU UART (from the `UART/I2C B` port) and battery power from the Holybro power module.
 A Molex microfit was added close to the Vista so that it could be easily disconnected without needing to gain access to the Pixhawk.
-The `msp_osd` module is used to stream telemetry to the Caddx Vista which can be seen on the DJI Goggles with the "custom OSD" feature enabled.
+
+The [msp_osd](../modules/modules_driver.md#msp-osd) module is used to stream telemetry to the Caddx Vista which can be seen on the DJI Goggles with the "custom OSD" feature enabled.
```
hamishwillee, line 180: @chris1seto As per comment in original PR it is a complete surprise to me that FPV has any integration with PX4 - I obviously did not read this properly. Let's get that documented separately please in peripherals.

This seems to have open questions to me.
1. For the custom cable, can you provide a pin map for the connections so someone could create this cable?
2. Did you need to do anything to start the msp_osd module or was it automatically started? 
3. Did you need to do any configuration? i.e. how does the module know what port to use and messages to send
   Normally you'd need to set up the serial link as in http://docs.px4.io/main/en/peripherals/mavlink_peripherals.html
4. The module documentation is a bit "thin": http://docs.px4.io/main/en/modules/modules_driver.html#msp-osd
   - "Does the things for the DJI Air Unit OSD" - I assume "MSP OSD" is a standard protocol? Is it used by anyone other than DJI? What googles does it work with?
     - or to put it another way "if I'm a new user, how can I work out whether PX4 can work with a particular pair of goggles".
   - Command arguments are not documented. Which is OK if there aren't any???

hamishwillee:
Note, this still doesn't have param file fixes. I assume that is why it is still in draft. If so, sorry for early review.

en/frames_plane/turbo_timber_evolution.md
chris1seto: Yeah, myself and another individual wrote msp_osd. That was actually one of the reasons this plane was built, to test it. There's no documentation for the module itself yet, but there's also nearly no configuration. It has a single param for which serial port to start on, and if that param is set then it will autostart and start streaming telemetry. 

I can document the cable pinout in this guide, but I think the rest of it should be documented in a more generic page?

hamishwillee: > I can document the cable pinout in this guide, but I think the rest of it should be documented in a more generic page?

Yes. Though you could, if you wish, put the cable pinout in the more generic page too under a section for the specific hardware used. That would be OK because the page won't be very big, and we can extend this out to other FPV systems as people discover them in the way way we have sections for other hardware.

> It has a single param for which serial port to start on, and if that param is set then it will autostart and start streaming telemetry.

That's cool. What we're after is making it clear what people have to do. That includes things that they "don't have to do but that they reasonably would expect to have to" - like setting the serial port and telemetry setup.

IF you write a generic page starting point that covers what your setup does then I can fluff it out to make it very generic. I.e. adding blurb about what FPV is, and that it isn't "generically supported" but we do support the XYZ protocol blah blah blah.
IN other words, something that others can use now, and we can extend the docs if other things get added.

Hope that makes sense.

hamishwillee:
Let me know when you're good for this to be merged. The generic FPV should be a separate PR that this links to (added later)

en/frames_plane/turbo_timber_evolution.md
chris1seto: The only thing that really differs about generic use of msp_osd vs this setup is that one i2c is combined with one uart into one connector on this fmu, whereas that may or may not be the case with other fmus. 

In this case, the param should actually be implicitly done in the config, do you think this needs to be mentioned explicitly?

chris1seto:
Hi @hamishwillee I released this from draft PR, though it could be the case that you have more items that you want to address. So just let me know if you want more changes or if you think it's good for the merge.

en/frames_plane/turbo_timber_evolution.md
hamishwillee: Yes, but as an aside. I think if there is something that might break if someone tries this with another FC it would be good to record.

hamishwillee:
@chris1seto I think this is excellent to get in now. We might come back to it when we have a generic FPV setup page and provide a cross link. Make sense?

chris1seto:
> @chris1seto I think this is excellent to get in now. We might come back to it when we have a generic FPV setup page and provide a cross link. Make sense?

Thanks very much @hamishwillee !! After the RD2 build is in I'll start the generic FPV page.

PR #2128, opened 2022-11-02
uorb: small correction regarding interrupt context

julianoes:


en/middleware/uorb.md
```diff
@@ -41,7 +41,7 @@ Then in the code, use them as topic id: `ORB_ID(offboard_mission)`.
 ## Publishing
 
 Publishing a topic can be done from anywhere in the system, including interrupt context (functions called by the `hrt_call` API).
-However, advertising a topic is only possible outside of interrupt context.
+However, the topic needs to be advertised and published once beforehand outside of interrupt context.
```
hamishwillee, line 44:
```suggestion
However, the topic needs to be advertised and published outside of an interrupt context (at least once) before it can be published in an interrupt context.
```

en/middleware/uorb.md
```diff
@@ -41,7 +41,7 @@ Then in the code, use them as topic id: `ORB_ID(offboard_mission)`.
 ## Publishing
 
 Publishing a topic can be done from anywhere in the system, including interrupt context (functions called by the `hrt_call` API).
-However, advertising a topic is only possible outside of interrupt context.
+However, the topic needs to be advertised and published outside of an interrupt context (at least once) before it can be published in an interrupt context.
 A topic has to be advertised in the same process as it's later published.
```
hamishwillee, line 45: @julianoes I've subedited that line. Would it be tidier to change the whole thing to:

```
A topic has to be advertised before it is published, and this must be done in the same process as it's later published.
Publishing a topic can be done from anywhere in the system.
Note however that the topic must be advertised and published outside of an interrupt context (at least once) before it can be published in an interrupt context.
```

julianoes: > A topic has to be advertised before it is published, and this must be done in the same process as it's later published.

Where do you take that from? What's a process? I don't think that's quite right for how we use it in NuttX.

hamishwillee: @julianoes Yup - all I did was a reword to make the whole thing flow better.

If the following line is incorrect we should fix it - but I don't know enough to know "how". Suggestions? Something for Beat?

julianoes: Right, well your suggestion is good then.

hamishwillee: Thanks. I'm actually going to merged your suggestion but let's check with Beat

@bkueng Julian had a query about the line that follows this that a topic must be advertised in the same process as it is published - but thre are no processes as such - see https://github.com/PX4/PX4-user_guide/pull/2128#discussion_r1014882914 - is what we're saying OK?

bkueng: You could argue that process==task, just w/o memory separation in our case.
As we use process/task-independent advertisement handles it does not need to happen in the same process/task.
The change here is correct though.

hamishwillee: @bkueng @julianoes so to be very clear, we delete the line "A topic has to be advertised in the same process as it's later published."?

hamishwillee: Thanks - done in https://github.com/PX4/PX4-user_guide/commit/134be1fa961a6aa119c71f57c2a49d0d6c2fbb36

PR #2150, opened 2022-11-21
Add ESC comment in the X500V2 guide

akshata-01:


en/frames_multicopter/holybro_x500v2_pixhawk6c.md
```diff
@@ -76,12 +76,17 @@ Although the design is well built such that you can do this later as well.
 
 ### Power
 
-- The Pixhawk 6C gets powered by a power module PM02 (in this case).
-- This power module is supplied by a battery (4S 16.8V 5200 mAh)
-- The motors get power through the power distribution board.
+The Pixhawk 6C gets powered by a power module PM02 (in this case).
+This power module is supplied by a battery (4S 16.8V 5200 mAh)
```
hamishwillee, line 80: When you split the docs up like this it is obvious that we don't have an image of the power module stuff. If you have one it would be good to add. I'll merge this, but if you can add it later that would b egood.

hamishwillee:
HI @akshata-01
1. I assume that this was actually finished and you just forgot to move it from Draft to Ready for Review? I hope so, I have moved it now.
2. The image was more than 1 MB and had lots of white space that was irrelevant. I have cropped it to the relevant information and shrunk it.
3. The image had "JPG" upper case in the filename. Because the build toolchain might run on an OS that is filename case-sensitive best to make all filenames lower case. That way they build on both WIndows and Linux. It shouldn't matter, but it is so easy to accidentally get the case wrong when you reference a file.

Merging. Thanks!

PR #2152, opened 2022-11-23
Update px4_vision_kit.md

jhiggason:
Uncommented out the section that shows how to install the USB image to the internal eMMC.  I followed these steps on my own device and they worked...not sure why they were commented out.


hamishwillee:
@vincentpoont2 - any idea why the sections on installing the USB image might have been commented out?

vincentpoont2:
If I recalled correctly, there was user having issue following these step. If @jhiggason followed this without issue, I think we can put it back.

hamishwillee:
Awesome! Thanks @jhiggason

PR #2153, opened 2022-11-24
Fix DShot and servo mixing note

chris1seto:
Previously, the DShot notes mentioned that PWM and DShot could not be mixed on FMU outputs. In recent versions of PX4, both V5X and v6X boards support two separate blocks of DShot output, and PWM is supported on any channels which are not inside DShot enabled blocks.

### DIFF
--- a/en/peripherals/dshot.md
+++ b/en/peripherals/dshot.md
@@ -34,13 +34,11 @@ Developers might alternatively modify the [airframe AUX mixer](../dev_airframes/
 :::note
 FMUv5-based boards (e.g. Pixhawk 4 or CUAV Pixhawk V5+) support DShot only on the first four FMU pins due to hardware conflicts.
 The other pins cannot be used as motor/servo outputs.
-FMUv5x-based boards support DShot only on the first six FMU pins.
-:::
 
-:::tip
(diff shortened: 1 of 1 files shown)


en/peripherals/dshot.md
```diff
@@ -34,14 +34,12 @@ Developers might alternatively modify the [airframe AUX mixer](../dev_airframes/
 :::note
 FMUv5-based boards (e.g. Pixhawk 4 or CUAV Pixhawk V5+) support DShot only on the first four FMU pins due to hardware conflicts.
 The other pins cannot be used as motor/servo outputs.
-FMUv5x-based boards support DShot only on the first six FMU pins.
-:::
 
-:::tip
-You can't mix DShot ESCs/servos and PWM ESCs/servos on the FMU (DShot is enabled/disabled for *all* FMU pins on the port). 
+FMUv5x and FMUv6x based boards support DShot only on the a block of channels 1 to 4, and a block of channels 5 and 6.
```
bkueng, line 38: I would use 'group' instead of 'block' to be consistent with what is used elsewhere.

chris1seto: Sounds good, I will change this verbiage!

bkueng (approved):

PR #2163, opened 2022-11-27
Update autotune params and test steps

chris1seto:
As per https://github.com/PX4/PX4-Autopilot/pull/20069, the autotuner module has been updated to support being enabled/disabled by an aux channel and allow RC stick inputs during autotuning. This PR is to update the docs to support the new functionality and behavior.

### DIFF
--- a/en/config/autotune.md
+++ b/en/config/autotune.md
@@ -9,7 +9,7 @@ Tuning only needs to be done once, and is recommended unless unless you're using
 Auto-tuning is performed while flying.
 The airframe must fly well enough handle moderate disturbances, and should be closely attended:
 - Test that your vehicle is [stable enough for autotuning](#pre-tuning-test).
-- Be ready to abort the autotuning process by moving the remote control sticks.
+- Be ready to abort the autotuning process by switching flight modes or toggling the aux auto-tune enable switch to the disabled position
 - Verify that the vehicle flies well after tuning.
 :::
 
@@ -51,7 +51,13 @@ It takes about 40 seconds ([between 19 and 68 seconds](#how-long-does-autotuning
 For best results, we recommend running the test in calm weather conditions.
 
 :::note
-The sequence can be aborted at any time by the operator by moving the roll/pitch stick on the RC controller.
+The auto-tuning sequence can be aborted at any time by the operator by switching flight modes.
+If an [aux switch was used](#Parameters) to enable the auto-tune sequence, the switch can be toggled back to the disabled position to instantly stop autotuning.
+:::
+
+:::note
+During auto tuning, the RC sticks can *still* be used to fly the vehicle.
+[Hold mode](../flight_modes/hold.md) is useful for keeping a fixed wing vehicle circling in one place, but any other flight mode will work as well.
 :::
 
 The test steps are:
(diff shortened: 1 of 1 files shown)


bresch:
Thanks, we should probably also add this function in the MC autotune as well

hamishwillee:
> Thanks, we should probably also add this function in the MC autotune as well

@bresch Is there a plan to do so? I will rewrite this as FW only.

en/config/autotune.md
```diff
@@ -131,6 +143,21 @@ This behaviour can be configured using the [MC_AT_APPLY](../advanced_config/para
...
+
+A remote control switch can be configured to enable/disable autotune (in any mode) using an RC AUX channel.
+
+To map a switch:
+
+1. Select an RC channel on your controller to use for the autotune enable/disable switch.
+1. Set [RC_MAP_AUX1](../advanced_config/parameter_reference.md#RC_MAP_AUX1) to match the RC channel for your switch (you can use any of `RC_MAP_AUX1` to `RC_MAP_AUX6`).
+1. Set [FW_AT_MAN_AUX](../advanced_config/parameter_reference.md#FW_AT_MAN_AUX) to the selected channel (i.e. `1: Aux 1` if you mapped `RC_MAP_AUX1`).
+
+The auto tuner will be disabled when the switch is below `0.5` (on the manual control setpoint range of of `[-1, 1]` and enabled when the switch channel is above `0.5`.
```
hamishwillee, line 156: Note, you had some info about the possible switch end points. I don't _think" that was needed, in particular because switch values are likely to vary/be configurable.

en/config/autotune.md
```diff
@@ -23,7 +24,7 @@ This test ensures that the vehicle can fly safely in position controlled modes.
 
 :::note
 During [Airframe Setup](../config/airframe.md) you should have selected the frame that most closely matches your vehicle.
-This will usually be tuned well enough to fly, and it _may_ also be sufficiently well tuned to run autotuning.
+This may fly well enough to run autotuning.
```
hamishwillee, line 27: Changed, as most airframes that people will select now are generic, so less likely good enough to fly "out of the box". Not sure what people are supposed to do in this case.

hamishwillee:
I have subedited this to:
1. My taste - now the params options has grown I've split it up a bit.
2. To make it clear the switch is only for FW.

Please review the changed version. I will push it in anyway because I think it is likely correct.

en/config/autotune.md
chris1seto: @hamishwillee Yeah this is where I need help determining what level of information is necessary to get the point across. I don't know if it's missing information that when an RC channel is mapped to an aux, it's end point is translate from 1000-2000 to [-1, 1], so the point here is that a user needs to make sure their switch has endpoints correctly configured. What do you think?

en/config/autotune.md
chris1seto: I would generally encourage people to start the bringup process with stock tunings and immediately autotune after a normal takeoff in manual mode. I know this isn't mainstream PX4 workflow, but this process has yet to fail me.

chris1seto:
@hamishwillee Your edits look good and concise, thank you much for the review and merge!

en/config/autotune.md
hamishwillee: @bresch 

- How risky would it be to advise this? I.e. use a stock tuning "generic X quad", "generic standard VTOL", configure, take off, and just run autotune without the pre-test? 
- What do you do if you can't get off the ground in the first place - i.e the stock tuning doesn't allow that? Does that mean control allocation is wrong?

@chris1seto Thanks. I suspect it does work in most cases. But then you have experience in brining up sane frames. I'm trying to mitigate risk for newbies who might not try again if their vehicle dies.

en/config/autotune.md
hamishwillee: @chris1seto I don't know what the end points refer to so I can't answer.

My thinking
- an RC switch with two switch positions will output different PWM values in each. 
- A channel has some range it can output, and the switch position values will fall between those.
  - Is this what you mean by "end points" - the range of possible values in the channel?
  - Does the channel range value vary by radio or are you saying it is always  1000-2000 us?
- If PX4 maps the range to -1 and +1 it must have some way of knowing/setting the input range?

I think you are saying that we can make the end points such that the two values actually emitted by the switch are on either side of the mapped +0.5 value.

PR #2194, opened 2022-12-08
QAV R5 kiss fixes

hamishwillee:
Fixes up the P4 configuration section for consistency with current approach - also makes it look more up to date, since all that was PX4 v7.

Also shrink some images.


en/frames_multicopter/qav_r_5_kiss_esc_racer.md
```diff
@@ -217,30 +222,62 @@ I mounted it like you can see in the next picture. The two screws and nuts to lo
...
+  You will need to select the *Generic 250 Racer* airframe (**Quadrotor x > Generic 250 Racer**).
+
+  ![QGC airframe selection of generic 250 racer](../../assets/airframes/multicopter/qav_r_5_kiss_esc_racer/qgc_airframe_generic_250_racer.png)
+
+Then set the geometry and outputs
+
+- [Actuators](../config/actuators.md)
+  - Update the vehicle geometry to match the QAV-R 5" KISS ESC Racer
+  - Assign actuators functions to outputs as wired.
+  - Test the configuration using the sliders.
```
hamishwillee, line 254:
```suggestion
- [Actuators](../config/actuators.md)
  - You should not need to update the vehicle geometry (as this is a preconfigured airframe).
  - Assign actuator functions to outputs to match your wiring
  - Test the configuration using the sliders. 
```

en/frames_multicopter/qav_r_5_kiss_esc_racer.md
```diff
@@ -217,30 +222,62 @@ I mounted it like you can see in the next picture. The two screws and nuts to lo
...
+  You will need to select the *Generic 250 Racer* airframe (**Quadrotor x > Generic 250 Racer**).
+
+  ![QGC airframe selection of generic 250 racer](../../assets/airframes/multicopter/qav_r_5_kiss_esc_racer/qgc_airframe_generic_250_racer.png)
+
+Then set the geometry and outputs
+
+- [Actuators](../config/actuators.md)
+  - You should not need to update the vehicle geometry (as this is a preconfigured airframe).
+  - Assign actuator functions to outputs to match your wiring
+  - Test the configuration using the sliders.
```
hamishwillee, line 254: Hi @bkueng I've been updating all the build logs to have a more or less common PX4 configuration section. 

Originally I just put you should not have to configure Actuators for these things that have a canned airframe. However trying this out it looks like the geometry is set, but the outputs are never mapped to specific motors/servos by default - you always have to do that.

So my understanding is that this is what I should say in all of the "specific" airframes. For a "generic airframe" I would have to also mention setting the geometry.

That about right?

bkueng: Partially, we're not using a specific airframe here. It's still generic in the sense that it's for a class of vehicles.

hamishwillee: @bkueng Ah, curses - this is "generic". But for [X500 v2](http://docs.px4.io/main/en/frames_multicopter/holybro_x500V2_pixhawk5x.html#px4-configuration) you would not need to update the geometry because it should be set.

bkueng: Yes that one should include both, geometry & outputs.

PR #2197, opened 2022-12-10
Version 1.13: Update PX4-ROS2 and ROS(1) over ROS2 bridge instructions

beniaminopozzan:
- Instructions and links updated from `main` to `release/1.13` branches.
- Instructions for working on ROS (1) over the ROS2 bridge updated with message definitions synchronization step.

### DIFF
--- a/en/ros/ros1_via_ros2.md
+++ b/en/ros/ros1_via_ros2.md
@@ -79,14 +79,16 @@ To use ROS (1) **and** ROS 2 (you need both for this!):
 
 1. [Setup your PX4 Ubuntu Linux development environment](../dev_setup/dev_env_linux_ubuntu.md) - the default instructions get the latest version of PX4 source and install all the needed tools.
 
-:::warning Note
-The above instructions must be adjusted to get version v1.13 of PX4 source.
-```sh
-git clone -b <tag> https://github.com/PX4/PX4-Autopilot.git --recursive
-```
-where `<tag>` can be `release/1.13` or any of the `v.13.*` tags depending on the user specific scenarios.
-Otherwise, please refer to the [latest](https://docs.px4.io/main/en/contribute/) documentation.
-:::
+   :::warning Note
+   The above instructions must be adjusted to get version v1.13 of PX4 source.
+   
+   ```sh
+   git clone -b <tag> https://github.com/PX4/PX4-Autopilot.git --recursive
+   ```
+   
+   where `<tag>` can be `release/1.13` or any of the `v.13.*` tags depending on the user specific scenarios.
+   Otherwise, please refer to the [latest](https://docs.px4.io/main/en/contribute/) documentation.
+   :::
 
 1. Open a new terminal in the root of the **PX4 Autopilot** project, and then start a PX4 Gazebo simulation using:
    ```sh
--- a/en/ros/ros1_via_ros2.md
+++ b/en/ros/ros1_via_ros2.md
@@ -51,6 +51,12 @@ To create and build the workspace:
    $ git clone https://github.com/PX4/px4_ros_com.git ~/px4_ros_com_ros1/src/px4_ros_com -b ros1 # clones the 'ros1' branch
    $ git clone https://github.com/PX4/px4_msgs.git ~/px4_ros_com_ros1/src/px4_msgs -b ros1
    ```
+1. Update the message definitions of the `ros1` branch
+   <!-- px4_msgs/ros1 is not synchronized with release/1.13 -->
+   ```sh
+   $ rm -f ~/px4_ros_com_ros1/src/px4_msgs/msg/*.msg
+   $ cp ~/px4_ros_com_ros2/src/px4_msgs/msg/*.msg ~/px4_ros_com_ros1/src/px4_msgs/msg/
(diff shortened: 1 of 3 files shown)


dagar:
I think we'd need to find a proper way to do this automatically. Do you want to limit this to just the ROS stuff and we could discuss the versioning approach with @hamishwillee separately?

beniaminopozzan:
> Do you want to limit this to just the ROS stuff?

Yes, of course, I've removed everything not ROS-related and adjusted the ROS pages to be point to the right versions.

hamishwillee:
@beniaminopozzan Thanks for this. Looks good. FYI for alignment of text in list it needs to be indented to the list item level.

I think we need to simplify this "for next time" by having a section in the build guide that covers getting a tag release. I'll add that now. So if you had to do this again we'd at least be able to link to a section on how to.

I'll tag you for review (and you can comment)

PS W.r.t. more generally keeping file paths up to date with a release this is not as easy as you might think. With gitbook legacy I could update links using a variable that I changed for the path, but vuepress does not support this.

I am (very slowly) looking at docsify for a possible alternative. That renders markdown on the fly, and would allow me to search for, and dynamically update the links based on the URL path. I think it will be possible, just hard for me to find time to do this as a "side job". FYI, my playpen here: https://hamishwillee.github.io/multidocsify_test/#/main/en/config/README

en/ros/ros1_via_ros2.md
```diff
@@ -72,6 +78,18 @@ As discussed in [ROS 2 User Guide > Sanity Check the Installation](../ros/ros2_c
 To use ROS (1) **and** ROS 2 (you need both for this!):
 
 1. [Setup your PX4 Ubuntu Linux development environment](../dev_setup/dev_env_linux_ubuntu.md) - the default instructions get the latest version of PX4 source and install all the needed tools.
+
+   :::warning Note
```
hamishwillee, line 82:
```suggestion
   :::note
```

en/ros/ros1_via_ros2.md
```diff
@@ -72,6 +78,18 @@ As discussed in [ROS 2 User Guide > Sanity Check the Installation](../ros/ros2_c
 To use ROS (1) **and** ROS 2 (you need both for this!):
 
 1. [Setup your PX4 Ubuntu Linux development environment](../dev_setup/dev_env_linux_ubuntu.md) - the default instructions get the latest version of PX4 source and install all the needed tools.
+
+   :::note
+   The above instructions must be adjusted to get version v1.13 of PX4 source.
+   
+   ```sh
+   git clone -b <tag> https://github.com/PX4/PX4-Autopilot.git --recursive
```
hamishwillee, line 86: @beniaminopozzan What you have as `tag` here is actually a branch - https://github.com/PX4/PX4-Autopilot/branches/all?query=release

Releases are tagged as well, but don't have the `release/` prefix: https://github.com/PX4/PX4-Autopilot/releases?q=release&expanded=true

What is it that we recommend people use?

hamishwillee: Note, I THINK that the release branch will reflect latest-and-greatest, while the tag comes of that and reflects a specific point build. So we should suggest the branch "first".

beniaminopozzan: I was a bit unsure about how to phrase that. My understanding is that the tags are the stable versions while the `release/` branches are where relevant commits are backported from `main` in preparation of an update of the releases, right?

So as this piece of doc is just for the release `1.13.*` I just listed all tags and branches which are compatible to it. If I had to choose, I'll say that `v1.13.2` it currently the recommended one.

beniaminopozzan:
> @beniaminopozzan Thanks for this. Looks good. FYI for alignment of text in list it needs to be indented to the list item level.

Thanks @hamishwillee  for the information, I'll be more careful next time. 

> I think we need to simplify this "for next time" by having a section in the build guide that covers getting a tag release. I'll add that now. So if you had to do this again we'd at least be able to link to a section on how to.

Yeah, a section about checking out specific releases is a good idea!

> PS W.r.t. more generally keeping file paths up to date with a release this is not as easy as you might think. With gitbook legacy I could update links using a variable that I changed for the path, but vuepress does not support this.

I'm sure of it, I was actually thinking that just search-replace all links and paths was quite naive.

> I am (very slowly) looking at docsify for a possible alternative. That renders markdown on the fly, and would allow me to search for, and dynamically update the links based on the URL path. I think it will be possible, just hard for me to find time to do this as a "side job". FYI, my playpen here: https://hamishwillee.github.io/multidocsify_test/#/main/en/config/README

You mean moving the whole documentation to a different platform?

hamishwillee:
> I'm sure of it, I was actually thinking that just search-replace all links and paths was quite naive.

It's fairly naïve, but not hopelessly so. That can help fix paths permanently to branch-specific locations, which is OK for links which can be tested, but not so much for text in code examples, where you might actually mean "mainline".
Upshot I probably should do more automation and updates of the old releases at release time, but its a maintenance cost, and there is only one of me.

> You mean moving the whole documentation to a different platform?

Yes. There are lots of reasons. No urgency though, which is why its a side project.

PR #2204, opened 2022-12-14
Apply equal code snippet format

GeneralMine:
Adapt every code section by removing unnecessary spaces and disturbing dollar signs in order to provide faster copy pasting of code snippets


hamishwillee (approved):
Thanks. Yes, this is better.

PR #2206, opened 2022-12-15
Update info on EKF tuning and baro usage

hamishwillee:
This falls out of discord discussion on how to disable baros and also how they are selected: https://discord.com/channels/1022170275984457759/1044214748050370630

PR #2207, opened 2022-12-15
Add barometers page in peripherals.

hamishwillee:
This adds a page on barometers to the Peripherals/hardware section. This is complementary to the #2206 update because it wasn't clear to me what you can and cannot do with barometers, or that people would find this information easily.

I hope this makes the following things clear
- Generally you don't need to think about barometers - they are in the FC and pre-calibrated
- What parts are supported (pointers to drivers)
- That calibration is not needed (right?)
- But if you do want to disable them, how you do it.

### DIFF
--- a/en/sensor/barometers.md
+++ b/en/sensor/barometers.md
@@ -1,0 +1,40 @@
+# Barometers
+
+Barometers measure atmospheric pressure, and are used in drones as altitude sensors.
+
+Most flight controller hardware on which PX4 incudes a barometer.
+By default PX4 will select the barometer with the highest priority (if any are present), and configure it as a data source for [Height estimation](../advanced_config/tuning_the_ecl_ekf.md#height).
+If a sensor fault is detected, PX4 will fall back to the next highest priority sensor.
+
+Generally barometers require no user configuration (or thought)!
+
+## Hardware Options
+
+[Pixhawk standard](../flight_controller/autopilot_pixhawk_standard.html) flight controllers include a barometer, as do [many others](../flight_controller/README.md).
+
+They are also present in other hardware:
+
+- [CUAV NEO 3 Pro GNSS module](https://doc.cuav.net/gps/neo-series-gnss/en/neo-3-pro.html#key-data) ([MS5611](../modules/modules_driver_baro.md#ms5611))
+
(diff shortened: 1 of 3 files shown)


en/sensor/barometers.md
```diff
@@ -0,0 +1,40 @@
...
+Generally barometers require no user configuration.
+If needed, you can:
+
+- Enable/Disable barometers as data source for [Height estimation](../advanced_config/tuning_the_ecl_ekf.md#height) using the [EKF2_BARO_CTRL](../advanced_config/parameter_reference.md#EKF2_BARO_CTRL) parameter.
+- Change the selection order of barometers using the [CAL_BAROx_PRIO](../advanced_config/parameter_reference.md#CAL_BARO0_PRIO) parameters for each baro.
+- Disable a baro by setting its [CAL_BAROx_PRIO](../advanced_config/parameter_reference.md#CAL_BARO0_PRIO) value to `0`.
+
+## Calibration
+
+Barometer parts are generally pre-calibrated.
```
hamishwillee, line 34: This is correct right? Anything else people need to know?

bresch: I'm not sure we can say that they are pre-calibrated, but:
- The absolute value isn't really important since we just use the difference in altitude between "now" and the value when initializing EKF2
- There is usually a scale factor error but it's compensated by the GNSS altitude using a bias estimator in EKF2 (we don't provide a way to calibrate that). This method is fine as long as the height change of the drone isn't too fast (below 200-300km/h probably; don't have real data on that).
- The baro readings can be corrected using a [QNH](https://docs.px4.io/main/en/advanced_config/parameter_reference.html#SENS_BARO_QNH) (https://en.wikipedia.org/wiki/Altimeter_setting) parameter, but again, it is only necessary to adjust it if the absolute barometric altitude is required by the pilot.

So I would rather say that "they don't require calibration".

hamishwillee: Thanks !
```suggestion
Barometers don't require calibration.

<!-- Notes:
- Absolute value isn't important since we just use the difference in altitude between "now" and the value when initializing EKF2
- There is usually a scale factor error but it's compensated by the GNSS altitude using a bias estimator in EKF2 (we don't provide a way to calibrate that). This method is fine as long as the height change of the drone isn't too fast (below 200-300km/h probably; don't have real data on that).
- The baro readings can be corrected using a [QNH](https://docs.px4.io/main/en/advanced_config/parameter_reference.html#SENS_BARO_QNH) (https://en.wikipedia.org/wiki/Altimeter_setting) parameter, but again, it is only necessary to adjust it if the absolute barometric altitude is required by the pilot.
-->
```

hamishwillee:
Thanks @bresch - updated as you suggested.

PR #2209, opened 2022-12-15
first fixes for battery

hamishwillee:
This is WIP
- Need to work out how the battery source works - then perhaps split out the battery chemistry section.
- Specifically, power modules, mavlink batteries, uavcan batteries, and whatever else.


hamishwillee:
This work was done in a separate PR.

PR #2215, opened 2022-12-21
Clarify difference between magnetometer fusion modes

hamishwillee:
This attempts to better clarify when/how you should use the different EKF magnetometer settings, as discussed here: https://discord.com/channels/1022170275984457759/1052231376008663070/1053665360945098885

IN particular capture "Setting '3-axis' mode all the time (`EKF2_MAG_TYPE=2`) is more error-prone, and requires that all the IMUs are well calibrated."

### DIFF
--- a/en/advanced_config/tuning_the_ecl_ekf.md
+++ b/en/advanced_config/tuning_the_ecl_ekf.md
@@ -136,8 +136,9 @@ Magnetometer data can be used in two ways:
   - This method is less accurate and does not allow for learning of body frame field offsets, however it is more robust to magnetic anomalies and large start-up gyro biases.
   - It is the default method used during start-up and on ground.
 - The XYZ magnetometer readings are used as separate observations.
-  - This method is more accurate when the vehicle is moving, as it allows body frame offsets to be learned (when the vehicle is moving the heading or the magnetometer biases are observable).
-    The method is less robust, and the accuracy benefits are not available when the vehicle is not moving.
+  - This method is more accurate but requires that the magnetometer biases are correctly estimated.
+    - The biases are observable while the drone is rotating and the true heading is observable when the vehicle is accelerating (linear acceleration).
+    - Since the biases can change and are only observable when moving, it is safer to switch back to heading fusion when not moving.
   - It assumes the earth magnetic field environment only changes slowly and performs less well when there are significant external magnetic anomalies.
   - This is the default method used when the vehicle is moving.
 
--- a/en/advanced_config/tuning_the_ecl_ekf.md
+++ b/en/advanced_config/tuning_the_ecl_ekf.md
@@ -128,16 +128,22 @@ This minimum data set is required for all EKF modes of operation. Other sensor d
 ### Magnetometer
 
 Three axis body fixed magnetometer data (or external vision system pose data) at a minimum rate of 5Hz is required.
+
 Magnetometer data can be used in two ways:
 
-* Magnetometer measurements are converted to a yaw angle using the tilt estimate and magnetic declination.
-  This yaw angle is then used as an observation by the EKF.
-  This method is less accurate and does not allow for learning of body frame field offsets, however it is more robust to magnetic anomalies and large start-up gyro biases.
-  It is the default method used during start-up and on ground.
-* The XYZ magnetometer readings are used as separate observations.
-  This method is more accurate and allows body frame offsets to be learned, but assumes the earth magnetic field environment only changes slowly and performs less well when there are significant external magnetic anomalies.
+- Magnetometer measurements are converted to a yaw angle using the tilt estimate and magnetic declination.
+  The yaw angle is then used as an observation by the EKF.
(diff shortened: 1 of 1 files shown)


en/advanced_config/tuning_the_ecl_ekf.md
```diff
@@ -128,16 +128,22 @@ This minimum data set is required for all EKF modes of operation. Other sensor d
...
-  This method is less accurate and does not allow for learning of body frame field offsets, however it is more robust to magnetic anomalies and large start-up gyro biases.
-  It is the default method used during start-up and on ground.
-* The XYZ magnetometer readings are used as separate observations.
-  This method is more accurate and allows body frame offsets to be learned, but assumes the earth magnetic field environment only changes slowly and performs less well when there are significant external magnetic anomalies.
+- Magnetometer measurements are converted to a yaw angle using the tilt estimate and magnetic declination.
+  The yaw angle is then used as an observation by the EKF.
+  - This method is less accurate and does not allow for learning of body frame field offsets, however it is more robust to magnetic anomalies and large start-up gyro biases.
+  - It is the default method used during start-up and on ground.
+- The XYZ magnetometer readings are used as separate observations.
+  - This method is more accurate when the vehicle is moving, as it allows body frame offsets to be learned (when the vehicle is moving the heading or the magnetometer biases are observable).
```
hamishwillee, line 139: This is correct?

en/advanced_config/tuning_the_ecl_ekf.md
```diff
@@ -128,16 +128,22 @@ This minimum data set is required for all EKF modes of operation. Other sensor d
...
+    The method is less robust, and the accuracy benefits are not available when the vehicle is not moving.
+  - It assumes the earth magnetic field environment only changes slowly and performs less well when there are significant external magnetic anomalies.
+  - This is the default method used when the vehicle is moving.
```
hamishwillee, line 142: Is the trigger to use this "when moving" or "when in-air"?

After you have moved around a bit you have learned the static biases so does that mean that you could just as reliably use the 3-axis approach all the time? Or is it that the start up biases will screw everything up?

en/advanced_config/tuning_the_ecl_ekf.md
bresch: Not exactly. The method is more accurate (no need to move to be more accurate) but it requires to have the mag bias correctly estimated. And to estimate the biases, you need to move. Once they are estimated and if they don't change anymore, the 3D fusion is still more accurate even while hovering. However, since we cannot guarantee that the bias are not changing and that the bias are no longer observable while not moving, it is safer to switch back to heading fusion when the motion stops.

en/advanced_config/tuning_the_ecl_ekf.md
bresch: > Is the trigger to use this "when moving" or "when in-air"?

When moving. The mag biases are observable while the drone is rotating and the true heading is observable when the vehicle is accelerating (linear acceleration).

> After you have moved around a bit you have learned the static biases so does that mean that you could just as reliably use the 3-axis approach all the time? Or is it that the start up biases will screw everything up?

Yes, you could, as long as they are not changing anymore. But it is safer to fallback to heading fusion.

bresch (changes requested):

en/advanced_config/tuning_the_ecl_ekf.md
hamishwillee: THanks! OK. Then I think this bit is OK.

en/advanced_config/tuning_the_ecl_ekf.md
```diff
@@ -128,16 +128,22 @@ This minimum data set is required for all EKF modes of operation. Other sensor d
...
-  It is the default method used during start-up and on ground.
-* The XYZ magnetometer readings are used as separate observations.
-  This method is more accurate and allows body frame offsets to be learned, but assumes the earth magnetic field environment only changes slowly and performs less well when there are significant external magnetic anomalies.
+- Magnetometer measurements are converted to a yaw angle using the tilt estimate and magnetic declination.
+  The yaw angle is then used as an observation by the EKF.
+  - This method is less accurate and does not allow for learning of body frame field offsets, however it is more robust to magnetic anomalies and large start-up gyro biases.
+  - It is the default method used during start-up and on ground.
+- The XYZ magnetometer readings are used as separate observations.
+  - This method is more accurate when the vehicle is moving, as it allows body frame offsets to be learned (when the vehicle is moving the heading or the magnetometer biases are observable).
+    The method is less robust, and the accuracy benefits are not available when the vehicle is not moving.
```
hamishwillee, line 140: Thanks @bresch . I'll steal all your words and make this more precise then. 
I'm going to merge, but would appreciate a final check when you have a moment.

```suggestion
  - This method is more accurate but requires that the magnetometer biases are correctly estimated.
    - The biases are observable while the drone is rotating and the true heading is observable when the vehicle is accelerating (linear acceleration).
    - Since the biases can change and are only observable when moving, it is safer to switch back to heading fusion when not moving.
```

PR #2217, opened 2022-12-21
Add ARKV6X

bradenwagstaff:


hamishwillee:
The content looks good - I did a tidy just to fix the vuepress rendering. Images are very large for the "information content" - can you shrink them? I'm away on holiday until Jan 2ish.

AlexKlimaj:
Need to add pinout charts to the ARK PAB page.

en/flight_controller/arkpab.md
```diff
@@ -0,0 +1,70 @@
...
+  - 5V
+  - TBD
+- Dimensions (Without Flight Controller Module)
+  - 74mm x 43.5mm x 12mm
+
+![ARKPAB Top Down Photo](../../assets/flight_controller/arkpab/arkpab_top.jpg)
+
+![ARKPAB Bottom Photo](../../assets/flight_controller/arkpab/arkpab_back2.jpg)
+
+![ARK Logo](../../assets/flight_controller/arkpab/ark_logo.png)
```
hamishwillee, line 70: We don't normally allow heavy logo-ing like this. Mostly because it adds nothing to the reader. If you feel you need it we can leave it here, as it does not push down any actual content. I would prefer we delete though.

en/flight_controller/pixhawk_autopilot_bus.md
```diff
@@ -0,0 +1,25 @@
+# Pixhawk Autopilot Bus & Carriers
```
hamishwillee, line 1: @bkueng I was wondering if you could have a quick look at this page? (below). This adds a page for PAB explaining what it is and offers. The main thing being that it helps users understand that they can plug and play FCs and boards between vendors, allowing flexibility for integrating with companions, but also to choose cheaper or more compact carrier designs. 

It also serves as a place to put compliant carrier boards - these shouldn't really be below a particular FC.

The question for you is whether this should live in the hierarchy 
a) where I have it (in FC section, after the FCs)
b) Higher up in FC section - specifically before FC listings, and after "Pixhawk Series" 
c) Inside FC peripherals
d) As a peer of Flight Controllers and FC peripherals.

Wasn't entirely sure of the right level of visibility.

hamishwillee:
@bradenwagstaff @AlexKlimaj I have subedited this. Specifically:
1. Made the headings more consistent with similar pages
2. Cropped images and shrunk them
3. Made all images lower case names - this enables safer linking on linux builds which are case sensitive.
4. Deleted images that are not used in your pages
5. Added to SUMMARY.md (needed if you want this in the sidebar).
6. Created a PAB page and put the ARK PAB under it. Right now the PAB is still shown in other pages as associated with the ARKv6x controller, but I am thinking perhaps reducing this link and making a link to "PAB" be a key feature of anything that is compliant - e.g. in the "Manufacturer Supported FCs" page, instead of having 

   > - [ARK Electronics ARKV6X](../flight_controller/arkv6x.md) (and [ARK Electronics Pixhawk Autopilot Bus Carrier](../flight_controller/arkpab.md))

   have
   
   > - [ARK Electronics ARKV6X](../flight_controller/arkv6x.md) ([PAB compliant](../flight_controller/whateverthepabdociscalled.md))

I will wait and see if there are comments by Tuesday/Weds and merge then if not. (yes, aware this is waiting on some pinouts - those could be added as a post process).

AlexKlimaj:
Added the pinouts of the connectors on the ARK PAB and made a few other minor edits.

en/flight_controller/arkpab.md
AlexKlimaj: I removed them.

hamishwillee:
FYI Removed the logos from the source, and did minor layout/cross linking fixes.

This looks great. Thanks.

en/flight_controller/pixhawk_autopilot_bus.md
bkueng: Not quite sure, probably most users will just use the bundled package. Keeping it here makes sense, maybe also one level up (d)).
Longer-term I think it makes sense to have a separate section for FC+mission computer bundled boards as these add more capabilities.

hamishwillee: @bkueng Thanks very much. That's pretty much my take on it.

> Longer-term I think it makes sense to have a separate section for FC+mission computer bundled boards as these add more capabilities.

I agree, though this is a challenge. For now I just have a section under companion computers: http://docs.px4.io/main/en/companion_computer/#integrated-companion-flight-controller-boards

PR #2218, opened 2022-12-22
List Safe Landing world and depth camera vehicles

DanMesh:
1. Add vehicles with depth cameras to the simulation vehicles list.
   - https://github.com/PX4/PX4-SITL_gazebo/pull/933
2. Add safe landing world to the simualtion worlds list.
   - https://github.com/PX4/PX4-SITL_gazebo/pull/932
3. Update references to the `sitl_gazebo` branch `master` to branch `main` in the sim worlds list.


en/simulation/gazebo_vehicles.md
```diff
@@ -40,7 +57,7 @@ make px4_sitl gazebo_solo
 <a id="typhoon_h480"></a>
 ### Typhoon H480 (Hexrotor)
 
-```
+```sh
```
hamishwillee, line 60: FYI the "sh" is entirely optional. All good though.

en/simulation/gazebo_vehicles.md
```diff
@@ -27,6 +27,23 @@ make px4_sitl gazebo
 make px4_sitl gazebo_iris_opt_flow
 ```
 
+<a id="quadrotor_depth_camera"></a>
```
hamishwillee, line 30: FYI only you only need an explicit anchor if you are linking to the section from somewhere AND you expect the title is likely to change at some point. I tend to do this for longer headings that get linked a lot of times. But it is better to use the auto-generated heading anchor if you can because when you press the link navigation works better.

```suggestion
```

en/simulation/gazebo_vehicles.md
```diff
@@ -27,6 +27,22 @@ make px4_sitl gazebo
 make px4_sitl gazebo_iris_opt_flow
 ```
 
+### Quadrotor with Depth Camera
+
+These models have a depth camera attached, modelled on the Intel® RealSense™ D455.
```
hamishwillee, line 32: Is it worth including some of the detail you put in the autopilot PR? Specifically the frame size, and FOV? I'm not saying it is - I am "wondering" if this would make any difference to consumers/users of the model.

hamishwillee:
@DanMesh This looks excellent. If you think it useful you might (as a post process) consider adding a section on "using depth cameras" (or whatever you think might be most useful for people wanting to discover this) in the parent doc: http://docs.px4.io/main/en/simulation/gazebo.html

So for example, you can see that in that topic there are sections on using video, and changing simulation speed, etc. I presume a depth camera is useful for providing data to some kind of obstacle avoidance system - so it might be nice to explain how to do that in the top level doc. I could be overthinking this, but I assume you would not have added this feature if it was not useful.

PR #2220, opened 2022-12-23
Add Holybro Dronecan M8N GPS

vincentpoont2:


hamishwillee:
I couldn't fix this git history - this was failing all tests due to those odd ids. Replaced the PR under my own name in https://github.com/PX4/PX4-user_guide/pull/2229

PR #2226, opened 2022-12-29
add a mission feasibility check section

sfuhrer:
Mainly driven by the takeoff/landing requirement changes that came in with https://github.com/PX4/PX4-Autopilot/pull/20588

### DIFF
--- a/en/config/safety.md
+++ b/en/config/safety.md
@@ -230,7 +230,7 @@ Parameter | Description
 A number of checks are run to ensure that a mission can only be started if it is _feasible_.
 For example, the checks ensures that the first waypoint isn't too far away, and that the mission flight path doesn't conflict with any geofences.
 
-As these are not strictly speaking "failsafes" they are documented in [Flying > Mission> Mission Feasibility Checks](../flying/missions.html#mission-feasibility-checks).
+As these are not strictly speaking "failsafes" they are documented in [Mission Mode > Mission Feasibility Checks](../flight_modes/mission.md#mission-feasibility-checks).
 
 ### Traffic Avoidance Failsafe
 
--- a/en/config/safety.md
+++ b/en/config/safety.md
@@ -225,6 +225,12 @@ Parameter | Description
 [COM_OBL_ACT](../advanced_config/parameter_reference.md#COM_OBL_ACT) | Failsafe action if no RC is available: Land mode, Hold mode, Return mode.
 [COM_OBL_RC_ACT](../advanced_config/parameter_reference.md#COM_OBL_RC_ACT) | Failsafe action if RC is available: Position mode, Altitude mode, Manual mode, Return mode, Land mode, Hold mode.
 
+### Mission Feasibility Checks
+
+A number of checks are run to ensure that a mission can only be started if it is _feasible_.
+For example, the checks ensures that the first waypoint isn't too far away, and that the mission flight path doesn't conflict with any geofences.
+
+As these are not strictly speaking "failsafes" they are documented in [Flying > Mission> Mission Feasibility Checks](../flying/missions.html#mission-feasibility-checks).
 
 ### Traffic Avoidance Failsafe
 
--- a/en/config/safety.md
+++ b/en/config/safety.md
@@ -226,19 +226,6 @@ Parameter | Description
 [COM_OBL_RC_ACT](../advanced_config/parameter_reference.md#COM_OBL_RC_ACT) | Failsafe action if RC is available: Position mode, Altitude mode, Manual mode, Return mode, Land mode, Hold mode.
 
 
-### Mission Failsafe
-
-The Mission Failsafe checks prevent a previous mission being started at a new takeoff location or if it is too big (distance between waypoints is too great).
-The failsafe action is that the mission will not be run.
-
-The relevant parameters are shown below:
-
-Parameter | Description
---- | ---
-[MIS_DIST_1WP](../advanced_config/parameter_reference.md#MIS_DIST_1WP) | The mission will not be started if the current waypoint is more distant than this value from the home position. Disabled if value is 0 or less.
-[MIS_DIST_WPS](../advanced_config/parameter_reference.md#MIS_DIST_WPS) | The mission will not be started if any distance between two subsequent waypoints is greater than this value.
-
-
 ### Traffic Avoidance Failsafe
 
 The Traffic Avoidance Failsafe allows PX4 to respond to transponder data (e.g. from [ADSB transponders](../advanced_features/traffic_avoidance_adsb.md)) during missions.
--- a/en/flight_modes/mission.md
+++ b/en/flight_modes/mission.md
@@ -6,11 +6,11 @@
 The mission is typically created and uploaded with a Ground Control Station (GCS) application like [QGroundControl](https://docs.qgroundcontrol.com/master/en/) (QGC).
 
 :::note
-* This mode requires 3d position information (e.g. GPS).
-* The vehicle must be armed before this mode can be engaged.
-* This mode is automatic - no user intervention is *required* to control the vehicle.
-* RC control switches can be used to change flight modes on any vehicle.
-* RC stick movement in a multicopter (or VTOL in multicopter mode) will [by default](#COM_RC_OVERRIDE) change the vehicle to [Position mode](../flight_modes/position_mc.md) unless handling a critical battery failsafe.
+- This mode requires 3d position information (e.g. GPS).
+- The vehicle must be armed before this mode can be engaged.
+- This mode is automatic - no user intervention is *required* to control the vehicle.
+- RC control switches can be used to change flight modes on any vehicle.
+- RC stick movement in a multicopter (or VTOL in multicopter mode) will [by default](#COM_RC_OVERRIDE) change the vehicle to [Position mode](../flight_modes/position_mc.md) unless handling a critical battery failsafe.
 :::
 
 ## Description
@@ -28,13 +28,15 @@ Missions are uploaded onto a SD card that needs to be inserted **before** bootin
 At high level all vehicle types behave in the same way when MISSION mode is engaged:
 
 1. If no mission is stored, or if PX4 has finished executing all mission commands, or if the [mission is not feasible](#mission-feasibility-checks):
-   * If flying the vehicle will loiter.
-   * If landed the vehicle will "wait".
+
+   - If flying the vehicle will loiter.
+   - If landed the vehicle will "wait".
 1. If a mission is stored and PX4 is flying it will execute the [mission/flight plan](../flying/missions.md) from the current step.
+   - On copters PX4 will treat a takeoff item as a normal waypoint if already flying.
 1. If a mission is stored and PX4 is landed:
-   * On copters PX4 will execute the [mission/flight plan](../flying/missions.md).
+   - On copters PX4 will execute the [mission/flight plan](../flying/missions.md).
      If the mission does not have a `TAKEOFF` command then PX4 will fly the vehicle to the minimum altitude before executing the remainder of the flight plan from the current step.
-   * On fixed-wing vehicles PX4 will not automatically take off (the autopilot will detect the lack of movement and set the throttle to zero).
+   - On fixed-wing vehicles PX4 will not automatically take off (the autopilot will detect the lack of movement and set the throttle to zero).
      The vehicle may start executing the mission if hand- or catapult- launched while in mission mode.
 1. You can manually change the current mission command by selecting it in *QGroundControl*.
 
@@ -59,12 +61,12 @@ We recommend you centre the control sticks before switching to any other mode.
 :::
  
 For more information about mission planning, see:
-* [Mission Planning](../flying/missions.md)
-* [Plan View](https://docs.qgroundcontrol.com/master/en/PlanView/PlanView.html) (*QGroundControl* User Guide)
+- [Mission Planning](../flying/missions.md)
+- [Plan View](https://docs.qgroundcontrol.com/master/en/PlanView/PlanView.html) (*QGroundControl* User Guide)
 
 ## Mission Feasibility Checks
 
-PX4 runs some basic sanity checks to determine if a mission is feasible when it is uploaded, and whenever there is an attempt to run the mission.
+PX4 runs some basic sanity checks to determine if a mission is feasible when it is uploaded, and when the vehicle is first armed.
 If any of the checks fail, the user is notified and it is not possible to start the mission.
 
 A subset of the most important checks are listed below:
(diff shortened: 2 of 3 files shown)


en/flying/missions.md
```diff
@@ -18,6 +18,28 @@ For more information see the [QGroundControl User Guide](https://docs.qgroundcon
 
 ![planning-mission](../../assets/flying/planning_mission.jpg)
 
+### Mission Feasibility Checks
+
+Upon receiving an updated mission file from the ground station, the flight controller runs some feasibility checks that are meant to ensure the safety of the vehicle and catch unintended mission settings. If one of the checks fails, the user is notified and it is not possible to start the mission.
```
hamishwillee, line 23: My understanding is that checks are run on more than just upload. I.e. I could upload a mission and the vehicle fails the checks, I could then drive to the place where the mission is supposed to be used and it passes the checks.

Point being that the checks must be run again at some point - I am assuming they are run every time you switch to mission mode? We should make it clear.

en/config/safety.md
```diff
@@ -225,19 +225,12 @@ Parameter | Description
 [COM_OBL_ACT](../advanced_config/parameter_reference.md#COM_OBL_ACT) | Failsafe action if no RC is available: Land mode, Hold mode, Return mode.
 [COM_OBL_RC_ACT](../advanced_config/parameter_reference.md#COM_OBL_RC_ACT) | Failsafe action if RC is available: Position mode, Altitude mode, Manual mode, Return mode, Land mode, Hold mode.
 
+### Mission Feasibility Checks
```
hamishwillee, line 228: While not strictly a "safety section" it is quite closely related, and I think this is a reasonable place to cross link the information. Reasonable?

en/flying/missions.md
```diff
@@ -18,6 +18,28 @@ For more information see the [QGroundControl User Guide](https://docs.qgroundcon
 
 ![planning-mission](../../assets/flying/planning_mission.jpg)
 
+### Mission Feasibility Checks
```
hamishwillee, line 21: I'm not sure if this section is better here or in http://docs.px4.io/main/en/flight_modes/mission.html. 

Generally we use this mission planning page for planning - it has always been a bit "thin" because we assume that QGC tells you most of what you need to know - i.e. what commands you can use. It is only recently I have expanded it a bit. We use the flight mode page (http://docs.px4.io/main/en/flight_modes/mission.html) for more detailed information that you might use both for manually planning missions and to understand the mechanics / debugging / companion missions etc.

I think that where you have put it is the right place since you probably should know this if you're planning a mission. But then probably you also need it cross linked from the mode page?

Note, I need to put some hard thought into this whole page. Never been a priority because most of the time you can pick up what you need to know in QGC quite easily.

en/flying/missions.md
```diff
@@ -18,6 +18,28 @@ For more information see the [QGroundControl User Guide](https://docs.qgroundcon
...
+Upon receiving an updated mission file from the ground station, the flight controller runs some feasibility checks that are meant to ensure the safety of the vehicle and catch unintended mission settings. If one of the checks fails, the user is notified and it is not possible to start the mission.
+
+The checks involve (selection of most likely ones):
+- First mission item too far away
+- Distance between two subsequent items is too large
+- Any mission item outside of defined geofence
+- More than one land start defined
+- Fixed-wing landing infeasible slope angle
+- Land start item before RTL item
+- Missing takeoff and/or land item (see parameter list below for configuration)
```
hamishwillee, line 32:
```suggestion
A subset of the most important checks are listed below:
- First mission item too far away from vehicle
- Distance between two subsequent items is too large
- Any mission item conflicts with a plan or safety geofence
- More than one land start mission item defined ([MAV_CMD_DO_LAND_START](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_LAND_START))
- A fixed-wing landing has an infeasible slope angle
- Land start item (`MAV_CMD_DO_LAND_START`) appears in mission before an RTL item ([MAV_CMD_NAV_RETURN_TO_LAUNCH](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_RETURN_TO_LAUNCH))
- Missing takeoff and/or land item when these are configured as a requirement (see parameter list below for configuration)

```

hamishwillee:
@sfuhrer Looks good. Minor subedit and also a few comments added.

hamishwillee:
PS, And now I've read the associated PR I see the trigger was allowing control over mission behaviour with RTL=1 

I think what this is saying is that a fixed wing aircraft currently requires a landing pattern if RTL=1 is set - right? But this should not be a requirement because if you read http://docs.px4.io/main/en/flight_modes/return.html#mission-landing-rally-point-return-type-rtl-type-1 it might not even use the mission landing. 
So the idea is that you allow takeoff and landing requirements on missions to be separately configured from the RTL mode - is that right?

In that case, it might be worth adding a note in http://docs.px4.io/main/en/flight_modes/return.html#mission-landing-rally-point-return-type-rtl-type-1 linking to mission feasibility docs in general, and MIS_TKO_LAND_REQ  in particular. More generally though this is what made me think that maybe the general feasibility info should be in mission mode doc, but we need to guide users in the planning section to what is "recommended" or "how to debug the message that tells them they need a landing pattern". Also explain when and why that landing pattern is needed :-)

Last of all, not completely clear on when each of these should be set and why?

- 0: No requirements
- 1: Require a takeoff (but not a landing?) (when would you set for FW? When would you set for MC? - ditto for others)
- 2: Require a landing (but not a takeoff?)
- 3: Require a takeoff and a landing.
- 4: Require a takeoff and a landing, or neither of both - Does this mean "Require both a takeoff and a landing, or neither"? As above, what is the use case where this makes sense to set?

en/config/safety.md
sfuhrer: Yes perfectly reasonable :+1:

en/flying/missions.md
sfuhrer: True, the checks are also run once after boot as soon as the home position is valid. Switching to Mission doesn't seem to be necessary.

I'm though not even sure if a rejected mission is stored on the vehicle, so if your example of the checks failing at location A, then passing automatically at location B would work. 

@Igor-Misic  weren't you involved in the building the logic around there? Would you know if this example is possible like that?

en/flying/missions.md
sfuhrer: >My leaning is to put this in http://docs.px4.io/main/en/flight_modes/mission.html and have a small section in this page.

Ah yes, that makes sense. The checks are clearly a "PX4 thing" and thus should live in the flight mode section. And here only reference that one, and mention that there are checks being run at mission upload/vehicle boot that need to pass to fly the mission. But really keep the flying/mission page to the bare minimum. 

The link from the flying/mission page to flight_modes/mission is missing atm.

>what are we hoping a user to know that they have to do when mission planning with this information?

I think it is important for users to know that there is some mission checking going on, and give them roughly an idea what the checks involve. Plus what to change if they have failures.

sfuhrer:
> PS, And now I've read the associated PR I see the trigger was allowing control over mission behaviour with RTL=1
> 
> I think what this is saying is that a fixed wing aircraft currently requires a landing pattern if RTL=1 is set - right? But this should not be a requirement because if you read http://docs.px4.io/main/en/flight_modes/return.html#mission-landing-rally-point-return-type-rtl-type-1 it might not even use the mission landing. So the idea is that you allow takeoff and landing requirements on missions to be separately configured from the RTL mode - is that right?
> 
> In that case, it might be worth adding a note in http://docs.px4.io/main/en/flight_modes/return.html#mission-landing-rally-point-return-type-rtl-type-1 linking to mission feasibility docs in general, and MIS_TKO_LAND_REQ in particular. More generally though this is what made me think that maybe the general feasibility info should be in mission mode doc, but we need to guide users in the planning section to what is "recommended" or "how to debug the message that tells them they need a landing pattern". Also explain when and why that landing pattern is needed :-)
> 
> Last of all, not completely clear on when each of these should be set and why?
> 
>     * 0: No requirements
> 
>     * 1: Require a takeoff (but not a landing?) (when would you set for FW? When would you set for MC? - ditto for others)
> 
>     * 2: Require a landing (but not a takeoff?)
> 
>     * 3: Require a takeoff and a landing.
> 
>     * 4: Require a takeoff and a landing, or neither of both - Does this mean "Require both a takeoff and a landing, or neither"? As above, what is the use case where this makes sense to set?

The dependency with the RTL_TYPE is a whole other story..so yes,

>So the idea is that you allow takeoff and landing requirements on missions to be separately configured from the RTL mode - is that right?

that sounds right.

en/flight_modes/mission.md
```diff
@@ -58,26 +58,56 @@ For more information about mission planning, see:
 * [Mission Planning](../flying/missions.md)
 * [Plan View](https://docs.qgroundcontrol.com/master/en/PlanView/PlanView.html) (*QGroundControl* User Guide)
 
+## Mission Feasibility Checks
```
hamishwillee, line 61: @sfuhrer FYI, I moved the more detailed docs into mission mode as discussed.

I guess we wait on @Igor-Misic before we merge? Also , I assume that an unfeasible mission on arming will not be deleted?

en/flight_modes/mission.md
```diff
@@ -58,26 +58,56 @@ For more information about mission planning, see:
...
 * [Plan View](https://docs.qgroundcontrol.com/master/en/PlanView/PlanView.html) (*QGroundControl* User Guide)
 
+## Mission Feasibility Checks
+
+PX4 runs some basic sanity checks to determine if a mission is feasible when it is uploaded, and when the vehicle is first armed.
+If any of the checks fail, the user is notified and it is not possible to start the mission (on upload the mission may not be saved).
+
+A subset of the most important checks are listed below:
+
+- First mission item too far away from vehicle ([MIS_DIST_1WP](#MIS_DIST_1WP))
```
hamishwillee, line 68: FYI only, I moved the parameters into a separate section below alongside the other relevant parameters, they are just linked here.

en/flying/missions.md
hamishwillee: @sfuhrer I reduced the section to be "high level" and linked.

hamishwillee:
> > PS, And now I've read the associated PR I see the trigger was allowing control over mission behaviour with RTL=1
> 
> The dependency with the RTL_TYPE is a whole other story..so yes,

OK, so I think RTL behaviour is indeed a mess so yes, let's wait on that before cross linking.
But can you please ping me when this is done so we can properly document the new behaviour?

> > 4: Require a takeoff and a landing, or neither of both - Does this mean "Require both a takeoff and a landing, or neither"? As above, what is the use case where this makes sense to set?
> 
> Yes, that's better english:)

Fixed in https://github.com/PX4/PX4-Autopilot/pull/20849

> It's basically to distinguish "Auto" flights (that have an automatic takeoff) from flights that are at least partially manual - in the latter the user is expected to always land the vehicle manually, which you can't ask for if the flight is started with an auto takeoff (assuming that only mission auto takeoffs are supported but not takeoff commands - it's the case currently for FW and partially for VTOL).

I think you might be saying that if I always have my vehicle under full auto control then I would set value to 3 to require a takeoff and landing.
However if I sometimes land manually then I might set 4 and I would then be able to choose to have missions without takeoff and landing or with takeoff and landing - up to me? Is that right? So its to make it easier for users who sometimes manually land.

en/flying/missions.md
```diff
@@ -18,6 +18,17 @@ For more information see the [QGroundControl User Guide](https://docs.qgroundcon
 
 ![planning-mission](../../assets/flying/planning_mission.jpg)
 
+### Mission Feasibility Checks
+
+PX4 runs some basic sanity checks to determine if a mission is feasible.
+For example, whether the mission is close enough to the vehicle, if the mission will conflict with a geofence, or if a mission landing pattern is required but is not present.
+
+The checks are run when it is uploaded and when it is first armed.
+If any of the checks fail, the user is notified and it is not possible to start the mission (on upload the mission may not even be saved).
```
hamishwillee, line 27:
```suggestion
The checks are run when the mission is uploaded and immediately before it is run.
If any of the checks fail, the user is notified and it is not possible to start the mission.

```

en/flying/missions.md
hamishwillee: @sfuhrer @Igor-Misic THanks very much. I have fixed it up to say this (I think) in the commit https://github.com/PX4/PX4-user_guide/pull/2226/commits/a3376e34612faff32f1a024c8265de55ebe8108a

There is a bit of a restructure, but the main effect is that it now says this:

> The checks are run when the mission is uploaded and immediately before it is run.
> If any of the checks fail, the user is notified and it is not possible to start the mission.

It also says that if the mode is switched to but not allowed to start due to being infeasible, then this is the same as if there is no mission when you switch to the mode - the vehicle holds/loiters.

en/flight_modes/mission.md
```diff
@@ -27,22 +27,26 @@ Missions are uploaded onto a SD card that needs to be inserted **before** bootin
...
 
-1. If a mission is stored and PX4 is flying it will execute the [mission/flight plan](../flying/missions.md) from the current step. 
-1. If a mission is stored and PX4 is landed:
-   * On copters PX4 will execute the [mission/flight plan](../flying/missions.md). If the mission does not have a `TAKEOFF` command then PX4 will fly the vehicle to the minimum altitude before executing the remainder of the flight plan from the current step.
-   * On fixed-wing vehicles PX4 will not automatically take off (the autopilot will detect the lack of movement and set the throttle to zero). The vehicle may start executing the mission if hand- or catapult- launched while in mission mode.  
-1. If no mission is stored, or if PX4 has finished executing all mission commands:
+1. If no mission is stored, or if PX4 has finished executing all mission commands, or if the [mission is not feasible](#mission-feasibility-checks):
    * If flying the vehicle will loiter.
    * If landed the vehicle will "wait".
+1. If a mission is stored and PX4 is flying it will execute the [mission/flight plan](../flying/missions.md) from the current step.
```
hamishwillee, line 33: @sfuhrer @Igor-Misic What if the current step is 0 and the first item is takeoff and you're already flying? On MC the vehicle will just skip takeoff and start at the next item. On FW I assume the same thing?

Otherwise I assume it just starts from the first item after takeoff if you are already flying?

hamishwillee:
I think we're nearly there - just need confirmation my update was OK and hopefully also an answer to https://github.com/PX4/PX4-user_guide/pull/2226/files#r1061174104

en/flight_modes/mission.md
sfuhrer: I don't think it will skip the Takeoff WP for MC, but consider it as a normal WP. And it currently does the same for FW. 
I will propose to change it on the PX4 side, as I would also expect it to not go to the Takeoff if already in air.

sfuhrer: It actually behaves differently for FW and MC flight: in MC it treats the Takeoff WP as a normal WP, in FW it only has to be above the Takeoff altitude to consider it reached. No special logic is in place for already taken off or not, so it will do the same whether it already is in air when the mission is started or not.

en/flying/missions.md
hamishwillee: Thanks @sfuhrer - I will revert that then. Probably in most cases that will be sufficient - i.e. you tend to boot when you're ready to fly, so that is when making the check makes most sense.

en/flight_modes/mission.md
hamishwillee: Hmmm. Don't know what that actually means for FW. I assume it means it will fall out of the air for a vehicle configured to catapult launch since it won't have acceleration required to ramp up the throttle. I guess we could test it. 

I'll just add the not for MC

en/flying/missions.md
hamishwillee: Actually, I see you already reverted.

hamishwillee:
@sfuhrer @Igor-Misic I think this is more than "good enough". Merging. Thank you.

PR #2231, opened 2022-12-30
Add imu_factory_calibration docs

nicovanduijn:
I noticed the IMU factory calibration is a little known, mostly undocumented PX4 feature. I've been met with a lot of confusion (including myself) about how it works, so I decided to document it.

Given my own confusion on how it actually works, I would insist on a fact-check by someone more knowledgeable prior to merging this. Great candidates would be one of @cmic0 @bkueng or @bresch


hamishwillee (approved):
@nicovanduijn Thanks very much for this - I did not know the feature existed.

I have moved this out of "standard/basic configuration" into the advanced configuration section. Generally the basic configuration is for those things that an end user with an RTF airframe would need to do. The advanced config is for everything else.

I have also subedited this a little "to my taste". Essentially that means cross linking to parameters and other docs and summarizing/removing repetition. There isn't any more (or less) information than you had.

I've approved, but let's get technical confirmation from the other people your recommended.

en/advanced_config/README.md
```diff
@@ -1,10 +1,32 @@
...
+Note that the [peripheral hardware](../peripherals/README.md) documentation contains configuration that is specific to particular sensors and actuators.
  
 :::tip
-The topics are useful if you are creating a new airframe type or significantly modifying a supported frame.
-Generally if you're using a [supported airframe](../airframes/airframe_reference.md#copter) (e.g. from [QGroundControl > Airframe](../config/airframe.md)) the default tuning and configuration should be acceptable. 
+If working with an existing kit or ready-to-fly airframe you should only need the [Basic Configuration](../config/README.md). 
 :::
 
-For the standard configuration appropriate to every vehicle see: [Basic Configuration](../config/README.md).
+Topics include:
```
hamishwillee, line 12: As an aside, I'm open to suggestions on how to better integrate the configuration information. At the moment we have 
- Basic Config 
  - the stuff everyone building an airframe will need to do - even if building a standard airframe reference frame.
  - Some of this would even need to be done for RTF or kit vehicle - such as compass calibration, which might drift
- Advanced config - vehicle specific stuff, factory imu configuration, serial port/mavlink config
- Peripherals - individual sensors and busses (e.g. UAVCAN) have some setup info. 

This structure is pretty good if you're unfamiliar with drones and you're starting with a canned multicopter airframe or a kit frame - because you only see what you need to understand the FC assembly and wiring. But if you come in with interest in VTOL or FC, or if you come in as an OEM some of this stuff is more hidden.

I am wondering if we might be better, at least for config, having

- Configuration
  - Basic Configuration
  - MC Config
  - VTOL
  - FW
  - Factory?
  - Other categories?
  
My motivation is to reduce the sidebar and keep all configuration and calibration together. I'd still keep specific sensor/actuator configuration and setup in the respective hardware docs.

Is this better/worse? Are there any other topics you'd group in "factory/OEM"? Should we consider doing it more like ArduPilot and dividing up by vehicle type (harder!)

nicovanduijn: I don't think my opinion carries a lot of weight, but I would personally not split these up by airframe types. I think the chances of having to duplicate some information is pretty high.

I'd prefer not to stall this PR for such a restructuring discussion though, unless you think those are tightly coupled?

en/advanced_config/imu_factory_calibration.md
```diff
@@ -0,0 +1,30 @@
+# IMU Factory Calibration
+
+PX4 OEM manufacturers can perform an IMU factory calibration in order to store persistent values for the accelerometer, gyroscope and magnetometer biases into persistent EEPROM.
+This ensures that end users can always reset vehicle configurations and tuning to a safe, flying state.
+
+This procedure will write the following parameters to `/fs/mtd_caldata`: [CAL_ACC*](../advanced_config/parameter_reference.md#CAL_ACC0_ID), [CAL_MAG*](../advanced_config/parameter_reference.md#CAL_MAG0_ID), [CAL_GYRO*](../advanced_config/parameter_reference.md#CAL_GYRO0_ID).
+This data will then be used when the parameters are set (or reset) to their default values.
+
+:::warning
+This feature is dependent on the availability of storage at `/fs/mtd_caldata`.
```
potaito, line 10: So if the board has EEPROM and the board config is properly set up, the directory `/fs/mtd_caldata` should automatically be created, right? No user intervention required?

hamishwillee: I very much doubt that creation or not of the file would be board-specific. I've updated the text to note this.

hamishwillee:
I'm going to merge this as it is useful. Would appreciate an after-the-fact sanity check from the designated reviewers - e.g. @bkueng 

In addition, I would quite like to know the following:

- The PR contains the text:

  > This feature relies on the FMU having a dedicated EEPROM chip or an accompanying IMU PCBA that has sufficient space for the data.
  > PX4 will store the data EEPROM at `/fs/mtd_caldata`, creating the file if necessary.

  - Is `/fs/` always mapped to EEPROM?
    - I know `etc/logging/` is on an SD card so I am wondering if this is a "standard mapping", and if there are other standard mappings we should document (again, I assume "/etc/ tells PX4 that we mean "SD card".
    - Can SD card be used?
  - Where are other "default values" for parameters stored? i.e. parameters have a default value. I assume this is stored in "ROM" and is unwritable, which is why we need to separately store these "writable default values" in writable persistent memory that is guaranteed to be present (EEPROM).

en/advanced_config/imu_factory_calibration.md
```diff
@@ -0,0 +1,30 @@
...
+
+PX4 OEM manufacturers can perform an IMU factory calibration in order to store persistent values for the accelerometer, gyroscope and magnetometer biases into persistent EEPROM.
+This ensures that end users can always reset vehicle configurations and tuning to a safe state for flying.
+
+This procedure will write the following parameters to `/fs/mtd_caldata`: [CAL_ACC*](../advanced_config/parameter_reference.md#CAL_ACC0_ID), [CAL_MAG*](../advanced_config/parameter_reference.md#CAL_MAG0_ID), [CAL_GYRO*](../advanced_config/parameter_reference.md#CAL_GYRO0_ID).
+This data will then be used when the parameters are set (or reset) to their default values.
+
+:::warning
+This feature relies on the FMU having a dedicated EEPROM chip or an accompanying IMU PCBA that has sufficient space for the data.
+PX4 will store the data EEPROM at `/fs/mtd_caldata`, creating the file if necessary.
```
bkueng, line 11: The file exists either way. It's a good way to check if the board supports it if in doubt.

bkueng (commented):
>  Is /fs/ always mapped to EEPROM?

No, it's virtual.

> Can SD card be used - e.g. as alternative or fallback?

No, but it could be implemented if anyone wanted to.

> Where are other "default values" for parameters stored?

They are built into the PX4 binary.

en/advanced_config/imu_factory_calibration.md
hamishwillee: @bkueng You're saying that if there is an EEPROM (or somewhere to store the file) this file will already be created?

Do you think it is necessary to go to the level of explaining how to check for board support? IF so, what's the best way - just run a file system exists() check?

hamishwillee: FYI, did https://github.com/PX4/PX4-user_guide/commit/5addb362935bc4902b720b3ac36d679d96e26064 to remove the "creating the file ..." bit.

bkueng: Yes I think it's a good check to be sure. You can do `ls /fs` on the shell, and the output should contain `mtd_caldata`

hamishwillee: Thanks very much. Added that in commit https://github.com/PX4/PX4-user_guide/commit/0d1571cab1cd6364a496e1b687d2f8fd47c2f2bc (meant to PR it, but oh well)

PR #2232, opened 2022-12-30
FW takeoff docs improvements

sfuhrer:
FW takeoff (and landing) docs are outdated. Major changes on the PX4 side came in lately through https://github.com/PX4/PX4-Autopilot/pull/20537 and https://github.com/PX4/PX4-Autopilot/pull/20557. This PR reflects the changes on the Takeoff side, the Landing side is still todo. 
I've moved all the takeoff docs/hints/param description to the Takeoff mode section (currently have also a section under the "flying 101" that is to a big part duplicated info). Maybe we can make it even clearer that this section applies to both takeoffs via mission and without mission in manually entered Takeoff flight mode. Also up for other ideas on where to place it @hamishwillee. 

@tstastny please check if the current controller logic is correctly resembled in the docs, and if you agree with the level of depth I propose here (there are ofc many more parameters that affect takeoffs, but listing them all removes focus from the important ones).


en/flight_modes/takeoff.md
```diff
@@ -31,47 +31,80 @@ Parameter | Description
...
 
 <span id="fixed_wing"></span>
-## Fixed Wing (FW)
+## Fixed-wing (FW)
+
+The aircraft takes off in the current direction (heading) using either *catapult/hand-launch mode* or *runway takeoff mode*.
+The mode defaults to catapult/hand launch, but can be set to runway takeoff by setting [RWTO_TKOFF](#RWTO_TKOFF) to 1.
+
+There are generally two ways to start an automatic takeoff on fixed-wing vehicles: either by planning a mission takeoff and starting the mission, or by switching to the Takeoff flight mode and arm the vehicle. Both ways lead to the same behavior, with the difference that in a mission takeoff, the user can specify the takeoff course (vehicle will track this course after being clear of the runway) and takeoff altitude (vehicle will climb to this altitude while staying on the takeoff course). For the non-mission takeoff, the course is set to the vehicle heading on arming, and the altitude is set to [MIS_TAKEOFF_ALT](#MIS_TAKEOFF_ALT).
```
hamishwillee, line 39: Layout to my taste
```suggestion
There are generally two ways to start an automatic takeoff on fixed-wing vehicles: either by planning a mission takeoff and starting the mission, or by switching to the Takeoff flight mode and arm the vehicle.
Both methods lead to very similar behavior, with the difference being that in a mission takeoff, the user can specify the takeoff course (vehicle will track this course after being clear of the runway) and takeoff altitude (vehicle will climb to this altitude while staying on the takeoff course).
For the non-mission takeoff, the course is set to the vehicle heading on arming, and the altitude is set to [MIS_TAKEOFF_ALT](#MIS_TAKEOFF_ALT).

```

en/flight_modes/takeoff.md
```diff
@@ -31,47 +31,83 @@ Parameter | Description
...
 <span id="fixed_wing"></span>
-## Fixed Wing (FW)
+## Fixed-wing (FW)
+
+The aircraft takes off in the current direction (heading) using either *catapult/hand-launch mode* or *runway takeoff mode*.
+The mode defaults to catapult/hand launch, but can be set to runway takeoff by setting [RWTO_TKOFF](#RWTO_TKOFF) to 1.
+
+There are generally two ways to start an automatic takeoff on fixed-wing vehicles: either by planning a mission takeoff and starting the mission, or by switching to the Takeoff flight mode and arm the vehicle.
+Both methods lead to very similar behavior, with the difference being that in a mission takeoff, the user can specify the takeoff course (vehicle will track this course after being clear of the runway) and takeoff altitude (vehicle will climb to this altitude while staying on the takeoff course).
+For the non-mission takeoff, the course is set to the vehicle heading on arming, and the altitude is set to [MIS_TAKEOFF_ALT](#MIS_TAKEOFF_ALT).
```
hamishwillee, line 41: Does this mean on reaching the altitude the vehicle will continue forever on the arming-set course in Takeoff mode? Or will it switch to some other mode on reaching the altitude (e.g. loiter/hold)?

hamishwillee: I see below it switches to hold. We should say that at this level.

I also "wonder" if we should pull the mission-mode takeoff information here into a note or its own section. I.e. I think this is useful to state, but we should try keep this whole doc as much about FW takeoff mode as possible and therefore keep the mission mode comparisons as an aside.

en/flight_modes/takeoff.md
```diff
@@ -31,47 +31,83 @@ Parameter | Description
...
 
 <span id="hand_launch"></span>
 ### Catapult/Hand Launch
 
-In *catapult/hand launch mode* the vehicle waits to detect launch (based on acceleration trigger).
-On launch it ramps up to full throttle ([RWTO_MAX_THR](#RWTO_MAX_THR)) in about 2 seconds and then performs a full throttle climbout, with *minimum* 10 degree takeoff pitch. 
-Once it reaches [FW_CLMBOUT_DIFF](#FW_CLMBOUT_DIFF) it will transition to [Hold mode](../flight_modes/hold.md) and loiter.
+In *catapult/hand-launch mode* the vehicle waits to detect launch (based on acceleration trigger).
+On launch it ramps up to full throttle [FW_THR_MAX](#FW_THR_MAX) and climbs with the maximum climb rate [FW_T_CLMB_MAX](#FW_T_CLMB_MAX) while keeping the pitch setpoint above [FW_TKO_PITCH_MIN](#FW_TKO_PITCH_MIN).
+Once it reaches [MIS_TAKEOFF_ALT](#MIS_TAKEOFF_ALT) (or altitude defined in the mission item) it will automatically switch to [Hold mode](../flight_modes/hold.md) and loiter.
```
hamishwillee, line 62: This is the takeoff mode doc - if you have that caveat this is incorrect because in a mission it will continue to the next mission item before continuing.
I think if you can have a catapult launch in a mission mode, we need to have that information in the mission mode doc or "below" it. 

```suggestion
Once it reaches [MIS_TAKEOFF_ALT](#MIS_TAKEOFF_ALT) it will automatically switch to [Hold mode](../flight_modes/hold.md) and loiter.

```

en/flight_modes/takeoff.md
```diff
@@ -31,47 +31,83 @@ Parameter | Description
...
+
+All RC stick movement is ignored during the full takeoff sequence.
+
+To launch in this mode:
+- Arm the vehicle.
+- Put the vehicle into *Takeoff mode* or *Start mission with current waypoint being a Takeoff*
```
hamishwillee, line 68: Again, this is the takeoff mode doc. 

```suggestion
- Put the vehicle into *Takeoff mode*

```

en/flight_modes/takeoff.md
```diff
@@ -31,47 +31,83 @@ Parameter | Description
...
 <span id="runway_launch"></span>
 ### Runway Takeoff
 
-The *runway takeoff mode* has the following phases:
-
-1. **Throttle ramp**: Clamped to the runway (pitch fixed, no roll, and heading hold) until reach the minimum airspeed for takeoff ([FW_AIRSPD_MIN](#FW_AIRSPD_MIN) x [RWTO_AIRSPD_SCL](#RWTO_AIRSPD_SCL)).
-1. **Takeoff**: Increase pitch and continue until vehicle altitude > navigation altitude ([RWTO_NAV_ALT](#RWTO_NAV_ALT)).
-1. **Climbout**: Climb until altitude above ground level > [FW_CLMBOUT_DIFF](#FW_CLMBOUT_DIFF).
-   In this phase roll and heading restrictions are removed.
+For vehicles with a landing gear and steerable wheel. Enable wheel controller via [FW_W_EN](#FW_W_EN). Vehicle should be centered and aligned with runway when takeoff is initiated (either by arming in takeoff mode or by starting a mission with a takeoff mission item as active waypoint). The operator can "nudge" the vehicle while on the runway to help keeping it centered and aligned (see [RWTO_NUDGE](../advanced_config/parameter_reference.md#RWTO_NUDGE)).
```
hamishwillee, line 84:
```suggestion
For vehicles with a landing gear and steerable wheel. Enable wheel controller via [FW_W_EN](#FW_W_EN).
Vehicle should be centered and aligned with runway when takeoff is initiated (either by arming in takeoff mode or by starting a mission with a takeoff mission item as active waypoint).
The operator can "nudge" the vehicle while on the runway to help keeping it centered and aligned (see [RWTO_NUDGE](../advanced_config/parameter_reference.md#RWTO_NUDGE)).

```

en/flight_modes/takeoff.md
```diff
@@ -31,47 +31,86 @@ Parameter | Description
...
-:::note
-In addition to the behaviour discussed above there is also a launch detector that may block the launch sequence from starting until some condition is met.
-For catapult launch this is some acceleration threshold.
-:::
 
 <span id="runway_launch"></span>
 ### Runway Takeoff
 
-The *runway takeoff mode* has the following phases:
+For vehicles with a landing gear and steerable wheel. Enable wheel controller via [FW_W_EN](#FW_W_EN).
```
hamishwillee, line 84:
```suggestion
For vehicles with a landing gear and steerable wheel.
Enable wheel controller via [FW_W_EN](#FW_W_EN).

```

en/flight_modes/takeoff.md
```diff
@@ -31,47 +31,88 @@ Parameter | Description
...
-In addition to the behaviour discussed above there is also a launch detector that may block the launch sequence from starting until some condition is met.
-For catapult launch this is some acceleration threshold.
-:::
 
 <span id="runway_launch"></span>
 ### Runway Takeoff
 
-The *runway takeoff mode* has the following phases:
+For vehicles with a landing gear and steerable wheel.
+Enable wheel controller via [FW_W_EN](#FW_W_EN).
```
hamishwillee, line 85: I think you are saying...
```suggestion
Runway takeoffs can be used by vehicles with landing gear and and steerable wheel (only).
You will first need to enable the wheel controller using the parameter [FW_W_EN](#FW_W_EN).

```

en/flight_modes/takeoff.md
```diff
@@ -31,47 +31,88 @@ Parameter | Description
...
+1. **Clamped to runway**: Pitch fixed, no roll and course hold until the rotation airspeed ([RWTO_ROT_AIRSPD](../advanced_config/parameter_reference.md#RWTO_ROT_AIRSPD)) is reached. The operator is able to nudge the vehicle left/right via yaw stick.
+1. **Climbout**: Increase pitch setpoint and climb to takeoff altitude. To prevent wingstrikes, the controller will keep the roll setpoint locked to 0 when close to the ground, and then gradually allow more roll while climbing. It is based on the vehicle geometry as configured in [FW_WING_SPAN](#FW_WING_SPAN) and [FW_WING_HEIGHT](#FW_WING_HEIGHT).
 
-Takeoff is affected by the following parameters:
+:::note
+For a smooth takeoff, the runway wheel controller possibly needs to be tuned. It consists of a rate controller (P-I-FF-controller with the parameters [FW_WR_P](../advanced_config/parameter_reference.md#FW_WR_P), [FW_WR_I](../advanced_config/parameter_reference.md#FW_WR_I), [FW_WR_FF](../advanced_config/parameter_reference.md#FW_WR_FF)) and an outerloop that calculates heading setpoints from course errors and can be tuned via [RWTO_L1_PERIOD](#RWTO_L1_PERIOD).
```
hamishwillee, line 97:
```suggestion
For a smooth takeoff, the runway wheel controller possibly needs to be tuned.
It consists of a rate controller (P-I-FF-controller with the parameters [FW_WR_P](../advanced_config/parameter_reference.md#FW_WR_P), [FW_WR_I](../advanced_config/parameter_reference.md#FW_WR_I), [FW_WR_FF](../advanced_config/parameter_reference.md#FW_WR_FF)) and an outer loop that calculates heading setpoints from course errors and can be tuned via [RWTO_L1_PERIOD](#RWTO_L1_PERIOD).

```

hamishwillee:
@sfuhrer Excellent. I am very happy with the removal of the FW landing standalone doc and the integration into Takeoff mode doc. 

I am not so sure about the mixing of mission mode takeoff information in the takeoff mode doc. It certainly isn't "clean" right now. From this I cannot tell 
- how to plan a mission takeoff
- whether I can do a runway takeoff in a mission
- whether I can do a catapult takeoff in a mission - and how I plan
- What configuration is required that is common to both mission and takeoff mode.

I think we at least need a "planning a mission takeoff" section to cover this, which we can cross link from mission mode, and which might cross link to the takeoff mode you have written - depending on how much duplication there is.

We can have a call to chat through if you want.

en/flight_modes/takeoff.md
sfuhrer: >Does this mean on reaching the altitude the vehicle will continue forever on the arming-set course in Takeoff mode? Or will it switch to some other mode on reaching the altitude (e.g. loiter/hold)?

It will switch to Hold and circle around Home (Takeoff without mission), proceed with the mission otherwise

sfuhrer: Yes please feel free to add the info about Hold here.

Separating all the Mission-takeoff related things to a note sounds reasonable to me indeed. We anyway should add some very rough bullet points on how to plan a Mission Takeoff also to the Mission section, and then reference to the docs here (like "place a Takeoff item in the direction of the runway, and set the altitude higher than obstacles in close proximity".

en/flight_modes/takeoff.md
sfuhrer: Yes this should be cleanly separated: Loiter in case of mission-less takeoff, proceed to next WP in Mission.

en/flight_modes/takeoff.md
```diff
@@ -31,47 +31,91 @@ Parameter | Description
...
-RC stick movement is ignored in both cases.
 
 <span id="hand_launch"></span>
 ### Catapult/Hand Launch
 
-In *catapult/hand launch mode* the vehicle waits to detect launch (based on acceleration trigger).
-On launch it ramps up to full throttle ([RWTO_MAX_THR](#RWTO_MAX_THR)) in about 2 seconds and then performs a full throttle climbout, with *minimum* 10 degree takeoff pitch. 
-Once it reaches [FW_CLMBOUT_DIFF](#FW_CLMBOUT_DIFF) it will transition to [Hold mode](../flight_modes/hold.md) and loiter.
+In *catapult/hand-launch mode* the vehicle waits to detect launch (based on acceleration trigger).
+On launch it ramps up to full throttle [FW_THR_MAX](#FW_THR_MAX) and climbs with the maximum climb rate [FW_T_CLMB_MAX](#FW_T_CLMB_MAX) while keeping the pitch setpoint above [FW_TKO_PITCH_MIN](#FW_TKO_PITCH_MIN).
```
tstastny, line 61: No ramp on catapult/hand launch. Also delegates throttle command to TECS, no explicit setting of max throttle.

https://github.com/PX4/PX4-Autopilot/blob/00bfd5436a6f743fe79d6c300ab37982fcf2987f/src/modules/fw_pos_control_l1/FixedwingPositionControl.cpp#L1538-L1583

en/flight_modes/takeoff.md
tstastny: there's also some repeat info in the paragraph from before and this one. maybe could do another revision pass through and even just describe each mode separately (with less general description beforehand)

en/flight_modes/takeoff.md
```diff
@@ -31,47 +31,91 @@ Parameter | Description
...
+
+Parameters that apply to both catapult/hand-launch as well as runway takeoffs:
+
+Parameter | Description
+--- | ---
+<span id="MIS_TAKEOFF_ALT"></span>[MIS_TAKEOFF_ALT](../advanced_config/parameter_reference.md#MIS_TAKEOFF_ALT) | Altitude setpoint above Home that the vehicle will climb to during takeoff if not in mission mode.
+<span id="FW_TKO_AIRSPD"></span>[FW_TKO_AIRSPD](../advanced_config/parameter_reference.md#FW_TKO_AIRSPD) | Takeoff airspeed (is set to [FW_AIRSPD_MIN](../advanced_config/parameter_reference.md#FW_AIRSPD_MIN) if not defined by operator)
+<span id="FW_TKO_PITCH_MIN"></span>[FW_TKO_PITCH_MIN](../advanced_config/parameter_reference.md#FW_TKO_PITCH_MIN) | This is the minimum pitch angle setpoint during the climbout phase.
+<span id="FW_THR_MAX"></span>[FW_THR_MAX](../advanced_config/parameter_reference.md#FW_THR_MAX) | Maximum throttle during the whole takeoff phase.
+<span id="FW_T_CLMB_MAX"></span>[FW_T_CLMB_MAX](../advanced_config/parameter_reference.md#FW_T_CLMB_MAX) | Climb rate setpoint during climbout phase.
```
tstastny, line 52: maybe good to indicate that this SHOULD NOT be changed specifically for a different desired climb rate, as this is a tuning parameter for TECS. -- or just mention it as the value that is used in paragraph, but then only list in the table here "configurable" parameters for adjusting the takeoff behavior?

en/flight_modes/takeoff.md
```diff
@@ -31,47 +31,91 @@ Parameter | Description
...
 
+Vehicle should be centered and aligned with runway when takeoff is initiated (either by arming in takeoff mode or by starting a mission with a takeoff mission item as active waypoint).
+The operator can "nudge" the vehicle while on the runway to help keeping it centered and aligned (see [RWTO_NUDGE](../advanced_config/parameter_reference.md#RWTO_NUDGE)).
 
-### Fixed Wing Takeoff Parameters
 
-Takeoff is affected by the following parameters:
+The *runway takeoff mode* has the following phases:
+1. **Throttle ramp**: Throttle is ramped up within [RWTO_RAMP_TIME](../advanced_config/parameter_reference.md#RWTO_RAMP_TIME) to [RWTO_MAX_THR](../advanced_config/parameter_reference.md#RWTO_MAX_THR).
+1. **Clamped to runway**: Pitch fixed, no roll and course hold until the rotation airspeed ([RWTO_ROT_AIRSPD](../advanced_config/parameter_reference.md#RWTO_ROT_AIRSPD)) is reached. The operator is able to nudge the vehicle left/right via yaw stick.
```
tstastny, line 94: not sure about calling it "course hold" - as it is actually following a line. maybe "controls to takeoff path"

en/flight_modes/takeoff.md
```diff
@@ -31,47 +31,91 @@ Parameter | Description
...
-<span id="RWTO_NAV_ALT"></span>[RWTO_NAV_ALT](../advanced_config/parameter_reference.md#RWTO_NAV_ALT) | Altitude above ground level (AGL) at which we have enough ground clearance to allow some roll. Until `RWTO_NAV_ALT` is reached the plane is held level and only rudder is used to keep the heading (see <span id="RWTO_HDG"></span>[RWTO_HDG](../advanced_config/parameter_reference.md#RWTO_HDG)). This should be below `FW_CLMBOUT_DIFF` if `FW_CLMBOUT_DIFF` > 0.
+<span id="RWTO_RAMP_TIME"></span>[RWTO_RAMP_TIME](../advanced_config/parameter_reference.md#RWTO_RAMP_TIME) | Throttle ramp up time.
+<span id="RWTO_ROT_AIRSPD"></span>[RWTO_ROT_AIRSPD](../advanced_config/parameter_reference.md#RWTO_ROT_AIRSPD) | Airspeed threshold to start rotation (pitching up). If not configured by operator is set to 0.9*FW_TKO_AIRSPD.
+<span id="FW_TKO_AIRSPD"></span>[FW_TKO_AIRSPD](../advanced_config/parameter_reference.md#FW_TKO_AIRSPD) | Airspeed setpoint during takeoff. If not configured by operator is set to FW_AIRSPD_MIN.
```
tstastny, line 112:
```suggestion
<span id="FW_TKO_AIRSPD"></span>[FW_TKO_AIRSPD](../advanced_config/parameter_reference.md#FW_TKO_AIRSPD) | Airspeed setpoint during the takeoff climbout phase (after rotation). If not configured by operator is set to FW_AIRSPD_MIN.

```

tstastny (changes requested):

en/flight_modes/takeoff.md
sfuhrer: Should we just remove it from the list here? As you're right, the last thing that should happen is that users tune this param for a specific takeoff climb rate. 
It's already mentioned above in the text that the vehicle will climb with this value, so listing it here as a "core parameter to tune for takeoff" is really not necessary.

sfuhrer: Done in 023be07d0c

en/flight_modes/takeoff.md
sfuhrer: removed it.

en/flight_modes/takeoff.md
sfuhrer: True, removed it.

en/flight_modes/takeoff.md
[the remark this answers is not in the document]
sfuhrer: Magically the rendering was correct (1,2,3) even with all 1. in the raw:) Anyway changed it now.

en/flight_modes/takeoff.md
sfuhrer: removed reference to the mission item altitude here.

sfuhrer:
@hamishwillee I've added a mini section in the Mission section as well now, let me know if that makes the operation clearer. Maybe we can extend it a bit more.

en/flight_modes/mission.md
```diff
@@ -162,3 +162,27 @@ Vehicles switch to the next waypoint as soon as they enter the acceptance radius
   - The equation is: 
     $$L_{1_{distance}}=\frac{1}{\pi}L_{1_{damping}}L_{1_{period}}\left \| \vec{v}_{ {xy}_{ground} } \right \|$$
 
+<span id="fixed_wing_mission_takeoff_landing"></span>
+## Fixed-wing Mission Takeoff/Landing
+
+Starting and ending flights with mission takeoff and landing is the recommended way of operating a plane autonomously. A detailed description of the executed logic during autonomous takeoffs and landings can be found
+in [Fixed-wing Takeoff](../config/flight_modes/takeoff.md#fixed_wing) and [Fixed-wing Takeoff](../config/flight_modes/land.md#fixed_wing).
```
sfuhrer, line 169: Need to fix the link once we get the landing part merged https://github.com/PX4/PX4-user_guide/pull/2245)

tstastny (approved):

en/flight_modes/mission.md
```diff
@@ -29,8 +29,11 @@ At high level all vehicle types behave in the same way when MISSION mode is enga
...
 1. If a mission is stored and PX4 is landed:
-   * On copters PX4 will execute the [mission/flight plan](../flying/missions.md). If the mission does not have a `TAKEOFF` command then PX4 will fly the vehicle to the minimum altitude before executing the remainder of the flight plan from the current step.
-   * On fixed-wing vehicles PX4 will not automatically take off (the autopilot will detect the lack of movement and set the throttle to zero). The vehicle may start executing the mission if hand- or catapult- launched while in mission mode.  
+
+   - On copters PX4 will execute the [mission/flight plan](../flying/missions.md).
+     If the mission does not have a `TAKEOFF` command then PX4 will fly the vehicle to the minimum altitude before executing the remainder of the flight plan from the current step.
+   - On fixed-wing vehicles PX4 will not automatically take off (the autopilot will detect the lack of movement and set the throttle to zero).
+     If the currently active waypoint is a Takeoff, the system will automatically takeoff (see [FW Takeoff/Landing in Mission](#fixed-wing-mission-takeoff-landing)).
```
hamishwillee, line 36: @sfuhrer @tstastny Just to be very clear here:
- In takeoff mode a catapult launch is triggered by acceleration detection - either throwing or shaking to start the throttle ramping up.
  - What this implies is that in a mission mode you do not need acceleration detection for catapult launch - is that correct? That does not make sense to me - I'd expect the vehicle to arm, but would still expect it to require throwing.
  - However for runway mode I'd expect it to just take off. Can you still "nudge" as in takeoff mode?
- I assume all the parameter setup for catapult/runway are required whether using mission or takeoff?
- If I do a catapult launch in mission I am throwing at a heading that might not match the takeoff item heading. Does the vehicle just track to that takeoff mission item heading as soon as it can?

en/flight_modes/mission.md
```diff
@@ -162,3 +166,27 @@ Vehicles switch to the next waypoint as soon as they enter the acceptance radius
...
+The section here is to highlight differences in mission takeoffs and landings compared to the stand-alone flight modes.
+
+### Takeoff
+
+Fixed-wing mission takeoffs can be configured by placing a Takeoff mission item, including an altitude setpoint, on the map.
+The vehicle will then takeoff towards this waypoint, and climb until the altitude specified is reached.
+After reaching the takeoff altitude, the mission will be automatically proceeded with the next waypoint.
+The configuration is the same for wheeled vehicles that do a runway takeoff, or hand-launched ones.
+In both cases, the vehicle should be placed facing the takeoff waypoint when the mission is started, 
+which will arm the vehicle and throttle up the motors.
```
hamishwillee, line 183: Again, so for catapult you start mission and you'd better be ready to throw the vehicle because it will be throttling up as soon as the takeoff item is reached? Do we need a warning?

en/flight_modes/takeoff.md
```diff
@@ -28,50 +28,96 @@ Parameter | Description
...
+## Fixed-wing (FW)
+
+Automatic takeoff has two modalities: *catapult/hand-launch* or *runway takeoff*.
+The mode defaults to catapult/hand launch, but can be set to runway takeoff by setting [RWTO_TKOFF](#RWTO_TKOFF) to 1.
+
+There are two ways to start an automatic takeoff on fixed-wing vehicles: either by [planning a mission takeoff](../flight_modes/mission.md#takeoff) and starting the mission, or by switching to the _Takeoff mode_ and arming the vehicle.
+
+In both cases, a flight path (starting point and takeoff course) and clearance altitude are defined.
+The flight path takes the vehicle's current position as starting point when the takeoff mode is first entered, and a straight line from this starting point continues in the direction of the defined course indefinitely.
+On takeoff, the aircraft will follow this line, climbing at the maximum climb rate ([FW_T_CLMB_MAX](../advanced_config/parameter_reference.md#FW_T_CLMB_MAX)) until reaching the clearance altitude.
```
hamishwillee, line 43: For a takeoff mode the heading comes from the launch direction. 

For mission mode I'm not certain. What I think happens is that the [TAKEOFF](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_TAKEOFF) item defines the lat, lon, altitude. 
- Is the altitude in the item the clearance altitude? 
- If you mission launch you'll initially be thrown or head off along runway. That direction might not match the mission item lat/lon. At what point do you head towards that lat/lon?

hamishwillee:
@tstastny @sfuhrer Getting pretty close. I've added some confirmations and then I'll do a final subedit. If you can answer the questions today I should be able to get this in tomorrow.

en/flight_modes/mission.md
```diff
@@ -162,3 +166,27 @@ Vehicles switch to the next waypoint as soon as they enter the acceptance radius
...
+
+If possible, always make the vehicle takeoff into the wind.
+
+### Land
+
+Currently the only way to land a vehicle autonomously is through a mission landing.
```
hamishwillee, line 189: Is this true? That's not what it says here: http://docs.px4.io/main/en/flight_modes/land.html#fixed-wing-fw

sfuhrer: Working on a fix, we should remove this line (or once the fix is in).

en/flight_modes/takeoff.md
sfuhrer: - Yes that's the clearance altitude.
- it will immediately go towards the mission Takeoff WP lat/lon (keep bearing from start of takeoff to Takeoff WP until past clearance altitude).

en/flight_modes/mission.md
sfuhrer: Actually the launch detection is only an additional feature, that's not necessary for hand-launch. If it's not enabled the motor will just immediately ramp up on arming. 

>What this implies is that in a mission mode you do not need acceleration detection for catapult launch - is that correct? That does not make sense to me - I'd expect the vehicle to arm, but would still expect it to require throwing.

Mission mode doesn't change the procedure - you arm the vehicle, and if launch detection is enabled it needs to detect an acceleration before it starts the motor. 

>However for runway mode I'd expect it to just take off. Can you still "nudge" as in takeoff mode?

You can only nudge if in runway takeoff mode.

>I assume all the parameter setup for catapult/runway are required whether using mission or takeoff?

Correct, there is no difference in required params.

en/flight_modes/mission.md
sfuhrer: It will throttle up as soon as you arm (start the mission) in case launch detection is disabled, and otherwise once it detects an acceleration. 
Not sure about needing a warning, this should be clear to expect when you arm a drone, no?

en/flight_modes/mission.md
hamishwillee: Do you have a PR for this fix?

en/flight_modes/mission.md
hamishwillee: I modified it a little to say that if you have runway it arms and ramps up throttle, but if you hand launch it will wait for the trigger. No need for a warning, but it is worth saying that the configuration is different.

en/flight_modes/takeoff.md
hamishwillee: No need to answer ^^^. What I've done here is more clearly separated the mission mode and takeoff mode differences. Just a paragraph reorg.

en/flight_modes/mission.md
```diff
@@ -197,3 +200,64 @@ Vehicles switch to the next waypoint as soon as they enter the acceptance radius
...
+A more detailed description of mission mode fixed-wing landing can be found in [Land mode > Fixed-wing](../flight_modes/land.md#fixed-wing-fw) (covering fixed wing landing in both mission mode and takeoff mode).
+:::
+
+Currently the only way to land a vehicle autonomously is through a mission landing.
+It is recommended that the landing is configured through a [landing pattern](https://docs.qgroundcontrol.com/master/en/PlanView/pattern_fixed_wing_landing.html).
+
+If possible, always plan the landing such that it does the approach into the wind.
+
+
+## Multicopter Mission Takeoff/Landing
```
hamishwillee, line 241: FYI, I added these bits for MC and VTOL takeoff landing for "symmetry". We could reasonably expand on them too.

en/flight_modes/takeoff.md
```diff
@@ -28,50 +28,100 @@ Parameter | Description
...
+There are two ways to start an automatic takeoff on fixed-wing vehicles: either by [planning a mission takeoff](../flight_modes/mission.md#fw-takeoff) and starting the mission, or by switching to the _Takeoff mode_ and arming the vehicle.
+
+In both cases, a flight path (starting point and takeoff course) and clearance altitude are defined.
+The flight path takes the vehicle's current position as starting point when the takeoff mode is first entered, and a straight line from this starting point continues in the direction of the defined course indefinitely.
+On takeoff, the aircraft will follow this line, climbing at the maximum climb rate ([FW_T_CLMB_MAX](../advanced_config/parameter_reference.md#FW_T_CLMB_MAX)) until reaching the clearance altitude.
+
+In _Takeoff mode_ (non-mission takeoffs), the course is set to the vehicle heading on arming, and the clearance altitude is set to [MIS_TAKEOFF_ALT](#MIS_TAKEOFF_ALT).
+Reaching the clearance altitude causes the vehicle to enter _Hold mode_.
+
+In [Mission mode](../flight_modes/mission.md) the operator defines the takeoff course and clearance altitude in the Takeoff mission item.
```
hamishwillee, line 48: FYI So here you see the mission mode all separated into own paragraph.

en/flight_modes/takeoff.md
```diff
@@ -28,50 +28,100 @@ Parameter | Description
...
 <span id="COM_RC_OVERRIDE"></span>[COM_RC_OVERRIDE](../advanced_config/parameter_reference.md#COM_RC_OVERRIDE) | Controls whether stick movement on a multicopter (or VTOL in MC mode) causes a mode change to [Position mode](../flight_modes/position_mc.md). This can be separately enabled for auto modes and for offboard mode, and is enabled in auto modes by default.
-<span id="COM_RC_STICK_OV"></span>[COM_RC_STICK_OV](../advanced_config/parameter_reference.md#COM_RC_STICK_OV) | The amount of stick movement that causes a transition to [Position mode](../flight_modes/position_mc.md) (if [COM_RC_OVERRIDE](#COM_RC_OVERRIDE) is enabled).
+<span id="COM_RC_STICK_OV"></span>[COM_RC_STICK_OV](../advanced_config/parameter_reference.md#COM_RC_STICK_OV) | The amount of stick movement that causes a transition to [Position mode](../flight_modes/position_mc.md) (if [COM_RC_OVERRIDE](#COM_RC_OVERRIDE) is enabled)
 
-<span id="fixed_wing"></span>
-## Fixed Wing (FW)
+<a id="fixed_wing"></a>
+## Fixed-wing (FW)
+
+Automatic takeoff has two modalities: *catapult/hand-launch* or *runway takeoff*.
```
hamishwillee, line 36: @sfuhrer I'm going to merge this now - I think it is OK.

My main concern is that we are mixing the takeoff mode and mission mode takeoff, as discussed. I think we could do a better job here, but maybe as a post process.

What we could do is:
1. Duplicate just this section in mission mode FW takeoff, and make it mission mode specific. 
2. Then reference the catapult and runway sections in takeoff mode for the "common stuff".
3. Remove the mission mode specific stuff from here.

We might then put the stuff currently in mission mode into the "Planning" doc. 

Let me know what you think?

sfuhrer: So where would the catapult and runway sections live? In the mission or the flight modes docs, or on a new, neutral ground?

PR #2241, opened 2023-01-09
WSL: add a step to check BIOS virtualization setting

MaEtUgR:
As I'm installing the toolchain on a new computer I add some hints I found useful.

### DIFF
--- a/en/dev_setup/dev_env_windows_wsl.md
+++ b/en/dev_setup/dev_env_windows_wsl.md
@@ -70,6 +70,7 @@ See the link for rollout information.
 
 To install WSL2 with the default Ubuntu distribution on a new installation of Windows 11:
 
+1. Make sure your computer your computer's virtualization feature is enabled in the BIOS. It's usually referred as "Virtualization Technology", "Intel VT-x" or "AMD-V" respectively.
 1. Open _cmd.exe_ as administrator.
    This can be done by pressing the start key, typing `cmd`, right klicking on the _Command prompt_ entry and selecting **Run as administrator**.
 1. Execute the command `wsl --install` to run the installation routine for WSL.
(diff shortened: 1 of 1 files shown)


hamishwillee:
@MaEtUgR How you going?

This is still in draft - did you want it reviewed?

I noted that WSL is now in the marketplace only right? I think I saw that it also provides the graphics support for Windows 10 now as well - is that correct? If so we can presumably update the docs to say that too!

Did you have a chance to play with the USB from VM stuff?

MaEtUgR:
> How you going?

Great thanks and you?

> This is still in draft - did you want it reviewed?

That was by accident sorry 🤦‍♂️ 

> I noted that WSL is now in the marketplace only right?

It can still be installed from the command line like before. I tried this on a new computer which also made me aware you have to check the BIOS setting. But I just checked now and the readme of WSLg graphics support was updated and recommends installing from the store: https://github.com/microsoft/wslg/commit/5886a29a21232dfd87bd50d14e431ddae24663a8

 > I think I saw that it also provides the graphics support for Windows 10 now as well - is that correct?
 
 I wasn't aware but the readme change also looks like you're right. It was probably packaged as stable in the latest Windows 10 major update which is now broadly deployed. Let's adapt.
 
 > Did you have a chance to play with the USB from VM stuff?
 
 No, it's not missing often enough in my workflow until now.

MaEtUgR:
image
👀

hamishwillee (approved):

hamishwillee:
@MaEtUgR Awesome.

Ok, so I've merged this. The docs I've read indicate that wsl on the store is going to be the path offered by microsoft going forward. Also that if the hardware supports it, this will allow graphics too. So we should update these docs:

1. Remove the Windows 10 limitations
2. Replace the command line instructions with link to https://aka.ms/wslstorepage and store instructions.

Do you want to do first draft of that? I can, but I'm likely to be stuck on ROS2 docs for some weeks.

MaEtUgR:
> Do you want to do first draft of that?

I would like to only write such a guide if I really tried it out which is at this moment not at the top of my list especially the Windows 10 part since I'm not using it anymore. Do you try it out? I should ramp up a VM for that and keep the test in the background although that might not be representative in terms of GPU support 👀

PR #2243, opened 2023-01-11
Updated ARK PAB and ARK V6X docs with updated product photos

bradenwagstaff:
We've added in our new product photos for the PAB and V6X. All of these photos are less than 200MB, have dimensions of about 500x400, and cropped similarly to the ones added last week. If 200MB is to large for the site we can compress them more, just let me know and I can do that


bradenwagstaff:
Made a mistake. Gonna redo this and make a new PR

PR #2244, opened 2023-01-12
Updated ARK PAB and ARK V6X docs with updated product photos

bradenwagstaff:
We've added in our new product photos for the PAB and V6X. All of these photos are less than 200MB, have dimensions of about 500x400, and cropped similarly to the ones added last week. If the photos don't size the way you need them to just let me know and I can get you what you need


hamishwillee (approved):
These look great thanks.

FWIW A page will shrink an image for display and you can click to zoom it to as many pixels wide as there are in the original image. There is no hard rule on the size you have to use for the pages - so if you have an image that would benefit from being wider in order to convey the information it demonstrates then you can.

As finger in the air though, if someone sends me an image 4000px wide I will generally:
1. Crop out the unnecessary information (i.e. white space on edges
2. Drop the width to about 1200px 
3. Run a near-lossless compressor
4. Check the size. If it is below 200Kb, move on.

Upshot, you don't have to kill yourself here, mostly trying to avoid stupidly large files.

PR #2245, opened 2023-01-18
update fixed-wing landing documentation

tstastny:
Updates fixed-wing landing docs following new logic in https://github.com/PX4/PX4-Autopilot/pull/19871 and https://github.com/PX4/PX4-Autopilot/pull/20537.

I removed the separate fixed-wing landing page and stuck things in the landing mode page, similar to @sfuhrer's PR #2232. Of course.. fixed-wing doesnt really have a "landing mode" commandable via mav command at the moment. Only possible to reach it through mission item. But as we start moving things into individual modes.. we'll hopefully get there soon.

@hamishwillee let me know if you would like to re-organize some of these topics.


en/flight_modes/land.md
```diff
@@ -32,24 +32,110 @@ Parameter | Description
...
+[FW_THR_IDLE](../advanced_config/parameter_reference.md#FW_THR_IDLE) | Idle throttle setting. The vehicle will retain this setting throughout the flare and roll out.
+
+### Abort
+
+#### Operator abort
+
+The landing may be aborted by the operator at any point during the final approach, resulting in a climb out to an orbit pattern centered above the land waypoint.
+The maximum of the aircraft's current altitude and [MIS_LND_ABRT_ALT](../advanced_config/parameter_reference.md#MIS_LND_ABRT_ALT) is set as the abort orbit alitutde -- height relative to (above) the landing waypoint.
+Landing configuration (e.g. flaps, spoilers, landing airspeed) is disabled during abort and the aicraft flies in cruise conditions.
+Aborting is not possible during the flare.
```
sfuhrer, line 91: Maybe worth a note here
```suggestion
Aborting is not possible during the flare. Operators can still manually abort the landing at this point by changing the flight mode to a manual mode (e.g. Position). 

```

en/flight_modes/land.md
```diff
@@ -32,24 +32,110 @@ Parameter | Description
...
+[FW_LND_FL_PMIN](../advanced_config/parameter_reference.md#FW_LND_FL_PMIN) | Minimum allowed pitch during the flare (often necessary to avoid negative pitch angles commanded to increase airspeed, as the throttle is reduced to idle setting.)
+[FW_LND_TD_TIME](../advanced_config/parameter_reference.md#FW_LND_TD_TIME) | The time after flare start when the vehicle should pitch the nose down.
+[RWTO_PSP](../advanced_config/parameter_reference.md#RWTO_PSP) | Pitch setpoint while on the runway. For tricycle gear, typically near zero. For tail draggers, positive.
+[FW_THR_IDLE](../advanced_config/parameter_reference.md#FW_THR_IDLE) | Idle throttle setting. The vehicle will retain this setting throughout the flare and roll out.
+
+### Abort
+
+#### Operator abort
+
+The landing may be aborted by the operator at any point during the final approach, resulting in a climb out to an orbit pattern centered above the land waypoint.
```
sfuhrer, line 88: Should we mention here how the operator abort can be triggered? (by pressing the "Abort landing" button on QGC, which appears when currently in a landing sequence). It's though ofc groundstation specific, thus not that sure if it belongs here.

sfuhrer (commented):
I couldn't find any inconsistency to the PX4 logic, looks all correct to me! And love the plots, they're very informative!

en/flight_modes/land.md
tstastny: https://github.com/PX4/PX4-user_guide/pull/2245/commits/99c66dcf609accb4564ef0c9ee71c47372e06019

en/flight_modes/land.md
tstastny: https://github.com/PX4/PX4-user_guide/pull/2245/commits/3f4a09ab88b7ce1cc71a736130ccbb442a2f2dc1

sfuhrer (approved):

en/flight_modes/land.md
```diff
@@ -32,24 +32,113 @@ Parameter | Description
 <span id="COM_RC_STICK_OV"></span>[COM_RC_STICK_OV](../advanced_config/parameter_reference.md#COM_RC_STICK_OV) | The amount of stick movement that causes a transition to [Position mode](../flight_modes/position_mc.md) (if [COM_RC_OVERRIDE](#COM_RC_OVERRIDE) is enabled).
 
 
-## Fixed Wing (FW)
+## Fixed-wing (FW)
 
-The vehicle will turn and land at the location at which the mode was engaged.
-RC stick movement is ignored.
+Fixed-wing landing via mav command is currently not supported.
```
hamishwillee, line 37: Are you saying that if you get [MAV_CMD_NAV_LAND](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_LAND) as a MAVLink command the command will be ignored - is that right? Does that mean the command will just be rejected?

Ditto for setting the mode to LAND using a mode setting command - this will just be rejected and you'll stay in your current mode?

en/flight_modes/land.md
```diff
@@ -26,30 +26,132 @@ Landing is affected by the following parameters:
...
+
+The flaring altitude is calculated during the final approach via "time-to-impact" ([FW_LND_FL_TIME](#FW_LND_FL_TIME)) and the approach descent rate.
+An additional safety parameter [FW_LND_FLALT](#FW_LND_FLALT) sets the minimum altitude at which the vehicle will flare (if the time based altitude is too low to allow a safe flare maneuver).
+
+If belly landing, the vehicle will continue in the flaring state until touchdown, land detection, and subsequent disarm. For runway landings, [FW_LND_TD_TIME](#FW_LND_TD_TIME) enables setting the time post flare start to pitch down the nose (e.g. consider tricycle gear) onto the runway ([RWTO_PSP](#RWTO_PSP)) and avoid bouncing. This time roughly corresponds to the touchdown post flare, and should be tuned for a given airframe during test flights only after the flare has been tuned.
+
+Parameter | Description
+--- | ---
+<a id="FW_LND_FL_TIME"></a>[FW_LND_FL_TIME](../advanced_config/parameter_reference.md#FW_LND_FL_TIME) | Time before impact (at current descent rate) at which the vehicle should flare.
```
hamishwillee, line 78: FYI, note I added anchors in all these. This pattern allows us to link from anywhere in the page to the table (maintaining context of the page) rather than linking to the parameter page directly.

en/flight_modes/land.md
```diff
@@ -26,30 +26,132 @@ Landing is affected by the following parameters:
...
+
+Parameter | Description
+--- | ---
+[FW_LND_ANG](../advanced_config/parameter_reference.md#FW_LND_ANG) | The maximum achievable landing approach slope angle. Note that smaller angles may still be commanded via the landing pattern mission item.
+[FW_LND_EARLYCFG](../advanced_config/parameter_reference.md#FW_LND_EARLYCFG) | Optionally deploy landing configuration during the landing descent orbit (e.g. flaps, spoilers, landing airspeed).
+[FW_LND_AIRSPD](../advanced_config/parameter_reference.md#FW_LND_AIRSPD) | Calibrated airspeed setpoint during landing.
+[FW_FLAPS_LND_SCL](../advanced_config/parameter_reference.md#FW_FLAPS_LND_SCL) | Flaps setting during landing.
+[FW_LND_THRTC_SC](../advanced_config/parameter_reference.md#FW_LND_THRTC_SC) | Altitude time constant factor for landing (overrides default [TECS tuning](../config_fw/advanced_tuning_guide_fixedwing.md)).
+
+### Flaring / Roll-out
```
hamishwillee, line 67: FY Note, we use first-letter capitalisation across docs.

en/flight_modes/land.md
```diff
@@ -26,30 +26,132 @@ Landing is affected by the following parameters:
...
-The vehicle will turn and land at the location at which the mode was engaged.
-RC stick movement is ignored.
+Fixed-wing landing via mav command is currently not supported.
+The landing mode is only accessible in missions.
+Please see QGroundControl docs for more information on [planning a fixed-wing landing pattern mission item](https://docs.qgroundcontrol.com/master/en/PlanView/pattern_fixed_wing_landing.html).
 
-Fixed wing landing logic and parameters are explained in the topic: [Landing (Fixed Wing)](../flying/fixed_wing_landing.md).
+### Landing Sequence
+
+The fixed-wing landing pattern has three components: a **landing descent orbit**, a **landing waypoint**, and a **landing approach slope**.
```
hamishwillee, line 43: I find this confusing. At first I thought it meant "three stages" but then I see that a waypoint is not a stage, and neither is a landing approach slope. 

Now I think these might relate to the landing pattern definition, but none of these terms are mentioned in https://docs.qgroundcontrol.com/master/en/PlanView/pattern_fixed_wing_landing.html#landing-point

Upshot, can we use consistent terminology or find a mapping between the two.

en/flight_modes/land.md
```diff
@@ -26,30 +26,132 @@ Landing is affected by the following parameters:
...
+
+On engaging a landing pattern, the following sequence is initiated:
+
+1. **Fly to landing location**: The aircraft flies at the current altitude to the landing descent orbit.
+2. **Descend to approach altitude**: On reaching the orbit, the vehicle will descend to the approach altitude.
+   Once reaching the approach altitude, the airframe will continue on the orbit until reaching the final approach exit (tangential exit towards landing waypoint).
+4. **Final approach**: The aircraft will follow the landing approach slope until the flare altitude is reached.
```
hamishwillee, line 50: It is much better if this title "Final approach" matches the heading it maps too, and even links. So I think this should be "Landing approach" (I.e. I am guessing this is what you talk about in the section below).

en/flight_modes/land.md
```diff
@@ -26,30 +26,132 @@ Landing is affected by the following parameters:
...
-Fixed wing landing logic and parameters are explained in the topic: [Landing (Fixed Wing)](../flying/fixed_wing_landing.md).
+### Landing Sequence
+
+The fixed-wing landing pattern has three components: a **landing descent orbit**, a **landing waypoint**, and a **landing approach slope**.
+
+On engaging a landing pattern, the following sequence is initiated:
+
+1. **Fly to landing location**: The aircraft flies at the current altitude to the landing descent orbit.
+2. **Descend to approach altitude**: On reaching the orbit, the vehicle will descend to the approach altitude.
+   Once reaching the approach altitude, the airframe will continue on the orbit until reaching the final approach exit (tangential exit towards landing waypoint).
```
hamishwillee, line 49: I think what happens is that when landing is engaged.
1. You fly at current altitude until you enter some radius "landing descent orbit".
   - I think this is the "loiter point radius" in the landing pattern - right? 
   - What is this in MAVLink - i.e. what item and param?
2. You then do a "descending orbit" to decrease your altitude to the approach altitude.
   - Is the approach altitude the "loiter point altitude" in the landing pattern? Or is it something else.
4. You then continue orbit descent until you get to the final approach - i.e. there is a tangential path to the landing waypoint. Now you're in final approach you descent along that path with some settings.
   - I think the final approach altitude might be calculated based on `FW_LND_ANGLE` - i.e. you orbit until you have a good tangent, and if the land angle is less than the max defined by  `FW_LND_ANGLE` you exit the loiter and proceed. Is that right?
5. Then you flare for final landing.

My concern here is that it might be hard to plan the landing if we don't make it clear at what point you decide to leave each stage/what altitude or conditions move between the stages. I might be being thick.

en/flight_modes/land.md
```diff
@@ -26,30 +26,132 @@ Landing is affected by the following parameters:
...
+4. **Final approach**: The aircraft will follow the landing approach slope until the flare altitude is reached.
+5. **Flare**: The vehicle flares until touching down.
```
hamishwillee, line 51: Ditto for this - good to link Flare to the section below it applies to.

hamishwillee:
Looks pretty good. I've added some comments.

It feels pretty odd to say "landings don't work in land mode", and then explain how mission landing works here. It might be worth moving this to mission mode section UNLESS you think that land mode will be supported again soon? I realize that is asymmetric with the takeoff docs. I'll think about it tonight.

en/flight_modes/land.md
sfuhrer: True, maybe enough to just speak about the **landing descent orbit** and the **landing waypoint** here, as those are the mission items that define a mission landing (the slope is defined implicitly from them).

en/flight_modes/land.md
sfuhrer: I'm currently working on enabling it properly. Currently it's not rejected but broken.

It then will circle down at the current location after getting a LAND command. The flaring part is the same as for the mission lading.

sfuhrer:
> It feels pretty odd to say "landings don't work in land mode", and then explain how mission landing works here. It might be worth moving this to mission mode section UNLESS you think that land mode will be supported again soon? I realize that is asymmetric with the takeoff docs. I'll think about it tonight.

I will fix the `land` mode. It will though have more difference to the `mission land `than `takeoff` has from `mission takeoff`. At this point I'm undecided what's better, to duplicate parts of the docs (majority of takeoff, maybe about 30% for land) or to have them only in the mission docs or the flight mode docs.

hamishwillee:
> I will fix the land mode. 

What's the PR for getting landing working? Just so it is easy to track when next update needed.

> It will though have more difference to the mission land than takeoff has from mission takeoff. At this point I'm undecided what's better, to duplicate parts of the docs (majority of takeoff, maybe about 30% for land) or to have them only in the mission docs or the flight mode docs.

I'm not sure either. My preference is to either fully separate or fully join the mission and land/takeoff modes.

So for example in the case of takeoff, we could easily have fixed wing takeoff docs linked from both, very much based on the current doc in land mode. We could then link it from both docs.

For the takeoff case I also suggested another "mixed" option in https://github.com/PX4/PX4-user_guide/pull/2232#discussion_r1092667080. Essentially the intro section for the land mode FW docs could be duplicated and then made specific for land mode and mission mode. The catapult and runway stuff would live in the land mode docs and get linked to.

So you can decide which you like. There is now enough info for me to format in either way without too much of your time wasted.

Might be worth finishing this doc, merging, then formatting both as we agree.


the mode docs as far as possible. If they are really very similar then I'm OK with a single doc Fixed Wing Landing which is referenced by Land mode and Mission mode.


But I have a suggestion in https://github.com/PX4/PX4-user_guide/pull/2232#discussion_r1092667080 for

en/flight_modes/land.md
hamishwillee: OK, so reading the docs I once wrote and should have remembered: https://docs.qgroundcontrol.com/master/en/PlanView/pattern_fixed_wing_landing.html#implementation

>This pattern creates three mission items:
> - [MAV_CMD_DO_LAND_START](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_LAND_START) - If you abort a landing it sends DO_GO_AROUND to the vehicle, which then causes the mission to return to this point and try to land again.
> - [MAV_CMD_NAV_LOITER_TO_ALT](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_LOITER_TO_ALT) - Start point for landing
> -[MAV_CMD_NAV_LAND](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_LAND) - End point for landing
> 
> The vehicle flares to landing using a flight path generated by the firmware between the NAV_LOITER_TO_ALT point and the NAV_LAND point.
>
> If those two locations violate the vehicle's flare constraints (e.g. descent angle is too steep) an error will be raised after you upload the invalid mission to the vehicle.

So based on the above (is it correct to your knowledge) and your comments:

I think what happens is that ...
1. Flying along in a mission, you see `MAV_CMD_DO_LAND_START`. 
   - From the comments in that command, in a mission this is just a placeholder so the system knows it is entering an FW landing sequence.
   - So that would immediately be accepted, ignoring its lat,lon,alt.
3. You then get a `MAV_CMD_NAV_LOITER_TO_ALT` which has a lat, lon, alt, radius
   - You fly towards this lat, lon 
   - When you reach the radius defined in item you do a loitering descent to MAV_CMD_NAV_LOITER_TO_ALT.alt = approach altitude.
4. YOu then continue to loiter down to **some** height where you exit the orbit (when vehicle has tangential path to the position (lat/lon/alt) in MAV_CMD_NAV_LAND)
   - The QGC docs kind of imply that after the loiter alt you have a path to flare that is up to the flight stack - which we have.
   - But it isn't clear how the vehicle decides it will exit the loiter
   - It isn't clear how the flare height is set?
   - Its a bit odd, because you're loitering when no longer in the loiter-to-alt - your're now in the NAV_LAND - right?
6. Then you flare for final landing at MAV_CMD_NAV_LAND

Can you confirm the behaviour and how flare height is set. If so, I think I get this :-)

sfuhrer: >Flying along in a mission, you see MAV_CMD_DO_LAND_START. 

DO_LAND_START is actually only used in RTL to find the land sequence. Not needed in Mission for the landing, there it just follows the WPs.

>2

Correct.

>YOu then continue to loiter down to some height

No, to **exactly** the approach altitude (altitude field of LOITER_TO_ALT). Once there you keep the altitude until the current course is aligned to the next waypoint, then exit. 

>YOu then continue to loiter down to some height

>It isn't clear how the flare height is set?

That's how it is formulated now, which parts would you want to have reformulated?
The flaring altitude is calculated during the final approach via "time-to-impact" ([FW_LND_FL_TIME](#FW_LND_FL_TIME)) and the approach descent rate. An additional safety parameter [FW_LND_FLALT](#FW_LND_FLALT) sets the minimum altitude at which the vehicle will flare (if the time based altitude is too low to allow a safe flare maneuver).

>Its a bit odd, because you're loitering when no longer in the loiter-to-alt - your're now in the NAV_LAND - right?

No you're no longer loitering then.... :thinking: 

>Then you flare for final landing at MAV_CMD_NAV_LAND

You flare once below the the flaring conditions as shown above, otherwise you simply keep descending at the constant descent speed along the final approach slope until ground contact. NAV_LAND is not a WP that has to be accepted in the normal sense- you need it to calculate your final approach (direction and gradient of slope).

sfuhrer:
> What's the PR for getting landing working? Just so it is easy to track when next update needed.

Will open it today.

> Might be worth finishing this doc, merging, then formatting both as we agree.

I would also prefer that. 

>Essentially the intro section for the land mode FW docs could be duplicated and then made specific for land mode and mission mode. The catapult and runway stuff would live in the land mode docs and get linked to.

I think that would be a good compromise indeed.

en/flight_modes/land.md
hamishwillee: OK, so at the MAV_CMD_NAV_LOITER_TO_ALT.alt you loiter until are aligned to next waypoint. Then you head towards MAV_CMD_NAV_LAND. You follow a glide slope down, then flare at end. 

Got it now. Will tidy up tomorrow based on this.

sfuhrer:
> Will open it today.

PR for fixing LAND outside of Mission: https://github.com/PX4/PX4-Autopilot/pull/21036

en/flight_modes/land.md
```diff
@@ -26,30 +27,26 @@ Landing is affected by the following parameters:
...
-This is because in LAND mode the vehicle may not know ground altitude and will assume it is at sea level.
-As ground level may be much higher, a vehicle will often reach the ground at an altitude above where flare logic would be engaged.
+Automated landing in missions is supported: [Mission mode > Fixed wing mission landing](../flight_modes/mission.md#fixed-wing-mission-landing).
 :::
 
-Landing is affected by the following parameters (also see [Landing (Fixed Wing)](../flying/fixed_wing_landing.md)):
+<!-- 
+After getting land command it will circle down at the current location.
+The flaring part is the same as for the mission lading.
+-->
```
hamishwillee, line 48: When the land-mode bug is fixed we might move some of the content around again. I am assuming that the abort, flaring, nudging and most of the configuration is the same.

If so, we could follow the same process as suggested for Takeoff: The introduction like "Landing sequence" is duplicated the made specific, but the abort etc stuff is in just one place and linked.

hamishwillee:
Hi @sfuhrer 

I've done two changes:
1. As discussed, moved the landing docs to mission mode doc from landing mode. Landing mode is now just a warning and link to the PR with fix. That was all done in one commit.
2. Minor rewrite of the landing sequence stuff

The changes are relatively minor, but make the sequence much more clear to me. Specifically, I have replaced terminology like "towards landing descent orbit" with "towards the loiter waypoint", and overall represented this in terms of the commands that are processed.

Merging, because IMO the sequence was the only thing that felt a bit unclear, and now it does not. Please check as a post process.

NEXT STEPS: I'll ping you on discord

PR #2246, opened 2023-01-20
Fix dependancy list and typos

svpcom:
Fix dependancy list and typos

### DIFF
--- a/en/tutorials/video_streaming_wifi_broadcast.md
+++ b/en/tutorials/video_streaming_wifi_broadcast.md
@@ -58,7 +58,7 @@ If you connect it to **USB3** port via **native USB3 cable** to a **Linux laptop
 
 For **Raspberry PI** (UAV or ground) it must be directly connected to 5V BEC (or high current power adapter in case of ground pi) in one of two ways:
 
-- Make a custom USB cable ([cut `+5V` wire from USB plug and connect it to BEC])(https://electronics.stackexchange.com/questions/218500/usb-charge-and-data-separate-cables)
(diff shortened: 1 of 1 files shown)


en/tutorials/video_streaming_wifi_broadcast.md
```diff
@@ -81,7 +81,9 @@ Be aware of [ground loop](https://en.wikipedia.org/wiki/Ground_loop_%28electrici
 ### Using a Linux Laptop as GCS (Harder than using a RasPi)
 1. On **ground** Linux development computer:
    ```
-   sudo apt install libpcap-dev libsodium-dev python3-all python3-twisted
+   sudo apt install python3-all libpcap-dev libsodium-dev python3-pip python3-pyroute2 \
+            python3-future python3-twisted python3-serial iw virtualenv \
```
hamishwillee, line 85: do you really need all these? virtualenv for example is recommended if you are using a virtual env, but none of the other things here require it or mention it - it is purely optional based on developer setup.

Ditto all the rest - the only reason I ever install libpcap is for wireshark - why is it needed here? Is future needed?

I know there is no particular harm in installing more dependencies, but it does annoy people if you ask for things that are not needed. If they are all essential then great - but we might split out those that are likely to already be installed into their own line just to make that clear.

svpcom: All dependencies are needed. I've added full list because some users continuously open bug reports when unable to build project with missing dependencies.
libpcap is core dependency - it used for packet injection.
virtual env is needed to setup `stdeb` (python deb package packer) which is not exists in debian/ubuntu.

hamishwillee: Thanks. Makes sense.

PR #2256, opened 2023-01-29
Docs: Fix typo in magnetomer to magnetometer

khanasif786:
Changed magnetomer to magnetometer


hamishwillee:
Thanks very much. Unfortunately this text is generated from source code, so we can't update it in this way. I have made the needed corresponding changes in https://github.com/PX4/PX4-Autopilot/pull/21024

PR #2259, opened 2023-02-02
ROS User Guide -  XRCE-DDS update

hamishwillee:
This is the ROS2 User Guide update for XRCE-DDS which replaces #2094

A few notes:
1. It just covers the ROS2 User Guide and XRCE-Middleware page. We can grow from this.
2. The ROS2 guide is a guided path - i.e. using the simulator, and installing from standalone build. There are many other ways to do things like install the agent - I have tried to make the XRCE-DDS doc show the options.
   So you read the guide to get the big picture and XRCE-DDS to get more info.
3. This uses foxy and Ubuntu 20.04. I trust PX4 on 22.04 very little.
4. There is a bit of repetition. What can you do? Hopefully nothing to annoying.
5. The XRCE doc no longer has example code - that's all in the ROS doc, because we only have a ROS story (not "ROS and/or DDS as we did in the old architecture).
6. I don't talk about microROS. We can build to that.

I refer to "ROS 2" pretty much everywhere. Anyone mind if I use "ROS2" - it looks better to me.

PS Next step would be the offboard tutorial being updated.


en/ros/ros2_comm.md
```diff
@@ -1,64 +1,35 @@
...
+The generator uses the uORB message definitions in the source tree: [PX4-Autopilot/msg](https://github.com/PX4/PX4-Autopilot/tree/main/msg) to create the message definitions.
 
-:::note
-Astute readers will note that the generated agent might not have been built with that same set of definitions (unless they were both built of the same 'master' commit).
 
-Right now this is not a problem because the PX4 message set/definitions are relatively stable and the updated/new messages get automatically deployed to `px4_msgs`.
+ROS 2 needs to have the same message definitions that were used to create the client.
+When working with PX4 main and normal releases you can clone the [PX4/px4_msgs](https://github.com/PX4/px4_msgs) package, which contains the definitions for each release in separate branches.
+If working on a custom build you could simply copy the msg definitoins directly out of PX4 source tree.
+These definitions can then be built in any ROS2 workspace that needs them.
```
hamishwillee, line 27: FMI I see in the examples that you just include these in the workspace and build:
```
cd ~/px4_ros_xrce_dds_ws/src
git clone https://github.com/eProsima/Micro-XRCE-DDS-Agent.git
colcon build
```
1. How does the "system" know about these definitions? I.e. the agent is pre-built and the files are just msgs, so how does ROS2 know that it can use them for message definitions?
2. Do they have to be in a particular location to be built or would any files in a directory src be built?
3. What does it actually do with them - build IDE files then ... what?

Basically its clear that the agent needs to know about the messages and so does anything in ROS2 that wants to use them. What isn't clear is how this works. It looks like magic.

In previous version you could have a "generic DDS message" or "ROS2 message". My understanding is we now just have one format which matches ROS2. What does this mean for "generic DDS"? Is it possible to create a system that uses PX4 topics without ROS? Is it something we want to support?

beniaminopozzan: @hamishwillee 

> Basically its clear that the agent needs to know about the messages and so does anything in ROS2 that wants to use them.

The Agent does not need the message definitions at build time, nor at run time as they are necessary only to the ROS2 nodes (and to commands like `ros2 topic echo`) that want to subscribe and publish to them.
My way to see this is that the pair client-agent (and I'm talking about both micro XRCE-DDS Agent and microROS Agent) acts a unique ROS2 node physically distributed between PX4 and the companion computer. Therefore you need the message definitions only where you use the messages which is the client side. The Agents are just bridging the transport protocols.

> How does the "system" know about these definitions? I.e. the agent is pre-built and the files are just msgs, so how does ROS2 know that it can use them for message definitions?

On the client side https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/microdds_client/dds_topics.yaml not only specifies which uorb topics have to be exposed to ROS2, it also tells which is their message type:
```yaml
  - topic: /fmu/out/vehicle_odometry
    type: px4_msgs::msg::VehicleOdometry
```
Then ROS2 will know that it has to look for the message `px4_msgs::msg::VehicleOdometry` when it wants to interact with the topic `/fmu/out/vehicle_odometry`. The package `px4_msgs` contains a message definition named `VehicleOdometry.msg`which according to the ROS2 convention is compiled into the message named `px4_msgs::msg::VehicleOdometry`... so everything works out of the box!
You only need to take care that the files `VehicleOdometry.msg` in `px4_msgs/msg/` and the one in `PX$-Autopilot/msg/` are identical.

> Do they have to be in a particular location to be built or would any files in a directory src be built?

`colcon` build everything (unless you tell it otherwise) that resides in `src/`. Therefore as long are you have the package `px4_msgs` somewhere inside your workspace `src` folder you are fine, it will be compiled. However, it is not necessary to have it in the same workspace of the Agent: workspace can be linked together. You can create a workspace with just the message package and build it. Then you can move to another workspace and have the new one extending the previous with `source workspace_with_onl_px4_msgs/install/local_setup.bash`.

> What does it actually do with them - build IDE files then ... what?

This is going above my knowledge, when a `*.msg` file is compiled you are generating all the code needed to serialize/deserialize it, access it, the structure that contains all the message fields, ecc such that your application will be able to include the message header and create publishers and subscribers.


> My understanding is we now just have one format which matches ROS2

You are right!

> What does this mean for "generic DDS"?

Generic DDS means that your applications directly interface with the DDS network without using ROS, which is basically what the client on the PX4 side does. Everything works because ROS2 is using the exact same network.

> Is it possible to create a system that uses PX4 topics without ROS?

I guess it is, but I'm not 100% sure.

>  Is it something we want to support?

I don't know.

hamishwillee: @beniaminopozzan Thanks. So I get how the client side works - I can see how it knows about what messages to generate client code from the yaml , and it has the definitions.

However the agent doesn't see the yaml, all it sees is data. ROS2 does not need the msg to get the listing of topics, so I infer from this that it can determine message name/type from the data it receives. However it does need the definitions to echo a particular topic "detail"

The bit that is hard is this:

> You only need to take care that the files VehicleOdometry.msg in px4_msgs/msg/ ...

What I think you are saying is that as long as the messages are in a folder named "px4_msgs/msg" within a sourced workspace that ROS can see it will build them - right? 

Further that it gets this from the msg type: `px4_msgs::msg::`

And that the type comes with the data that ROS2 can interpret without the messages.

That makes sense!

beniaminopozzan: @hamishwillee of course!

> However the agent doesn't see the yaml, all it sees is data. ROS2 does not need the msg to get the listing of topics, so I infer from this that it can determine message name/type from the data it receives. However it does need the definitions to echo a particular topic "detail"

Exactly!

> What I think you are saying is that as long as the messages are in a folder named "px4_msgs/msg" within a sourced workspace that ROS can see it will build them - right?

You are not wrong, but what I meant is that the the message definitions used in ROS2 must be identical to the ones used in PX4. And indeed `PX4-bot` makes sure that this happens. But if a user decides to implement its own uorb message and to expose it to ROS2 then it has to

- update the yaml file
- copy the new message definition in "px4_msgs/msg" and recompile the workspace

en/ros/ros2_comm.md
```diff
@@ -206,6 +217,28 @@ E.g. in the case of `sensor_combined` use `ros2 topic hz /fmu/sensor_combined/ou
...
+  ```yaml
+  - topic: /fmu/out/vehicle_odometry
+    type: px4_msgs::msg::VehicleOdometry
+  ```
+
+  The ROS toolchain uses the prefix in the message type (i.e. `px4_msgs::msg::`) to determine the search path for message definitions.
+  :::
+
+
 ## ROS 2 Example Applications
```
hamishwillee, line 242: @beniaminopozzan @Jaeyoung-Lim Well, I can run run @Jaeyoung-Lim [example](https://github.com/Jaeyoung-Lim/px4-offboard/) with the XRCE-DDS agent, which is nice. Unfortunately it is not quite as good as the examples currently used by this document, which are the advertiser and listener examples in px4_ros_com https://github.com/PX4/px4_ros_com/tree/main/src/examples - in the sense that these show data in/out of ROS2/PX4 - ie. both main use cases.

The problem is that I can't get these to work with XRCE-DDS agent. Any ideas on how they might be modified to work? I don't actually understand ROS and the launch behaviour :-(. 








https://github.com/PX4/px4_ros_com/tree/main/src/examples/advertisers

beniaminopozzan: @hamishwillee for the the listener examples the issue is about the ROS topic quality of service which does not matches between publishers and subscribers. This PR https://github.com/PX4/px4_ros_com/pull/174 fixes it for them.

This is actually a good point to bring up: ROS2 default QoS https://docs.ros.org/en/foxy/Concepts/About-Quality-of-Service-Settings.html#qos-profiles is different from the one used by PX4, therefore every time a user wants to create a subscriber it need to specify the right QoS otherwise it won't be able to listen to the topics published by PX4.
Relevant merged PR about this is https://github.com/PX4/PX4-Autopilot/pull/20624

Citing ROS2 documentation:

> By default, publishers and subscriptions in ROS 2 have “keep last” for history with a queue size of 10, “reliable” for reliability, “volatile” for durability, and “system default” for liveliness. Deadline, lifespan, and lease durations are also all set to “default”.

PX4 instead uses
```cpp
uxrQoS_t qos = {
  .durability = UXR_DURABILITY_TRANSIENT_LOCAL,
  .reliability = UXR_RELIABILITY_BEST_EFFORT,
  .history = UXR_HISTORY_KEEP_LAST,
  .depth = 0,
};
```
for publishers, and
```c++
uxrQoS_t qos = {
  .durability = UXR_DURABILITY_VOLATILE,
  .reliability = UXR_RELIABILITY_BEST_EFFORT,
  .history = UXR_HISTORY_KEEP_LAST,
  .depth = queue_depth,
};
```
for subscribers, which corresponds to the ROS2 predefined QoS _sensor data_ (the one that the PR chooses). Not all combinations publisher-subscriber are possible: https://docs.ros.org/en/foxy/Concepts/About-Quality-of-Service-Settings.html#qos-compatibilities .

en/middleware/xrce_dds.md
```diff
@@ -0,0 +1,469 @@
+# XRCE-DDS: PX4-FastDDS Bridge
```
hamishwillee, line 1: At moment just a scratchpad of topics I'll need.

en/ros/ros2_comm.md
```diff
@@ -1,218 +1,306 @@
...
+   nav_state: 4
+   failure_detector_status: 0
+   hil_state: 0
+   ...
+   ---
+   ```
+
+## ROS 2 Example Applications
+
+### ROS 2 Listener
```
hamishwillee, line 220: This depends on the qos and screen output fixes being made to the exmaple.

en/middleware/xrce_dds.md
```diff
@@ -0,0 +1,303 @@
...
+:::
+
+PX4 uses XRCE-DDS middlware to allow [uORB messages](../middleware/uorb.md) to be published and subscribed on a companion computer as though they were ROS 2 topics.
+This provides a fast and reliable integration between PX4 and ROS2, and makes it much easier for ROS 2 applications to get vehicle information and send commands. 
+
+PX4 uses an XRCE-DDS implementation that leverages [eProsima Micro XRCE-DDS](https://micro-xrce-dds.docs.eprosima.com/en/stable/introduction.html).
+
+The following guide describes the architecture and various options for setting up the client and agent.
+In particular it covers the options that are most important to PX4 users.
+
```
hamishwillee, line 14: FYI Point  above is that this is not an exhaustive guide - for that you see the eProsima guide. But we don't do everything the architecture allows - e.g. connection over CAN, so this is "PX4s take on the infrastructure".

en/middleware/xrce_dds.md
```diff
@@ -0,0 +1,303 @@
...
+
+## Architecture
+
+The XRCE-DDS middleware consists of a client running on PX4 and an agent running on the companion computer, with bi-directional data exchange between them over a serial, UDP, TCP or custom link.
```
hamishwillee, line 18: Do we support these options:  serial, UDP, TCP ? What about CAN and the other links that eProsima say they support?

en/middleware/xrce_dds.md
```diff
@@ -0,0 +1,303 @@
...
+
+To start the agent with settings for connecting to the XRCE-DDS client running on the simulator (note that the command name is different than if you build the agent locally):
+
+```sh
+micro-xrce-dds-agent udp4 -p 8888
+```
+
+:::note
+At time of writing the stable of version installed from snap connects to PX4 but reports errors creating topics.
+The development version, fetched using `--edge` above, does work.
```
hamishwillee, line 100: FYI - `stable` doesn't work for me.

en/middleware/xrce_dds.md
```diff
@@ -0,0 +1,303 @@
...
+:::
+
+### Build/Run within ROS2 Workspace
+
+The agent can be built and launched within a ROS2 workspace (or build standalone and launched from a workspace.
+You must already have installed ROS 2 following the instructions in: [ROS 2 User Guide > Install ROS 2](##TBDXXX)
```
hamishwillee, line 106: - [ ] Need to fix up this link

en/middleware/xrce_dds.md
```diff
@@ -0,0 +1,303 @@
...
+These are documented in the eProsima user guide: [Micro XRCE-DDS Agent > Agent CLI](https://micro-xrce-dds.docs.eprosima.com/en/latest/agent.html#agent-cli). <!-- what options work? i.e. does PX4 allow CAN? -->
+
+:::note
+You should create a single instance of the agent for each channel over which you need to connect.
+:::
+
+For example, the PX4 simulator runs the XRCE-DDS client over UDP on port 8888, so to connect to the simulator you would start the agent with the command:
+
+```sh
+MicroXRCEAgent udp4 -p 8888
```
hamishwillee, line 161: This is all a bit "thin". Can't really update until we have a working companion setup guide/example.

en/middleware/xrce_dds.md
```diff
@@ -0,0 +1,303 @@
...
+    type: px4_msgs::msg::VehicleOdometry
+  ```
+  :::
+
+
+### PX4 ROS2 QoS Settings
+
+PX4 QoS settings for publishers are incompatible with the default QoS settings for ROS2 subscribers.
+So if ROS2 code needs to subscribe to a uORB topic, it will need to use compatible QoS settings.
+One example of which is shown in [ROS 2 User Guide > ROS2 Subscriber QoS Settings]()
```
hamishwillee, line 217: - [ ] Fix link

en/middleware/xrce_dds.md
```diff
@@ -0,0 +1,303 @@
...
+## Helpful Resources
+
+- [ROS2 in PX4: ROS2 in PX4: Technical Details of a Seamless Transition to XRCE-DDS](https://www.youtube.com/watch?v=F5oelooT67E) - Pablo Garrido & Nuno Marques (youtube)
+- [PX4 ROS2 offboard tutorial](https://gist.github.com/julianoes/adbf76408663829cd9aed8d14c88fa29) (Github gist - JulianOes)
+- [ROS2 PX4 Offboard Tutorial](https://github.com/Jaeyoung-Lim/px4-offboard/blob/2d784532fd323505ac8a6e53bb70145600d367c4/doc/ROS2_PX4_Offboard_Tutorial.md) (Jaeyoung-Lim).
+
+
+
+<!---
+Some of this might be useful.
```
hamishwillee, line 254: Below here is stuff from the old architecture on uart config. I've left here - but probably would like in a "hardware tutorial".

en/ros/ros2_comm.md
```diff
@@ -348,116 +428,147 @@ int main(int argc, char *argv[])
...
+  Technically the messages must be in a folder defined by the type prefix in the yaml file:
+  As you can see below, this is `px4_msgs::msg::`:
+  
+  ```yaml
+  - topic: /fmu/out/vehicle_odometry
+    type: px4_msgs::msg::VehicleOdometry
+  ```
+  :::
+  
+## ros2 CLI
```
hamishwillee, line 462: @beniaminopozzan FYI, I thought it useful to have the ros2 cli information here, where people might discover it. I think it is useful because PX4 users might not know about these, and they are a quick way of working out if topics are published (perhaps I should say that!)

ROS people might find them useful because they make clear you need the px4_msgs to do much other than list the topics.

en/ros/ros2_comm.md
```diff
@@ -259,6 +331,12 @@ public:
 	}
 ```
 
+:::note
```
hamishwillee, line 334: @beniaminopozzan This highlights the subscriber code. Actual docs further down.

en/ros/ros2_comm.md
```diff
@@ -1,222 +1,292 @@
...
+The topics that you can use are defined in [dds_topics.yaml](https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/microdds_client/dds_topics.yaml), and you can get more information about their data in the [uORB Message Reference](../msg_docs/README.md).
+For example, [vehicle_global_position](../msg_docs/vehicle_global_position.md) can be used to get the vehicle global position, while [vehicle_command](../msg_docs/vehicle_command.md) can be used to command actions such as takeoff and land.
+
+The [ROS 2 Example applications](#ros-2-example-applications) examples below provide concrete examples of how to use these topics.
+
+## Compatibility Issues
+
+This section contains information that may affect how you write your ROS code.
+
+### ROS2 Subscriber QoS Settings
```
hamishwillee, line 256: @beniaminopozzan And this is the main qos settings topic. It isn't quite as good as yours, but it is more succinct and instructional - i.e. do this, it should just work. I ran out of puff - we can edit if you think it insufficient.

en/middleware/xrce_dds.md
```diff
@@ -0,0 +1,303 @@
+# XRCE-DDS (PX4-FastDDS Bridge)
+
+:::note
+XRCE-DDS replaces the [Fast-RTPS Bridge](https://docs.px4.io/v1.13/en/middleware/micrortps.html#rtps-dds-interface-px4-fast-rtps-dds-bridge) used in PX4 v1.13.
+:::
+
+PX4 uses XRCE-DDS middlware to allow [uORB messages](../middleware/uorb.md) to be published and subscribed on a companion computer as though they were ROS 2 topics.
```
hamishwillee, line 7:
```suggestion
PX4 uses XRCE-DDS middleware to allow [uORB messages](../middleware/uorb.md) to be published and subscribed on a companion computer as though they were ROS 2 topics.
```

en/middleware/xrce_dds.md
beniaminopozzan: @hamishwillee PX4 supports serial and UDP on the client side at the moment.

en/middleware/xrce_dds.md
```diff
@@ -0,0 +1,303 @@
...
+  As you can see below, this is `px4_msgs::msg::`:
+  
+  ```yaml
+  - topic: /fmu/out/vehicle_odometry
+    type: px4_msgs::msg::VehicleOdometry
+  ```
+  :::
+
+
+### PX4 ROS2 QoS Settings
```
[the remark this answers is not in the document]
beniaminopozzan: No, I agree that it is convenient to a quick reference on the QoS details used by the client!

en/ros/ros2_comm.md
beniaminopozzan: As you said, most of the users will be satisfied by that, it get the job done!

en/middleware/xrce_dds.md
```diff
@@ -0,0 +1,303 @@
...
+You should create a single instance of the agent for each channel over which you need to connect.
+:::
+
+For example, the PX4 simulator runs the XRCE-DDS client over UDP on port 8888, so to connect to the simulator you would start the agent with the command:
+
+```sh
+MicroXRCEAgent udp4 -p 8888
+```
+
+### Starting the Client
```
beniaminopozzan, line 164: There is one last method to start the client on real vehicles: using the parameter `XRCE_DDS_0_CFG` of the group [Micro XRCE-DDS](https://docs.px4.io/main/en/advanced_config/parameter_reference.html#micro-xrce-dds). With that you can start the client on serial or Ethernet or even WiFi (it seems). Then you have other parameters, like `XRCE_DDS_UDP_PRT` to setup the UDP port, `XRCE_DDS_DOM_ID` if you want to connect to a different `DOMAIN_ID` (https://docs.ros.org/en/foxy/Concepts/About-Domain-ID.html) and finally `XRCE_DDS_KEY` if you want to change the default `SESSION_ID` (needed for multi vehicle simulations). Anyway, I think the last two parameters can be put in the docs later on.

en/middleware/xrce_dds.md
hamishwillee:
```suggestion
The XRCE-DDS middleware consists of a client running on PX4 and an agent running on the companion computer, with bi-directional data exchange between them over a serial or UDP link.
```

en/ros/ros2_comm.md
```diff
@@ -348,116 +428,147 @@ int main(int argc, char *argv[])
...
+  - topic: /fmu/out/vehicle_odometry
+    type: px4_msgs::msg::VehicleOdometry
+  ```
+  :::
+  
+## ros2 CLI
 
-2. Build the workspace:
+The [ros2 CLI](https://docs.ros.org/en/foxy/Tutorials/Beginner-CLI-Tools.html) is a useful tool for working with ROS.
+You can use it, for example, to inspect topics, and launch complex ROS systems.
```
hamishwillee, line 465:
```suggestion
You can use it, for example, to quickly check whether topics are being published, and also inspect them in detail if you have `px4_msg` in the workspace. The command also lets you launch more complex ROS systems via a launch file.
```

en/middleware/xrce_dds.md
hamishwillee:
```suggestion
You must already have installed ROS 2 following the instructions in: [ROS 2 User Guide > Install ROS 2](../ros/ros2_comm.md#install-ros-2).
```

en/middleware/xrce_dds.md
hamishwillee:
```suggestion
One example of which is shown in [ROS 2 User Guide > ROS2 Subscriber QoS Settings](../ros/ros2_comm.md#ros2-subscriber-qos-settings).
```

en/middleware/xrce_dds.md
```diff
@@ -0,0 +1,335 @@
...
+
+```sh
+MicroXRCEAgent udp4 -p 8888
+```
+
+On RaspberryPi, if you're using a serial link and connected to the IO pins, you might connect using this command:
+
+```sh
+sudo MicroXRCEAgent serial --dev /dev/AMA0 -b 921600
+```
```
hamishwillee, line 168: @beniaminopozzan I added this - seem correct to you? Based on discussion in https://github.com/PX4/PX4-Autopilot/issues/21120#issuecomment-1428347091

en/middleware/xrce_dds.md
hamishwillee: @beniaminopozzan Thanks! Actually you should ALWAYS use the parameters on hardware. If you set the command you're setting yourself up for conflicts with other configuration. I have mentioned it still, but not as recommended.

I have rewritten this section, can you check?

en/middleware/xrce_dds.md
```diff
@@ -0,0 +1,335 @@
...
+  - [SER_TEL2_BAUD](../advanced_config/parameter_reference.md#SER_TEL2_BAUD), [SER_URT6_BAUD](../advanced_config/parameter_reference.md#SER_URT6_BAUD) (and so on):
+    If using a serial connection, the baud rate must be set using the `_BAUD` parameter associated with the serial port.
+    For example, you'd set a value for `SER_TEL2_BAUD` if you are connecting to the companion using `TELEM2`.
+    For more information see [Serial port configuration](../peripherals/serial_configuration.md#serial-port-configuration).
+
+  Some setups might also need these parameters to be set:
+
+  - [XRCE_DDS_KEY](../advanced_config/parameter_reference.md#XRCE_DDS_KEY): The XRCE-DDS key.
+    If you're working in a multi-client, single agent configuration, each client should have a unique key.
+  - [XRCE_DDS_DOM_ID](../advanced_config/parameter_reference.md#XRCE_DDS_DOM_ID): TBD.
```
hamishwillee, line 195: TBD could be added as a post process - but what is it for/how/when should it be set?

hamishwillee:
@beniaminopozzan I think I have addressed your concerns/points. Good enough to merge? (modulo the highlighted parts)?

@TSC21 Want to comment?

en/middleware/xrce_dds.md
beniaminopozzan: I didn't have time to go through all those  "issue" topics on the PX4 repository yet. In particular I'm not 100% sure that /dev/AMA0 are the TX RX pins of the raspberry. And if you are using Raspberry Pi 4, which has multiple serial ports things gets more complicated. I'll get back to you as soon as possible

en/middleware/xrce_dds.md
beniaminopozzan: - `XRCE_DDS_KEY` sets the `key` value of the client _session_. All clients connected to the same agent must have different keys (see https://micro-xrce-dds.docs.eprosima.com/en/stable/client_api.html#session , `uxr_init_session`). Moreover the key cannot have value zero. It is important to set this parameter only for multi vehicle setups and, more importantly, multi vehicle simulations, where all clients are connected in UDP to the same agent (multipla agents operating on different ports could be used but I don't se any advantages in doing that)

- `XRCE_DDS_DOM_ID` is the equivalent of the [`ROS_DOMAIN_ID`](https://docs.ros.org/en/foxy/Concepts/About-Domain-ID.html) and gives logical separation between DDS networks. Therefore clients configured with different `XRCE_DDS_DOM_ID` will belong to different networks. By default, ROS2 operates on ID 0

en/middleware/xrce_dds.md
beniaminopozzan: Thanks!!!

For serial connection I think it is perfect, for Ethernet connection it seems that @ozitoune managed to have it working on Pixhawk 6X with
```
microdds_client start -t udp -p 8888 -h 192.168.0.100
```
where `-h 192.168.0.100` tells the IP address of the _Agent_. As this is not a parameter in PX4, it is then required to start the client manually or from System Startup.

beniaminopozzan:
@hamishwillee absolutely! Thank you! I believe that at this point anyone that reads this docs can start using the bridge painlessly, test it and provide feedback :)

en/middleware/xrce_dds.md
ViniciusAbrao: Hi, I have a step-by-step tutorial that describe this raspberry setup. Maybe it can help somehow:

https://github.com/ViniciusAbrao/px4_ros2_xrcedds/blob/master/raspiberrypi_setup.pdf

hamishwillee: So the short answer is "Yes" we should link to the companion computer setup, and either expand it, or possibly have a RaPI setup page below it. This needs to cross link to the official RaPi docs.

I will think about minimal amount to put in here on Wednesday and merge this. Vince, I am sure your doc will be useful - I'll check it out Weds.

@beniaminopozzan Thanks so much for your help!

en/middleware/xrce_dds.md
hamishwillee: @beniaminopozzan That's horrible - can you create an issue to get a parameter for the host into PX4?

en/middleware/xrce_dds.md
```diff
@@ -205,17 +217,21 @@ To use these ports you must first disable the existing configuration:
...
 
 ```sh
-microdds_client start -t serial -d /dev/ttyS3 -b 921600
+microdds_client start -t udp -p 8888 -h 192.168.0.100
 ```
 
+:::note
+At time of writing there is no PX4 parameter for setting the address of the remote host on which the XRCE-DDS agent is running.
+Therefore for Ethernet connections you will _have_ to use a command to start the client.
+:::
```
hamishwillee, line 234: FYI @beniaminopozzan This "fixes" the docs to explain you need to use command line for Ethernet, as discussed in https://github.com/PX4/PX4-user_guide/pull/2259/commits/986995c6a4309cc496d17c67659a336ded449045#r1108235444

If you give me an issue for creating the host parameter for pointing to the agent I'll add that as a link from "at the time of writing" in the note.

en/middleware/xrce_dds.md
```diff
@@ -161,12 +163,18 @@ For example, the PX4 simulator runs the XRCE-DDS client over UDP on port 8888, s
...
+When working with real hardware, the setup depends on the hardware, OS, and channel.
+For example, if you're using the RasPi `UART0` serial port, you might connect using this command (based on the information in [Raspberry Pi Documentation > Configuring UARTS](https://www.raspberrypi.com/documentation/computers/configuration.html#configuring-uarts)):
 
 ```sh
 sudo MicroXRCEAgent serial --dev /dev/AMA0 -b 921600
 ```
 
+:::note
+For more information about setting up communications channels see [Pixhawk + Companion Setup > Serial Port setup](../companion_computer/pixhawk_companion.md#serial-port-setup), and sub-documents.
+:::
```
hamishwillee, line 175: @beniaminopozzan @ViniciusAbrao This generalizes the info about agent startup, making it clear that "it depends", linking to the RaPi configuring UARTs to help people understand a bit more about this case.
I also link, as suggested, to the companion computers page. That has some general Linux setup.

@ViniciusAbrao What would be great is if you could add your setup information as a topic below https://docs.px4.io/main/en/companion_computer/pixhawk_companion.html 

If this can be general it would be "Raspberry Pi Companion". If it has to be specific to a particular RaPi then ""Raspberry Pi Xxxx Companion", where Xxxx is the version. I think it can be general, but we'd need notes to highlight things to check in a particular system.

What would be great is more than just this specific part of setup around ROS. RasPi is very popular, so would be good to have images of the hardware connection showing the lines. Explanation of how it is powered, and why done that way, etc.
The setup of the connection could then be part of this.

Does it make sense/is it possible?

en/ros/ros2.md
```diff
@@ -1,27 +1,28 @@
...
+- [ROS 2 Offboard Control Example](../ros/ros2_offboard_control.md)
 
 :::note
 ROS 2 is officially supported only on Linux platforms.
 Ubuntu 20.04 LTS is the official supported distribution.
 :::
 
 :::note
-To use the [PX4-ROS 2 bridge](../ros/ros2_comm.md) effectively you must (at time of writing) have a reasonable understanding of the PX4 internal architecture and conventions.
+To use the [ROS2](../ros/ros2_comm.md) effectively you must (at time of writing) have a reasonable understanding of the PX4 internal architecture and conventions.
```
hamishwillee, line 25: @beniaminopozzan This statement (and the ones below it, not all of which are visible here) states that you need know a bit about PX4 conventions to use ROS2 properly. This was added when ROS 2 first came out because at that time MAVROS did all the conversions for conventions - e.g. FRD to FRU and visa versa between ROS2 and PX4.

Is this still true/an issue or are conventions now taken care of for ROS 2 users?

Assuming this is a problem, do we know what the issues are? If we're not going to provide translation layers we should document the different conventions.

hamishwillee:
@beniaminopozzan @ViniciusAbrao @TSC21 @mrpollo This is good enough to go in. I am merging it and will kick off a new one to start cleaning up some more bits.

I'm going to start with general cleanup, but we also need to update https://docs.px4.io/main/en/ros/ros2_offboard_control.html

mrpollo:
Thanks for leading the charge here @hamishwillee its a major milestone 🎉

hamishwillee:
Honestly, you should clone @beniaminopozzan - endlessly supportive and knowledgeable.

en/middleware/xrce_dds.md
beniaminopozzan: Here it is!
https://github.com/PX4/PX4-Autopilot/issues/21179

en/middleware/xrce_dds.md
beniaminopozzan: It definitively makes sense. The linked docs covers only for USB-to-serial adapter but most (not only Raspberry Pi) companion computers have hardware serial devices.

en/ros/ros2.md
beniaminopozzan: The statement still holds. In details, most of the uorb messages assume implicitly that PX4 conventions are used (after all, they where not supposed to be exposed to external components such as ROS nodes), some of them, which are quite relevant are https://github.com/PX4/PX4-Autopilot/blob/main/msg/TrajectorySetpoint.msg, https://github.com/PX4/PX4-Autopilot/blob/main/msg/VehicleLocalPositionSetpoint.msg.
Therefore it is necessary that the user converts properly the frames before using them (like taking off sending negative z-position setpoints)

On the other hand, at least https://github.com/PX4/PX4-Autopilot/blob/main/msg/VehicleOdometry.msg (and all the topics that uses this message) allows for different frames. Sadly, ROS2 conventions are still not supported, but there is a draft PR to fix that: https://github.com/PX4/PX4-Autopilot/pull/20737. Nevertheless, this will allow to send data in ROS2 convention to PX4, but PX4 will always reply using PX4 convention.

> Assuming this is a problem, do we know what the issues are?

I wouldn't call this a problem, it's more a _fundamental difference_ that has to be clarified :)
On the implementation POV, it is quite straightforward once you know which rotation to apply. Moreover, lets assume you have a ROS node, something completely unrelated to PX4, that publishes position setpoints in ROS frames using ROS standard messages. Regardless the frames mismatch, you still need to convert from standard ROS messages to PX4 messages, therefore an additional ROS node is needed, that node can also apply the rotations.

I would be happy to work on the docs for the different conventions (starting from https://docs.px4.io/main/en/tutorials/motion-capture.html#coordinate-frames)!

hamishwillee: > I would be happy to work on the docs for the different conventions (starting from https://docs.px4.io/main/en/tutorials/motion-capture.html#coordinate-frames)!

@beniaminopozzan That would be great. I think this information should be in the ROS2 User Guide because that is where  people will get a first view on conventions.
Possibly as a sub topic of [compatibility issues](https://docs.px4.io/main/en/ros/ros2_comm.html#compatibility-issues). Up to you.

I think having frames information would be good. Perhaps we should merge https://docs.px4.io/main/en/tutorials/motion-capture.html into https://docs.px4.io/main/en/ros/external_position_estimation.html then pull the https://docs.px4.io/main/en/ros/external_position_estimation.html#reference-frames-and-ros section into the ROS2 user guide and update it to be ROS2 rather than ROS1/MAVROS centric. Again, that's the direction I'd follow, but you're the expert.

en/middleware/xrce_dds.md
hamishwillee: Yes, it's a simplistic topic. In an ideal world you'd start with a doc that covers all connection scenarios for as much hardware and OS as possible. That requires someone with broad experience commit quite a lot of time. Getting support for that is hard.

So the way I approach these problems is to attempt to build out specific hardware scenarios, then generalize them.
So for example here if @ViniciusAbrao can create a RaPi Companion doc covering as much as possible about the connections then that might contain information about non-USB serial connections, which can then migrate to the parent doc. 

Thats my thinking, anyway.

From that we do things like potentially cross link to https://docs.px4.io/main/en/companion_computer/holybro_pixhawk_rpi_cm4_baseboard.html with more software setup instructions. The idea being that we make setup of any RaPi setup as seamless as possible. 

@ViniciusAbrao You onboard to create that doc? RaPi is a very popular option, so I think a "good start". One thing no one seems to cover on companion setups, at least in detail, is how to power the RaPi.

en/middleware/xrce_dds.md
hamishwillee: Thanks!!! I'll add to docs.

en/ros/ros2.md
beniaminopozzan: > I think this information should be in the ROS2 User Guide because that is where people will get a first view on conventions.

Yes, definitively!

> I think having frames information would be good. Perhaps we should merge https://docs.px4.io/main/en/tutorials/motion-capture.html into https://docs.px4.io/main/en/ros/external_position_estimation.html then pull the https://docs.px4.io/main/en/ros/external_position_estimation.html#reference-frames-and-ros section into the ROS2 user guide and update it to be ROS2 rather than ROS1/MAVROS centric. Again, that's the direction I'd follow, but you're the expert.

Yes, definitively x2!
And there is also https://docs.px4.io/main/en/advanced_config/tuning_the_ecl_ekf.html#external-vision-system :) !

en/middleware/xrce_dds.md
ViniciusAbrao: Yes @hamishwillee ... I am going to create the doc covering the connections and the setup. I let you know when I finish it !!

en/ros/ros2.md
hamishwillee: :-)

W.r.t. https://docs.px4.io/main/en/advanced_config/tuning_the_ecl_ekf.html#external-vision-system we should leave that there because it is the central place for all things EKF2. BUT we should also cross link to it from everywhere that needs information like this.

I'll keep an eye out for your PR on this! Do you need reminders?

en/middleware/xrce_dds.md
hamishwillee: Thanks @ViniciusAbrao - awesome!!!! Any kind of ETA?

ViniciusAbrao: > Thanks @ViniciusAbrao - awesome!!!! Any kind of ETA?

I am going to work on it until next week !!

en/ros/ros2.md
beniaminopozzan: Of course!

> Do you need reminders?

No thanks! I just have to find the time to work on it :)

en/middleware/xrce_dds.md
ViniciusAbrao: @hamishwillee I have finished the review. Please, find the PR at:

https://github.com/PX4/PX4-user_guide/pull/2318#issuecomment-1451006165

If it is not the best way to push it, please let me know. Also, maybe the English can be improved, so, feel free to edit with corrections or technical changes.

hamishwillee: Thanks @ViniciusAbrao - I've had a first look at it and asked some questions. Next week I'll look at your answers, restructure it to be more easy to extend later (for example if we add information about "how to connect via Ethernet") and link it in.

Would be good if you could push it as a branch other than main in the PR sometime in the next week - that makes editing much easier with my knowledge of git.

ViniciusAbrao: Thanks @hamishwillee The new PR is done at:

https://github.com/PX4/PX4-user_guide/pull/2323

I answered your questions... please, let me know if some points still not clear.

PR #2287, opened 2023-02-22
Micrortps cleanup1

hamishwillee:
REmoveal of 
- micrortps_manual_code_generation doc
- Reduce micrortps to a stub pointing to old versions of docs - also remove from sidebar - you get to it if you jump versions or if you see a link that we still have to remove.
- Remove installing fastrtps doc - put a comment that we don't cover it because it is now done automatically as part of everything else.
- A bunch of notes about other things that need testing.

FYI @beniaminopozzan  - nothing to do, just making you aware.


beniaminopozzan:
Thanks you @hamishwillee !
The

> **This section is out of date!**

are actually very helpful reminders for what has to be updated next!

PR #2318, opened 2023-03-01
pixhawk_rpi included

ViniciusAbrao:
Tutorial of how to setup the RPi with Pixhawk.


ViniciusAbrao:
@hamishwillee I have opened this PR to include the tutorial of how to setup the RPi with Pixhawk.

hamishwillee:
Excellent start - thank you.
1. Would it be possible for you to create a branch and add that to the PR instead of `ViniciusAbrao:main` - so for example
```
git checkout -b pr_rpi_pixhawk
git push upstream pr_rpi_pixhawk #assuming this repo is 'upstream'
```
Then replace main with pr_rpi_pixhawk 

Why? It make it much easier for me to submit changes to your PR.

en/companion_computer/pixhawk_rpi.md
```diff
@@ -0,0 +1,259 @@
...
+```
+sudo apt -y install gcc-arm-none-eabi make stlink-tools
+```
+
+Clone the last PX4 version and run the script setup:
+
+```
+git clone https://github.com/PX4/PX4-Autopilot.git --recursive 
+PX4-Autopilot/Tools/setup/ubuntu.sh
+```
```
hamishwillee, line 33: This is all basically building PX4 on Ubuntu Linux right? 
1. PX4 should be build using a specific build toolchain, as per https://github.com/PX4/PX4-Autopilot/blob/main/Tools/setup/ubuntu.sh#L162 - that's basically a link to http://docs.px4.io/main/en/dev_setup/dev_env_linux_ubuntu.html#simulation-and-nuttx-pixhawk-targets
2. What are stlink-tools for?

en/companion_computer/pixhawk_rpi.md
```diff
@@ -0,0 +1,259 @@
...
+
+Build it on the Pixhawk. On Linux, the default name of a USB connection is `/dev/ttyACM0`:
```
hamishwillee, line 35: This reads as "Build PX4 on Pixhawk" - you don't mean that right?

Re "On Linux, the default name of a USB connection is `/dev/ttyACM0`:" and then `sudo chmod a+rw /dev/ttyACM0`.

Why are you doing this? I mean on my laptop I just connect the USB port to PX4 and attach QGC and it magically works.
Is this required because the USB port might be doing something else on your laptop?

en/companion_computer/pixhawk_rpi.md
```diff
@@ -0,0 +1,259 @@
...
+
+Open the raspi-config:
+
+```
+sudo raspi-config
+```
+
+Go to the **Interface Option** and then click at **Serial Port**. Select **No** to disable serial login shell and select **Yes** to enable serial interface. Click **finish** and restart the RPi.
+
+In the sequence, in the RPi's terminal:
```
hamishwillee, line 69: Do you mean "follow these commands in RPi;s terminal, in the following order"?

en/companion_computer/pixhawk_rpi.md
```diff
@@ -0,0 +1,259 @@
...
+```
+
+Then, **ctrl+x**, **ctrl+y**, **Enter** and restart the RPi.
+
+To check if the serial port is available, in the RPi's terminal, confirm the list of serial devices by typing:
+
+```
+cd /
+ls /dev/ttyAMA0
+```
```
hamishwillee, line 89: If possible, what would this list look like?

en/companion_computer/pixhawk_rpi.md
```diff
@@ -0,0 +1,259 @@
...
+sudo chmod a+rw /dev/ttyACM0
+sudo mavproxy.py --master=/dev/ttyACM0 --baudrate 57600
+```
+
+The steps above are important to double check that the hardware communication is done. In the sequence, we are going to move to the XRCE_DDS communication.
+
+## ROS setup on RPi
+
+To install ROS 2 Humble you can follow the official tutorial, available in the website:
```
hamishwillee, line 155: I will probably replace this with link to http://docs.px4.io/main/en/ros/ros2_comm.html (or at least sections in it).

en/companion_computer/pixhawk_rpi.md
```diff
@@ -0,0 +1,259 @@
...
+sudo chmod a+rw /dev/ttyACM0
+sudo mavproxy.py --master=/dev/ttyACM0 --baudrate 57600
+```
+This is a particular setup when we are looking for both offboard control, via XRCE_DDS with the RPi, and simultaneously to keep monitoring the drone via MAVlink with QGroundControl.
+Since the Pixhawk is connected with the RPi, a remote control via QGroundControl (in a ground station laptop) can be done using SSH and the mavros package, running in a new terminal:
+
+```
+ssh -C -Y user@raspberrypi_IP
+sudo chmod a+rw /dev/ttyACM0 
+roslaunch mavros px4.launch fcu_url:=/dev/ttyACM0 gcs_url:=udp://@laptop_IP
```
hamishwillee, line 256: Why MAVROS? Is this because you want to use ROS1? It's fine if so, I just want to understand the setup.

en/companion_computer/pixhawk_rpi.md
```diff
@@ -0,0 +1,259 @@
...
+
+```
+sudo apt install python3-pip
+sudo pip3 install mavproxy
+sudo apt remove modemmanager
+```
+
+## MAVlink communication
+
+Connect the Pixhawk with the laptop via `USB` cable and change the following parameters in QGroundControl:
```
hamishwillee, line 103: So basically you're setting up the Pixhawk to use Telem1 for a telemetry radio, Telem2 for RaPi?

en/companion_computer/pixhawk_rpi.md
```diff
@@ -0,0 +1,259 @@
...
+
+It is important to identify the correct pins in the hardware. The image illustrates the `TELEM2` pins in the Pixhawk-6C Model.
+
+<img src="../../assets/companion_computer/pixhawk_rpi/pixhawk_pins.png" width="914" height="360" /> 
+
+Ps.: The pins are numbered according to the following:
+
+<img src="../../assets/companion_computer/pixhawk_rpi/pins_numbers.png" width="615" height="232" /> 
+
+The following image illustrates the correspondent pins at the RPi GPIO.
```
hamishwillee, line 127: Which RPi is this? I.e. I know there are a few RaPi's and these instructions will describe a particular config "mainly". But which one am I looking at?

I pretty much understand PX4 and Pixhawk. I know nothing much about RaPi, so this is where the holes are for me.

en/companion_computer/pixhawk_rpi.md
```diff
@@ -0,0 +1,259 @@
...
+
+<img src="../../assets/companion_computer/pixhawk_rpi/rpi_gpio.png" width="845" height="457" /> 
+
+To supply the Pixhawk with 5 Volts is possible to follow with the connection of `POWER1` port with a LiPO battery, via Power Module (PM02, for example). Another option, for this setup moment, is to supply the 5 Volts via the connection of the Pixhawk's `USB` with a smartphone charger, for example, or by keeping it in the laptop `USB`. 
+
+Also, for this setup moment, the RPi is charged with 5 Volts via Power Supply (external source/charger).
```
hamishwillee, line 133: Fair enough. What are you thinking for the real drone? 5V BEC? Separate battery?

en/companion_computer/pixhawk_rpi.md
hamishwillee: Or to put it another way, there should be no need to build PX unless you want to change the messages supported. Generally you could select the "main/master" firmware as in QGC right? http://docs.px4.io/main/en/config/firmware.html#installing-px4-master-beta-or-custom-firmware

But I'm still interested in why you would want stlink-tools :-)

ViniciusAbrao: > This is all basically building PX4 on Ubuntu Linux right?

 yes 

> 1. PX4 should be build using a specific build toolchain, as per https://github.com/PX4/PX4-Autopilot/blob/main/Tools/setup/ubuntu.sh#L162 - that's basically a link to http://docs.px4.io/main/en/dev_setup/dev_env_linux_ubuntu.html#simulation-and-nuttx-pixhawk-targets

I prefer to write here most of the steps, when possible, instead of links, so the user can follow all in the sequence. If you prefer I can change it for the link, but if possible, I can complement. 
What did I miss? Should it be:

PX4-Autopilot/Tools/setup/ubuntu.sh --no-sim-tools (since nuttx by default is installed?)

> 2. What are stlink-tools for?

I look I long time for it when doing the build. I am not sure why it is necessary, but I found in the PX4 forum that it is required, so after install it works.

en/companion_computer/pixhawk_rpi.md
ViniciusAbrao: > This reads as "Build PX4 on Pixhawk" - you don't mean that right?
 
Yes, I mean that. 

> Re "On Linux, the default name of a USB connection is `/dev/ttyACM0`:" and then `sudo chmod a+rw /dev/ttyACM0`.
> 
> Why are you doing this? I mean on my laptop I just connect the USB port to PX4 and attach QGC and it magically works. Is this required because the USB port might be doing something else on your laptop?

To build the PX4 on the Pixhawk we need to give permission to write on it .... Maybe in your Ubuntu it is the default, but usually I need to change that permission when working with USB devices in Ubuntu (PS.: the same with Arduino for example).

en/companion_computer/pixhawk_rpi.md
ViniciusAbrao: > Do you mean "follow these commands in RPi;s terminal, in the following order"?

No ... the "sudo raspi-config" will pop-up a window, like a GUI, in which you need to go and click, select and do everything.

en/companion_computer/pixhawk_rpi.md
ViniciusAbrao: > If possible, what would this list look like?

In this case, the answer is just "/dev/ttyAMA0".

en/companion_computer/pixhawk_rpi.md
ViniciusAbrao: If you see the section "Install ROS 2", 

http://docs.px4.io/main/en/ros/ros2_comm.html#install-ros-2

there is a link to:

https://docs.ros.org/en/foxy/Installation/Ubuntu-Install-Debians.html.

I only change it directly to the Humble link, instead of Foxy.

en/companion_computer/pixhawk_rpi.md
ViniciusAbrao: > Why MAVROS? Is this because you want to use ROS1? It's fine if so, I just want to understand the setup.

I did not find in the past another way to do this bridge in the same time from the Pixhawk to the RPi and to the Laptop. This is important because the offboard control is on the RPi, but when We are developing new codes, is very important to have multiple controls. So the QGroundStation, in my laptop, can send commands to land, for example, if something goes wrong with my offboard code. 

In this context, mavros have been the best choice to setup this bridge. Since this is a very common need for the users, I prefer to explain it in the end of the Tutorial.

en/companion_computer/pixhawk_rpi.md
ViniciusAbrao: > So basically you're setting up the Pixhawk to use Telem1 for a telemetry radio, Telem2 for RaPi?

Yes... We can leave the Telem1 for radio, Telem2 for mavlink (with RPi) and also USB to mavlink (with laptop in the setup moment or even with RPi, simultaneously latter in case of both mavlink and xrce_dds).

 At his moment of the tutorial we are building and testing the wire connections. So TELEM2 is going to be used to it.

en/companion_computer/pixhawk_rpi.md
ViniciusAbrao: > Which RPi is this? I.e. I know there are a few RaPi's and these instructions will describe a particular config "mainly". But which one am I looking at?
> 
> I pretty much understand PX4 and Pixhawk. I know nothing much about RaPi, so this is where the holes are for me.

Yes, I understand you. I only did not write specifically because, as you can see here:

https://www.raspberrypi.com/documentation/computers/raspberry-pi.html

"A 40-pin GPIO header is found on all current Raspberry Pi boards"

If you prefer we can list the models, no problem.

en/companion_computer/pixhawk_rpi.md
ViniciusAbrao: @hamishwillee This is a very good question. Let me tell you that I connect the `TELEM2` VCC to supply the RPi in the 5V power (pin 4). 

I am telling in tutorial that this is not recommended because I already read about it. Since I am not really confident about the risks to damage the Pixhawk, I did not write it in the tutorial. But in my case, I have never had any problems with this setup.

en/companion_computer/pixhawk_rpi.md
hamishwillee: > No, this is building the PX4 in the Pixhawk.

it is not possible to build PX4 "in" the Pixhawk. So I think you mean "for" the Pixhawk.

What we will do here is link to the normal PX4 instructions, then add a summary of the steps `PX4-Autopilot/Tools/setup/ubuntu.sh --no-sim-tools`.

Last of all perhaps we will add a note about  stlink-tools as well.

In general you should not need these tools. The main reason would be to do on target debugging. Though in some cases you can use them to manually flash bootloaders and stuff. I don't think most people would need them - what did you see when you did not add these?

en/companion_computer/pixhawk_rpi.md
hamishwillee: >> This reads as "Build PX4 on Pixhawk" - you don't mean that right?
> Yes, I mean that.

As above, no you don't - you mean "for Pixhawk, on your Ubuntu laptop". Building PX4 in Pixhawk would mean that you had Ubuntu running on the Pixhawk and then you built PX4 using the Pixhawk as the computer. Not possible.

> To build the PX4 on the Pixhawk we need to give permission to write on it .... Maybe in your Ubuntu it is the default, but usually I need to change that permission when working with USB devices in Ubuntu (PS.: the same with Arduino for example).

That's clear. Normally I do nothing special to write - I just install the toolchain and do upload. But this is good. What might be good is to try the upload step without this and see what the output is. Then we could say to users "if you get this error, then you need to do this to get the port to work?"

en/companion_computer/pixhawk_rpi.md
hamishwillee: OK. What I meant was "what do you mean by in the sequence". I think you mean "do these steps".

en/companion_computer/pixhawk_rpi.md
hamishwillee: I will probably refer to both.

en/companion_computer/pixhawk_rpi.md
hamishwillee: Do you need USB to mavlink if you have the Radio connection? Sure for setup, but "generally"? You won't be able to do both MAVLink and XRCE-DDS over the serial link unless you can add a protocol spitter.

en/companion_computer/pixhawk_rpi.md
hamishwillee: My concern here is the instructions for setup. What I want to say is "connect the wires to this port on this output that will be present on all RaPis. That will always map to this particular `dev/Xxxx` value.". In that case we can treat all the boards as the same because as long as you connect to that port you specify the same values for your agent setup.

However my understanding was that different RaPis have different `dev/` values for different ports?

en/companion_computer/pixhawk_rpi.md
hamishwillee: This is definitely risky. None of the ports on Pixhawk are rated for very high power supply. Generally they supply enough for an old-style 3DR telemetry radio, but if you're attaching a newer high power radio like a microhard setup then you need to power it separately.

What this means is that so far you have been lucky-ish. The RaPi you are using is not drawing that much more power than the old telemetry radio. 

So recommendation would definitely be a BEC. I'll tidy this up when I edit.

hamishwillee:
> @hamishwillee I included a new PR with the branch at:

Thanks so much. I'm back on this on Wednesday "properly" so I'll take the information you have provided and do a good subedit then. In the meantime, let's continue our open discussions here.

en/companion_computer/pixhawk_rpi.md
ViniciusAbrao: > OK. What I meant was "what do you mean by in the sequence". I think you mean "do these steps".

Yes, 'do these steps"

en/companion_computer/pixhawk_rpi.md
ViniciusAbrao: In my understanding, the RX/TX pins of RPi map the communication to the dev/serial0 (same as ttyAMA0). Since we say that the connection must be in the pin 8 and pin 10 (RX/TX), and this is a standard IO pinout, it would result in a default approach independent of the RPi model.

en/companion_computer/pixhawk_rpi.md
ViniciusAbrao: We can do like I mentioned before (https://github.com/PX4/PX4-user_guide/pull/2318#discussion_r1123810951). With USB responsible for mavlink and TELEM2 responsible for xrce.

en/companion_computer/pixhawk_rpi.md
ViniciusAbrao: Please let me know if my answer below (https://github.com/PX4/PX4-user_guide/pull/2318#discussion_r1123837322) have clarified it.

en/companion_computer/pixhawk_rpi.md
ViniciusAbrao: I fully agree with you, and yes, I believe I have been lucky-ish  :-)

hamishwillee:
Hi @ViniciusAbrao Got caught in other work! Still planning on looking at this. Not ignoring you! Tomorrow.

en/companion_computer/pixhawk_rpi.md
hamishwillee: This is clear thank you.

hamishwillee: However I think for the instructions we should update to just point to the existing setup instructions. IF there is a problem that you need to use the command `sudo chmod a+rw /dev/ttyACM0` then we need to update our general Ubuntu documentation to explain this and link to that. Otherwise other people will be stuck who don't read your instructions. Hope that makes sense.

en/companion_computer/pixhawk_rpi.md
hamishwillee: Ah, that is what is confusing - we need to very clearly separate our different config setups. 

The setup you're actually likely to want is:
- Pixhawk with RPi via TELEM2 (xrce)  - ROS2 companion control
- Pixhawk with XXXXX via MAVLink, where XXXX might be RPI or QGC or whatever. It really depends if you need your RPi to get MAVLink from QGC or cloud or whatever and it is the channel for passing through MAVLink to PX4. 

Let me see if I can pull the apart a bit more.

ViniciusAbrao: Yes, I agree that it can be improved to better clarify the different setups. Let me know if are still any confusion in what I meant.

hamishwillee:
Closing this now. All data mined for new branch.

PR #2322, opened 2023-03-02
ROS2 Offboard control example - update to XRCE-DDS

hamishwillee:
This updates the ROS2 offboard example docs to to XRCE-DDS.

This essentially just worked "out of the box" when run using the instructions in the ROS2 User Guide for setup. The only difference is that this uses `ros2 run` rather than `ros2 launch`. No launch file I guess.
Most of my confusion was why we had a disarm() command and no implementation.

This example could do with some love. Specifically, it would be nice to monitor for the command ack, and to actually land the vehicle. But anyhow ....

IN terms of structure and explanation this is much simpler. Some of the detail is already in the ROS2 user guide so did not have to be here. Some comments inline.

### DIFF
--- a/en/ros/ros2_offboard_control.md
+++ b/en/ros/ros2_offboard_control.md
@@ -22,7 +22,8 @@ To subscribe to data coming from nodes that publish in a different frame (for ex
 
 ## Trying it out
 
-Follow the instructions in [ROS 2 User Guide](..ros/ros2_comm.md) to install PX4 and run the simulator, install ROS2, and start the XRCE-DDS Agent.
+Follow the instructions in [ROS 2 User Guide](..ros/ros2_comm.md) to install PX and run the simulator, install ROS2, and start the XRCE-DDS Agent.
+
 After that we can follow a similar set of steps to those in [ROS 2 User Guide > Build ROS2 Workspace](..ros/ros2_comm.md#build-ros-2-workspace) to run the example.
 
 To build and run the example:
@@ -64,12 +65,6 @@ To build and run the example:
 1. Launch the example.
 
    ```
-   ros2 launch px4_ros_com offboard_control_launch.yaml
-   ```
-   
-   or
-   
-   ```
    ros2 run px4_ros_com offboard_control
    ```
 
@@ -80,10 +75,11 @@ The vehicle should arm, ascend 5 metres, and then wait (perpetually).
 The source code of the offboard control example can be found in [PX4/px4_ros_com](https://github.com/PX4/px4_ros_com) in the directory [/src/examples/offboard/offboard_control.cpp](https://github.com/PX4/px4_ros_com/blob/main/src/examples/offboard/offboard_control.cpp).
 
 :::note
-PX4 publishes and subscribes all the messages used in this example as ROS topics by default (see [dds_topics.yaml](https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/microdds_client/dds_topics.yaml)).
+PX4 publishes all the messages used in this example as ROS topics by default (see [dds_topics.yaml](https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/microdds_client/dds_topics.yaml)).
 :::
 
-PX4 requires that the vehicle is already receiving setpoints before switching to offboard mode, and will switch out of offboard mode if the stream rate drops below approximately 2Hz.
+PX4 requires that the vehicle is already receiving `OffboardControlMode` messages before it will arm in offboard mode, or before it will switch to offboard mode when flying.
+In addition, PX4 will switch out of offboard mode if the stream rate of `OffboardControlMode` messages drops below approximately 2Hz.
 The required behaviour is implemented by the main loop spinning in the ROS2 node, as shown below:
 
 ```cpp
@@ -111,7 +107,7 @@ timer_ = this->create_wall_timer(100ms, timer_callback);
 
 The loop runs on a 100ms timer.
 For the first 10 cycles it calls `publish_offboard_control_mode()` and `publish_trajectory_setpoint()` to send [OffboardControlMode](../msg_docs/OffboardControlMode.md) and [TrajectorySetpoint](../en/msg_docs/TrajectorySetpoint.md) messages to PX4.
-These messages are required in order for PX4 to switch to offboard mode, even though they are "ignored" because the vehicle is not in offboard mode.
+The `OffboardControlMode` messages are streamed so that PX4 will allow arming once it switches to offboard mode, while the `TrajectorySetpoint` messages are ignored (until the vehicle is in offboard mode).
 
 After 10 cycles `publish_vehicle_command()` is called to change to offboard mode, and `arm()` is called to arm the vehicle.
 After the vehicle arms and changes mode it starts tracking the position setpoints.
--- a/en/ros/ros2_offboard_control.md
+++ b/en/ros/ros2_offboard_control.md
@@ -22,7 +22,7 @@ To subscribe to data coming from nodes that publish in a different frame (for ex
 
 ## Trying it out
 
-Follow the instructions in [ROS 2 User Guide](..ros/ros2_comm.md) to install PX and run the simulator, install ROS2, and start the XRCE-DDS Agent.
+Follow the instructions in [ROS 2 User Guide](..ros/ros2_comm.md) to install PX4 and run the simulator, install ROS2, and start the XRCE-DDS Agent.
 After that we can follow a similar set of steps to those in [ROS 2 User Guide > Build ROS2 Workspace](..ros/ros2_comm.md#build-ros-2-workspace) to run the example.
 
 To build and run the example:
--- a/en/ros/ros2_offboard_control.md
+++ b/en/ros/ros2_offboard_control.md
@@ -64,6 +64,12 @@ To build and run the example:
 1. Launch the example.
 
    ```
+   ros2 launch px4_ros_com offboard_control_launch.yaml
+   ```
+   
+   or
+   
+   ```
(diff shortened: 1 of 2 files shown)


en/ros/ros2_offboard_control.md
```diff
@@ -187,22 +176,19 @@ void OffboardControl::publish_vehicle_command(uint16_t command, float param1,
...
-```sh
-$ source path_to_colcon_workspace/install/setup.bash
-$ ros2 run px4_ros_com offboard_control
-```
+<!--
 
 ## Demo with PX4 SITL and Gazebo Classic
 
 @[youtube](https://youtu.be/Nbc7fzxFlYo)
+-->
```
hamishwillee, line 194: MY VM does not have the grunt to run the simulation and record. So if someone wants to create a video and upload would be very pleased!

en/ros/ros2_offboard_control.md
```diff
@@ -187,22 +176,19 @@ void OffboardControl::publish_vehicle_command(uint16_t command, float param1,
...
 
-As the description suggests, the above code serves the purpose of sending `VehicleCommand` messages with commands to the flight controller.
-
-## Usage
-
-After building the colcon workspace, and after starting PX4 SITL (`make px4_sitl_rtps gazebo-classic`, which starts the microRTPS client automatically on UDP ports 2019 and 2020) and the microRTPS agent (`micrortps_agent -t UDP`, starting the agent connected to UDP ports 2020 and 2019):
+:::note
+[VehicleCommand](../msg_docs/VehicleCommand.md) is one of the simplest and most powerful ways to command PX4, and by subscribing to [VehicleCommandAck](../msg_docs/VehicleCommandAck.md) you can also confirm that setting a particular command was successful.
+The param and command fields map to [MAVLink commands](https://mavlink.io/en/messages/common.html#mav_commands) and their parameter values.
+:::
```
hamishwillee, line 187: FYI Here trying to explain the mapping to MAVLink.

en/ros/ros2_offboard_control.md
```diff
@@ -1,172 +1,162 @@
...
 
-```cpp
-timesync_sub_ = this->create_subscription<px4_msgs::msg::Timesync>("/fmu/timesync/out",
-    10,
-    [this](const px4_msgs::msg::Timesync::UniquePtr msg) {
-        timestamp_.store(msg->timestamp);
-    });
-```
+   ```
+   ros2 run px4_ros_com offboard_control
```
hamishwillee, line 67: This whole section is just a simplified user guide setup. The only difference is what is launched, and the fact that we use ros2 run.

There does appear to be a yaml launch file in the repo, but calling it does not launch this example, so don't know if there is something I am missing?

en/ros/ros2_offboard_control.md
```diff
@@ -1,172 +1,162 @@
...
+   ```
+1. Launch the example.
 
-```cpp
-timesync_sub_ = this->create_subscription<px4_msgs::msg::Timesync>("/fmu/timesync/out",
-    10,
-    [this](const px4_msgs::msg::Timesync::UniquePtr msg) {
-        timestamp_.store(msg->timestamp);
-    });
-```
```
hamishwillee, line 95: This wasn't in the source code, which just seems to get the time to send from the local clock. I believe ROS2 does everything it needs to sync "under the hood"?

en/ros/ros2_offboard_control.md
```diff
@@ -1,172 +1,162 @@
...
+   ```
+   ros2 run px4_ros_com offboard_control
+   ```
 
-The above is required in order to obtain a synchronized timestamp to be set and sent with the `offboard_control_mode` and `trajectory_setpoint` messages.
+The vehicle should arm, ascend 5 metres, and then wait (perpetually).
+
+## Implementation
```
hamishwillee, line 72: FYI the main change here is to explain a little more about what offboard mode requires, and link to that. The idea being that first you understand why this code writes setpoints before changing the mode, and also that you can go to offboard mode to understand how you might do velocity control etc.

beniaminopozzan: I went through the code and instructions, everything works just fine but
and you won't like it
1. It seems that you can switch to offboard mode anytime.
2. Arming is disabled until a valid stream of setpoints is received (as it should be).

To be more precise, what I did was

1. start the simulation and open QGC.
2. Without even starting the micro XRCE-DDS Agent I switched to offboard mode from QGC. The command was accepted. I also tried from mavlink console with `commander mode offboard` and the result was the same.
3. With mode set to offboard arming was denied. The reported description in QGC is _The offboard component is not sending setpoints or the required estimate (e.g. position) is missing._. This actually tells us that contrarily to what is written in the first note of [offboard-mode](https://docs.px4.io/main/en/flight_modes/offboard.html#offboard-mode), only an estimate of what we want to control is needed. For example, If I want to send attitude setpoints I can do it without GPS.
4. As soon as I started sending setpoints arming was allowed again.

Further testing made me realize also that you only need to provide a continuous stream of `offboard_control_mode` messages to allow arming. Therefore you can do weird things like

1. Start streaming `offboard_control_mode` messages (for example in offboard position mode)
2. Arm the vehicle.
3. After a while (before the auto disarm is triggered) start sending trajectory setpoints. The vehicle will take off.
4. Stop sending trajectory setpoints, the vehicle will use the last one available and keep flying. No error is raised.

beniaminopozzan (commented):
I like it in this simple but exhaustive format. There is only the issue with the new failsafe behavior. @hamishwillee , is it OK for you to settle this PR and "add love" later on?

en/ros/ros2_offboard_control.md
hamishwillee: I will recheck this.

en/ros/ros2_offboard_control.md
hamishwillee: PX4 should never allow you to switch to offboard mode without sending a stream of offboard_control_mode - that's like the heartbeat. If offboard mode messages stop it should switch to hold mode.

So your further testing is correct as per my understanding. I think using the last trajectory setpoint is OK for position setpoints because you'll fly to that position and then stop. If the same thing happens for velocity setpoints it is risky -though as above, I think ONLY the offboard control mode  is the heartbeat. It may be a "problem", but it is expected behaviour.

Re 
> he offboard component is not sending setpoints or the required estimate (e.g. position) is missing.. 

Hmmm. You're likely right. If you're in offboard mode the assumption is that position is provided by the companion - PX4 has no need of its own position/GPS.

@Jaeyoung-Lim Does this align with your understanding too?

en/ros/ros2_offboard_control.md
hamishwillee: FYI, got tied up in MAVLink. Hope to look at this tomorrow. Sorry!

hamishwillee: > Speaking about simplicity, do we still need all the scripts https://github.com/PX4/px4_ros_com/tree/main/scripts?

@beniaminopozzan I don't think so. From what I can tell these were required to satisfy the complicated RTPS build process. They aren't needed now, and the [v1.13](https://docs.px4.io/v1.13/en/ros/ros2_comm.html#build-ros-2-workspace) docs reference the specific branch.

Can you delete these? Also anything else from main that either doesn't work or doesn't make sense in XRCE-DDS land. IMO the 1.13 branch should remain relevant for that release, but otherwise this should just be a library of examples and useful helper scripts.

Make sense? Most of what I know about from this is from you and Nuno's presentations.

hamishwillee: > On my machine, running
> ```
> ros2 launch px4_ros_com offboard_control_launch.yaml
> ```

Yes, that works for me too - now. I wonder what I was trying. I might swap to this, as I think launch  files are very cool :-)

hamishwillee: Done - added both options with an "or".

en/ros/ros2_offboard_control.md
hamishwillee: @beniaminopozzan FYI a little more debug (some duplicated) which pretty much confirm my assumptions
1. `OffboardControlMode` must be streamed to arm in offboard mode _OR to switch to offboard mode when flying_. If this stops, you'll drop back to position mode by default (the fallback mode is configurable by parameters). What this means is that this is the heartbeat, and does block offboard operation if not sent (though you can enter the mode).
2. The setpoints make no difference to mode switching.

IN theory if you have RC required you should also need https://docs.px4.io/main/en/advanced_config/parameter_reference.html#COM_RCL_EXCEPT set to avoid a failsafe. Julian had this in his tutorial. My testing is that this does not seem to matter - it may be that RC checks are disabled on simulator for some reason.

I'll update this doc with minimal changes, and do some updates to offboard mode as a separate PR.

hamishwillee:
Hurrah, merged. Thanks for reviews. 

I'm looking at offboard now. Note, I said somewhere offboard will not require GPS. It most definitely requires a position source of some kind, because you're supplying setpoints, and these have to be relative to some frame that the flight stack knows. I guess in theory you might be able to fly with velocity setpoints without this, but I suspect "not supported".

en/ros/ros2_offboard_control.md
beniaminopozzan: @hamishwillee Thanks for confirming it!
Did you have QGC with virtual joystick opened? I would explain why it had no effects.

hamishwillee: No. I thought the same thing so I tried with COM_RC_IN_MODE = 0 (joystick only), and also checked that there was no virtual joystick. This was a completely clean installation on a fresh VM. I suspect the check is broken.

PR #2326, opened 2023-03-05
remove sudo from pip3 calls and add --user option

beniaminopozzan:


hamishwillee (approved):
Thanks @beniaminopozzan - I really don't get the Linux security model, but if this works, then great!

PR #2334, opened 2023-03-09
Offboard mode: Fix errors and make it less MAVLink-centric

hamishwillee:
This fixes a couple of errors and makes the document a bit less MAVLink-centric - which is important now that you can use ROS2.

The main error was that this previously indicated that you couldn't enter offboard mode without first streaming setpoints. That is incorrect - you can switch to offboard mode when landing, you just can't arm. So this now says PX4 won't let you arm in offboard mode, or switch to offboard mode while flying.

In addition this was pretty much MAVLink centric and now we can use ROS 2. The main difference between the two approaches is that MAVLink requires a stream of setpoints for proof of life. ROS 2 requires you set the offboard control topic for proof of life and separately set the setpoint. So for ROS 2 you could just send the setpoint once - it doesn't get overwritten.

I've tried to capture this, and also use words that make it clear that the streaming is a heartbeat/proof of life. 

What's missing is a list of the setpoint uorb topics that people might use, along with the values that they can actually set? I assume that like the matching MAVLink topics these are different for different vehicles. Or it might be that each vehicle has some subset of specific messages. I could infer this from the MAVLink messages, but was wondering if anyone already has a view on what set are relevant.

@Jaeyoung-Lim @beniaminopozzan Would appreciate your review.


beniaminopozzan:
@hamishwillee , I took the freedom to follow

> but was wondering if anyone already has a view on what set are relevant.

and pushed a draft of the relevant topics for MC offboard control along with the possible combinations of `OffboardControlMode`. I only worked with MC so I don't have the list for VTOL / FW / Rover yet.

Regarding the required estimates for each mode, I used https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/commander/HealthAndArmingChecks/checks/offboardCheck.cpp as reference. I'm still trying to simulate a situation in which the position estimate is invalid while the velocity one is valid... no luck for now.

en/flight_modes/offboard.md
```diff
@@ -2,39 +2,103 @@
...
 Operations like uploading, downloading missions can be performed in any mode.
 
-A stream of setpoint commands must be received by the vehicle prior to engaging the mode, and in order to remain in the mode (if the message rate falls below 2Hz the vehicle will stop). 
-In order to hold position while in this mode, the vehicle must receive a stream of setpoints for the current position.
+## ROS 2 Messages
+
+The following ROS 2 messages and their particular fields and field values are allowed for the specified frames.
+In addition to providing heartbeat functionality, `OffboardControlMode` has two other main purposes:
+
+1. Controls at which level of the [PX4 control architecture](../en/flight_stack/controller_diagrams) the offboard setpoints must be injected and disables the bypassed controllers.
```
hamishwillee, line 44:
```suggestion
1. Controls the level of the [PX4 control architecture](../en/flight_stack/controller_diagrams.md) at which offboard setpoints must be injected, and disables the bypassed controllers.

```

en/flight_modes/offboard.md
```diff
@@ -2,39 +2,106 @@
...
+bool acceleration
+bool attitude
+bool body_rate
+bool actuator
+```
+
+The fields are ordered in terms of priority such that `position` takes precedence over all the other, `velocity` takes precedence over `acceleration` and the other below it, etc.
+Depending on the first non-zero value from top to bottom, the valid estimate required and the right setpoint message(s) are defined.
+
+| desired control quantity | position field | velocity field | acceleration field | attitude field | body_rate field | acturator field | required estimate | required message                                                                                                                |
```
hamishwillee, line 65: acturator - a typo?

en/flight_modes/offboard.md
```diff
@@ -2,39 +2,106 @@
...
 
-A stream of setpoint commands must be received by the vehicle prior to engaging the mode, and in order to remain in the mode (if the message rate falls below 2Hz the vehicle will stop). 
-In order to hold position while in this mode, the vehicle must receive a stream of setpoints for the current position.
+## ROS 2 Messages
+
+The following ROS 2 messages and their particular fields and field values are allowed for the specified frames.
+In addition to providing heartbeat functionality, `OffboardControlMode` has two other main purposes:
+
+1. Controls the level of the [PX4 control architecture](../en/flight_stack/controller_diagrams.md) at which offboard setpoints must be injected, and disables the bypassed controllers.
+1. Determines which valid estimates (position or velocity) are required.
```
hamishwillee, line 45:
```suggestion
1. Determines which valid estimates (position or velocity) are required, and also which setpoint messages should be used.

```

en/flight_modes/offboard.md
```diff
@@ -2,39 +2,106 @@
...
+bool position
+bool velocity
+bool acceleration
+bool attitude
+bool body_rate
+bool actuator
+```
+
+The fields are ordered in terms of priority such that `position` takes precedence over all the other, `velocity` takes precedence over `acceleration` and the other below it, etc.
+Depending on the first non-zero value from top to bottom, the valid estimate required and the right setpoint message(s) are defined.
```
hamishwillee, line 63:
```suggestion
The fields are ordered in terms of priority such that `position` takes precedence over `velocity` and later fields, `velocity` takes precedence over `acceleration`, and so on.
The first field that has a non-zero value (from top to bottom) defines what valid estimate is required in order to use offboard mode, and the setpoint message(s) that can be used.
For example, if the `acceleration` field is the first non-zero value, then PX4 requires a valid `velocity estimate`, and the setpoint must be specified using the `TrajectorySetpoint` message.

```

en/flight_modes/offboard.md
```diff
@@ -2,39 +2,108 @@
...
+|--------------------------|:--------------:|:--------------:|:------------------:|:--------------:|:---------------:|:---------------:|:-----------------:|---------------------------------------------------------------------------------------------------------------------------------|
+| position (NED)           |     &check;    |        -       |          -         |        -       |        -        |        -        |      position     | TrajectorySetpoint                                                                                                              |
+| velocity (NED)           |     &cross;    |     &check;    |          -         |        -       |        -        |        -        |      velocity     | TrajectorySetpoint                                                                                                              |
+| acceleration (NED)       |     &cross;    |     &cross;    |       &check;      |        -       |        -        |        -        |      velocity     | TrajectorySetpoint                                                                                                              |
```
hamishwillee, line 71:
```suggestion
| position (NED)           |     &check;    |        -       |          -         |        -       |        -        |        -        |      position     | `TrajectorySetpoint`                                                                                                              |
| velocity (NED)           |     &cross;    |     &check;    |          -         |        -       |        -        |        -        |      velocity     | `TrajectorySetpoint`                                                                                                              |
| acceleration (NED)       |     &cross;    |     &cross;    |       &check;      |        -       |        -        |        -        |      velocity     | `TrajectorySetpoint`                                                                                                              |

```

en/flight_modes/offboard.md
```diff
@@ -2,39 +2,109 @@
...
+### Copter
 
-Offboard mode requires an active connection to a remote MAVLink system (e.g. companion computer or GCS). 
-If the connection is lost, after a timeout ([COM_OF_LOSS_T](#COM_OF_LOSS_T)) the vehicle will attempt to land or perform some other failsafe action. 
-The action is defined in the parameters [COM_OBL_ACT](#COM_OBL_ACT) and [COM_OBL_RC_ACT](#COM_OBL_RC_ACT).
+* [px4_msgs::msg::TrajectorySetpoint](https://github.com/PX4/PX4-Autopilot/blob/main/msg/TrajectorySetpoint.msg)
+  * The following input combinations are supported:
+    * Position setpoint (`position` different from `NaN`). Non-`NaN` values of velocity and acceleration are used as feedforward terms for the inner loop controllers.
+    * Velocity setpoint (`velocity` different from `NaN` and `position` set to `NaN`). Non-`NaN` values acceleration are used as feedforward terms for the inner loop controllers.
+    * Acceleration setpoint  (`acceleration` different from `NaN` and `position` and `velocity` set to `NaN`)
```
hamishwillee, line 85:
```suggestion
    * Acceleration setpoint (`acceleration` different from `NaN` and `position` and `velocity` set to `NaN`)

```

en/flight_modes/offboard.md
```diff
@@ -2,39 +2,109 @@
...
+    * `roll`, `pitch`, `yaw` and `thrust_body`.
+  - All the value are in the drone body FRD frame. The rates are in \[rad/s\] while thrust_body is normalized in \[-1, 1\].
+
+* [px4_msgs::msg::VehicleThrustSetpoint](https://github.com/PX4/PX4-Autopilot/blob/main/msg/VehicleThrustSetpoint.msg) + [px4_msgs::msg::VehicleTorqueSetpoint](https://github.com/PX4/PX4-Autopilot/blob/main/msg/VehicleTorqueSetpoint.msg)
+  * The following input combination is supported:
+    * `xyz` for thrust and `xyz` for torque.
+  - All the value are in the drone body FRD frame and normalized in \[-1, 1\].
 
-## Supported Messages
+
```
hamishwillee, line 103:
```suggestion
```

hamishwillee:
@beniaminopozzan This is excellent - just what we need. I did some minor tidy up, and there are a couple of suggestions, but I am tempted to get this in as-is. Can you do a quick sanity check and I'll push it in once that is done.

One question. The text implies that you can only have one setpoint at a time, based on the first non-zero value in the uorb topic. To confirm - you can't have a attempt to follow two setpoints at once right?

en/flight_modes/offboard.md
beniaminopozzan: Yes, sorry!

beniaminopozzan:
> The text implies that you can only have one setpoint at a time, based on the first non-zero value in the uorb topic. To confirm - you can't have a attempt to follow two setpoints at once right?

You do want to have only one setpoint at a time. But there is no check to make sure that only one setpoint is being provided.
For example, you could be in `OffboardControlMode.position = 1` and send both trajectory_setpoint and rates_setpoint messages. The trajectory setpoints would go into the position controller, then attitude setpoints would be generated, followed by rates setpoint. At that point those rates setpoints would mix with the one coming from offboard... potentially causing a failure (and the rate rate controller wouldn't even notice as for it they are all the same).

However, you can perfectly send both messaged if only `OffboardControlMode.body_rate = 1` as all the higher order controllers are disabled.

To sum it up, if _multiple_ setpoints are like position+velocity, where velocity is a feedforward term, then yes, you can have two setpoints at once and everything works. If _multiple_ setpoints are intended as different "setpoint"-type messages sent together, then the answer is "as long as the associated controller is active, then all setpoints entering it will be accepted".

hamishwillee:
Thanks very much. Decided to merge this and deal with possible case of multiple setpoints in https://github.com/PX4/PX4-user_guide/pull/2341 so as not to block this.

PR #2336, opened 2023-03-10
gazebosim: fix gz_env.sh.in path and name

beniaminopozzan:
The script name needed to be updated and the path fixed


hamishwillee (approved):
Sounds plausible :-)

I probably won't force a rebuilt until next week, but this will be in the next one. Thanks!

PR #2341, opened 2023-03-15
Offboard mode - use the setpoint matching your offboardcontrol value

hamishwillee:
This attempts to capture my understanding of the issue of "setting multiple setpoints" discussed in https://github.com/PX4/PX4-user_guide/pull/2334#issuecomment-1465403621

@beniaminopozzan This is slightly different advice than you gave. You essentially said that you might set position and velocity at the same time, because provided the velocity is feedforward it "could work ok". But the [controller diagram](http://docs.px4.io/main/en/flight_stack/controller_diagrams.html) shows that the velocity controller is getting velocity setpoints, so I'm not sure that these can't be in conflict with what is coming out of the position controller. 

More detail on the changes in line. THIS is for Discussion. We don't need to take all or any of it. Let's see what you think.


en/flight_modes/offboard.md
```diff
@@ -61,21 +61,26 @@ bool actuator
...
-| position (NED)           |     &check;    |        -       |          -         |        -       |        -        |        -        |      position     | `TrajectorySetpoint`                                                                                                              |
-| velocity (NED)           |     &cross;    |     &check;    |          -         |        -       |        -        |        -        |      velocity     | `TrajectorySetpoint`                                                                                                              |
-| acceleration (NED)       |     &cross;    |     &cross;    |       &check;      |        -       |        -        |        -        |      velocity     | `TrajectorySetpoint`                                                                                                              |
+| position (NED)           |     &check;    |        -       |          -         |        -       |        -        |        -        |      position     | `TrajectorySetpoint.position`                                                                                                              |
```
hamishwillee, line 70: Here I added `TrajectorySetpoint.position`.
IFF the advice I give below that you should only set the setpoint fields matching the control quantity is good, then it makes sense to show the specific fields you can set. We would change the "required message" heading as well perhaps. 

For discussion :-)

en/flight_modes/offboard.md
```diff
@@ -61,21 +61,26 @@ bool actuator
 ```
 
 The fields are ordered in terms of priority such that `position` takes precedence over `velocity` and later fields, `velocity` takes precedence over `acceleration`, and so on.
-The first field that has a non-zero value (from top to bottom) defines what valid estimate is required in order to use offboard mode, and the setpoint message(s) that can be used.
-For example, if the `acceleration` field is the first non-zero value, then PX4 requires a valid `velocity estimate`, and the setpoint must be specified using the `TrajectorySetpoint` message.
+The first field that has a non-zero value (from top to bottom) defines what valid estimate is required in order to use offboard mode, the setpoint message(s) that can be used, and the controllers that are enabled.
+For example, if the `acceleration` field is the first non-zero value, then PX4 requires a valid `velocity estimate`, the setpoint must be specified using the `TrajectorySetpoint` message `acceleration` field, and the position and velocity controllers are disabled (so that even if set, the setpoint position and velocity field will not be used).
```
hamishwillee, line 65: The statement about what controllers are disabled is quite significant. I thought worth expanding out here below in the example.

en/flight_modes/offboard.md
```diff
@@ -61,21 +61,26 @@ bool actuator
...
 | body_rate (RFD)          |     &cross;    |     &cross;    |       &cross;      |     &cross;    |     &check;     |        -        |        none       | [VehicleRatesSetpoint](../msg_docs/VehicleRatesSetpoint.md)                                                                     |
 | thrust and torque (RFD)  |     &cross;    |     &cross;    |       &cross;      |     &cross;    |     &cross;     |     &check;     |        none       | [VehicleThrustSetpoint](../msg_docs/VehicleThrustSetpoint.md) and [VehicleTorqueSetpoint](../msg_docs/VehicleTorqueSetpoint.md) |
 
 where &check; means that the bit is set, &cross; means that the bit is not set and `-` means that the bit is value is irrelevant.
 
+:::note
+Generally you should only use the setting message and fields corresponding to the bit set in `OffboardControlMode`.
+If you set a field that injects an offboard setpoint at a lower level than the highest active controller, it may conflict with setpoints generated by a higher controller.
```
hamishwillee, line 81: So this is my attempt to explain why you should always set the bit to match the thing that you are controlling.

Even if you consider a position controller feeding into a velocity controller, there are potential conflicts right if you were to set the velocity but not position - since that position controller is active if the offboardcontrol bit is alive, and will be sending _some_ setpoint.

beniaminopozzan:
From you comments @hamishwillee I realized I could have been more clear.
Unfortunately, the connection between `OffboardControlMode` and the active controller(s) is not that straightforward as I initially stated.
Position, velocity and acceleration controllers reside in the same module (the diagram is misleading in this sense), the `mc_pos_control` module. Restricting the discussion to offboard control only, this module is _enabled_ if **at least** one of the `position`, `velocity` or `acceleration` flags are set. When that happens the controller uses the `TrajectorySetpoint` messages as reference and it uses ALL available fields! The only way to not use one or more fields is to set them to `NaN`.

For example, let's say you want to control just the velocity. You then set just the velocty field of OffboardControlMode and then start sending trajectory setpoints with the desired velocity, and as you may assume that the position controller is switched off, you also set to zero the position field in the trajectory setpoint. But the position controller is not inactive and you will end up in a quite weird situation in which you are trying to track a zero reference position with a non zero reference velocity. In this situation your desired velocity is considered as a feedforward term.

This is to say that the most the most important role is played by how the fields of the trajectory setpoint message are filled. Only `NaN`s actually disable the associated controller. It is up to the used to correctly match not only setpoint message with offboard control mode fields, but also the specific fields of the setpoint message and the ones in the offboard control mode. 

You could ask "then why offboard control mode has the position, velocity and acceleration fields if at the end just one could do to enable/disable the `mc_pos_control` module?"  -> We want the distinction to be able to know which valid estimate is required.

To have an good idea of the cascade of the module involved in the mc control loops you can look at the setpoint messages that are used:

- `trajectory_setpoint` is the setpoint message for the `mc_pos_control` module. This module controls position, velocity and acceleration all together (unless `NaN`s) and outputs a `vehicle_attitude_setpoint` message.
- `vehicle_attitude_setpoint` is the setpoint message for the `mc_att_control` module. This module controls the attitude of the drone and outputs a `vehicle_rates_setpoint` message.
- `vehicle_rates_setpoint` is the setpoint message for the `mc_rate_control` module. This module controls the body rates of the drone and outputs a pair of `vehicle_torque_setpoint` and `vehicle_thrust_setpoint` messages.
- `vehicle_torque_setpoint` and `vehicle_thrust_setpoint` are finally fed to the control allocation module.

en/flight_modes/offboard.md
beniaminopozzan: In my original intention I didn't specify the setpoint fields here but I put the specific setpoint structure in the section below, where all the possible combinations are explained (and in particular the field that must be set to `NaN` are introduced).
I fear that stating the fields here will lead users

1. to correctly set them
2. at the same to forget to put `NaN` to the others

I think it is better to force them to read the section below to understand the appropriate structure

en/flight_modes/offboard.md
beniaminopozzan: My apologies, this

> the position and velocity controllers are disabled (so that even if set, the setpoint position and velocity field will not be used).

 is not correct, see https://github.com/PX4/PX4-user_guide/pull/2341#issuecomment-1469377993
 
- It needs just one of `position`, `velocity`, `acceleration` fields to be set to enable all controllers. Users need to use appropriately the setponint fields (set them to `NaN`) to have additional control.
- `attitude` set enables attitude controller and rate controller.
- `body_rate` set enables only the rate controller.

Jaeyoung-Lim:
@beniaminopozzan @hamishwillee Your comment is correct regarding the multicopter controls. I think as an example this would work great. 

However, I would refrain from constraining offboard just for the mc and specific modules. Offboard mode is an infrastructure to enable enabling different level setpoints for development that is agnostic to vehicle type or controller.

For example, I use offboard for developing fixedwing position and attitude controls, and also a custom flight controller.

Since the px4 implementation of the cascaded loop is "correct" if a lower level setpoint is specified the higher level controllers go into "idle". For combinations of different levels(e.g. pos, vel) the MC pos implementation takes a feedforward but fixedwing uses it as a path tangent since the vehicle does not have much autority over airspeed

beniaminopozzan:
@Jaeyoung-Lim you are absolutely correct!
I took mc control as use case because is the only one I'm used to.
In the other PR (the one just merged) I believe that it is clear that the setpoint explanation is tailored for mc, it's not that I don't want to include the other types of vehicle, I just don't know we'll enough how they are working internally

hamishwillee:
Thanks @beniaminopozzan @Jaeyoung-Lim 

I'm going to close this PR. The original text is correct, and everything in this is wrong.

There was a lot I wasn't sure about, but I think these things are answered by the copter section.

What is missing is equivalent documentation for other frames. @Jaeyoung-Lim Can you add the needed section for FW and/or VTOL? Or should I be chasing @sfuhrer ?

PR #2350, opened 2023-03-17
Update ROS naming convention

beniaminopozzan:
According to https://discourse.ros.org/t/should-i-put-a-space-between-ros-and-2/27949 the correct naming convention is **ROS 1** and **ROS 2**.


hamishwillee:
Thanks @beniaminopozzan - that's a change - originally the theory was ROS and ROS 2. This new convention is better, though it would be even better as ROS1 and ROS2.

beniaminopozzan:
@hamishwillee at least now we are consistent :)

hamishwillee:
For sure - it's a good thing. Just not as good as it would be if I was in charge. Yes, I'm serious :-)

hamishwillee:
And thanks for sorting this stuff out. I'll review the other stuff next week.

PR #2353, opened 2023-03-20
Bootloader Flashing for Kakute H7 v2 and mini

arguelle:
Added steps to manually flash the bootloader. The process varies from the standard method in [bootloader flashing](https://docs.px4.io/main/en/advanced_config/bootloader_update_from_betaflight.html) since the bootloader can be built without having to switch to the PX4-Bootloader repository.

The mass erase step has also been added since it seems to be necessary to fix [issues with params](https://github.com/PX4/PX4-Autopilot/issues/20881) not saving.

### DIFF
--- a/en/flight_controller/kakuteh7mini.md
+++ b/en/flight_controller/kakuteh7mini.md
@@ -77,6 +77,41 @@ The board comes pre-installed with [Betaflight](https://github.com/betaflight/be
 Before the PX4 firmware can be installed, the *PX4 bootloader* must be flashed.
 Download the [holybro_kakuteh7mini_bootloader.hex](https://github.com/PX4/PX4-user_guide/raw/main/assets/flight_controller/kakuteh7mini/holybro_kakuteh7mini_bootloader.hex) bootloader binary and read [this page](../advanced_config/bootloader_update_from_betaflight.md) for flashing instructions.
 
+### Building the bootloader
+
+:::note
+The bootloader may need to be flashed manually if the parameters do not save after the firmware setup.
+:::
+To build the bootloader for this target:
+
+```
+make holybro_kakuteh7mini_bootloader
+```
+
+### Flashing the bootloader
+
+#### Enter DFU mode
+
+To enter DFU mode, hold the boot button down while connecting the USB cable to your computer. The button can be released after the board is powered up.
+
+#### dfu-util
+
+Erase all flash params using a mass erase.
+:::note
+This command may generate an error that can be ignored.
+:::
+
+```
+dfu-util -a 0 --dfuse-address 0x08000000:force:mass-erase:leave -D ./build/holybro_kakuteh7mini_bootloader/holybro_kakuteh7mini_bootloader.bin
+```
+
+Enter DFU mode again and flash the board regularly.
+
+```
+dfu-util -a 0 --dfuse-address 0x08000000 -D ./build/holybro_kakuteh7mini_bootloader/holybro_kakuteh7mini_bootloader.bin
+
+```
+
 ## Building Firmware
 
 To [build PX4](../dev_setup/building_px4.md) for this target:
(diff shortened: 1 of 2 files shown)


arguelle:
@hamishwillee I agree. I will make changes to the bootloader documentation.

@julianoes Would it be safe to add the [mass erase option for dfu-util](https://github.com/PX4/PX4-Autopilot/pull/20545#issuecomment-1396330758) in general? I have had to use it with every Kakute H7 to get the params to save. We have to flash again without the option either way correct? Would it be best to just add a note stating that this is an additional step for Kakute H7 v2 & mini?

julianoes:
Thanks a lot for the docs contribution!

> Would it be best to just add a note stating that this is an additional step for Kakute H7 v2 & mini?

That would make sense. I would not that this is required for H7 v2 & mini but might not be required for others.

hamishwillee:
>> Would it be best to just add a note stating that this is an additional step for Kakute H7 v2 & mini?

> That would make sense. I would not that this is required for H7 v2 & mini but might not be required for others.

Yes, but that could be in the bootloader doc too, possibly with a cross link in the Kakute docs "note, for this case you will also need to do mass erase as linked here ..."

arguelle:
@hamishwillee I reverted the changes to the FC pages and added some **notes** to the [bootloader update](https://docs.px4.io/main/en/advanced_config/bootloader_update_from_betaflight.html) instuctions.

- The bootloader for Kakute H7 v2 and mini is built from the PX4 source
- Additional instruction for the mass-erase and some context regarding why it may be required.

The FC pages already refer to the bootloader page for flashing instructions.

en/advanced_config/bootloader_update_from_betaflight.md
```diff
@@ -24,7 +24,39 @@ You should now be able to install PX4 firmware on the board.
 
 #### Download Bootloader Source
 
-Download and build the [Bootloader](https://github.com/PX4/Bootloader) via:
+Flight controllers that have bootloader PX4-Autopilot `make` targets, can build the bootloader from the PX4-Autopilot source.
```
hamishwillee, line 27: FYI I have reworded this to be first because there turns out to be quite a lot of FCs you can flash in this way. OK?

en/advanced_config/bootloader_update_from_betaflight.md
```diff
@@ -50,12 +82,25 @@ The button can be released after the board is powered up.
...
 
+:::note
+The [Holybro Kakute H7 v2](../flight_controller/kakuteh7v2.md) and mini flight controllers may require that you first run an additional command to erase flash parameters (in order to fix problems with parameter saving):
+
+```
+dfu-util -a 0 --dfuse-address:force:mass-erase:leave 0x08000000 -D  build/<target>/<target>.bin
+```
+
+The command may generate an error which can be ignored.
+Once completed, continue to flash the bootloader as previously mentioned.
```
hamishwillee, line 101: This reads as though for the problem FCs you must erase then flash. Is that right? Perhaps we should have this step first then?

hamishwillee:
Thanks! I subedited and added a question or two - hope this is OK.

en/advanced_config/bootloader_update_from_betaflight.md
arguelle: This looks great. I originally wanted to try this layout since it's the only one that's worked for the FCs I have used (including matek).
I am glad you included the 'make list_config_targets'

en/advanced_config/bootloader_update_from_betaflight.md
arguelle: Correct, the erase operation must be completed first. I just didn't want to place the exception first, but placing it in a note seems to keep it out of the way.

arguelle:
I placed the note about the problem FCs above as you mentioned in your question.
I am still new to this general workflow btw, so please excuse (and mention it) if I do something out of the ordinary.

Also, when is it appropriate to label a code block as bash? I see you added it to a line.

en/advanced_config/bootloader_update_from_betaflight.md
hamishwillee: Yes, I see your point. I think this works and worse case people just do what is in the note, it does not harm.

Thanks again, merging now :-)

hamishwillee:
> Also, when is it appropriate to label a code block as bash? I see you added it to a line.

There are no real rules about adding bash to the code blocks - no harm in adding or removing. I tend to add it more often than not, because it makes `#` lines into more clear comments in the rendered version.

PR #2355, opened 2023-03-21
Update autopilot_manufacturer_supported.md

jwarnergithub:
Moved the MRobotics FC's up on the list and added ThePeach FC's to the list. This list should now match the left side navigation menu.


hamishwillee (approved):
Thanks @jwarnergithub . The list is supposed to be alphabetic by manufacturer, so mRobotics was correct in the list and incorrect in the sidebar. It is not alphabetic by FC version, because I wanted to make the more recent FCs appear first - so that's a bit random. We probably should just make it all completely consistent.

But anyhow, tidied this up now. Thanks for the change.

PR #2364, opened 2023-03-22
Update some PX4 param references

sfuhrer:
Updates parameter references in preparation for a new release v1.14

### DIFF
--- a/en/config/safety.md
+++ b/en/config/safety.md
@@ -241,8 +241,8 @@ Parameter | Description
 
 ### Quad-chute Failsafe
 
-Failsafe for when a VTOL vehicle can no longer fly in fixed-wing mode, perhaps because a pusher motor, airspeed sensor or control surface failed.
-If triggered, the vehicle will immediately switch to hover mode and execute action defined in parameter [COM_QC_ACT](../advanced_config/parameter_reference.md#COM_QC_ACT) .
+Failsafe for when a VTOL vehicle can no longer fly in fixed-wing mode, perhaps because a pusher motor, airspeed sensor, or control surface, failed.
+If triggered, the vehicle will immediately switch to Hold mode and execute action defined in parameter [COM_QC_ACT](#COM_QC_ACT) .
 
 :::note
 The quad-chute can also be triggered by sending a MAVLINK [MAV_CMD_DO_VTOL_TRANSITION](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_VTOL_TRANSITION) message with `param2` set to `1`.
@@ -252,14 +252,14 @@ The parameters that control when the quad-chute will trigger are listed in the t
 
 Parameter | Description
 --- | ---
-[COM_QC_ACT](../advanced_config/parameter_reference.md#COM_QC_ACT) | Defines automatic action taken after a quad-chute. Can be set to: Warning only, Return, Land, Hold.
-[VT_FW_QC_HMAX](../advanced_config/parameter_reference.md#VT_FW_QC_HMAX) |Maximum height above the ground (if available, otherwise above home if available, otherwise above the local origin) where triggering a quad-chute is possible. At high altitudes there is a big risk to deplete the battery and therefore crash if quad-chuting there.
-[VT_QC_HR_ERROR_I](../advanced_config/parameter_reference.md#VT_QC_HR_ERROR_I) | Quad-chute uncommanded descent threshold. This is the threshold for integrated height rate error to trigger a uncommanded-descent quad-chute. Only checked in altitude-controlled fixed-wing flight. Additional conditions that have to be met for uncommanded descent detection are a positive (climbing) height rate setpoint and a negative (sinking) current height rate estimate.
-[VT_QC_T_ALT_LOSS](../advanced_config/parameter_reference.md#VT_QC_T_ALT_LOSS) | Altitude loss threshold for quad-chute triggering during VTOL transition to fixed-wing
+<a id="COM_QC_ACT"></a>[COM_QC_ACT](../advanced_config/parameter_reference.md#COM_QC_ACT) | Defines automatic action taken after a quad-chute. Can be set to: Warning only, Return, Land, Hold.
+<a id="VT_FW_QC_HMAX"></a>[VT_FW_QC_HMAX](../advanced_config/parameter_reference.md#VT_FW_QC_HMAX) |Maximum height above the ground (if available, otherwise above home if available, otherwise above the local origin) where triggering a quad-chute is possible. At high altitudes there is a big risk to deplete the battery and therefore crash if quad-chuting there.
+<a id="VT_QC_HR_ERROR_I"></a>[VT_QC_HR_ERROR_I](../advanced_config/parameter_reference.md#VT_QC_HR_ERROR_I) | Quad-chute uncommanded descent threshold. This is the threshold for integrated height rate error to trigger a uncommanded-descent quad-chute. Only checked in altitude-controlled fixed-wing flight. Additional conditions that have to be met for uncommanded descent detection are a positive (climbing) height rate setpoint and a negative (sinking) current height rate estimate.
+<a id="VT_QC_T_ALT_LOSS"></a>[VT_QC_T_ALT_LOSS](../advanced_config/parameter_reference.md#VT_QC_T_ALT_LOSS) | Altitude loss threshold for quad-chute triggering during VTOL transition to fixed-wing
  flight. Quad-chute triggers the current altitude is more than this value below the altitude at the beginning of the transition.
-[VT_FW_MIN_ALT](../advanced_config/parameter_reference.md#VT_FW_MIN_ALT) | Minimum altitude above Home for fixed-wing flight. When the altitude drops below this value in fixed-wing flight the vehicle a quad-chute is triggered.
-[VT_FW_QC_R](../advanced_config/parameter_reference.md#VT_FW_QC_R) | Absolute roll threshold for quad-chute triggering in FW mode. 
-[VT_FW_QC_P](../advanced_config/parameter_reference.md#VT_FW_QC_P) | Absolute pitch threshold for quad-chute triggering in FW mode.
+<a id="VT_FW_MIN_ALT"></a>[VT_FW_MIN_ALT](../advanced_config/parameter_reference.md#VT_FW_MIN_ALT) | Minimum altitude above Home for fixed-wing flight. When the altitude drops below this value in fixed-wing flight the vehicle a quad-chute is triggered.
+<a id="VT_FW_QC_R"></a>[VT_FW_QC_R](../advanced_config/parameter_reference.md#VT_FW_QC_R) | Absolute roll threshold for quad-chute triggering in FW mode. 
+<a id="VT_FW_QC_P"></a>[VT_FW_QC_P](../advanced_config/parameter_reference.md#VT_FW_QC_P) | Absolute pitch threshold for quad-chute triggering in FW mode.
 
 
 ## Failure Detector
--- a/en/config/safety.md
+++ b/en/config/safety.md
@@ -242,8 +242,7 @@ Parameter | Description
 ### Quad-chute Failsafe
 
 Failsafe for when a VTOL vehicle can no longer fly in fixed-wing mode, perhaps because a pusher motor, airspeed sensor or control surface failed.
-If triggered, the vehicle will immediately switch to multicopter mode.
-If the vehicle was in [Mission mode](../flight_modes/mission.md) it enters failsafe [Return mode](../flight_modes/return.md).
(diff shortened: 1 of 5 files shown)


en/config/safety.md
```diff
@@ -239,24 +239,27 @@ Parameter | Description
 --- | ---
 [NAV_TRAFF_AVOID](../advanced_config/parameter_reference.md#NAV_TRAFF_AVOID) | Set the failsafe action: Disabled, Warn, Return mode, Land mode.
 
-### QuadChute Failsafe
+### Quad-chute Failsafe
 
 Failsafe for when a VTOL vehicle can no longer fly in fixed-wing mode, perhaps because a pusher motor, airspeed sensor or control surface failed.
```
hamishwillee, line 244:
```suggestion
Failsafe for when a VTOL vehicle can no longer fly in fixed-wing mode, perhaps because a pusher motor, airspeed sensor, or control surface, failed.
```

en/config/safety.md
```diff
@@ -239,24 +239,27 @@ Parameter | Description
...
+[VT_QC_T_ALT_LOSS](../advanced_config/parameter_reference.md#VT_QC_T_ALT_LOSS) | Altitude loss threshold for quad-chute triggering during VTOL transition to fixed-wing
+ flight. Quad-chute triggers the current altitude is more than this value below the altitude at the beginning of the transition.
+[VT_FW_MIN_ALT](../advanced_config/parameter_reference.md#VT_FW_MIN_ALT) | Minimum altitude above Home for fixed-wing flight. When the altitude drops below this value in fixed-wing flight the vehicle a quad-chute is triggered.
+[VT_FW_QC_R](../advanced_config/parameter_reference.md#VT_FW_QC_R) | Absolute roll threshold for quad-chute triggering in FW mode. 
+[VT_FW_QC_P](../advanced_config/parameter_reference.md#VT_FW_QC_P) | Absolute pitch threshold for quad-chute triggering in FW mode.
```
hamishwillee, line 262: FYI generally I like to keep context on the page, so I link to the in-page definition rather than the parameters. Hence an anchor for each.

```suggestion
<a id="COM_QC_ACT"></a>[COM_QC_ACT](../advanced_config/parameter_reference.md#COM_QC_ACT) | Defines automatic action taken after a quad-chute. Can be set to: Warning only, Return, Land, Hold.
<a id="VT_FW_QC_HMAX"></a>[VT_FW_QC_HMAX](../advanced_config/parameter_reference.md#VT_FW_QC_HMAX) |Maximum height above the ground (if available, otherwise above home if available, otherwise above the local origin) where triggering a quad-chute is possible. At high altitudes there is a big risk to deplete the battery and therefore crash if quad-chuting there.
<a id="VT_QC_HR_ERROR_I"></a>[VT_QC_HR_ERROR_I](../advanced_config/parameter_reference.md#VT_QC_HR_ERROR_I) | Quad-chute uncommanded descent threshold. This is the threshold for integrated height rate error to trigger a uncommanded-descent quad-chute. Only checked in altitude-controlled fixed-wing flight. Additional conditions that have to be met for uncommanded descent detection are a positive (climbing) height rate setpoint and a negative (sinking) current height rate estimate.
<a id="VT_QC_T_ALT_LOSS"></a>[VT_QC_T_ALT_LOSS](../advanced_config/parameter_reference.md#VT_QC_T_ALT_LOSS) | Altitude loss threshold for quad-chute triggering during VTOL transition to fixed-wing
 flight. Quad-chute triggers the current altitude is more than this value below the altitude at the beginning of the transition.
<a id="VT_FW_MIN_ALT"></a>[VT_FW_MIN_ALT](../advanced_config/parameter_reference.md#VT_FW_MIN_ALT) | Minimum altitude above Home for fixed-wing flight. When the altitude drops below this value in fixed-wing flight the vehicle a quad-chute is triggered.
<a id="VT_FW_QC_R"></a>[VT_FW_QC_R](../advanced_config/parameter_reference.md#VT_FW_QC_R) | Absolute roll threshold for quad-chute triggering in FW mode. 
<a id="VT_FW_QC_P"></a>[VT_FW_QC_P](../advanced_config/parameter_reference.md#VT_FW_QC_P) | Absolute pitch threshold for quad-chute triggering in FW mode.
```

en/config/safety.md
```diff
@@ -239,24 +239,27 @@ Parameter | Description
 --- | ---
 [NAV_TRAFF_AVOID](../advanced_config/parameter_reference.md#NAV_TRAFF_AVOID) | Set the failsafe action: Disabled, Warn, Return mode, Land mode.
 
-### QuadChute Failsafe
+### Quad-chute Failsafe
 
 Failsafe for when a VTOL vehicle can no longer fly in fixed-wing mode, perhaps because a pusher motor, airspeed sensor or control surface failed.
-If triggered, the vehicle will immediately switch to multicopter mode.
-If the vehicle was in [Mission mode](../flight_modes/mission.md) it enters failsafe [Return mode](../flight_modes/return.md).
+If triggered, the vehicle will immediately switch to hover mode and execute action defined in parameter [COM_QC_ACT](../advanced_config/parameter_reference.md#COM_QC_ACT) .
```
hamishwillee, line 245: @sfuhrer I plan to merge all this, but note this change from hover mode to hold mode. I don't know what hover mode is, so if this is actually correct, then let me know and we can revert.
```suggestion
If triggered, the vehicle will immediately switch to Hold mode and execute action defined in parameter [COM_QC_ACT](#COM_QC_ACT) .
```

en/config_vtol/vtol_without_airspeed_sensor.md
```diff
@@ -62,7 +62,7 @@ Set the minimum front transition time ([VT_TRANS_MIN_TM](../advanced_config/para
 
 ### Optional Recommended Parameters
 
-Because the risk of stalling is real, it is recommended to set the 'fixed wing minimum altitude' aka 'QuadChute' ([VT_FW_MIN_ALT](../advanced_config/parameter_reference.md#VT_FW_MIN_ALT)). 
+Because the risk of stalling is real, it is recommended to set the 'fixed wing minimum altitude' aka 'Quad-chute' ([VT_FW_MIN_ALT](../advanced_config/parameter_reference.md#VT_FW_MIN_ALT)).
```
hamishwillee, line 65: @sfuhrer Does this sentence make sense - specifically it looks like it is missing a word:
```suggestion
Because the risk of stalling is real, it is recommended to set the 'fixed wing minimum altitude' aka 'quad-chute' threshold ([VT_FW_MIN_ALT](../advanced_config/parameter_reference.md#VT_FW_MIN_ALT)). 

```

en/config_vtol/vtol_without_airspeed_sensor.md
```diff
@@ -71,7 +71,7 @@ The position estimator tested for this mode is EKF2, you can set this by changin
 ## First Flight Without Airspeed Sensor
 
 The values apply to a position controlled flight (like [Hold mode](../flight_modes/hold.md) or [Mission mode](../flight_modes/mission.md) or Mission mode). 
-It is therefore recommended that a mission is configured at a safe altitude, approximately 10m above the QuadChute threshold. 
+It is therefore recommended that a mission is configured at a safe altitude, approximately 10m above the Quad-chute threshold.
```
hamishwillee, line 74:
```suggestion
It is therefore recommended that a mission is configured at a safe altitude, approximately 10m above the quad-chute threshold. 

```

en/flight_modes/hold.md
```diff
@@ -28,7 +28,7 @@ The behaviour can be configured using the parameters below.
 
 Parameter | Description
 --- | ---
-<span id="MIS_LTRMIN_ALT"></span>[MIS_LTRMIN_ALT](../advanced_config/parameter_reference.md#MIS_LTRMIN_ALT) | Minimum height for loiter mode (vehicle will ascend to this altitude if mode is engaged at a lower altitude).
+<span id="NAV_MIN_LTR_ALT"></span>[NAV_MIN_LTR_ALT](../advanced_config/parameter_reference.md#NAV_MIN_LTR_ALT) | This is the minimum altitude above Home the system will always obey in Loiter (Hold) mode if switched into this mode without specifying an altitude (e.g. through Loiter switch on RC).
```
hamishwillee, line 31:
```suggestion
<a id="NAV_MIN_LTR_ALT"></a>[NAV_MIN_LTR_ALT](../advanced_config/parameter_reference.md#NAV_MIN_LTR_ALT) | This is the minimum altitude above Home the system will always obey in Loiter (Hold) mode if switched into this mode without specifying an altitude (e.g. through Loiter switch on RC).

```

en/flight_modes/takeoff.md
```diff
@@ -79,7 +79,7 @@ The _launch detector_ is affected by the following parameters:
 
 Parameter | Description
 --- | ---
-<span id="LAUN_ALL_ON"></span>[LAUN_ALL_ON](../advanced_config/parameter_reference.md#LAUN_ALL_ON) | Enable automatic launch detection. If disabled motors spin up on arming already
+<span id="FW_LAUN_DETCN_ON"></span>[FW_LAUN_DETCN_ON](../advanced_config/parameter_reference.md#FW_LAUN_DETCN_ON) | Enable automatic launch detection. If disabled motors spin up on arming already
 <span id="FW_LAUN_AC_THLD"></span>[FW_LAUN_AC_THLD](../advanced_config/parameter_reference.md#FW_LAUN_AC_THLD) | Acceleration threshold (acceleration in body-forward direction must be above this value)
 <span id="FW_LAUN_AC_T"></span>[FW_LAUN_AC_T](../advanced_config/parameter_reference.md#FW_LAUN_AC_T) | Trigger time (acceleration must be above threshold for this amount of seconds)
 <span id="FW_LAUN_MOT_DEL"></span>[FW_LAUN_MOT_DEL](../advanced_config/parameter_reference.md#FW_LAUN_MOT_DEL) | Delay from launch detection to motor spin up
```
hamishwillee, line 85: Just improve link anchors. Ignore.

```suggestion
<a id="FW_LAUN_DETCN_ON"></a>[FW_LAUN_DETCN_ON](../advanced_config/parameter_reference.md#FW_LAUN_DETCN_ON) | Enable automatic launch detection. If disabled motors spin up on arming already
<a id="FW_LAUN_AC_THLD"></a>[FW_LAUN_AC_THLD](../advanced_config/parameter_reference.md#FW_LAUN_AC_THLD) | Acceleration threshold (acceleration in body-forward direction must be above this value)
<a id="FW_LAUN_AC_T"></a>[FW_LAUN_AC_T](../advanced_config/parameter_reference.md#FW_LAUN_AC_T) | Trigger time (acceleration must be above threshold for this amount of seconds)
<a id="FW_LAUN_MOT_DEL"></a>[FW_LAUN_MOT_DEL](../advanced_config/parameter_reference.md#FW_LAUN_MOT_DEL) | Delay from launch detection to motor spin up

```

hamishwillee:
@sfuhrer Thanks very much for this, and for being so fast. I have made some mostly minor changes. The only two which need concern you were merged, but I have unresolved them above so you can easily check them - such as https://github.com/PX4/PX4-user_guide/pull/2364#pullrequestreview-1353529877

en/config/safety.md
sfuhrer: yeah that hover was correct, Hold is wrong. We need to settle on a consistent naming - I usually refer to the VTOL states to "Hover" (instead of "MC") and "Aerodynamic forward flight" (instead of "FW"). For "Hover" we are quite well aligned in the team, though for how we want to call the fixed-wing/forwrad/cruise/aerodynamic flight we're not aligned. 
Also currently we mostly refer to them as MC and FW modes in the docs, so for now I guess we should keep to that.

en/config_vtol/vtol_without_airspeed_sensor.md
sfuhrer: I would actually not refer to a specific quad-chute threshold here, just add a link to the quad-chute sections where all the different thresholds are defined.

en/config/safety.md
hamishwillee: So we are absolutely consistent in the docs - Hold mode for position locked in MC and for path locked in FW. In MAVLink the mode is called cruise mode for FW I think, and we should perhaps use that.

Hover would be different, because to me that does not imply "locked in place". But up to you guys to decide and I can make it so. But it has to happen end to end and affect QGC too.

sfuhrer: No, for me "Hover" = MC. We also have VTOLs with only one one propeller, so they obviously can never be in "Multicopter" mode. 
For the change here: lets just change it to:
If triggered, the vehicle will immediately switch to **MC** mode and execute action defined in parameter [COM_QC_ACT](#COM_QC_ACT) .

hamishwillee: If you can't hold position in your MC I'd call it altitude hold mode: http://docs.px4.io/main/en/flight_modes/altitude_mc.html

I guess the point is what the user will see for the mode in QGC so they don't have a confusing perception.
Calling it MC mode here is fine though.

Can you make this and the other change?

hamishwillee: If not, ping me - might get to it before next week. Might now!

PR #2385, opened 2023-03-26
Update bootloader_update_from_betaflight.md

jwarnergithub:
Verified this command syntax was incorrect when flashing an H7 Mini v1.3 today. Updated with the working syntax.


hamishwillee:
Thanks very much. Closing as I merged another PR with this same information #2389 . YOu were first, but looking at my queue in reverse order.

PR #2386, opened 2023-03-27
holybro x500v2: add note about actuator assignment

bkueng:
From https://github.com/PX4/PX4-Autopilot/pull/21349


hamishwillee (approved):
Thanks @bkueng

PR #2392, opened 2023-03-28
Mission: fix L1 param names (replaced by NPFG with same meaning)

sfuhrer:
@tstastny I guess this documentation is also valid for NPFG?


hamishwillee (approved):
Thanks very much!

PR #2393, opened 2023-03-28
Update and rename adding_a_new_frame.md babyshark config

sfuhrer:


hamishwillee:
@sfuhrer FYI, this broke the build, because the summary links to it.

More important, was it it intentional? The topic is "# Adding a Frame Configuration" - why would you move it?

hamishwillee:
PS I'm moving it back!

sfuhrer:
The renaming wasn't intentional, thanks for fixing it!

hamishwillee:
No worries. Easily done. Thanks so much for all these fixes - docs starting to look accurate again.

PR #2395, opened 2023-03-28
FW/VTOL param fixes

sfuhrer:
- Flaps/Spoiler trimming has changed on PX4
- VTOL backtransition tuning was massively outdated.


en/config_fw/trimming_guide_fixedwing.md
```diff
@@ -43,14 +43,10 @@ Given that the downward pitch moment induced by an asymmetric airfoil increases
 For this purpose, a bilinear curve (see figure below) function of airspeed and a pitch trim increment function of the flaps state can be defined using the following parameters:
 
 - [FW\_DTRIM\_\[R/P/Y\]_\[VMIN/VMAX\]](../advanced_config/parameter_reference.md#FW_DTRIM_R_VMIN) are the roll/pitch/yaw trim value added to `TRIM_ROLL/PITCH/YAW` at min/max airspeed (defined by [FW_AIRSPD_MIN](../advanced_config/parameter_reference.md#FW_AIRSPD_MIN) and [FW_AIRSPD_MAX](../advanced_config/parameter_reference.md#FW_AIRSPD_MAX)).
-- [FW_DTRIM_P_FLPS](../advanced_config/parameter_reference.md#FW_DTRIM_R_FLPS) and [FW_DTRIM_P_FLPS](../advanced_config/parameter_reference.md#FW_DTRIM_P_FLPS) are the roll/pitch trim value added to `TRIM_ROLL/PITCH/YAW` when the flaps are deployed.
+- [CA_SV_CSx_FLAP](../advanced_config/parameter_reference.md#CA_SV_CS0_FLAP) and [CA_SV_CSx_SPOIL](../advanced_config/parameter_reference.md#CA_SV_CS0_SPOIL) are the trimming values that are applied to this controls surfaces if the flaps respectively the spoilers are fully deployed.
```
hamishwillee, line 46:
```suggestion
- [CA_SV_CSx_FLAP](../advanced_config/parameter_reference.md#CA_SV_CS0_FLAP) and [CA_SV_CSx_SPOIL](../advanced_config/parameter_reference.md#CA_SV_CS0_SPOIL) are the trimming values that are applied to these control surfaces if the flaps or the spoilers are fully deployed, respectively. 
```

en/config_fw/trimming_guide_fixedwing.md
```diff
@@ -43,14 +43,10 @@ Given that the downward pitch moment induced by an asymmetric airfoil increases
 For this purpose, a bilinear curve (see figure below) function of airspeed and a pitch trim increment function of the flaps state can be defined using the following parameters:
```
hamishwillee, line 43:
```suggestion
For this purpose, a bilinear curve function of airspeed and a pitch trim increment function of the flaps state (see figure below) can be defined using the following parameters:
```

hamishwillee:
@sfuhrer Thank so much for taking care of this, and the other open issues.

This is the reason I love parameter links - they highlight old content as it gets stripped out. I did a minor subedit and removed another note about PX4v1.7. Generally speaking I think we should remove all notes prior to v1.10, and consider even if the v1.10 ones are needed.

One question, you took a lot of old crap about the backtransition out - is it just not needed? I.e. is this all that needs to be said?

Merging in any case.

sfuhrer:
>you took a lot of old crap about the backtransition out - is it just not needed

The big part of that was removing the docs about reversing thrust during backtransition, which is a feature that we no longer support. The new way of having the vehicle pitch up to break proved to be sufficient for all the vehicles I know. And it comes with less hardware complexity and safety risk. 
The "typical setup" section wasn't adding any value - we put the most reasonable parameter settings already as defaults.

en/config_vtol/vtol_back_transition_tuning.md
```diff
@@ -1,57 +1,19 @@
 # VTOL Back-transition Tuning
 
 When a VTOL performs a back-transition (transition from fixed wing mode to multicopter) it needs to slow down before the multicopter can take proper control.
```
hamishwillee, line 3: > The big part of that was removing the docs about reversing thrust during backtransition, which is a feature that we no longer support. The new way of having the vehicle pitch up to break proved to be sufficient for all the vehicles I know. And it comes with less hardware complexity and safety risk.

@sfuhrer Thanks for the detailed explanation - makes sense.

Just a though, but it seems to me that some people might wonder **how** you back transition, and possibly why you don't use reverse thrust.

Do you think it is worth adding a note above this line "like":

> VTOL supports back transition to multicopter using vehicle pitch-up to brake.
> This has proved sufficient for all vehicle types tested.

sfuhrer: Currently I kind of covered that with

> To help with breaking, the controller will pitch up the vehicle if the current deceleration is below what is set in expected deceleration ([VT_B_DEC_MSS]

But please feel free to add your comment, it's correct.

hamishwillee: It is sufficient ... and yet I missed it. I've proposed an alternative in https://github.com/PX4/PX4-user_guide/pull/2422 that might be OK.

PR #2400, opened 2023-03-29
Remove UAVCAN_ESC_IDLT

hamishwillee:
This removes the UAVCAN_ESC_IDLT param, gone in PX4 v1.14.

@bkueng I have assumed that instead of using this you simply run actuator tests and set the minimum output values appropriately - right?


bkueng:
Correct

PR #2404, opened 2023-03-29
Typo on reference frame in desired control quantity

federicociresola:
After some research, I think I discovered a typo on this page on the reference frame of the column "desired control quality".  I noticed it when I read the page of PX4 terminology (https://docs.px4.io/main/en/contribute/notation.html#acronyms) because I wasn't understanding what RFD meant. 
Then I changed the typo from RFD to FRD on the whole page.

I hope It is right
 
Signed-off-by: Federico Ciresola <cirefede@gmail.com>


beniaminopozzan:
thanks @federicociresola !

hamishwillee:
These sorts of typos creep in so easily - thanks for taking the time to fix it for everyone.

PR #2405, opened 2023-03-29
Big fat warnings about using XRCE-DDS to PX4 v1.13

hamishwillee:
We're getting support requests. If people come to this from PX4 v1.13 docs they will get a warning.


mrpollo:
Thanks @hamishwillee

PR #2408, opened 2023-03-29
Move compass config under calibration to make it easier to find

hamishwillee:
This covers disabling internal compasses.

@AlexKlimaj Can you add a suggestion to correct this for multi EKF?


hamishwillee:
Merging anyway as structure is better

PR #2410, opened 2023-03-30
Update uorb top documentation re: Mavlink Console

msberk:
Edit the `uorb top` documentation to reflect that the command does not work reliably in the Mavlink Console. This is in response to @AlexKlimaj's comment on Discord recently after I was unable to run `uorb top` over the Mavlink Console on a Pixhawk 6X running PX4 1.13.2.

If this is still applicable to main then this change should likely be applied there, too.

### DIFF
--- a/en/middleware/uorb.md
+++ b/en/middleware/uorb.md
@@ -102,6 +102,10 @@ For more information see: [Sensor/Topic Debugging](../debug/sensor_uorb_topic_de
 
 ### uorb top Command
 
+:::note
+`uorb top` has been known to have issues when operating over the Mavlink Console, and may not start. The [PX4 System Console](../debug/system_console.md) is generally more reliable when using `uorb top`
+:::
+
 The command `uorb top` shows the publishing frequency of each topic in real-time:
 
 ```sh
@@ -122,7 +126,6 @@ sensor_baro                          0    1   42     0 1
 sensor_combined                      0    6  242   636 1
 ```
 The columns are: topic name, multi-instance index, number of subscribers, publishing frequency in Hz, number of lost messages per second (for all subscribers combined), and queue size.
-
 
 ## Multi-instance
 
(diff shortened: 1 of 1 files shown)


mrpollo:
Hey @dagar is this a bug or expected behavior?

msberk:
I guess whether it's a bug or expected behavior, it still seems like it'd be useful to note that it does not work over the Mavlink console.

hamishwillee:
@msberk I agree this is worth having a note :-), but whether or not it is a bug or not defines what the comment looks like.

For example, I like to add a bug number so that this kind of note can easily be checked and removed when the bug is fixed. The [conversation with alex](https://discord.com/channels/1022170275984457759/1032055573753118742/1090763402076487720) just says it isn't working in this case, but I can't find an open issue or bug report I can link against.

I did try this locally and it works for me over USB to a CUAV Pixhawk 6x with latest main.

Can you provide more detail. If not we can either create an issue for tracking and link it - or perhaps water-down this to "anecdotally".

hamishwillee:
What I don't want to do is have an instruction that means people have to run into this problem forever. MAVLink console is too useful for this not to work.

msberk:
@hamishwillee sorry if I seemed short in the reply - I misinterpreted the comment to imply that if it was expected it may not need to be edited. The bug report linking makes sense for being able to track and remove notes such as this.

Glad to hear the issue doesn't seem to exist in `main` on the CUAV. The hardware is a Holybro 6X, I am chasing down another issue re: low servo output rates using SBus and going to be uploading a modified 1.13.3 firmware in the next couple days anyway, so while I'm doing that I can check to see if `uorb top` works on `main` on my hardware and comment back here.

hamishwillee:
That would be great. If we can get a specific reproducible case then we can create an issue and move forward.

hamishwillee:
Hi @msberk Did you manage to get any further testing done on this?

hamishwillee:
THanks, and no worries. I'm on other things too :-)

hamishwillee:
Closing as stale. Happy to discuss again if it can be reproduced, but perhaps we would start with a PX4 issue report.

PR #2411, opened 2023-03-31
Update pixhawk4.md

jwarnergithub:
Added (link broken) next to 3 links that don't work after Holybro updated their web site on 3-16-23. Pinged Holybro in Discord so they are aware.


jwarnergithub:
Found a list of files. Will create a new PR with updated links.

PR #2416, opened 2023-04-03
Update Cube Orange Page Detail

chrisksl:
Correct processor from STM32H743ZI to STM32H753VI
Remove Debug Port as it is not available on the latest ADSB Carrier Board


julianoes (approved):
Great, thanks for that.

proficnc:
looks good.

hamishwillee:
Hi @chrisksl,

The text changes are good but you have also added two (very large) images that you then do not use. Please delete the images.

Note, if you were to add these images then you would need to move them into the `\assets\flight_controller\cube\orangeplus\` directory and shrink them significantly in terms of file size.

I attached some reduced size versions FYI
ports_usb_sdcard
ports_top_main

Cheers H

hamishwillee:
PS, I'd fix this myself but your fork appears to be blocked against allowing changes, even in PRs.

hamishwillee:
I've taken the content changes in #2425. The images are too big and not used, so not needed in this PR. Closing.

PR #2422, opened 2023-04-05
vtol_back_transition_tuning - fix typo restructure

hamishwillee:
@sfuhrer This is a slight restructure to the back transition tuning in response to https://github.com/PX4/PX4-user_guide/pull/2395/files#r1152822464

This moves discussion of braking first ahead of transition duration in both the intro and the sections. This makes it more obvious that you pitch to brake.

This might not be optimal. For example, if tuning the duration is far more important than the braking it should go first, and my changes rejected.

Either way though, if you hate this, I will have to fix the typo `breaking` > `braking`..


sfuhrer (approved):
No I do like that, this is better than before, thanks!

PR #2427, opened 2023-04-06
gdb command line not working unless given a model

afwilkin:
I noticed that the current commands to build and attach gdb were not working. I tried to change the PX4-Autopilot build process, but it kept breaking it. I thought for the mean time this might be a quick way to make sure users can do what they need to do

### DIFF
--- a/de/debug/simulation_debugging.md
+++ b/de/debug/simulation_debugging.md
@@ -31,16 +31,12 @@ make px4_sitl_default jmavsim___valgrind
 
 ## Start combinations
 
-SITL can be launched with and without debugger attached and with either jMAVSim or Gazebo Classic as simulation backend. This results in the start options below:
+SITL can be launched with and without debugger attached with Gazebo Classic as simulation backend. This results in the start options below:
 
 ```sh
-make px4_sitl_default jmavsim
-make px4_sitl_default jmavsim___gdb
-make px4_sitl_default jmavsim___lldb
-
 make px4_sitl_default gazebo-classic
-make px4_sitl_default gazebo-classic___gdb
-make px4_sitl_default gazebo-classic___lldb
+make px4_sitl_default gazebo-classic_iris_gdb
+make px4_sitl_default gazebo-classic_iris_lldb
 ```
 
 where the last parameter is the &lt;viewer\_model\_debugger&gt; triplet (using three underscores implies the default 'iris' model). This will start the debugger and launch the SITL application. In order to break into the debugger shell and halt the execution, hit `CTRL-C`:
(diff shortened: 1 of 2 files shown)


de/debug/simulation_debugging.md
```diff
@@ -31,16 +31,12 @@ make px4_sitl_default jmavsim___valgrind
 
 ## Start combinations
 
-SITL can be launched with and without debugger attached and with either jMAVSim or Gazebo Classic as simulation backend. This results in the start options below:
+SITL can be launched with and without debugger attached with Gazebo Classic as simulation backend. This results in the start options below:
```
hamishwillee, line 34: See this file path is /de. That means it is a translation file. Unfortunately that means this can't be merged. We will need to change the same file in the /en path.

en/debug/simulation_debugging.md
```diff
@@ -31,22 +31,15 @@ make px4_sitl_default jmavsim___valgrind
...
 
 ```sh
-make px4_sitl_default jmavsim
-make px4_sitl_default jmavsim___gdb
-make px4_sitl_default jmavsim___lldb
-
 make px4_sitl_default gazebo-classic
-make px4_sitl_default gazebo-classic___gdb
-make px4_sitl_default gazebo-classic___lldb
+make px4_sitl_default gazebo-classic_iris_gdb
```
hamishwillee, line 38: @afwilkin Thanks for this. We definitely need to do something, but exactly what depends a bit on what is going to happen in future and what is supported now. 

@beniaminopozzan @Jaeyoung-Lim  Do you know if
- there an intent to bring back the vehicle-less build targets, and 
- whether it is going to happen for both Gazebo Classic and JMAVSim (and Gazebo?).
- there are JMAVSim gdb/lldb targets for individual vehicles? 
- Do all vehicles have them?

@beniaminopozzan I'd just check if `list_vmd_make_targets` if it still existed.

If we can answer some of that, knowing if this is a good approach will be a lot easier. I'm hesitant to delete options without knowing if they are coming back, and at least being able to say what is going to happen in the docs.

de/debug/simulation_debugging.md
afwilkin: thanks for explaining that to me. I was slightly confused by that. Will back it out

en/debug/simulation_debugging.md
afwilkin: totally makes sense. I just ran into the issue when trying to do it, so I thought I would add some information. 

What about a note that at the current moment you need to add a vehicle name in order for these gazebo sim to work?

afwilkin: when i tried to ripgrep I could not find anything on `list_vmd_make_targets` command

`atlas:~/px4/main/PX4-Autopilot$ rg list_vmd_make_targets
atlas:~/px4/main/PX4-Autopilot$ rg list_vmd
atlas:~/px4/main/PX4-Autopilot$ 
`
But then again I do not 100% know how that command is formed

hamishwillee: Ah, sorry about that. I was whinging to @beniaminopozzan about the fact that this command USED to exist, and was very very useful. It got removed silently (from my perspective) .

hamishwillee: > What about a note that at the current moment you need to add a vehicle name in order for these gazebo sim to work?

Yes, that would be simpler/safer for now.

The `list_vmd_make_targets` can also be removed as it isn't planned to come back (yet, though I haven't given up).

Would still be good to know the plan for these other targets, but the note would be enough to help people for now.

afwilkin:
@hamishwillee I just reverted everything, and added a note. Let me know if this makes sense or if you need me to change anything more

afwilkin:
> @afwilkin I'm going to close this because I can't edit it. On further consideration I think your original instincts were good here so I have made the edits I think are required in #2446 . Essentially to document the situation as it is now.
> 
> Can you please sanity check that?

Totally. Apologies as I made this PR from a business account instead of personal, so it wont let you edit. In the future I will make it from my personal account

hamishwillee:
Thanks very much - all good, and thanks for your patience with me changing tack a bit.

PR #2435, opened 2023-04-06
Gripper.md: remove robocloaw reference

sfuhrer:
As the roboclaw driver was removed from the firmware


hamishwillee:
Thanks very much - also gone in PX4 v1.14? If not, then we should leave this in until I take the version cut for v1.14.

sfuhrer:
It's also in 1.14.

hamishwillee:
Great. And thanks for taking the time to remove this.

PR #2438, opened 2023-04-06
Update gimbal_control.md

danielhonies:
I think the command was renamed. vmount is not available anymore and 'gimbal' command returns the expected behaviour.

### DIFF
--- a/en/advanced/gimbal_control.md
+++ b/en/advanced/gimbal_control.md
@@ -85,7 +85,7 @@ To run it, use:
 make px4_sitl gazebo-classic_typhoon_h480
 ```
 
-To just test the mount driver on other models or simulators, make sure the driver runs (using `gimbal start`), then configure its parameters.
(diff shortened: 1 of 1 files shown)


en/advanced/gimbal_control.md
```diff
@@ -85,11 +85,11 @@ To run it, use:
 make px4_sitl gazebo-classic_typhoon_h480
 ```
 
-To just test the mount driver on other models or simulators, make sure the driver runs (using `vmount start`), then configure its parameters.
+To just test the mount driver on other models or simulators, make sure the driver runs (using `gimbal start`), then configure its parameters.
```
hamishwillee, line 88:
```suggestion
To just test the [gimbal driver](../modules/modules_driver.md#gimbal) on other models or simulators, make sure the driver runs (using `gimbal start`), then configure its parameters.
```

hamishwillee:
Yes it does. Thank you very much!

PR #2442, opened 2023-04-11
Update Pixhawk_series.md to add new FC

vincentpoont2:
Add Holybro FMUv6X and FMUv6C FC to Pixhawk_series.md


hamishwillee (commented):
Thank! We might need to restructure this a bit so that all these things only need to get updated in  one place - too easy to miss these things.

hamishwillee:
Actually, I have deleted this whole list in https://github.com/PX4/PX4-user_guide/pull/2445

PR #2444, opened 2023-04-12
Add missing driver modules into sidebar

hamishwillee:


hamishwillee:
PS Note to me, going to add a flaw checker to catch these earlier.

PR #2446, opened 2023-04-12
Simulation debugging fixes - only able to do debugger with model

hamishwillee:
Replaces #2427

Many thanks to @afwilkin


hamishwillee:
Merging this as I know the changes to be "more accurate" and less frustrating than the ones previously. Note that I did not regenerate the terminal output you'd get actually doing the debugging.

PR #2448, opened 2023-04-12
Mission.md: remove reference to MIS_DIST_WPS

sfuhrer:
Was removed on the PX4 side with https://github.com/PX4/PX4-Autopilot/pull/21262.

### DIFF
--- a/en/flight_modes/mission.md
+++ b/en/flight_modes/mission.md
@@ -76,7 +76,6 @@ If any of the checks fail, the user is notified and it is not possible to start 
 A subset of the most important checks are listed below:
 
 - First mission item too far away from vehicle ([MIS_DIST_1WP](#MIS_DIST_1WP))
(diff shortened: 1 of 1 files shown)


hamishwillee:
Thanks @sfuhrer - appreciate you fixing this stuff!

BTW, would it be a (lot/little) (easier/harder/the same) for you to do this kind of fix if the docs source code was inside PX4-Autopilot repo? There is some pull for moving the source across. I can see how if you're doing a search replace on all instances of "some parameter" you might be more likely to at least notice there is a docs impact if it is present.

I'm not opposed in theory. Just quite a bit of work for me, so I want to get a sense of the benefit.

sfuhrer:
Yes I generally like the idea, but haven't thought through all the side effects of it. Would enable is to even run a CI check to catch removed PX4 params in the docs part. And we could block the merging of new features without having docs in parallel.

hamishwillee:
Yes, that is a possible benefit - though I suspect no one has ever blocked code going in based on missing docs before, and we won't be the first.

PR #2449, opened 2023-04-12
PX4Flow Warning

mrpollo:
From a Community Q&A

https://discuss.px4.io/t/px4-community-q-a-april-12-2023/31552

### DIFF
--- a/en/sensor/px4flow.md
+++ b/en/sensor/px4flow.md
@@ -1,5 +1,9 @@
 # PX4FLOW Smart Camera
 
+:::warning
+The PX4 Team does not recomend the PX4FLOW for new designs.
+:::
+
 PX4FLOW is an [optical flow](../sensor/optical_flow.md) smart camera.
 It has a native resolution of 752x480 pixels and calculates optical flow on a 4x binned and cropped area at 400 Hz, giving it a very high light sensitivity.
 


en/sensor/px4flow.md
```diff
@@ -1,5 +1,9 @@
 # PX4FLOW Smart Camera
 
+:::warning
+The PX4 Team does not recomend the PX4FLOW for new designs.
```
hamishwillee, line 4: @mcsauder @mrpollo A note is great, but this isn't actually all that helpful because as a reader I have no idea whether this is a transient problem, or something I might fix in another way. For example, if this is a PX4 problem I might want to revert to PX4 v1.13 - would it work then? Will my vehicle stop working if I upgrade to this version?

What I normally do in this case is reference a PX4-Autopilot issue - that makes it clear what the problem is, and users can see from that what versions it applies to, and whether it is likely to be fixed. If it gets closed then they know the answer is "no". 

Can we have that kind of link here. Also fix the typo "recommend"?

mcsauder: Hey @hamishwillee ! Please advise on this.  The consensus after I worked with this sensor for the past few weeks and everyone on yesterday's call and last week's was that this sensor should no longer be supported but we need a polite way to indicate this.  The firmware on the sensor appears to be broken, and the firmware on the PX4 side might also not be quite working.  Either way, at present, it is entirely dysfunctional.  Your advice is very welcomed!

hamishwillee: @mcsauder Thanks very much. 

My take on this is that if we know it doesn't work we try to help people not waste time on it by making that clear.
Further, if we know now that even if the firmware was fixed it probably wouldn't work and we won't do anything to make it work, then we should removed the docs.

So what I have done in https://github.com/PX4/PX4-user_guide/pull/2463 is replace the content in the page with a link to the legacy docs. It looks like this:

image

I assume it might work if you can get legacy firmware and older PX4? If not, I'd just say "not supported".

If the situation was that we intended to support this in future then I'd have a note that indicates it does not work and a link to the PX4 Autopilot issue tracking it. 

Does that make sense?

Note that older docs won't make this clear, which is a pity. But we can't maintain everything. I have a possible solution for this that doesn't require "much" maintenance. Not imminent though.

PR #2450, opened 2023-04-12
Actuators.md: add flap/spoiler confiugration docs

sfuhrer:
To cover the changes introduced in https://github.com/PX4/PX4-Autopilot/pull/20815.


en/config/actuators.md
```diff
@@ -140,32 +140,59 @@ Note that you will need to also ensure that the ESC associated with bidirectiona
...
 - `Lock control surfaces in hover`:
   - `Enabled`: Most vehicles do not use control surfaces in hover. Use this setting to lock them so that they don't affect vehicle dynamics.
   - `Disabled`: Set this for vehicles that use control surfaces in hover, such as the duo tailsitter (which uses elevons for pitch and yaw control). It should also be set for vehicles that use control surfaces to provide additional stabilization in hover mode when moving at speed or in high winds.
 
+#### Flaps and Spoilers Configuration
+
+`Flap-control` and `spoiler-control` are aerodynamic configurations that can either be commanded manually by the pilot through a switch on the RC, or are set automatically by the controller, e.g. to reduce the airspeed before landing. 
+The flap and spoiler configuration controls are of the range of [0, 1] and can then be arbitrarily allocated to the control surfaces.
+It's important to understand the difference: with `flap/spoiler-control` we refer to an abstract metric of "how much flap/spoiler configuration" the controller currently demands, while the actual flaps or spoiler actuator setpoint can vary from that value.
```
hamishwillee, line 181:
```suggestion
The flap and spoiler configuration controls are of the range of [0, 1] and can then be arbitrarily allocated to the control surfaces.

It's important to understand the difference: with `flap/spoiler-control` we refer to an abstract metric of "how much flap/spoiler configuration" the controller currently demands, while the actual flaps or spoiler actuator setpoint can vary from that value.
```

en/config/actuators.md
```diff
@@ -140,32 +140,60 @@ Note that you will need to also ensure that the ESC associated with bidirectiona
...
 - `Lock control surfaces in hover`:
   - `Enabled`: Most vehicles do not use control surfaces in hover. Use this setting to lock them so that they don't affect vehicle dynamics.
   - `Disabled`: Set this for vehicles that use control surfaces in hover, such as the duo tailsitter (which uses elevons for pitch and yaw control). It should also be set for vehicles that use control surfaces to provide additional stabilization in hover mode when moving at speed or in high winds.
 
+#### Flaps and Spoilers Configuration
+
+`Flap-control` and `spoiler-control` are aerodynamic configurations that can either be commanded manually by the pilot through a switch on the RC, or are set automatically by the controller, e.g. to reduce the airspeed before landing. 
+The flap and spoiler configuration controls are of the range of [0, 1] and can then be arbitrarily allocated to the control surfaces.
+
+It's important to understand the difference: with `flap/spoiler-control` we refer to an abstract metric of "how much flap/spoiler configuration" the controller currently demands, while the actual flaps or spoiler actuator setpoint can vary from that value.
```
hamishwillee, line 182: @sfuhrer I don't understand what you're saying here - the difference "to what".

I _think_ you might be trying to point out that the deflections indicate the maximum deflection that would be requested in the flap-control or spoiler-control configuration, BUT that this does not mean that this will be the actual deflection of the control surface. Instead, the deflection will be a combination of the requested deflection for flap/spoiler control and whatever other things are requested by the control loops for the control surface - such as yaw. 

Can you modify this to clarify, unless I'm just be thick?

en/config/actuators.md
```diff
@@ -140,32 +140,60 @@ Note that you will need to also ensure that the ESC associated with bidirectiona
...
+Only the most common settings are displayed by default.
+Select the **Advanced** checkbox in the top right corner of the view to display all settings.
+:::
+
 The fields are:
 
 - `Control Surfaces`: The number of control surfaces (set this first!)
 - `Type`: The type of each control surface: `LeftAileron`, `RightAileron`, `Elevator`, `Rudder`, `Left Elevon`, `Right Elevon`, `Left V-Tail`, `Right V-Tail`, `Left Flap`, `Right Flap`, `Airbrakes`, `Custom`.
-- `Roll scale`: Effectiveness of actuator around roll axis (normalised: -1 to 1).
+- `Roll Torque`: Effectiveness of actuator around roll axis (normalised: -1 to 1).
```
hamishwillee, line 157: I changed these from scale to Torque, as that is what I am now seeing in QGC and your image below. Is the definition still correct?

en/config/actuators.md
```diff
@@ -140,32 +140,60 @@ Note that you will need to also ensure that the ESC associated with bidirectiona
 The control surfaces section of the geometry panel lets you set the number and types of control surfaces that are present on the vehicle.
 You may also need to set trim and slew rate values in some cases.
 More advanced users can also configure the roll scale, yaw scale, and pitch scale (generally the defaults are acceptable, and this is not needed).
-
 An "example" control surface section for a vehicle with two ailerons is shown below.
 Note that ailerons only affect roll, so the pitch and yaw fields are disabled.
 
 ![Control Surface Setup Example](../../assets/config/actuators/control_surfaces_geometry.png)
```
hamishwillee, line 146: I updated this image (see above) to show both the advanced and basic cases. 
1. I couldn't get it to show lock control surfaces in hover - is that only for VTOL? What frame would I use to show this?
2. FYI the daily QGC hides any fields not needed by any of the displayed surfaces - so we'll have to update the images when daily becomes a release, and some of the text.

en/config/actuators.md
```diff
@@ -140,32 +140,60 @@ Note that you will need to also ensure that the ESC associated with bidirectiona
...
   [Generally you should use the default actuator value](#actuator-roll-pitch-and-yaw-scaling).
 - `Trim`: An offset added to the actuator so that it is centered without input.
   This might be determined by trial and error.
-- `Slew Rate`: Minimum time allowed for the motor/servo signal to pass through the full output range, in seconds.
+- (Advanced) `Slew Rate`: Minimum time allowed for the motor/servo signal to pass through the full output range, in seconds.
   - The setting limits the rate of change of an actuator (if not specified then no rate limit is applied).
      It is intended for actuators that may be damaged if they move too fast — such as the tilting actuators on a tiltrotor VTOL vehicle.
   - For example, a setting of 2.0 means that the motor/servo will not be commanded to move from 0 to 1 at a rate that completes the operation in less than 2 seconds (in case of reversible motors, the range is -1 to 1).
+- (Advanced) `Flap Scale`: How much this actuator is deflected at the "full flaps configuration" [0, 1] (see [Flaps and Spoilers Configuration](#flaps-and-spoilers-configuration) below).
```
hamishwillee, line 169: Is this in PX4 v1.14? I can't get them to display on whatever I have on my hardware (that won't flash), though I was able to get them in simulator.

en/config/actuators.md
```diff
@@ -140,32 +140,60 @@ Note that you will need to also ensure that the ESC associated with bidirectiona
...
+  Can be used to configure aerodynamic surface as flap or to compensate for generated torque through main flaps.
+- (Advanced) `Spoiler Scale`: How much this actuator is deflected at the "full spoiler configuration" [0, 1] (see [Flaps and Spoilers Configuration](#flaps-and-spoilers-configuration) below).
+  Can be used to configure aerodynamic surface as spoiler or to compensate for generated torque through main spoiler.
 - `Lock control surfaces in hover`:
   - `Enabled`: Most vehicles do not use control surfaces in hover. Use this setting to lock them so that they don't affect vehicle dynamics.
   - `Disabled`: Set this for vehicles that use control surfaces in hover, such as the duo tailsitter (which uses elevons for pitch and yaw control). It should also be set for vehicles that use control surfaces to provide additional stabilization in hover mode when moving at speed or in high winds.
 
+#### Flaps and Spoilers Configuration
```
hamishwillee, line 177: This looks good, except that you're configuring aileron, rudder/elevator, flaps.

We also have the option to configure spoiler as its own actuator (at least in my UI). SHould we at least mention this since we're talking about spoilers?

hamishwillee:
@sfuhrer Thanks very much. Looks good, but I have some questions, and there might need to be a few other updates as parts of the UI have changed.

en/config/actuators.md
sfuhrer: 1. Yes that's VTOL only. eg `make px4_sitl_default gazebo_standard_vtol`
2. okay :+1:

en/config/actuators.md
sfuhrer: yeah, "roll torque scale" would be better but too long, but I think people understand what it means with the help of the docs here. And ofc being consistent is the most important thing.

en/config/actuators.md
sfuhrer: >except that you're configuring aileron, rudder/elevator, flaps.

I'm not talking about flaps/spoilers actuators here, but flaps/spoiler control configuration..that's the thing I tried to make clear below.

And yes, mentioning here that dedicated spoiler actuators are also available would be good.

en/config/actuators.md
sfuhrer: I expected that this was hard to understand.

What I want to say:
"flap control" is an abstract way of the controller to tell the allocator, from 0 to 1, how much it should adjust the aerodynamic property of the wings to the "full flaps control configuration". Then the allocator would can use all the available control surfaces it wants (usually flaps, ailerons, elevator). 

I hope this makes it a bit clearer, and if you'd have ideas of how to properly formulate/find fitting terms for that then let me know!

en/config/actuators.md
hamishwillee: Thanks. Rather than regenerate this now, I added a "VTOL only" note like this:
image

en/config/actuators.md
```diff
@@ -140,32 +140,60 @@ Note that you will need to also ensure that the ESC associated with bidirectiona
...
   - The setting limits the rate of change of an actuator (if not specified then no rate limit is applied).
      It is intended for actuators that may be damaged if they move too fast — such as the tilting actuators on a tiltrotor VTOL vehicle.
   - For example, a setting of 2.0 means that the motor/servo will not be commanded to move from 0 to 1 at a rate that completes the operation in less than 2 seconds (in case of reversible motors, the range is -1 to 1).
+- (Advanced) `Flap Scale`: How much this actuator is deflected at the "full flaps configuration" [0, 1] (see [Flaps and Spoilers Configuration](#flaps-and-spoilers-configuration) below).
+  Can be used to configure aerodynamic surface as flap or to compensate for generated torque through main flaps.
+- (Advanced) `Spoiler Scale`: How much this actuator is deflected at the "full spoiler configuration" [0, 1] (see [Flaps and Spoilers Configuration](#flaps-and-spoilers-configuration) below).
+  Can be used to configure aerodynamic surface as spoiler or to compensate for generated torque through main spoiler.
 - `Lock control surfaces in hover`:
```
hamishwillee, line 173:
```suggestion
- (VTOL only) `Lock control surfaces in hover`:
```

en/config/actuators.md
```diff
@@ -140,32 +140,60 @@ Note that you will need to also ensure that the ESC associated with bidirectiona
...
+
+- The flaps have both `Flap Scale` set to 1, meaning that they will be fully deflected with the flap-control at 1.
+  They also have a slew rate of 0.5/s, meaning that it will take 2s to fully deflect them (a slew rate on the flaps is generally recommended to reduce the disturbances their movement creates).
+- The ailerons are primarily tasked to provide the commanded roll torque.
+  They also have `Spoiler Scale` set to 0.5, and will additionally be deflected upwards 50% if the controller demands full spoiler configuration.
+  The aileron deflection is thus the sum of the (asymmetrical) deflection for the roll torque, plus the (symmetrical) deflection for the spoiler setpoint.
+- The elevator is primarily tasked to provide pitch torque.
+  It also has non-zero entries in the `Flap Scale` and `Spoiler Scale` fields.
+  These are the elevator deflections added to compensate for the pitching moments generated by the flaps and spoiler actuators.
+  In the case here the elevator would be deflected 0.3 up when the flaps are fully deployed to counteract the pitching down moment caused by the flaps.
```
hamishwillee, line 197: Note, all of the above explains what they are, which is good. Say I want to set the flaps scale for an aileron though, how do I decide "0.3" would be a good value? Are there rules of thumb, or tuning guides?

en/config/actuators.md
hamishwillee: As it is now, the heading is opposite to your intent, which is why it is confusing - it makes me thing this is about configuring flaps, and configuring spoilers. What you mean is a title "Flap Scale and Spoiler Scale Configuration". 

What I think you are saying is that all types of controls might have an _effect_ like flaps (which affect lift) and spoilers (which slow the aircraft). What this section about is defining how affect much each type of control has, specifically for each of those scales. Obviously you would expect dedicate flaps to have a very big effect, so a large flap scale value. Ditto for dedicated spoilers and the spoiler scale value.

So something like:


```suggestion
#### Flap Scale and Spoiler Scale Configuration

Control surfaces such as ailerons (which primarily affect roll) and elevators (which primarily affect pitch), may also have an effect on lift or vehicle speed, in the same way as dedicated flaps and spoiler controls.
The flap scale and spoiler scale allow you to specify how much each control should be deflected when the controller is demanding "full flaps" or "full spoiler".
Then control allocator can then use all the available control surfaces it wants for flap and spoiler control (usually flaps, ailerons, elevator).

```

en/config/actuators.md
hamishwillee: I think so, but my suggestion above should tell you if I am or not: https://github.com/PX4/PX4-user_guide/pull/2450/files#r1170857584

I've tried to rewrite this bit below so it makes sense following my suggestion above.

en/config/actuators.md
```diff
@@ -140,32 +140,60 @@ Note that you will need to also ensure that the ESC associated with bidirectiona
...
+- (VTOL only) `Lock control surfaces in hover`:
   - `Enabled`: Most vehicles do not use control surfaces in hover. Use this setting to lock them so that they don't affect vehicle dynamics.
   - `Disabled`: Set this for vehicles that use control surfaces in hover, such as the duo tailsitter (which uses elevons for pitch and yaw control). It should also be set for vehicles that use control surfaces to provide additional stabilization in hover mode when moving at speed or in high winds.
 
+#### Flaps and Spoilers Configuration
+
+`Flap-control` and `spoiler-control` are aerodynamic configurations that can either be commanded manually by the pilot through a switch on the RC, or are set automatically by the controller, e.g. to reduce the airspeed before landing. 
+The flap and spoiler configuration controls are of the range of [0, 1] and can then be arbitrarily allocated to the control surfaces.
+
+It's important to understand the difference: with `flap/spoiler-control` we refer to an abstract metric of "how much flap/spoiler configuration" the controller currently demands, while the actual flaps or spoiler actuator setpoint can vary from that value.
```
hamishwillee, line 182: Let me pull this apart and just talk about spoilers. Because the terms configuration and control are scattered here and so ambigous.

> `Flap-control` and `spoiler-control` are aerodynamic configurations that can either be commanded manually by the pilot through a switch on the RC, or are set automatically by the controller, e.g. to reduce the airspeed before landing. 

The pilot or come controller says "I want to set spoiler control" because I am landing. 

> The flap and spoiler configuration controls are of the range of [0, 1] and can then be arbitrarily allocated to the control surfaces.

Or are you saying there is an abstract spoiler control setting that has a range [0,1] that a controller can request. SO say it requests 50% you are saying that 50% is the "abstract setpoint" but the allocator can take this 50% and assign it to different control surfaces?

> It's important to understand the difference: with `flap/spoiler-control` we refer to an abstract metric of "how much flap/spoiler configuration" the controller currently demands, while the actual flaps or spoiler actuator setpoint can vary from that value.

This is still confusing but again, I think you're saying that "spoiler" control is the abstract amount - say 30% of full. This is like the "abstract spoiler actuator setpoint". However in reality the behaviour is spread across a bunch of controls, so there is no one "spoiler actuator setpoint" but more a bunch"

hamishwillee: Excellent. I'll reword tomorrow. Note comment https://github.com/PX4/PX4-user_guide/pull/2450#discussion_r1170850045

en/config/actuators.md
```diff
@@ -140,32 +140,64 @@ Note that you will need to also ensure that the ESC associated with bidirectiona
...
+For example, a  pilot or the landing system might engage "Spoiler-control" in order to reduce the airspeed before landing.
+
+The configurations are an _abstract_ way for the controller to tell the allocator how much it should adjust the aerodynamic properties of the wings relative to the "full flaps" or "full spoiler" configuration (between `[0,1]`, where "1" indicates the full range).
+The allocator then uses any of the available control surfaces it wants in order to achieve the requested configuration: usually flaps, ailerons, and elevator.
+
+The `flap scale` and `spoiler scale` settings in the actuator UI inform the allocator how much ailerons, elevators, flaps, spoilers, and other control surfaces, contribute to a requested "Flap-control" and/or "Spoiler-control" value.
+Specifically, they indicate how much each control surface should be deflected when the controller is demanding "full flaps" or "full spoiler".
+
+In the following example, the vehicle has two ailerons, one elevator, one rudder and two flaps as control surfaces:
```
hamishwillee, line 188: @sfuhrer I have re-written this section in line with your confirmation, mostly just chopping and changing your words.
I think it is clear now?

It would be useful to state how you decide that say, a `flap scale` of .3 is reasonable, but perhaps we should merge this and discuss that as a post process?

en/config/actuators.md
sfuhrer: There is no easy rule of thumb, though we could mention that one should be careful with not deflecting the ailerons too much to keep good effectiveness for roll control

en/config/actuators.md
sfuhrer: Yes your rewording makes it much clearer, thank you!

And yes I would go ahead and merge it. Setting the appropriate scales is mostly an iterative process based on flight data, we could add this comment if you think it helps.

hamishwillee: @sfuhrer Thanks very much - this was a hard one!

>Setting the appropriate scales is mostly an iterative process based on flight data, we could add this comment if you think it helps.

I'm in two minds - on one hand it would mean people would know we thought about it and can't provide this - that's good because it reduces some sorts of questions.

The problem is, people are still stuck. This is really nothing you could provide a "script" on, or guidance for reading the log? I'm thinking might be an interesting dev conference presentation for how you approach this on a couple of slightly different vehicles.

sfuhrer: I've never fine tuned a fixed-wing runway vehicle, so I'm probably not the expert you think I am ;) I just enable the options and hope they're intuitively enough to follow. 
@tstastny tagging you as FW maintainer, something to maybe keep in mind to add docs at some point.

hamishwillee: Thanks! @tstastny , I'm watching you :-)

PR #2456, opened 2023-04-16
gps_compass: fix param used for GPS yaw

julianoes:
The previous param option is deprecated.

FYI @vincentpoont2.


hamishwillee:
Thanks, that's cool! I'll fix the other 34 hits in 9 files recommending use of EKF2_AID_MASK.

PR #2458, opened 2023-04-17
Improve the CAN bus docs

pavel-kirienko:
- Link the article that concisely describes the differences between DroneCAN and Cyphal.
- Link a post that answers frequently asked questions about CAN bus termination and topology.
- Fix a minor naming mistake.
- Use Cyphal/CAN instead of Cyphal where pertinent.


mrpollo:
thanks @pavel-kirienko

PR #2463, opened 2023-04-19
Deprecate PX4FLow and instead link to old docs

hamishwillee:
The PX4Flow optical sensor is no longer supported (firmware doesn't work and may also be PX4 problems). This adds a clear note that this is the case, removes the docs, and points back to old docs for those who have old PX4 and old PX4 flow firmware.

After this goes in and docs sync, also remove the orphan files here: /assets/hardware/sensors/px4flow/

This flows (ha ha) out of #2449


en/sensor/px4flow.md
```diff
@@ -1,467 +1,10 @@
-# PX4FLOW Smart Camera
+# PX4FLOW Smart Camera (Deprecated)
```
hamishwillee, line 1: Page renders like this. @mcsauder I assume that we're indicating that PX4 v1.13 is the last version it might work on with old PX4Flow firmware (not the current lot)?


image

hamishwillee:
FYI @mrpollo This does a more permanent removal of PX4 Flow. Based on discussion with @mcsauder it is gone for good/no longer supported in any version. Given that, best to make it very clear in our docs.

If there is no response from anyone I plan to merge this tomorrow.

PR #2465, opened 2023-04-19
simulation: fix last rtps in hyperlink

beniaminopozzan:
This should be the last remaining RTPS reference to be removed.


hamishwillee:
Excellent!

PR #2487, opened 2023-05-03
Update dev_env_windows_vm.md

Ionstream:
PR for WiFi telemetry configuration on virtual machines.

### DIFF
--- a/en/dev_setup/dev_env_windows_vm.md
+++ b/en/dev_setup/dev_env_windows_vm.md
@@ -95,3 +95,7 @@ You should also be able to build and upload firmware using a command like:
 ```
 make px4_fmu-v5_default upload
 ```
+
+### Telemetry over WiFi for QGroundControl
+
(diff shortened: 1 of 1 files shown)


en/dev_setup/dev_env_windows_vm.md
```diff
@@ -95,3 +95,7 @@ You should also be able to build and upload firmware using a command like:
 ```
 make px4_fmu-v5_default upload
 ```
+
+### Telemetry over WiFi for QGroundControl
+
+If you cannot receive telemetry from your vehicle while QGroundControl in your virtual machine, it may be due to your host machine's networking setup. For instance, VirtualBox 7 running Ubuntu 20.04 LTS will default to Network Address Translation (NAT), which blocks outbound UDP packets that QGroundControl uses to communicate with the vehicle. Changing this network setting to "Bridged Adapter" mode resolves this issue by giving the guest OS direct access to networking hardware on the host.
```
hamishwillee, line 101: @Ionstream Thanks very much. I have reworded a little. The subtle difference is that it is the VM's networking setup that is the problem - since that's where you set the networking mode, which is slightly different meaning to "the host machine's networking setup".

Does this look OK?

```suggestion
If using _QGroundControl_ within a virtual machine you should set the VM networking settings to  "Bridged Adapter" mode.
This gives the guest OS direct access to networking hardware on the host.
If you use the Network Address Translation (NAT), which is set by default for VirtualBox 7 running Ubuntu 20.04 LTS, this will block the outbound UDP packets that QGroundControl uses to communicate with the vehicle.
```

hamishwillee:
I think this looks good. Merging now.

PR #2488, opened 2023-05-03
Flaw checker workflow - remove translations stuff

hamishwillee:


hamishwillee:
Note, flaws above are "global" - i.e. they won't show up on most PRs. Also most of the InternalLinkToHTML are invalid - we're forced to use HTML by the toolchain in those places. I'll squash that eventually.

PR #2535, opened 2023-05-19
EKF2: update link to state and cov

bresch:
Those got moved quite some time ago...


hamishwillee:
Thanks @bresch - looks good to me!

PR #2539, opened 2023-05-22
Update bootloader_update_from_betaflight.md

vincentpoont2:
Add Note about Betaflight configurator 
Add Kakute bootloader details

### DIFF
--- a/en/advanced_config/bootloader_update.md
+++ b/en/advanced_config/bootloader_update.md
@@ -1,11 +1,11 @@
 # Bootloader Update
 
-The [PX4 bootloader](https://github.com/PX4/Bootloader) is used to load firmware for Pixhawk boards (PX4FMU, PX4IO).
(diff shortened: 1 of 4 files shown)


vincentpoont2:
@hamishwillee Seems like some of the Newer Betaflight Configurator does not allow flashing PX4 bootloader. I added a note, but I'm unsure if this is the best way. 

I haven't figure out starting which version it does not work, it is very finicky.

hamishwillee:
Thanks @vincentpoont2 . This is the right way, but I'm hesitant to add a note "use other methods" unless we have other methods to recommend (do we?). 

Is it possible to download old versions that we know to work? Or is this going to be an ongoing problem - i.e. we recommend a version but it disappears or stops working for PX4?

vincentpoont2:
@hamishwillee I don't exactly know which version works at the moment. I was thinking "other methods" would refer to the other methods below like dfu-util & dfuse.

hamishwillee:
Thanks very much. I did a few further updates to the page just to better name the sections. If you find out what version works/stops working let me know. At some point we probably will have to remove the section altogether if it doesn't work - or at least move it to the bottom and show dfutil first.

PR #2541, opened 2023-05-24
Dev env mac - improve structure

### LINKED ISSUE #2021: The affiliation of the catalog of books
Why do the titles of this guide, including chapter and section titles, not explicitly indicate affiliation? The cluttered entries make it difficult for readers to read. Should this be a bad habit of programmers? ?

hamishwillee:
The structure was confused. This partially fixes #2021 (though it will not satisfy the author).


hamishwillee:
@beniaminopozzan @Jaeyoung-Lim We probably need to add instructions to install Gazebo (gz) on macOS. Do we know if it works/who can we get to work out how?

It might be too early, but would be good to have something in place for when it becomes necessary.

Merging this now but your advice welcome.

PR #2547, opened 2023-05-27
add holybro pmw3901 and edit Holybro Unicore UM982 GPS

vincentpoont2:
add holybro pmw3901 and edit Holybro Unicore UM982 GPS


hamishwillee:
Thanks Vince, looks great.

PR #2548, opened 2023-05-27
add Holybro Rm3100

vincentpoont2:
add Holybro Rm3100


hamishwillee:
Thanks @vincentpoont2 

I decided to split out the stand alone compasses. Arguably I could split out the GNSS and RTK as well, all in this top page (interested to know what you think).  Having a monolithic table seemed sensible when there weren't so many of them, but now it is quite a lot to take in.

Stand along pages appreciated if possible.

PR #2549, opened 2023-05-27
add holybro DronCAN F9P

vincentpoont2:
add holybro DronCAN F9P


en/gps_compass/README.md
```diff
@@ -46,8 +46,12 @@ Device | GPS | Compass | [RTK](../gps_compass/rtk_gps.md) | [GPS Yaw Output](#co
 [Holybro Nano Ublox M8 5883 GPS](https://holybro.com/products/nano-m8-5883-gps-module) | UBX-M8030 | QMC5883 | | | |
 [Holybro M8N GPS](../gps_compass/gps_holybro_m8n_m9n.md) | M8N | IST8310 | | | |
 [Holybro M9N GPS](../gps_compass/gps_holybro_m8n_m9n.md) | M9N | IST8310 | | | |
+[Holybro DroneCAN M8N GPS](https://holybro.com/products/dronecan-m8n-gps) | M8N | BMM150 | | | |
```
hamishwillee, line 49: Better to have PX4 docs pages and link to those. But either way, this page already exists above linked to our page.
```suggestion
```

hamishwillee:
Thanks @vincentpoont2 ! I appreciate that it is duplication for you, but it is better for the docs if we have standalone pages.
The reason is that historically hardware is available and used after OEMs lose interested. If we have a page then the instructions remain sane for quite a while after the original vendor stops shipping. From my perspective I also have one place to link to for a product, and then when you change your URLs I only have to change in one place ... or mark that one page as "discontinued".

PR #2573, opened 2023-06-12
Update rtk_gps.md

Ymhahah:
Add DroneCAN Check mark on Holybro DroneCAN GPS.


hamishwillee:
Thank you @Ymhahah . This fix is correct, but is made in the wrong place. We use crowdin for translation from English and that tool then synchronizes updates to translated versions. So we never update the translation directly in github.

What I have done abovue in #2577 is made the same change to the English documents. This will eventually appear in the Chinese translation.

For more information on how things work see: http://docs.px4.io/main/en/contribute/translation.html

PR #2579, opened 2023-06-15
Updated quadchute params

RomanBapst:
Changes reflecting this PR:
https://github.com/PX4/PX4-Autopilot/pull/21598


RomanBapst:
@sfuhrer Let me know if you think the text is too complicated for the user guide. We could summarize it simpler without going into the details too much.

sfuhrer (approved):
For me this amount of description is necessary.

en/config/safety.md
```diff
@@ -251,11 +251,10 @@ The parameters that control when the quad-chute will trigger are listed in the t
...
-<a id="VT_QC_T_ALT_LOSS"></a>[VT_QC_T_ALT_LOSS](../advanced_config/parameter_reference.md#VT_QC_T_ALT_LOSS) | Altitude loss threshold for quad-chute triggering during VTOL transition to fixed-wing
- flight. Quad-chute triggers the current altitude is more than this value below the altitude at the beginning of the transition.
+<a id="COM_QC_ACT"></a>[COM_QC_ACT](../advanced_config/parameter_reference.md#COM_QC_ACT) | Quad-chute action after switching to multicopter flight. Can be set to: [Warning](#action_warning), [Return](#action_return), [Land](#action_land), [Hold](#action_hold).
+<a id="VT_FW_QC_HMAX"></a>[VT_FW_QC_HMAX](../advanced_config/parameter_reference.md#VT_FW_QC_HMAX) | Maximum quad-chute height, below which the quad-chute failsafe cannot trigger. This prevents high altitude quad-chute descent, which can drain the battery (and itself cause a crash). The height is relative to ground, home, or the local origin (in preference order, depending on what is available).
```
hamishwillee, line 255: I modified this: VT_FC_QC_HMAX, please check.

I understand that this prevents quadchute triggering at height. My assumption is that this is a bit like an interlock - so if you have a quadchute fail at heigh altitude it gets ignored. When you get below this height if you are still in some other circumstance that causes a failsafe it will then trigger. However if you somehow recovered then it would not. Right? Do we need to capture this nuance?

en/config/safety.md
```diff
@@ -251,11 +251,10 @@ The parameters that control when the quad-chute will trigger are listed in the t
...
+<a id="VT_QC_T_ALT_LOSS"></a>[VT_QC_T_ALT_LOSS](../advanced_config/parameter_reference.md#VT_QC_T_ALT_LOSS) | Altitude loss threshold for quad-chute triggering during VTOL transition to fixed-wing flight. Quad-chute triggers if the current altitude is more than this value below the altitude at the beginning of the transition.
```
hamishwillee, line 257: Is this more clear/better?

```suggestion
<a id="VT_QC_T_ALT_LOSS"></a>[VT_QC_T_ALT_LOSS](../advanced_config/parameter_reference.md#VT_QC_T_ALT_LOSS) | Altitude loss threshold for quad-chute triggering during VTOL transition to fixed-wing flight. The quad-chute is triggered if the vehicle falls this far below its initial altitude before completing the transition.
```

en/config/safety.md
```diff
@@ -251,11 +251,10 @@ The parameters that control when the quad-chute will trigger are listed in the t
...
+<a id="VT_QC_ALT_LOSS"></a>[VT_QC_ALT_LOSS](../advanced_config/parameter_reference.md#VT_QC_ALT_LOSS) | Uncommanded descent quad-chute altitude threshold. An uncommanded descend quad-chute event will be triggered as soon as the vehicle drops below this value measured relative to the current altitude reference. The altitude reference is computed based on the vehicle altitude and the altitude setpoint as follows. If the vehicle flies above the current altitude sepoint, then the altitude reference is set to the current altitude setpoint. No event will be triggered as the altitude error is positive. If the vehicle flies below the current altitude setpoint, then the altitude reference is set to the maximum altitude the vehicle has achieved while flying below the current altitude setpoint. This is to make sure that an uncommanded quad-chute event is not triggered in case the altitude setpoint increases faster than the vehicle can follow. On the other hand an event will still be triggered if the vehicle continuously loses altitude.
```
hamishwillee, line 256: The discussion of implementation is more than a user needs. I think all they need to know is that you've thought about the fact that not being able to track when the altitude setpoint is increasing too fast, could look like an uncontrolled descent. So this bit:

> Note that the quad-chute is not triggered if the commanded altitude setpoint increases faster than the vehicle can follow (it is only triggered if the vehicle continuously loses altitude below the commanded setpoint). 

The question is, whether the "how this is done" information is useful to someone else, and whether or not we should keep it here, or somewhere else. 

How about something like this. It defines the terms, and uses a little puncutation (`<br>`) to improve the table layout.


```suggestion
<a id="VT_QC_ALT_LOSS"></a>[VT_QC_ALT_LOSS](../advanced_config/parameter_reference.md#VT_QC_ALT_LOSS) | Uncommanded descent quad-chute altitude threshold.<br><br>In altitude controlled modes, such as [Hold mode](../flight_modes/hold.html), [Position mode](../flight_modes/position_fw.md), [Altitude mode](../flight_modes/altitude_fw.md), or [Mission mode](../flight_modes/mission.md), a vehicle should track its current "commanded" altitude setpoint. The quad chute failsafe is triggered if the vehicle falls too far below the commanded setpoint (by the amount defined in this parameter).<br><br>Note that the quad-chute is is only triggered if the vehicle continuously loses altitude below the commanded setpoint; it is not triggered if the commanded altitude setpoint increases faster than the vehicle can follow.
```

hamishwillee:
Thanks, I've added some changes and some suggestions/comments.

en/config/safety.md
RomanBapst: @hamishwillee If the vehicle goes below the threshold but the quadchute condition does not apply anymore, then it will not trigger anymore. So it's not like the FC remembers that it would have triggered and then triggers it once it's below the altitude.

en/config/safety.md
RomanBapst: @hamishwillee Yeah, this is great!

en/config/safety.md
RomanBapst: @hamishwillee Yep, that was exactly my thought as well and I was prepare to shorten it drastically. You've already done a great job here. Only one mistake you have "...is is...". Maybe that's how it's said in Australia though...

en/config/safety.md
RomanBapst: @hamishwillee 

> if you are still in some other circumstance that causes a failsafe it will then trigger.

Only if the circumstances are related to the conditions which trigger quadchute. Any other failsafe conditions like low battery has nothing to do with quadchute and thus will not trigger it.

PR #2585, opened 2023-06-20
Update cubepilot_cube_orange.md for carrier board pinouts

afwilkin:
Had to look for these for a bit. thought it might be useful to add them here


hamishwillee:
Thanks very much. I have decided that you are right that the pinouts are too hard to find "as is", but if that is worth doing, it is worth doing in the docs. So I am going to close this in favour of https://github.com/PX4/PX4-user_guide/pull/2586

PR #2586, opened 2023-06-21
Cube Orange Plus - add pinouts

hamishwillee:
This replaces #2585 by @afwilkin , adding the pinouts.

Many controllers follow the Pixhawk standard for ports now, but not the Cube. So it is worth adding pinouts, because it is really hard to infer them from the schematics as we suggested previously. I also don't want to cross link to ArduPilot here - either it is worth doing, or it is not.

I have updated this from https://github.com/proficnc/The-Cube/blob/master/CB_REV_C_Altium/PCB%203D%20Print%20from%20Above.PDF and https://docs.cubepilot.org/user-guides/autopilot/the-cube-module-overview , and sanity checked against the ArduPilot docs.

@julianoes My understanding is that the main differentiator for Cube Orange + is that it comes with a new board that has ADSB-IN receiver on the new carrier board - is that correct? It also lacks a debug port, but that isn't documented here, so not too worried about it
1. Is that right? In which case we should add the info to make it findable on search, and also perhaps cross link from ADSB page. 
2. If it does work with ADSB-In presumably PX4 needs to configure this specially, or would the board be set up already  ?
3. Are these pinouts likely the same for all the other boards? i.e. can I copy this to Black, Orange etc?

### DIFF
--- a/en/flight_controller/cubepilot_cube_orange.md
+++ b/en/flight_controller/cubepilot_cube_orange.md
@@ -12,6 +12,10 @@ The [Cube Orange](https://www.cubepilot.com/#/cube/features) flight controller i
 The controller is designed to be used with a domain-specific carrier board in order to reduce the wiring, improve reliability, and ease of assembly.
 For example, a carrier board for a commercial inspection vehicle might include connections for a companion computer, while a carrier board for a racer could includes ESCs for the frame of the vehicle.
 
+The ADS-B carrier board includes a customized 1090MHz ADSB-In receiver from uAvionix.
+This provides attitude and location of commercial manned aircraft within the range of Cube.
+This is automatically configured and enabled in the default PX4 firmware.
+
 Cube includes vibration isolation on two of the IMU's, with a third fixed IMU as a reference / backup.
 
 :::tip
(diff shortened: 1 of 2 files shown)


julianoes:
> 1. Is that right? In which case we should add the info to make it findable on search, and also perhaps cross link from ADSB page.

No, Cube Orange already came with the Ping ADSB-In module.

> 2. If it does work with ADSB-In presumably PX4 needs to configure this specially, or would the board be set up already ?

It works automatically, I've tested it.

> 3. Are these pinouts likely the same for all the other boards? i.e. can I copy this to Black, Orange etc?

I'm pretty sure the pinout is identical for Orange and Orange+. For Black I'd assume so, but I'm not too sure.

The main change of the Orange+ over the Orange is that it is based on the STM32H757 (dual core), and has a few changes in terms of the used sensors.

julianoes (approved):

hamishwillee:
Thanks very much @julianoes . I pushed the same pinouts to Cube Orange, added a note about the ADSB-In receiver on the carrier board, and added a note about the main differences.

PR #2598, opened 2023-06-27
Update holybro_pm02d.md

igalloway:
Add chip used section under specifciations to make it clearer what to look for in software.


mrpollo:
thanks @igalloway

PR #2599, opened 2023-06-27
esc_calibration: document latest functionality

MaEtUgR:
For the 1.14 release because we already changed the actuator configuration and users will need to go through their setups I also corrected the PWM calibration and the default PWM values.

See https://github.com/PX4/PX4-Autopilot/pull/21711 and https://github.com/PX4/PX4-Autopilot/pull/21513

This is the documentation change with it. And needs to be referenced by the release notes.

- [ ] Depends on  https://github.com/PX4/PX4-Autopilot/pull/21513
- [ ] Release note also required


en/advanced_config/esc_calibration.md
```diff
@@ -2,20 +2,33 @@
...
-Subsequently all the ESCs/motors will respond to flight controller input in the same way (across the whole input range).
+This calibration updates all the ESCs with a fixed maximum (2000us) and minimum (1000us) PWM input from the flight controller.
+Subsequently all the ESCs/motors will respond to flight controller input in the same way across the whole input range.
 
-Calibration is recommended for all ESCs, and in particular for low cost models.
+Calibration is recommended for all PWM/OneShot ESCs that support it, and in particular for low cost models.
+
+:::note
+High-quality controllers come with a factory calibration and should all respond the same way.
+In practice the input range may differ even on high quality controllers because you cannot be sure if they were manually recalibrated before and if that was done correctly. Also once calibrated they might not respond the same way as from the factory anymore.
```
hamishwillee, line 18:
```suggestion
In practice the input range may differ even on high quality controllers because they may be have been manually recalibrated after leaving the factory, and that might not have been done correctly. Following manual calibration they might not respond in the same way as they did after factory calibration.

```

en/advanced_config/esc_calibration.md
```diff
@@ -2,20 +2,34 @@
...
+
+Using this calibration tool is recommended, as it ensures that all ESCs behave exactly the same way.
+:::
+
+:::note
+Some ESCs do not support such a calibration. They must come factory calibrated and should respond consistently out of the box.
+This has to be verified using [Actuator Testing](../config/actuators.md#actuator-testing). Jump to step 7 which is still important.
+:::
+
+OneShot ESCs should be [configured to use OneShot](../peripherals/oneshot.md#px4-configuration) before calibration otherwise calibration steps should be repeated after switching.
```
hamishwillee, line 28:
```suggestion
OneShot ESCs should be [configured to use OneShot](../peripherals/oneshot.md#px4-configuration) before calibration (you should calibrate the ESCs after switching, even if you have previously calibrated).

```

en/advanced_config/esc_calibration.md
```diff
@@ -2,20 +2,34 @@
...
 
 ## Preconditions
 
-The system must include a power module (PX4 uses the measured voltage to determine whether or not a battery is connected).
+You must be able to (un)power ESCs and flight controller independently. If the flight controller is powered through USB and there is a power module the procedure can detect battery connection. If battery power can't be detected e.g. because there's no power module then the calibration sequence is still performed based on timeouts.
```
hamishwillee, line 32:
```suggestion
You must be able to (un)power ESCs and flight controller independently.
If the flight controller is powered through USB and there is a power module the procedure can detect battery connection.
If battery power can't be detected e.g. because there's no power module then the calibration sequence is still performed based on timeouts.

```

en/advanced_config/esc_calibration.md
```diff
@@ -24,33 +38,42 @@ To calibrate the ESCs:
 1. Remove the propellers.
 
    :::warning
-   Never attempt ESC calibration with props on.
-   
+   Never attempt ESC calibration with propellers on!
+
    The motors should not spin during ESC calibration.
-   However if an ESC doesn't properly support/detect the calibration sequence then it will respond to the PWM input by running the motor at maximum speed.
+   However if the calibration starts when the ESC is powered already or the ESC doesn't properly support/detect the calibration sequence then it will respond to the PWM input by running the motor at maximum speed.
```
hamishwillee, line 44:
```suggestion
   However if calibration starts when the ESC is powered already, or if the ESC doesn't properly support/detect the calibration sequence, then it will respond to the PWM input by running the motor at maximum speed.

```

en/advanced_config/esc_calibration.md
```diff
@@ -24,33 +38,42 @@ To calibrate the ESCs:
...
-   Never attempt ESC calibration with props on.
-   
+   Never attempt ESC calibration with propellers on!
+
    The motors should not spin during ESC calibration.
-   However if an ESC doesn't properly support/detect the calibration sequence then it will respond to the PWM input by running the motor at maximum speed.
+   However if the calibration starts when the ESC is powered already or the ESC doesn't properly support/detect the calibration sequence then it will respond to the PWM input by running the motor at maximum speed.
    :::
-1. Disconnect the battery and connect the flight controller via USB (only).
+1. Map the ESCs you're calibrating as Motors in the vehicle's [Actuator Configuration](../config/actuators.md). Only mapped actuators get an output and hence only ESCs mapped as motors get calibrated.
```
hamishwillee, line 46:
```suggestion
1. Map the ESCs you're calibrating as Motors in the vehicle's [Actuator Configuration](../config/actuators.md).
   Only mapped actuators get an output, and hence only ESCs mapped as motors get calibrated.

```

en/advanced_config/esc_calibration.md
```diff
@@ -24,33 +38,42 @@ To calibrate the ESCs:
...
+1. Unpower the ESCs by unplugging the battery but keep the flight controller powered and connected to the ground station best by plugging it via USB (only).
```
hamishwillee, line 47:
```suggestion
1. Unpower the ESCs by unplugging the battery.
   The flight controller should ideally still be powered via USB, and remain connected to the ground station.

```

en/advanced_config/esc_calibration.md
```diff
@@ -24,33 +38,42 @@ To calibrate the ESCs:
...
 1. Open the *QGroundControl* **Settings > Power**, then press the **Calibrate** button.
 
    ![ESC Calibration step 1](../../assets/qgc/setup/esc/qgc_esc_calibration.png)
 
-1. Connect the battery when prompted:
+1. After successfully starting the calibration immediately power the ESCs:
```
hamishwillee, line 52:
```suggestion
1. After starting the calibration sequence, power the ESCs (you should be prompted):

```

en/advanced_config/esc_calibration.md
```diff
@@ -24,33 +38,42 @@ To calibrate the ESCs:
...
 
    ![ESC Calibration step 2](../../assets/qgc/setup/esc/esc_calibration_step_2.png)
 
    The calibration will begin automatically:
 
    ![ESC Calibration step 3](../../assets/qgc/setup/esc/esc_calibration_step_3.png)
 
-1. Once the calibration complete you will be prompted to disconnect the battery.
+1. You're informed once the calibration sequence finished. During the calibration you should hear model specific beeping from the ESC that indicates the individual steps of the calibration.
```
hamishwillee, line 60:
```suggestion
1. During the calibration you should hear model specific beeping from the ESC, which indicates the individual steps of the calibration.

   You will be prompted when the calibration completes. 

```

en/advanced_config/esc_calibration.md
```diff
@@ -24,33 +38,42 @@ To calibrate the ESCs:
...
+   - The maximum value for a motor (default: 1900us, parameters: [PWM_MAIN_MAXn](../advanced_config/parameter_reference.md#PWM_MAIN_MAX1)/[PWM_AUX_MAXn](../advanced_config/parameter_reference.md#PWM_AUX_MAX1)) should be chosen such that if the value is further increased from there the motor doesn't actually spin faster anymore.
+   This can again be done by changing the disarmed value e.g. to 1900us and 1950us and 1975us and observing e.g. audibly if the motor is still going faster.
+   - The disarmed value for a motor (default: 1000us, parameters: [PWM_MAIN_DISn](../advanced_config/parameter_reference.md#PWM_MAIN_DIS1)/[PWM_AUX_DISn](../advanced_config/parameter_reference.md#PWM_AUX_DIS1)) should make the motor stop and stay stopped. When calibrated properly this value should be able to stay 1000us. If the ESC was previously miscalibrated then the motor might already spin with 1000us.
```
hamishwillee, line 69:
```suggestion
1. Go back to the [Actuator Configuration](../config/actuators.md).
   All motors with the same (re)calibrated ESCs should behave the same and work with the same configuration.
   Verify the following things:
   
   - The minimum value for a motor (default: 1100us, parameters: [PWM_MAIN_MINn](../advanced_config/parameter_reference.md#PWM_MAIN_MIN1)/[PWM_AUX_MINn](../advanced_config/parameter_reference.md#PWM_AUX_MIN1)) should make the motor spin slowly but reliably and also spin up reliably after it was stopped.
     Use a bit of margin here to make sure it doesn't stop in air when it shouldn't.
     To find the right value you can (still without propellers) test by changing the disarmed value e.g. to 1100us and 1050us and 1025us which makes the motor spin immediately and when you found the right value reset the disarmed value.
     After setting the desired minimum value you can test it using [Actuator Testing](../config/actuators.md#actuator-testing) by moving the test output slider just above the initial bump.
   - The maximum value for a motor (default: 1900us, parameters: [PWM_MAIN_MAXn](../advanced_config/parameter_reference.md#PWM_MAIN_MAX1)/[PWM_AUX_MAXn](../advanced_config/parameter_reference.md#PWM_AUX_MAX1)) should be chosen such that if the value is further increased from there the motor doesn't actually spin faster anymore.
     This can again be done by changing the disarmed value e.g. to 1900us and 1950us and 1975us and observing e.g. audibly if the motor is still going faster.
   - The disarmed value for a motor (default: 1000us, parameters: [PWM_MAIN_DISn](../advanced_config/parameter_reference.md#PWM_MAIN_DIS1)/[PWM_AUX_DISn](../advanced_config/parameter_reference.md#PWM_AUX_DIS1)) should make the motor stop and stay stopped.
     When calibrated properly this value should be able to stay 1000us.
     If the ESC was previously mis-calibrated then the motor might already spin with 1000us.

```

en/advanced_config/esc_calibration.md
```diff
@@ -24,33 +38,42 @@ To calibrate the ESCs:
...
 
-:::note
-High-quality controllers come with a factory calibration.
-In *theory* this means that the outputs might be can be configured by setting the values provided in the ESC technical specification for each output during [Actuator Configuration](../config/actuators.md) (under the hood this sets the [PWM_MAIN_MINn](../advanced_config/parameter_reference.md#PWM_MAIN_MIN1)/[PWM_AUX_MINn](../advanced_config/parameter_reference.md#PWM_AUX_MIN1) and [PWM_MAIN_MAXn](../advanced_config/parameter_reference.md#PWM_MAIN_MAX1)/[PWM_AUX_MAXn](../advanced_config/parameter_reference.md#PWM_AUX_MAX1) parameters).
+   :::note
+   VTOL or fixed-wing motors do not need any special PWM configuration. They will automatically stop during flight when desired by the autopilot with the PWM configuration as described.
```
hamishwillee, line 72: @MaEtUgR Is this more correct? Most of my changes in this update are layout only, but this one is a content change.

```suggestion
   VTOL and fixed-wing motors do not need any special PWM configuration.
   They will automatically stop during flight when commanded by the autopilot with the default PWM configuration.

```

en/advanced_config/esc_calibration.md
```diff
@@ -24,33 +38,42 @@ To calibrate the ESCs:
...
+   :::
 
-In practice the input range may differ even on high quality controllers.
-Using this calibration tool is recommended, as it ensures that all ESC behave exactly the same way.
-:::
+## Troubleshooting
+Possible problems:
+1. If you do not plug any ESC at the right time or the ESCs don't support calibration you'll still get the message that the calibration successfully completed because the system cannot possibly know if it actually worked.
+1. If you have a power module configured and connected the system will disallow starting the calibration for safety reasons. Unplug power to the ESCs first. If you're blocked because a power module is necessary to keep your flight controller alive but can (un)power the ESCs separately make sure the power module is disabled in PX4 just for the ESC calibration e.g. by [PWM_AUX_MINn](../advanced_config/parameter_reference.md#PWM_AUX_MIN1).
+1. If there is a power module and the system detects an increase in current consumption directly after successfully initiating calibration it will be aborted for safety reasons.
```
hamishwillee, line 79:
```suggestion

Possible problems:

1. If you do not plug any ESC at the right time or the ESCs don't support calibration you'll still get the message that the calibration successfully completed because the system cannot possibly know if it actually worked.
1. If you have a power module configured and connected the system will disallow starting the calibration for safety reasons.
   Unplug power to the ESCs first. If you're blocked because a power module is necessary to keep your flight controller alive but can (un)power the ESCs separately make sure the power module is disabled in PX4 just for the ESC calibration e.g. by [PWM_AUX_MINn](../advanced_config/parameter_reference.md#PWM_AUX_MIN1).
1. If there is a power module and the system detects an increase in current consumption directly after successfully initiating calibration it will be aborted for safety reasons.

```

en/advanced_config/esc_calibration.md
```diff
@@ -24,33 +47,90 @@ To calibrate the ESCs:
...
+
+1. Calibration can state that it has succeeded even when it has failed.
+
+   This happens happen if you do not power the ESC at the right time, or the ESCs don't support calibration.
+   This occurse because PX4 has has no way to know whether or not calibration was successful.
+
+1. Calibration cannot be started if you have a power module configured and connected (for safety reasons).
+
+   Unplug power to the ESCs first.
+   If you're blocked because a power module is necessary to keep your flight controller alive, but you can (un)power the ESCs separately, you can disable the motor just for the ESC calibration using parameters such as [PWM_AUX_MINn](../advanced_config/parameter_reference.md#PWM_AUX_MIN1).
```
hamishwillee, line 133: Are you saying that in order to calibrate the motor must be unpowered, and you can do this by setting the minimum value to (all) motors to some low value - e.g. PWM_AUX_MIN1 to 900us?

FYI, I updated all these items to state "problem first". This usually a better approach.

en/advanced_config/esc_calibration.md
```diff
@@ -24,33 +47,90 @@ To calibrate the ESCs:
...
 
-1. Once the calibration complete you will be prompted to disconnect the battery.
+1. During the calibration you should hear model-specific beeping from the ESC, which indicates the individual steps of the calibration.
+
+   You will be prompted when the calibration completes.
 
+   <a id="actuatorconfig_step"></a>
    ![ESC Calibration step 4](../../assets/qgc/setup/esc/esc_calibration_step_4.png)
 
+1. Go back to the [Actuator Configuration](../config/actuators.md) section.
```
hamishwillee, line 81: I've modified this a bit to remove the parameter names (you shouldn't need to know them when using the actuators config). I've also modified the steps so you use the slider in the current position, rather than messing with the disarmed value and having to set it back afterwards. I just think this is more logical.

NOTE, http://docs.px4.io/main/en/config/actuators.html#actuator-configuration-and-testing is the same process. This has two implications:

1. We should fix up the images/defaults in http://docs.px4.io/main/en/config/actuators.html#actuator-configuration-and-testing to note the new defaults.
2. Should we just link to this section rather than reproducing>"
3. Does http://docs.px4.io/main/en/config/actuators.html#actuator-configuration-and-testing need a bullet point for setting the _maximum_ value as you have here?

en/advanced_config/esc_calibration.md
MaEtUgR: Completely correct. This is a huge help!

en/advanced_config/esc_calibration.md
MaEtUgR: It's a copy paste mistake on my side. I copied the parameter link but didn't change to the correct name 🤦‍♂️ 
Thanks for pointing it out it only occurred to me now when reading your very helpful improvements.

en/advanced_config/esc_calibration.md
MaEtUgR: You are finding the rough spots immediately 👍 

> you shouldn't need to know them when using the actuators config

Completely true for going through the procedure. I only mentioned them for completeness e.g. for people who want to take note and put these settings into an airframe configuration. Not sure if it's appropriate.

> so you use the slider in the current position, rather than messing with the disarmed value

That sounds more logical yes. But I prefer the more hacky-sounding way with the disarmed value. The goal is to find an explicit value by trial and observation and you might realistically end up trying 10 times including all motors. If you do that with the slider the workflow is:
- type a value in the minimum field
- turn the sliders on
- move the slider over the bump
- observe the result and repeat

I'm using the disarmed field because then it's just:
- type a value in the disarmed field
- observe the result and repeat

Additionally, I can't find the maximum value any other way because I do it by listening if the motor spins any faster while changing the command. With the slider workflow I cannot hear the direct difference because the motor gets stopped and spun up again in between tries.

> is the same process

There's especially overlap with https://docs.px4.io/main/en/config/actuators.html#motor-configuration I'm aware this is not ideal. My goal is to have people not skim through the large actuators page and miss the fact that correctly configuring the limits is crucial and in principle more important than the ESC calibration itself.

> Should we just link to this section rather than reproducing

Yes but what's the best way to do that? Making sure users with short attention span go through the important steps with the least amount of confusion. Probably my approach is too local.

> need a bullet point for setting the maximum value as you have here?

Yes, it's part of a bring-up. It usually gets neglected because it's not important for the happy path flying. But ESCs commonly also have a dead zone at the top of the command range. Not configuring it leads to worse performance in corner cases. The controller/allocation relies on that high thrust actuation range being available. If there's no authority for the motor to actually spin faster that's unexpectedly non-linear and tracking gets unpredictably worse whenever that command is hit during flight. Can for example result in wobbles and attitude overshoot when flying with full thrust.

en/advanced_config/esc_calibration.md
MaEtUgR: > flight controller should ideally still be powered

Maybe that's already covered but the flight controller must stay powered. That's easiest (= ideal?) achieved with the classic pixhawk boards because they stay powered with USB only.

For e.g. Skynode USB alone is not enough, you need to keep the power module connected and so it's more tricky. I'm not sure but wouldn't be surprised if there are other setups that work this way. Hence it's also mentioned in the troubleshooting. Hopefully, I can improve the workflow for these cases.

hamishwillee: I will tidy this bit.

en/advanced_config/esc_calibration.md
```diff
@@ -24,33 +48,92 @@ To calibrate the ESCs:
...
+
+     To find the "optimal" minimum value, move the slider to the bottom (disarmed).
+     Then increase the PWM output's `disarmed` setting in small increments (e.g. 1025us, 1050us, etc), until the motor starts to spin reliably (it is better to be a little too high than a little too low).
+     Enter this value into the `minimum` setting for all the motor PWM outputs, and restore the `disarmed` output to `1100us`.
+
+   - The maximum value for a motor (default: `1900us`) should be chosen such that increasing the value doesn't make the motor spin any faster.
+
+     You can confirm that the motor spins quickly at the maximum setting in [Actuator Testing](../config/actuators.md#actuator-testing), by moving the associated test output slider to the top snap position.
+
+     To find the "optimal" maximum value, first move the slider to the bottom (disarmed).
```
hamishwillee, line 108: My assumption here is that props are still off - correct?

en/advanced_config/esc_calibration.md
```diff
@@ -24,33 +48,92 @@ To calibrate the ESCs:
...
+     Then increase the PWM output's `disarmed` setting to near the default maximum (`1900`) - the motors should spin up.
```
hamishwillee, line 109: Should we suggest a lower value as starting point - i.e. is it possible that the 1900 value will already be too high?

en/advanced_config/esc_calibration.md
hamishwillee: You're most welcome!

> so you use the slider in the current position, rather than messing with the disarmed value

Thanks for the explanation. I have rewritten this as you would prefer in https://github.com/PX4/PX4-user_guide/pull/2599/commits/5b748488c917a07fcc2e1756fb8ffd0670b30433 - needs to be checked, and I have a question.

> My goal is to have people not skim through the large actuators page and miss the fact that correctly configuring the limits is crucial and in principle more important than the ESC calibration itself.

I'm fine with reproducing it here - the actuator doc is a monolith. I want to split it a bit, but will probably do so in a following release (a scary job). At that point maybe we revisit whether we want to duplicate this or not.

What we will need now is to update the Actuators doc to show the new defaults for all the images. I guess that will have to wait for your PR to go in?

> need a bullet point for setting the maximum value as you have here?

Great. I added one. Will add a separate discussion point there.

en/config/actuators.md
```diff
@@ -489,7 +500,7 @@ For each motor:
 
      ![PWM Minimum Output](../../assets/config/actuators/pwm_minimum_output.png)
      :::note
-     For DShot output, this is not required <!-- any, or just the minimum check? -->
+     For DShot output, this is not required.
```
hamishwillee, line 503: As above, what is required for dshot and what is not? Ditto CAN.

en/config/actuators.md
```diff
@@ -470,11 +470,22 @@ To assign an actuator:
 
 ### Motor Configuration
 
+:::note
+If using PWM or OneShot ESCs, you should first perform [ESC Calibration](../advanced_config/esc_calibration.md) (this topic also covers motor configuration).
+
+[DShot](../peripherals/dshot.md) and [DroneCAN](../dronecan/escs.md)/Cyphal ESCs do not require this configuration.
```
hamishwillee, line 476: What do DShot/Dronecan need to do. My understanding is that outputs for disarmed/min/max would not need to be done.

What do we need to check? My assumption is that the motors are spinning in correct direction and the correct motor is mapping to the expected output. But those are covered in a different section of this doc.

en/config/actuators.md
```diff
@@ -470,11 +470,22 @@ To assign an actuator:
...
+:::
+
+:::warning
+Remove propellors!
```
hamishwillee, line 480: I though a good idea to say here!

hamishwillee:
No worries - great to see this being tidied up. It makes a lot more sense now. I've done what I can so hands off until you tell me that you need something else, or the PR goes in and we can update the actuators images for the new defaults.

MaEtUgR:
@hamishwillee I rebased and went through everything again adding my suggestions and small corrections.
The only thing missing now is merging the two to some degree redundant motor configuration parts. If it's ok for you I'd do that in the next iteration such that we have at least a working up-to-date PWM guide to point to when merging https://github.com/PX4/PX4-Autopilot/pull/21513

en/config/actuators.md
```diff
@@ -471,32 +471,32 @@ To assign an actuator:
 ### Motor Configuration
 
 :::note
-If using PWM or OneShot ESCs, you should first perform [ESC Calibration](../advanced_config/esc_calibration.md) (this topic also covers motor configuration).
+If using PWM or OneShot ESCs, you should first perform [ESC Calibration](../advanced_config/esc_calibration.md) (this topic also covers PWM specific motor configuration).
 
-[DShot](../peripherals/dshot.md) and [DroneCAN](../dronecan/escs.md)/Cyphal ESCs do not require this configuration.
+[DShot](../peripherals/dshot.md) ESCs do not require configuration of the command limits but only rotation direction.
```
hamishwillee, line 476: @MaEtUgR So we still need to configure max/min/dicarmed etc for CAN ESCs?

en/advanced_config/esc_calibration.md
```diff
@@ -24,33 +48,94 @@ To calibrate the ESCs:
...
+
+   The motors _should_ not spin during ESC calibration.
+   However if calibration starts when the ESC is already powered, or if the ESC doesn't properly support/detect the calibration sequence, then it will respond to the PWM input by running the motor at maximum speed.
    :::
-1. Disconnect the battery and connect the flight controller via USB (only).
+1. Map the ESCs you're calibrating as motors in the vehicle's [Actuator Configuration](../config/actuators.md).
+   Only mapped actuators get an output and only ESCs mapped as motors will be calibrated.
+
+1. Unpower the ESCs by unplugging the battery.
+   The flight controller must stay powered e.g. through USB remaining connected to the ground station.
```
hamishwillee, line 60:
```suggestion
   The flight controller must stay powered, for example, by keeping the USB connected to the ground station.

```

en/advanced_config/esc_calibration.md
hamishwillee: I will assume not.

hamishwillee:
> @hamishwillee I rebased and went through everything again adding my suggestions and small corrections.
> The only thing missing now is merging the two to some degree redundant motor configuration parts. If it's ok for you I'd do that in the next iteration such that we have at least a working up-to-date PWM guide to point to when merging [PX4/PX4-Autopilot#21513](https://github.com/PX4/PX4-Autopilot/pull/21513)

@MaEtUgR Thanks for sorting the details out. Merging redundant motor configuration parts can certainly wait, perhaps forever.

Do you have any ETA on  [PX4/PX4-Autopilot#21513](https://github.com/PX4/PX4-Autopilot/pull/21513) being merged?

en/config/actuators.md
MaEtUgR: Yes, because in the cases I've seen a rather [generalized motor command message](https://github.com/dronecan/DSDL/blob/6537918298be25eb285dff3428604e078b00f9a9/uavcan/equipment/actuator/Command.uavcan) is used and each ESC model/configuration can map it differently. That's also why the option to configure the limit commands exists e.g. for DroneCAN: https://github.com/PX4/PX4-Autopilot/blob/39e04d971229f3ca0d1e080d55bf7ac2f7894562/src/drivers/uavcan/module.yaml#L15-L25

MaEtUgR:
@hamishwillee Thank you so much for your help!

> Merging redundant motor configuration parts can certainly wait, perhaps forever.

Not forever but not be blocking. The reason for my strategy of the 7th step is that if I was the reader the existing motor configuration is in the middle of a very long page. And I might miss it not knowing how important that step is. This can and should for sure be solved with a reference and merging the two motor configuration guides. PWM is the main example for the entire procedure anyways.

> Do you have any ETA on https://github.com/PX4/PX4-Autopilot/pull/21513 being merged?

I would have merged it already but wanted to notify users about the change and make sure we can reference the updated documentation page on how to do things correctly now. So if this page is understandable I would move on, merge the change both in main and 1.14 and notify the people we reach with the documentation reference.

hamishwillee:
> Not forever but not be blocking. 

The reason I say forever is the way I want to fix this in actuators page is to split it up for the various frames - it is just too big and intimidating, and what you need to know for MCs is a lot less than all of that. But doing so is a big job, so I am waiting to get courage.

I don't mind the duplication at this level in the ESC page. But yes, better long term to have one source of instructions.

> le I would move on, merge the change both in main and 1.14 and notify the people we reach with the documentation reference.

I think the page is understandable. I count myself as the "average uninformed user", so unless you want someone else to take a look, I am OK for it to be merged.

MaEtUgR:
> it is just too big and intimidating, and what you need to know for MCs is a lot less than all of that

I thought exactly that but wasn't sure about a good solution to that problem. What you suggest makes sense 👍 

> unless you want someone else to take a look, I am OK for it to be merged

You are the best tech doc writer and reviewer I know. I think we can merge then.

hamishwillee:
> You are the best tech doc writer and reviewer I know. I think we can merge then.

Am I the only tech writer you know? :-)

Thanks for merging, and for making this whole area much more robust "out of the box".

PR #2615, opened 2023-07-07
fix tip display format in README.md

chinanewer:


hamishwillee:
Thank you, but changes cannot be made in this way.

The sources are English,  and they go through crowdin to generate this translation, as covered here: http://docs.px4.io/main/en/contribute/translation.html#translation

So changes need to be made either in English version on Github OR in Crowdin.

PR #2616, opened 2023-07-09
Add information for the logger module error behavior and file size limitations

lvanasse:
Add information from a [forum thread](https://discuss.px4.io/t/how-does-people-deal-with-large-ulog-file/32918/11) regarding the limitations and behavior on error of the logger module.


hamishwillee (approved):
Excellent, thank you.

PR #2617, opened 2023-07-10
Update dev_env_mac.md

bousborne:
tbb@2020 has been disabled, and therefore cannot be linked or installed as it currently is. Therefore, this addition just adds a simple "sed" command to edit the brew formula and comment out the "disable date" line. This will then allow the tbb@2020 to be installed and linked through homebrew.


julianoes (approved):
I guess that's a needed workaround, thanks for that!

hamishwillee:
Thanks for taking the time to work this out and document.

PR #2620, opened 2023-07-10
RC loss trigger defaults

zeroos:
Hello,

It seems like the defaults for RC loss trigger have changed in the meantime: the default value of [COM_RCL_EXCEPT](https://docs.px4.io/main/en/advanced_config/parameter_reference.html#COM_RCL_EXCEPT) is 0, so the failsafe is not ignored in any modes.

### DIFF
--- a/en/config/safety.md
+++ b/en/config/safety.md
@@ -66,27 +66,21 @@ Battery Emergency Level | [BAT_EMERGEN_THR](../advanced_config/parameter_referen
 
 The RC Loss failsafe may be triggered if the RC transmitter link is lost.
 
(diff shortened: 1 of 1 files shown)


hamishwillee:
@zeroos Thank you very much. I have a nasty feeling that this did not change, and this document has been in error for some time.

I have modified this a bit because disabling the timeout in auto modes is something that users commonly want to do, and which must therefore be highlighted. IMO the QGC safety screens should actually add this option, and possibly remove the failsafe timeout.

Much appreciate you taking the time to highlight this issue.

zeroos:
Awesome, thank you for your help! :)

PR #2644, opened 2023-07-19
Frames entry points and experimental frames rework

hamishwillee:
Currently the structure of the docs is quite MC-centric - you go through the basic assembly and configuration, and you probably have a working MC. You may well then find your way to specific advanced tuning config, and/or stuff for FW and VTOL.

We do have the [Airframes Builds](http://docs.px4.io/main/en/airframes/) section, but that is mostly about "this is a set of frames we support, and some build logs".

What this does is the first step in turning Airframes Builds into "Vehicle Types & Setup". Essentially this section will be where you come to set up a particular frame. The entry point has been changed to say this and to make some vehicles as "Experimental".

I've created a helicopter top level page @bkueng and put the config page beneath that - this is the model I will also do for MC and so on. 

Hope this makes sense. @bkueng @MaEtUgR @sfuhrer @RomanBapst This is first step. It will affect you though. I hope to merge this tomorrow after some more thinking and moving things around. Appreciate any suggestions for improvement.


en/frames_sub/README.md
```diff
@@ -1,7 +1,8 @@
-# Underwater Vehicles (UUV)
+# Submarines (Underwater Vehicles - UUV)
```
sfuhrer, line 1: could just as well also say what the first U stands for;)

sfuhrer (approved):
Looks very reasonable to me.

en/config_heli/README.md
```diff
@@ -1,31 +1,29 @@
...
-We would welcome your [contribution](../contribute/README.md) of new features, new frame configurations, or other improvements.
-:::
+This section contains topics related to [helicopter](../frames_helicopter/README.md) configuration and tuning.
 
 ## Supported Configurations
 
 Supported helicopter configurations:
 
 - Single main rotor with swash-plate controlled by up to 4 swash-plate servos and a mechanically uncoupled tail rotor driven by an ESC.
+- Mechanically coupled tails controlled by a servo can be flown by mapping the tail servo to the tail motor output and setting an adequate range and disarmed value for the servo.
```
bkueng, line 10: It's a separate `CA_AIRFRAME` type which gives you a tail servo

en/frames_helicopter/README.md
```diff
@@ -0,0 +1,48 @@
...
+  For example, PX4 does not support helicopters with coaxial or dual rotor types, and features such as RPM governor and autorotation.
+
+Maintainer volunteers, [contribution](../contribute/README.md) of new features, new frame configurations, or other improvements would all be very welcome!
+:::
+
+<!-- image here please of PX4 helicopter -->
+
+## Helicopter Types
+
+PX4 supports helicopters with a single main rotor with a swash-plate controlled by up to 4 swash-plate servos, and a mechanically uncoupled tail rotor driven by an ESC.
```
bkueng, line 18: Or tail servo

bkueng (commented):
Makes sense

en/config_heli/README.md
hamishwillee: Thanks @bkueng : 

10: Helicopter (tail ESC)
11: Helicopter (tail Servo)

But how do you select the frames? I.e. there is only one airframe shown for Helicopter in http://docs.px4.io/main/en/airframes/airframe_reference.html#helicopter

bkueng: Yes there's currently only 1 helicopter airframe. We could add a generic one for tail servo's as well, but I'm not sure if it's worth it at this point.

hamishwillee: @bkueng Just to be very clear, to access the tail servo option they would select the generic helicopter airframe then manually modify the CA_AIRFRAME to 11? A that point it will work as a tail servo? Will the actuator UI change at all?

bkueng: Correct, yes the UI will reflect that.

en/config_heli/README.md
```diff
@@ -1,31 +1,37 @@
...
+   At time of writing there is only _Generic Helicopter (Tail ESC)_ in the Helicopter group.
+   This will configure the helicopter frame with a mechanically uncoupled tail (tail ESC).
+
+   ![QGC - helicopter airframe](../../assets/config/airframe/airframe_heli_generic.png)
+
+   :::note
+   There is no separate airframe for the helicopter with tail servo.
+   To select this configuration, set the parameter [CA_AIRFRAME](../advanced_config/parameter_reference.md#CA_AIRFRAME) to _Helicopter (tail Servo)_.
+   The actuator configuration screen will then change to support this frame type.
+   :::
```
hamishwillee, line 33: @bkueng I updated this to refer to changing CA_AIRFRAME for the tail servo case - can't check it - having trouble getting my firmware to update to main. 

Can you read this section lines 10 to 333, and in particular the paragraph on line 11. I am not sure if that is correct, or has been superseded.

I have made an assumption here that most of the config using helicopter with a tail servo is the very similar to existing setup.

I'm going to merge. I hope this is OK and can easily be fixed if not.

en/config_heli/README.md
```diff
@@ -1,31 +1,37 @@
...
+This section contains topics related to [helicopter](../frames_helicopter/README.md) configuration and tuning.
 
 ## Supported Configurations
 
 Supported helicopter configurations:
 
 - Single main rotor with swash-plate controlled by up to 4 swash-plate servos and a mechanically uncoupled tail rotor driven by an ESC.
+- Single main rotor with swash-plate controlled by up to 4 swash-plate servos and a mechanically coupled tail controlled by a servo.
+  (This can be flown by mapping the tail servo to the tail motor output and setting an adequate range and disarmed value for the servo.
+  The system currently assumes it's a separate motor.)
```
bkueng, line 12: This should be removed (and mabe refer to CA_AIRFRAME below)

PR #2654, opened 2023-07-21
Update pixhawk_autopilot_bus.md

vincentpoont2:
Add Holybro Baseboards


hamishwillee (approved):
Thanks! Even better if we had docs pages on these ...

PR #2656, opened 2023-07-25
Update ros2_comm.md

pyromania99:
Mention the possibility of keeping previous simulation software while following guide


hamishwillee (approved):
Thanks @pyromania99 

I wasn't sure about this - it's odd to say "you only need to install PX4 for the simulator" and then to say "this is how you install PX4 without the simulator". In addition, the tutorial is only going to be reliable if you do follow the instructions and install the recommended simulator.

Anyway, what I have done is put this at the end "out of the way". 

Thanks again.

PS I fixed the typo separately while I was thinking about this update, which is why you won't see it here.

pyromania99:
[like]  Aniketh Parkala reacted to your message:
________________________________

PR #2665, opened 2023-07-27
Docs(zh): fix tip block error

veaba:
Tip element block should be new line.


veaba:
It shouldn't be here.

image

hamishwillee:
Thank you, but we can't fix the problem in this way. The translations have to go through crowdin - as described here: http://docs.px4.io/main/en/contribute/translation.html

Otherwise we will end up with lots of sync errors.

You're welcome to join those projects and fix it there!

hamishwillee:
I've updated the section, losing some translation, but fixing this error.

hamishwillee:
Sorry you're confused. Crowdin is a tool designed for managing translations. It takes the English docs, you provide translation, and it spits out the translated docs. What is cool about this is that it provides you with hints, and if the English docs update, then the translations that were incorrect are lost. That's a good thing - it means that there can be incomplete translations, but not "old translations that are wrong".

veaba:
I will try how use it.

PR #2673, opened 2023-08-02
add ulog_cpp library

bkueng:
Adds https://github.com/PX4/ulog_cpp to the list of ULog libraries.

I needed a C++ reader and writer library, so I'm making this generally available. For most use-cases the python library will probably still be the preferred solution.


hamishwillee (approved):
Thanks @bkueng - much appreciated.

PR #2681, opened 2023-08-09
[Merge once PX4 v1.14 docs have branched] Mission: edit mission stop and resume part

sfuhrer:
Changed PX4 feature:
- during a survey (with camera triggers), the active waypoint is reset to the previous one
- speed and camera settings are re-applied to resumed missions

Related PX4 PRs: 
- https://github.com/PX4/PX4-Autopilot/pull/21710
- https://github.com/PX4/PX4-Autopilot/pull/21714


hamishwillee:
Thanks @sfuhrer . This is not in v1.14 right?

I'm so confused by this text :-( I'm going to make a guess at what you're saying based on just this text:

> Camera configuration settings and vehicle speed setpoint settings are also stored in mission items as they can be configured to change within a mission.
> Upon resuming a mission the last camera and speed settings from the already flown part of the mission are automatically applied. 

I think you're saying that the current speed/camera settings are now stored alongside/after every waypoint, instead of just at the position in which they were downloaded.

So when you pause a mission you set the current active waypoint to the previous active waypoint. When you resume a mission vehicle start from wherever it is and flys to the current active waypoint at the system/mission default speed. When it reaches that point it picks up appropriate speed and camera settings, because they are defined with the point.

From the end user perspective, what this does is fly the last leg of the mission in which you paused, in its entirety. This differs from previously, where you'd resume from wherever you happened to be towards the current active waypoint.

If you upload a new mission, then the current mission item is set to 1. Resuming a mission would then be the same thing - you'd fly to the first mission item and continue as per usual. 

From an end user perspective this is different. Previously on switching back to the mission you'd have the same mission number (I think). Certainly if in an active mission you would have continued to the current mission item, then switched to the next sequence number in the new missino.

It sounds like you're no longer allowing mission upload if the mission is active.

Is that all correct?

----

If not, can you please write this out again in bullet points, separating out the three cases:
- active mission paused (changed to any other mode) and then resumed
- active mission paused, then upload, then resumed.
- active mission with upload occurring while active.

sfuhrer:
>Thanks @sfuhrer . This is not in v1.14 right?

Correct. We haven't yet branched off the docs?

>From an end user perspective this is different. Previously on switching back to the mission you'd have the same mission number (I think). Certainly if in an active mission you would have continued to the current mission item, then switched to the next sequence number in the new mission.

Yes this new now, it sets the current index to the previous one. Only if you had camera trigger items though.

>It sounds like you're no longer allowing mission upload if the mission is active.

This was never allowed. You always had to switch out of mission to upload a new one. And then the "current item" was reset to 1. Nothing new there.

The reset of what you wrote is all correct. 

> 
    active mission paused (changed to any other mode) and then resumed
    active mission paused, then upload, then resumed.
    active mission with upload occurring while active.

We could indeed split it up, though only the first two as the last is not possible. And the first (resume) could be split up for the normal and the "mapping/camera is triggering" case.

hamishwillee:
Thanks very much. NO, not branched yet.

1. Small update to v1.14 behaviour in https://github.com/PX4/PX4-user_guide/pull/2686 following this discussion - can you check and I'll rebase this PR on that.

> We could indeed split it up, though only the first two as the last is not possible. And the first (resume) could be split up for the normal and the "mapping/camera is triggering" case.

I meant "write it "**for me**" in bullet points separated, so I separate two cases". Not needed now as I think I get it.

Does this work as new proposal?

> Missions can be paused by switching out of mission mode to any other mode (such as [Hold mode](../flight_modes/hold.md) or [Position mode](../flight_modes/position_mc.md)), and resumed by switching back to mission mode.
> If the vehicle does not have camera trigger items, it will head from its _current position_ towards the same waypoint as it as was heading towards originally.
> If the vehicle has camera trigger items it will head from its current position towards the last waypoint it travelled through before pausing, and then retrace its path at the same speed and with the same camera triggering behaviour.
> This ensures that in survey/camera missions the planned path is captured.
> A mission can be uploaded while the vehicle is paused, in which which case the current active mission item is set to 1.
  
> **Note** When a mission is paused that contains camera trigger items, PX4 sets the current active mission item to the previous waypoint, so that when the mission is restarted the vehicle will retrace its planned path.
> In addition, PX4 stores mission items for speed setting and camera triggering for every waypoint to which they apply, even though the uploaded mission may only specify the items once (to apply to all subsequent items).
> This ensures that if the mission is paused, it will return to the planned speed from the first waypoint it reaches.


Is that right? I.e. my assumptions are:
- all speeds and camera triggers duplicated at all points in missions "where they make sense"
- decision to retrace path depends on ANY camera trigger being present anywhere in the mission. I.e. this is not smart enough to know that the current leg of a mission doesn't capture images, so it does not need to retrace its steps.

sfuhrer:
>Small update to v1.14 behaviour in https://github.com/PX4/PX4-user_guide/pull/2686 following this discussion - can you check and I'll rebase this PR on that.

Saw it and approved it, thanks!

So we wait with merging this PR here till the docs are branched off.

>Does this work as new proposal?

I like the first part a lot. About the seconds part ("Note"): I think that part could be made more clear by just writing:

Note When a mission is paused while the camera on the vehicle was triggering, PX4 sets the current active mission item to the previous waypoint, so that when the mission is restarted the vehicle will retrace its last mission leg. 

In addition, PX4 stores the last applied mission items for speed setting and camera triggering (from the already covered mission plan), and re-applies those settings on resuming the mission.

en/flight_modes/mission.md
```diff
@@ -54,11 +54,18 @@ At high level all vehicle types behave in the same way when MISSION mode is enga
...
-When you switch back to mission mode the vehicle will continue the mission, heading from the _current vehicle position_ to the current active mission item (the same waypoint it as heading towards originally).
-Note that if you moved the vehicle while the mission was paused you will no longer be following the original track towards the waypoint.
+Missions can be paused by switching out of mission mode to any other mode (such as [Hold mode](../flight_modes/hold.md) or [Position mode](../flight_modes/position_mc.md)), and resumed by switching back to mission mode.
+If the vehicle does not have camera trigger items, it will head from its _current position_ towards the same waypoint as it as was heading towards originally.
+If the vehicle has camera trigger items it will head from its current position towards the last waypoint it travelled through before pausing, and then retrace its path at the same speed and with the same camera triggering behaviour.
+This ensures that in survey/camera missions the planned path is captured.
```
hamishwillee, line 60:
```suggestion
Missions can be paused by switching out of mission mode to any other mode (such as [Hold mode](../flight_modes/hold.md) or [Position mode](../flight_modes/position_mc.md)), and resumed by switching back to mission mode.
If the vehicle was not capturing images when it was paused, on resuming it will head from its _current position_ towards the same waypoint as it as was heading towards originally.
If the vehicle was capturing images (has camera trigger items) it will instead head from its current position towards the last waypoint it travelled through (before pausing), and then retrace its path at the same speed and with the same camera triggering behaviour.
This ensures that in survey/camera missions the planned path is captured.

```

en/flight_modes/mission.md
```diff
@@ -54,11 +54,19 @@ At high level all vehicle types behave in the same way when MISSION mode is enga
...
+If the vehicle was capturing images (has camera trigger items) it will instead head from its current position towards the last waypoint it travelled through (before pausing), and then retrace its path at the same speed and with the same camera triggering behaviour.
+This ensures that in survey/camera missions the planned path is captured.
+
 A mission can be uploaded while the vehicle is paused, in which which case the current active mission item is set to 1.
 
+:::note
+When a mission is paused while the camera on the vehicle was triggering, PX4 sets the current active mission item to the previous waypoint, so that when the mission is restarted the vehicle will retrace its last mission leg.
+
+In addition, PX4 stores the last applied mission items for speed setting and camera triggering (from the already covered mission plan), and re-applies those settings on resuming the mission.
+:::
```
hamishwillee, line 68:
```suggestion
:::note
When a mission is paused while the camera on the vehicle was triggering, PX4 sets the current active mission item to the previous waypoint, so that when the mission is restarted the vehicle will retrace its last mission leg.
In addition, PX4 stores the last applied mission items for speed setting and camera triggering (from the already covered mission plan), and re-applies those settings on resuming the mission.
:::

```

en/flight_modes/mission.md
```diff
@@ -54,11 +54,19 @@ At high level all vehicle types behave in the same way when MISSION mode is enga
...
+If the vehicle was capturing images (has camera trigger items) it will instead head from its current position towards the last waypoint it travelled through (before pausing), and then retrace its path at the same speed and with the same camera triggering behaviour.
+This ensures that in survey/camera missions the planned path is captured.
+
 A mission can be uploaded while the vehicle is paused, in which which case the current active mission item is set to 1.
 
+:::note
+When a mission is paused while the camera on the vehicle was triggering, PX4 sets the current active mission item to the previous waypoint, so that when the mission is restarted the vehicle will retrace its last mission leg.
+In addition, PX4 stores the last applied mission items for speed setting and camera triggering (from the already covered mission plan), and re-applies those settings on resuming the mission.
+:::
+
```
hamishwillee, line 68:
```suggestion
:::note
When a mission is paused while the camera on the vehicle was triggering, PX4 sets the current active mission item to the previous waypoint, so that when the mission is restarted the vehicle will retrace its last mission leg.
In addition, PX4 stores the last applied mission items for speed setting and camera triggering (from the already covered mission plan), and re-applies those settings on resuming the mission.
:::
```

hamishwillee:
Thanks @sfuhrer . I have updated to a slight refinement of my text plus your text (which is good thanks). This can merge as soon as I have branched off PX4 v1.14. Expecting that in 2 weeks. Maybe next week. Depends on the candidate.

> I don't really get you here, it just stores and re-applies the last camera items and speed setpoints.

That's why you didn't like my text - my assumption of how things worked was wrong. I thought you were modifying the whole stored mission on upload to include applicable camera trigger and speed setpoints on every waypoint (which would work and be valid, but is unnecessary). Instead you're just fixing up the points for the last mission item on pausing.

It's a good implementation, and thanks for your patience in helping me understand it.

sfuhrer:
@hamishwillee sounds good, thanks for the help here!

hamishwillee:
@sfuhrer FYI v1.14 branch created, so merging this.

PR #2689, opened 2023-08-16
mission with local position estimate

hamishwillee:
This comes from the discussion [here](https://discord.com/channels/1022170275984457759/1134467913324187769) about planning missions with only a local estimate.

I have updated in two places - mission mode and ros/external_position_estimation.html, which are the two places you might try find out about it.

@bresch Is the global position generated by EFK from local position and global position origin used in **all auto modes**, or just mission? I.e. can I use hold mode if I set this? What about orbit and so on?

FYI @beniaminopozzan having this in the ROS topic isn't strictly needed, but when we create our "master" topic on VIO/MOCAP I want this information to be included, and this is a good way to make that more likely).

### DIFF
--- a/en/flight_modes/mission.md
+++ b/en/flight_modes/mission.md
@@ -1,12 +1,12 @@
 # Mission Mode
 
-[<img src="../../assets/site/position_fixed.svg" title="Position fix required (e.g. GPS)" width="30px" />](../getting_started/flight_modes.md#key_position_fixed)
+[<img src="../../assets/site/position_fixed.svg" title="Global position fix required (e.g. GPS)" width="30px" />](../getting_started/flight_modes.md#key_position_fixed)
 
 *Mission mode* causes the vehicle to execute a predefined autonomous [mission](../flying/missions.md) (flight plan) that has been uploaded to the flight controller. 
 The mission is typically created and uploaded with a Ground Control Station (GCS) application like [QGroundControl](https://docs.qgroundcontrol.com/master/en/) (QGC).
 
 :::note
-- This mode requires 3d position information (e.g. GPS).
+- This mode requires a global 3d position estimate (from GPS or a [local position](#missions-using-a-local-position-estimate)).
 - The vehicle must be armed before this mode can be engaged.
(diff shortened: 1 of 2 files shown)


hamishwillee:
Merging. Pretty sure this is correct, though it might need to be extended to other modes based on what @bresch says.

beniaminopozzan:
Thanks @hamishwillee !
I did not test it yet, but this should also already available for plain ROS 2 users (no mavlink) as the `SET_GPS_GLOBAL_ORIGIN` is managed by the `VEHICLE_CMD_SET_GPS_GLOBAL_ORIGIN` vehicle_command :)
The format and the units of measure are a bit different though, see the MAVLINK to uorb function:
https://github.com/PX4/PX4-Autopilot/blob/b21ad6af14dfc3c95c0d912748fcfaf23e7c7a93/src/modules/mavlink/mavlink_receiver.cpp#L1190-L1212

hamishwillee:
@beniaminopozzan Cool. I'm happy to wait until it is tested before documenting :-).

FYI, the values are different because MAVLink encodes GPS positions as an integer by multiplying by 10E7. This prevents rounding errors you might get if you were to shove the value into a float for transmission. 

The tricky bit would be that the implementation should emit https://mavlink.io/en/messages/common.html#GPS_GLOBAL_ORIGIN after setting the values. I guess that is so anything else looking at the vehicle can tell it is using local co-ordinates - i.e. I really don't know what relies on that information, so perhaps you could avoid setting it without actually breaking something.

bresch:
> Merging. Pretty sure this is correct, though it might need to be extended to other modes based on what @bresch says.

Sorry, I haven't seen this earlier. Yes, it looks correct to me. This is at least how it should work in theory, I actually never tried it myself (yet).

> Is the global position generated by EFK from local position and global position origin used in all auto modes, or just mission? I.e. can I use hold mode if I set this? What about orbit and so on?

By all auto modes; you also need a valid global position for orbit, RTL , hold and all the other auto modes.

hamishwillee:
Thanks @bresch - attempted to fix in https://github.com/PX4/PX4-user_guide/pull/2707

PR #2690, opened 2023-08-16
Add info about the CUAV debug adapter

hamishwillee:
This adds info about the [CUAV debug adapter](https://store.cuav.net/shop/cuav-c-adb/) to the docs. This comes with an STLINK-MINIE so is a good alternative if you want to use that with the 10-pin Debug port and don't want to solder your own.

FYI @niklaut do you know if the M2 port on the adapter conforms to a standard? I.e. is this something you could buy a cable for if you wanted to connect to something other than the STLINK? Fine if you don't know.

image

Going to mere either way!


niklaut:
Thanks for adding this, I was on holidays.

> do you know if the M2 port on the adapter conforms to a standard?

Yes, it's based on the [MIPI10 debug standard](https://en.wikipedia.org/wiki/MIPI_Debug_Architecture), which is an industry wide standard:


That STMicro extended to 14-pins by adding one row left and right to add serial rx/tx pins:



So it's a non-standard connector, but the cable comes packaged with the STLinkv3.


There is also a [MIPI20 debug pinout, which provides tracing data](https://wiki.segger.com/19-pin_JTAG/SWD_and_Trace_Connector). This will be interesting for the future when I get tracing to work with J-Trace/ORBTrace:

niklaut:
These are just RM1.27 cables btw, nothing special, you can use any 10/14/20 pin ribbon cable for this.

hamishwillee:
Thanks very much! I added connector name and link to manual in the CUAV adaptor docs - so now someone can easily find this out if they want.

PR #2707, opened 2023-08-24
Update docs on using auto modes with local estimate

hamishwillee:
This docs from #2689 to account for the fact that all automatic modes require a position estimate - not just mission mode.

To do this, the topic in the external position topic changed from "Planning/Executing Missions" to "Enabling Auto Modes with a Local Position". I thought about "Global Position from Local Position" (or similar) but I think people are more likely to find this.

Then in all the auto modes I linked back from the high level description box:
image

So if you are interested in local positions you can find out about this from either the flight modes or external_position_estimation topic

@beniaminopozzan Note that as discussed, this being in "external_position_estimation" is OK, but the topic should not be living under ROS as such. It's a placeholder :-).

This OK for you @bresch

### DIFF
--- a/en/flight_modes/land.md
+++ b/en/flight_modes/land.md
@@ -6,7 +6,8 @@ The *Land* flight mode causes the vehicle to land at the position where the mode
 After landing, vehicles will disarm after a short timeout (by default).
 
 :::note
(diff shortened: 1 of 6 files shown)


bresch (approved):
Yes, that looks good to me, or as the bot said: "No flaws found".

PR #2709, opened 2023-08-24
Simulation/README: fix gazebo classic make commands

sfuhrer:


hamishwillee (approved):
Thanks @sfuhrer .  Damn that copy replace replace :-)

PR #2717, opened 2023-09-01
Fix location of `rootfs` in SITL

DanMesh:
The `rootfs` directory for SITL builds is no longer in `build/px4_sitl_[target]/tmp/`, but rather directly under `build/px4_sitl_[target]`.

- Changed in https://github.com/PX4/PX4-Autopilot/pull/20090


hamishwillee (approved):
Thanks @DanMesh !

PR #2718, opened 2023-09-01
Update code.md

tstastny:
One more exception to google style. @hamishwillee @bkueng 

FYI - i can remove the part about "tab-ability" .. that was an insight from @bkueng :) but in general, i think it will be quite hard to get folks to switch to postfixed underscores.. and a lot of code to change.

### DIFF
--- a/en/contribute/code.md
+++ b/en/contribute/code.md
@@ -42,6 +42,10 @@ If you update an existing file you are not required to make the whole file compl
 
 - `lowerCamelCase()` is used for functions and methods to *visually* distinguish them from `ClassConstructors()` and `ClassNames`.
 
+### Private Class Member Variable Names
+
+- `_underscore_prefixed_snake_case` is used for private class member variable names, as oppose to `underscore_postfixed_` due to "tab-ability" (see all class members with one press of tab) and PX4 style legacy (hard to teach old dogs new tricks).
+
 ### Class Privacy Keywords
 
 - *zero* spaces before `public:`, `private:`, or `protected:` keywords.
@@ -60,15 +64,15 @@ public:
          */
         float doSomething(const float input_param) const {
                 const float in_scope_variable = input_param + kConstantFloat;
-                return in_scope_variable * private_member_variable_;
+                return in_scope_variable * _private_member_variable;
         }
 
-        void setPrivateMember(const float private_member_variable) { private_member_variable_ = private_member_variable; }
+        void setPrivateMember(const float private_member_variable) { _private_member_variable = private_member_variable; }
 
         /**
          * @return Whatever we are "getting" [units]
          */
-        float getPrivateMember() const { return private_member_variable_; }
+        float getPrivateMember() const { return _private_member_variable; }
 
 private:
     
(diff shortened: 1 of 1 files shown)


hamishwillee (commented):
Sounds reasonable - was this broadly agreed and communicated with the dev team?

tstastny:
Came out of a discussion with @bkueng. Just threw it up so I didn't forget. We should discuss in Tuesday maintainer call before pulling trigger.

hamishwillee:
Ok, keep me posted. FWIW I must be an old dog, because I don't see why you would want post underscores - the prefix more immediately tells you the variable is private.

dagar:
Underscore prefix is a little controversial because there are some other rules that are a bit close like in any scope underscore + uppercase is reserved, double underscore is reserved, and in a global namespace anything with with underscore prefix could be reserved.

So even with all that I personally still have a mild preference for it, but let's just stop talking about the google style guide if we're going to carry all of these exceptions.

hamishwillee:
Appreciate the explanation.

> stop talking about the google style guide if we're going to carry all of these exceptions.

I think you're saying "lets not make this change because we want to stick to the guide as closely as possible"? If so, this is supposed to be discussed in the maintainer call.

I don't care what the rules are, but I think we should lint to enforce them.

hamishwillee:
@tstastny Did this get discussed/resolved in any way?

MaEtUgR (approved):
I couldn't agree more. I was close to suggesting this as well but didn't want to introduce any opinionated extras.

tstastny:
> @tstastny Did this get discussed/resolved in any way?

shoot - i keep forgetting to bring it up in the dev calls. will put on the discussion list for next one.

hamishwillee:
Don't forget @tstastny :-)

hamishwillee:
@tstastny Looking for that comment in the dev call notes .... I guess you've forgotten that we'll both be at the PX4 developer day soon right and I can apply some menaces :-)

hamishwillee:
@tstastny Grrr :-)

tstastny:
From maintainer call: https://discuss.px4.io/t/px4-maintainers-call-november-07-2023/35031
@dagar please feel free to correct anything i got wrong here. again, I'm not terribly opinionated.. just want to lock it down so we don't have to think about it anymore, and have consistency.

- Abandon attempts to square with Google Style
- Update guide for what we are actually doing
  - e.g. some stragglers like file names (headers)
- Get linting set up FIRST so that we can actually enforce automatically (@MaEtUgR started this, we will need to find time to revisit and get it up and running)

@hamishwillee ok for you? i can revise the style guide for the main things we actually have traditionally cared about and intend to lint.

hamishwillee:
@tstastny Thanks for the update.

Sounds sensible, in particular to get linting working first and have it be the canonical definition of our style guide. Once the PR is accepted would be great if you could update this guide and I can review.

As an aside, it doesn't appear to be "quite" _Abandon attempts to square with Google Style_ since the clang setup is based on google. It is a good thing to be based as close to a documented style as possible.

FYI (and to you too @MaEtUgR ) we might also want to look at https://github.com/reviewdog/reviewdog . MDN use this tool to auto-lint changes made though github and add fixes as suggestions. It might not be useful for this, but I plan to integrate it into docs at some point.

MaEtUgR:
This clarification is more correct than before since almost all code has this style right now. We still need to enforce by linting but this change has no disadvantage.

hamishwillee:
Works for me!

PR #2722, opened 2023-09-06
Vehicle-specific config under vehicle top level topics

hamishwillee:
This is second step in reorganizing docs around vehicle-specific topics - moving the config topics out of advanced config under each section. No topics have been added, so this only changes the nesting location of these topics in the sidebar

Along with this I have made some changes to the Standard/Basic Config, and Advanced Config entry points to section up what topics are present. This is helpful because it makes clear that some "advanced topics" in particular might not actually be all that advanced. 
As part of this, for example, I moved ESC Calibration from Advanced to Standard - FYI @MaEtUgR !

I have not yet changed the entry points for config - that's a follow on step because it is likely to require some "help". But my intent is that if you were in MC, say, you would not have to go to the standard config entry point. This will include links for the standard stuff as well.

I'm going to add some comments for @sfuhrer / @MaEtUgR and merge this. Bigger changes ahead.


en/config/README.md
```diff
@@ -1,53 +1,77 @@
...
+- [Joystick Setup](../config/joystick.md)
+- [Flight Mode Configuration](../config/flight_mode.md)
 
-If your [selected airframe configuration](../config/airframe.md) is a _specific vehicle model_ (e.g. [Holybro s500](../frames_multicopter/holybro_s500_v2_pixhawk4.md)) then it may benefit from fine tuning, but this is not generally required.
+Safety configuration:
 
-If you're using a new airframe, a "generic" airframe, significantly modifying an airframe, or using less common sensors then see:
-- [Advanced Configuration](../advanced_config/README.md): Vehicle-specific tuning, fine tuning, factory-level configuration.
-- [Flight Controller Peripherals](../peripherals/README.md) - Hardware and software setup related to specific hardware (in particular lesser-used sensors).
+- [Battery and Power Module Setup](../config/battery.md) (optional)
```
hamishwillee, line 49: FYI, I put battery under safety, because to some extent that is what it is. Happy for suggestions of another name that captures "general important feature config"

en/SUMMARY.md
```diff
@@ -45,10 +45,18 @@
   * [Battery](config/battery.md)
   * [Safety](config/safety.md)
     * [Failsafe Simulation](config/safety_simulation.md)
+  * [ESC Calibration](advanced_config/esc_calibration.md)
```
hamishwillee, line 48: FYI @MaEtUgR Note this moved here under standard config, because it is useful in all cases. This really should be in its own section in QGC.

en/SUMMARY.md
```diff
@@ -115,38 +132,21 @@
...
-    * [VTOL w/o Airspeed Sensor](config_vtol/vtol_without_airspeed_sensor.md)
-    * [VTOL Weather Vane](config_vtol/vtol_weathervane.md)
-  * [ESC Calibration](advanced_config/esc_calibration.md)
+  * [Full Parameter Reference](advanced_config/parameter_reference.md)
   * [ECL/EKF Overview & Tuning](advanced_config/tuning_the_ecl_ekf.md)
   * [Flight Termination Configuration](advanced_config/flight_termination.md)
-  * [Bootloader Update](advanced_config/bootloader_update.md)
   * [Bootloader Flashing onto Betaflight Systems](advanced_config/bootloader_update_from_betaflight.md)
   * [Land Detector Configuration](advanced_config/land_detector.md)
+  * [Prearm/Arm/Disarm Configuration](advanced_config/prearm_arm_disarm.md)
```
hamishwillee, line 140: I wonder if a simplified version of this should go up in the safety setup in standard configuration (ideally in its own QGC screen). Essentially to control the settings of whether a safety button should be used if present, and whether prearm is the default equivalent of disarmed.

en/advanced_config/README.md
```diff
@@ -1,32 +1,43 @@
 # Advanced Configuration
 
-This section contains configuration topics that are relevant when bringing up a new airframe (or significantly modifying a supported frame), and/or for OEM manufacturers.
-It covers vehicle-specific tuning, factory/OEM calibration, and more advanced/less-common configuration.
+This section contains contains more advanced configuration topics, and those for features that are more rarely modified, or that are specific only to manufacturers/OEMs.
 
-Note that the [peripheral hardware](../peripherals/README.md) documentation contains configuration that is specific to particular sensors and actuators.
+Parameter config:
```
hamishwillee, line 5: Note, I split this advanced config into sections, which leads me to the conclusion that they are not all that advanced - see rendered [here](https://github.com/PX4/PX4-user_guide/blob/9c18a38e722f155d281f85eabbb633f02549f2c3/en/advanced_config/README.md). They are "OEM specific" or "Config for features you might not touch, like serial ports or landing". 

I am wondering if it might be better to have a parent config, put standard config under that, then make advanced a peer (or split up and the parts are peers). Still thinking about that - depends a bit on whether I add features as an organisation structure.

en/config/README.md
```diff
@@ -1,53 +1,77 @@
...
+## Preconditions
 
-To access **Vehicle Setup** screens, open the QGC application menu ("Q" icon in the top-left corner) and then choose **Vehicle Setup** in the *Select Tool* popup:
+Before starting this section you should [Download QGroundControl](http://qgroundcontrol.com/downloads/) and install it on your **desktop** computer.
+Then open the QGC application menu ("Q" icon in the top-left corner) and choose **Vehicle Setup** in the _Select Tool_ popup:
 
 ![QGC Main Menu Popup: highlighting Vehicle Setup](../../assets/qgc/setup/menu_setup.png)
-:::
 
-The sub topics cover each of the steps in detail (first install the PX4 firmware and set your vehicle airframe):
```
hamishwillee, line 13: Are there any bits of this config you would do if you got an RTF vehicle from a manufacturer? My assumption is "maybe compass" if you got a warning about it. The rest should be setup in an RTF correct right?

Everything else you would probably do once when bringing up your vehicle first time. If you changed something that affected weight you'd then do the autotune again, but the rest probably doesn't change right?

en/SUMMARY.md
MaEtUgR: I don't think it should be in QGC. It's quite autopilot-specific and if I were able to adjust QGC accordingly it would get much easier for the user...

en/SUMMARY.md
MaEtUgR: Totally, many of these options would get very simple to use intuitively if the UI was up to speed 😬

en/advanced_config/README.md
MaEtUgR: Maybe there are just mandatory and optional steps mapping to things you have to have/go through and setup of additional features someone might need for his vehicle or use case.

en/config/README.md
MaEtUgR: It really depends. RTF could be everything from
- a PX4-based product you don't have to configure anything except follow the wizard
all the way to
- some product that was made compatible and you flash and set everything up yourself possibly even with some workarounds for specifics

> If you changed something that affected weight you'd then do the autotune again

I'd generally expect this to end in a worse result than keeping a good manufacturer tuning even if the weight changed. What you might want to adjust is the hover thrust but even that is learned automatically for climb rate-controlled modes.

en/SUMMARY.md
hamishwillee: > I don't think it should be in the QGC docs. It's quite autopilot-specific and if I were able to adjust QGC accordingly it would get much easier for the user...

@MaEtUgR I'm not suggesting it should be in the QGC _docs_ - though I guess I did not provide enough context sorry. 

I'm suggesting that ESC calibration should not be part of Batter/Power configuration in QGC but instead its own specific view in QGC. It might be even be better as part of actuator config, though that interface is already quite "full"

For this docs change I mostly wanted you to be aware that this is now part of the "standard" instructions.

en/SUMMARY.md
hamishwillee: Do you have a fair idea of what you'd like the UI to look like? If we can draw it we can suggest it as an intern job or similar.

In the meantime, it might be worth having a docs section in safety setup for choosing the safety switch/arm behaviour - perhaps an extension of http://docs.px4.io/main/en/config/safety.html#arm-disarm-switch

en/advanced_config/README.md
hamishwillee: Yes - that's what we do now with standard and advanced. That makes advanced very large and "can't see the wood for the trees". If you split it then perhaps you can provide more directed guidance - i.e. factory needs this, but most people don't. 

Right now I'm compromising by putting heading levels in the top level page http://docs.px4.io/main/en/advanced_config/ but the sections below are just a random set. 

I'm still thinking about it :-)

en/config/README.md
hamishwillee: THanks for clarifying - a good reminder to me that I need to be precise. I meant [fully assembled PX4 installed](http://docs.px4.io/main/en/complete_vehicles/#fully-assembled-px4-preinstalled).

@MaEtUgR So what I think you are saying is that for such a vehicle, even if you modify it then probably you should stick with manufacturer configuration - there is no need for tuning. W.r.t. compass I've been told that generally a reasonable compass part should not "drift" but it can do so under some circumstances - so this is a "maybe" if you get warnings.

I'd expect the [UAV Kits](http://docs.px4.io/main/en/complete_vehicles/#uav-kits) with PX4 might also fit into this category?

If however you have the RTF which can run PX4 but has not been set up for PX4 you might be starting from scratch. I guess if QGC provides a particular airframe config you could rely on it being able to fly out of the box.

en/SUMMARY.md
hamishwillee: PS, thanks for all these comments. Very helpful. It is scary to make major changes to structure without the team being involved.

en/config/README.md
MaEtUgR: > even if you modify it then probably you should stick with manufacturer configuration - there is no need for tuning

Yes, if there is a manufacturer tuning for this exact geometry and drive train then it's usually better to not start retuning except if you see issues that are clearly tuning-related. By tuning I mean controller gains and limits and IMU filtering.

> W.r.t. compass I've been told that generally a reasonable compass part should not "drift" but it can do so under some circumstances

Yeah, the compass is susceptible to many factors e.g. magnetization of parts, rewiring, environment, ... We have by now quite some mechanisms to try and make the best out of it (will be the topic of Mathieu's dev summit talk) but it's never wrong to calibrate the compass and verify against the map the absolute heading being stable in all cardinal directions.

> I guess if QGC provides a particular airframe config you could rely on it being able to fly out of the box.

We have to be very specific. In the past most cases weren't so clear. e.g. the F450 airframe, was it tuned with the original drive train? does a user build it with those components? Many people including me just used the frame which only gives size information but doesn't allow a good out-of-the-box tuning.

PR #2734, opened 2023-09-14
MC Config - start work to make it a complete overview

hamishwillee:
I'm moving towards trying to make the docs possible to navigate in a more vehicle centric way. It's a bit iterative so there is some duplication at this point.

This pr  provides (draft) for the MC config page. The intent is to provides an overview of all the config you need to do for MC with explanation. Unlike current standard config is descriptive - it tells you what needs to happen in each section and why before providing the links. So it is more introductory. I like this because it allows me to explain why things are so, rather than just how to.

@MaEtUgR This is incomplete, but I'd be interested in getting your opinion on the approach - is this useful? Just the first few sections are done, but you can see the pattern. 
FWIW I think this demonstrates that we are going to need to split the Actuator page for MC out.  It would be really awesome to embed ESC calibration in the actuator screen and have all this in one topic.

@sfuhrer FYI too, since if this goes in I'll try do something similar for the winged frames.


en/config_mc/README.md
```diff
@@ -1,10 +1,141 @@
...
+## Safety Configuration
+
+- [Battery/Power Module Setup](../config/battery.md)
+- [Safety Configuration (Failsafes)](../config/safety.md)
+
+## Tuning
+
+<!-- 
+- Explain what you have to tune on PX4, what you can tune, and what each topic covers
+- I expect we should start with an exhaustive list of the tuning you could want to do - such as position tuning, etc. Do we have one?
```
hamishwillee, line 105: Note, so here we have rate tuning (autotune), rate/position etc in the basic manual, +airmode tuning in the advanced, + setpoint tuning. Also latency/filters. Is there a complete list of the tuning that someone might want to do "of this type"?

en/config_mc/README.md
```diff
@@ -1,10 +1,152 @@
...
+  
+The frames named with "Generic" generally just define the vehicle type, number of rotors, and "placeholder" rotor positions; almost everything needs to be configured.
+:::
+
+The next step is to define your vehicle [geometry](../config/actuators.md#motor-geometry-multicopter) (the number of motors and their relative positions) and [assign those motors](../config/actuators.md#actuator-outputs) to the physical outputs that they are wired to on your flight controller (both of these are covered in [Actuator Configuration and Testing](../config/actuators.md)).
+
+If using PWM ESCs and OneShot ESCs (but not DShot and DroneCAN/Cyphal ESC) you should then perform [ESC Calibration](../advanced_config/esc_calibration.md) before proceeding to [Motor Configuration](../config/actuators.md#motor-configuration).
+This ensures that all ESC provide exactly the same output for a given input (ideally we'd calibrate ESCs first, but you can't calibrate your ESCs until outputs are mapped).
+
+The final step is [Motor Configuration](../config/actuators.md#motor-configuration), which sets the directions of motors, and their disarmed, armed and high limits.
```
sfuhrer, line 49:
```suggestion
The final step is [Motor Configuration](../config/actuators.md#motor-configuration), which for PWM ESCs sets the disarmed, low and high PWM limits.
```

sfuhrer: The direction cannot be configured for normal PWM ESC.

en/config_mc/README.md
```diff
@@ -1,10 +1,152 @@
...
+
+Relevant topics:
+
+- [Vehicle (Frame) Selection](../config/airframe.md) - Select vehicle type to match your frame.
+- [Actuator Configuration and Testing](../config/actuators.md) - Vehicle geometry, output mapping, motor configuration, testing.
+- [ESC Calibration](../advanced_config/esc_calibration.md) - Do between output mapping and motor configuration (topic above) for PWM and OneShot ESC.
+
+## Sensor Setup and Calibration
+
+PX4 most commonly relies on a magnetometer (compass) for direction information, a barometer for altitude, a gyroscope for attitude, an accelerometer for velocity calculations, and a GPS/GNSS for global position.
```
sfuhrer, line 59:
```suggestion
PX4 most commonly relies on a magnetometer (compass) for direction information, a barometer for altitude, a gyroscope for body rates, an accelerometer for attitude and a GPS/GNSS for global position.
```

en/config_mc/README.md
```diff
@@ -1,10 +1,152 @@
...
+Pixhawk flight controllers (and many others) have inbuilt magnetometer, accelerometer, gyroscope, and barometer.
+The inbuilt compass usually isn't particularly reliable, and it is common to also add an external compass (usually with a GPS for position).
```
sfuhrer, line 61:
```suggestion
The inbuilt compass usually isn't particularly reliable, and it is common to also add an external compass (usually combined with a GNSS receiver in the same device).
```

en/config_mc/README.md
```diff
@@ -1,10 +1,152 @@
...
+- [Flight Controller Peripherals](../peripherals/README.md) - Setup specific sensors, optional sensors, actuators, and so on.
+
+:::note
+Barometers don't need calibration, and if you've mounted the flight controller level, there isn't much value in a [Level Horizon](../config/level_horizon_calibration.md) calibration either.
+
+Obviously you don't need to calibrate or configure anything that isn't present or used by PX4.
+For multicopter, that includes [airspeed sensors](../config/airspeed.md), which are not used for flight control.
+:::
+
+## GPS Setup
```
sfuhrer, line 85: I don't think the GPS requires a dedicated section.

en/config_mc/README.md
sfuhrer: I think we need to organize the tuning pages a bit more, but for now I find it okay to just link them all here.

sfuhrer (approved):
I think this is going it the right direction, and could already go in like this as a landing page and as a good iteration towards splitting up many parts of the docs into MC/FW/VTOL.

en/config_mc/README.md
hamishwillee: Agreed. I'm going to remove this. I will include it above as "a sensor that is not covered"

en/config_mc/README.md
```diff
@@ -1,10 +1,152 @@
...
+
+<!-- TBD
+
+Note that it is plugnplay except for RTK. Worth saying.
+
+-->
```
hamishwillee, line 91:
```suggestion
```

en/config_mc/README.md
hamishwillee: Yes. For me the most important things right now are to:
- Clarify when Autotune can be used (and what to do if not)
- Have a list of all the tuning we might reasonable want to do, so we can design our tuning approach.

en/config_mc/README.md
```diff
@@ -1,10 +1,152 @@
...
+
+PX4 can use other peripherals, such as distance sensors, optical flow sensors, traffic avoidance alarms, cameras, and so on.
+
+- [Flight Controller Peripherals](../peripherals/README.md) - Setup specific sensors, optional sensors, actuators, and so on.
+
+:::note
+Barometers don't need calibration, and if you've mounted the flight controller level, there isn't much value in a [Level Horizon](../config/level_horizon_calibration.md) calibration either.
+
+Obviously you don't need to calibrate or configure anything that isn't present or used by PX4.
+For multicopter, that includes [airspeed sensors](../config/airspeed.md), which are not used for flight control.
```
hamishwillee, line 82:
```suggestion
Sensors that you don't need to calibrate/configure include:

- [Level Horizon](../config/level_horizon_calibration.md) calibration isn't usually needed if you have mounted the flight controller level.
- Sensors that are not present, or that are not used by PX4 multicopter for flight control, such as [Airspeed sensors](../config/airspeed.md).
- Sensors that don't need calibration, including: Barometers and GNSS.
```

en/config_mc/README.md
hamishwillee: Yes, but my understanding is that you still need to set the limits even for non-PWM ESC. Please correct me if I'm wrong. Assuming I am not, how about:

```suggestion
The final step is [Motor Configuration](../config/actuators.md#motor-configuration), which sets the motor input values for  disarmed, low and high speed.
For CAN and D-Shot it can also set motor direction.
```

sfuhrer: I'm not sure. Also the GUI implies that setting the direction there is not possible. 
image
image

@bkueng can you enlighten us? In https://docs.px4.io/main/en/config/actuators.html#reversing-motors it says there is a "Set Spin Direction " button, where is that?

bkueng: > That seems very unintuitive to me

It's a bit of a compromise for an efficient UI layout. I see that point though. Maybe adding 'Permanently' is enough?

> It feels wrong because this will change the directions of all ESCs presumably

No, you can set them individually (there's a dropdown list when clicking the button).

> you can set the direction of CAN or DSHOT ESC through the UI but not PWM.

Only DShot, CAN would still need to be implemented.

> You should ESC calibrate for PWM and One SHot but not Dshot and Can

Yes

> For all ESC you still need to set disarmed, armed, and high values.

It depends. For dshot you don't need to set them, but there's DSHOT_MIN if needed.

hamishwillee: Thanks very much @bkueng - I'll adjust the text appropriately.

>>It feels wrong because this will change the directions of all ESCs presumably
> No, you can set them individually (there's a dropdown list when clicking the button).

I don't think this is ever going to be intuitive, but it might be convenient. if it were me I'd have toggle button next to each affected motor - the UI would have to be special though in that the button would hold not have visible - you press it, that changes the state, but it isn't a radio button or similar that shows selected/not selected.

hamishwillee: Modifying THIS text to be correct. Will separately update actuator config. @sfuhrer FYI

```suggestion
The final step is [Motor Configuration](../config/actuators.md#motor-configuration):
- [Reverse any motors](../config/actuators.md#reversing-motors) that don't match the spin direction configured in the Geometry.
  For DShot ESC you can do this through the Acuator Testing UI.
- PWM, OneShot, and CAN ESC, set the motor input limits for disarmed, low and high speed (not needed for DShot ESC)
```

en/config/actuators.md
```diff
@@ -591,8 +592,14 @@ If any motors do not turn in the correct direction they must be reversed.
 
 There are several options:
 
-- If the ESCs are configured as [DShot](../peripherals/dshot.md) you can reverse the direction via UI (**Set Spin Direction** buttons).
-  Note that the current direction cannot be queried, so you might have to try both options.
+- If the ESCs are configured as [DShot](../peripherals/dshot.md) you can permanently reverse the direction via UI.
```
hamishwillee, line 595: FYI only @bkueng I added diagram here to make it more obvious how the DShot ESC are set. We can fix if a way is found to improve the UI.

PR #2737, opened 2023-09-19
Add Holybro RemoteID to  remote_id.md

vincentpoont2:


hamishwillee (approved):
Thanks @vincentpoont2 . I've moved into alphabetic order and fixed a few typos. 

I also added an image of the pinouts. This is mostly because I wanted an image of the unit. Eventually I hope to have a slightly more complete doc for each unit - because often the units outlast manufacturer support.

PR #2740, opened 2023-09-20
add page for PX4 ROS library

bkueng:
Documentation for https://github.com/PX4/PX4-Autopilot/pull/20707.

Some bits are still missing & it needs integration with the other pages.

@hamishwillee if you want you can already take a first pass. @tstastny will go through the steps to see if anything is missing.


hamishwillee:
@bkueng Housekeeping
1. I've done some formatting changes on this to my taste.
   - Ran prettier and did my own normal layout changes. FYI now using VSCode and prettier to format by default - makes things a little less arbitrary.
   - Moved doc to a /ros2/ sub folder. I'm intending to separate/move ROS2 sources into own folder over time, but might as well start with this. It just makes things a little easier to manage.
   - Images are shrunk using lossless compressor and all grouped in same ros2 folder. Again, easier to manage.
   - Took the typo error (interpretted) into main and rebased
2. @beniaminopozzan Should review this doc as well when it is done.
3. As an aside, can we also have a separate additions on PX4 modes outside of ROS2? I can probably do this if I know:
   - What standard modes we support: https://mavlink.io/en/messages/development.html#MAV_STANDARD_MODE
   - What old style modes these map to (though that should be blindingly obvious
   - Confirm that the [AVAILABLE_MODES](https://mavlink.io/en/messages/development.html#AVAILABLE_MODES) and [CURRENT_MODE](https://mavlink.io/en/messages/development.html#CURRENT_MODE) etc are emitted as per the spec by PX4, and that https://mavlink.io/en/messages/development.html#MAV_CMD_DO_SET_STANDARD_MODE will change the mode.
   - Are the names in QGC matching standard modes or PX4 modes? Is there a translation repo yet for the standard and custom modes?
   - What I would like to do is get QGC to use the standard mode name as defined in MAVLink spec IF the standard mode is being used or the overloaded mode name/translation if not. Where I'm heading here is that if you're in Plane then you'd show Cruise rather than Position mode etc.

hamishwillee:
Top level comments on this feature.
1. It's pretty amazing - but then you know that. 
2. I hate the name: "PX4 SDK" - that's an SDK containing tools and libraries for developing PX4 modules. This is the "PX4 ROS SDK" or "PX4 ROS2 SDK" - a library for developing PX4 ROS2 applications.
3. My criteria for success on a doc is "Do I think I might be able to do this task". I don't think this is far off, though we could do a lot more.
4. It's quite monolithic. We might consider splitting it when the docs are complete. 
5. The tutorial is not a tutorial - it's an overview. I think a tutorial might be very useful for this particular case as well - perhaps adding a custom smart shot a'la solo - but you could make it much simpler than that.
   - Note we don't have that many tutorials in the docs because they create duplication and require additional maintenance. But inherently they serve a useful purpose in showing someone who doesn't know anything the things that we think are important to them. This is perhaps a case where an exception could be made.
6. There is some versioning here about QGC daily - is there a build we know that will have this coming up for PX4 v1.14. We need such a build anyway, so if we know what it will be, good to list it.
7. This suggests taking PX4 main. I _believe_ that the ROS2 integration intends to version by PX4 release. Upshot:
   - Good to mention the compatibility requirement for main is "right now" but when this is out in PX4 release you'll take the PX4 messages for that release.
   - How will you manage compatibility for the SDK depending on ROS? Are there topics that can't be modified in PX4 because of this?
   - How will you manage compatibility for users of the SDK? Are we saying that this is stable by PX4 release, or we'll maintain backwards compat?
8. This currently doesn't explain how you can set the name of your mode and translate it.

I'll add some other comments/suggestions in line.

en/ros2/px4_sdk.md
```diff
@@ -0,0 +1,336 @@
+# PX4 SDK
```
hamishwillee, line 1: The name :-(

en/ros2/px4_sdk.md
```diff
@@ -0,0 +1,336 @@
...
+
+The [PX4 SDK](https://github.com/PX4/px4_sdk) is a C++ library to simplify controlling PX4 from ROS 2.
+It allows to write external modes that are dynamically registered with PX4 and behave the same way as internal ones.
+A mode can send different types of setpoints, ranging from high-level navigation tasks all the way down to direct actuator controls.
+
```
hamishwillee, line 6: Does it also provide co-ordinate system mappings? I.e. if you write a setpoint, do you need to think about translation into PX4 land?

en/ros2/px4_sdk.md
```diff
@@ -0,0 +1,336 @@
...
+## Overview
+
+This diagram provides a conceptual overview:
+
+![ROS2 modes overview diagram](../../assets/middleware/ros2/px4_sdk/ros2_modes_overview.svg)
```
hamishwillee, line 11: The diagram is great. Say you're in payload delivery mode. Does PX4 have the concept that it is in a payload delivery mode, and if so is this worth showing? Or is it more that PX4 thinks it is under external control/some special mode.

It might not matter if the story is "ROS2 modes are first-class modes ". But you could put a dashed box to show that it has some presence in PX4. Just a thought.

en/ros2/px4_sdk.md
```diff
@@ -0,0 +1,336 @@
...
+
+<!-- Source: https://docs.google.com/drawings/d/1WByCfgcytnaow7r41VhYJL8OGrw1RjFO51GoPMQBCNA/edit -->
+
+The following sections define and explain the terms used in the diagram.
+
+### Definitions
+
+#### Mode
+
```
hamishwillee, line 20: This might be where you could say something like "A mode running in ROS2 is a first class peer of other modes in the system ..."

en/ros2/px4_sdk.md
```diff
@@ -0,0 +1,336 @@
...
+- A mode is a component that can send setpoints to the vehicle, which control its motion (such as velocity or direct actuator commands).
```
hamishwillee, line 21: Good to list the setpoints somewhere below and link to them.

en/ros2/px4_sdk.md
```diff
@@ -0,0 +1,336 @@
...
+- It is passive, meaning it does not activate other modes (by sending commands), and it must be activated by someone else (either the user through RC/GCS, the FMU in a failsafe situation or by a _mode executor_).
```
hamishwillee, line 22:
```suggestion
- A mode can't activate other modes (by sending commands), and must be activated by the user (through RC/GCS), the flight controller in a failsafe situation, a _mode executor_, or some other external system.
```

en/ros2/px4_sdk.md
```diff
@@ -0,0 +1,336 @@
...
+- It selects a setpoint type and sends it while it is active.
+  A mode can switch between multiple setpoint types.
+- Has a name displayed by the GCS.
```
hamishwillee, line 25: Only a name, or other information?

en/ros2/px4_sdk.md
```diff
@@ -0,0 +1,336 @@
...
+- Can configure its mode requirements (for example that it requires a valid position estimate).
```
hamishwillee, line 26: should we link to the requirements that can be set?

en/ros2/px4_sdk.md
```diff
@@ -0,0 +1,336 @@
...
+- A mode can consist of different tasks (for example flying to a target, lowering a winch, releasing a payload and then fly back).
```
hamishwillee, line 27: Do tasks have any specific meaning in code, as they do in PX4? If not, then:
```suggestion
- A mode can perform different tasks, such as flying to a target, lowering a winch, releasing a payload and then fly back.
```

en/ros2/px4_sdk.md
```diff
@@ -0,0 +1,336 @@
...
+
+#### Mode Executor
+
```
hamishwillee, line 30: It would be good to pull out one of these as an introductory sentence. I move the topic "When would you use a mode executor?" - something like this:

```suggestion
A mode executor is an optional component, associated with a mode, for putting the vehicle into a state where the mode can be activated.
For example, the mode executor for a payload delivery mode might first trigger a take-off, then switch to the mode to deliver the payload, and when that completes trigger an RTL.
```

FMI is the state machine flexible? I.e. my executor for above case might be called when already in the air - can it skip the takeoff state?

en/ros2/px4_sdk.md
```diff
@@ -0,0 +1,336 @@
...
+:::note
+- These definitions guarantee that a user can take away control from a custom mode or executor at any point in time by commanding a mode switch through RC or a GCS.
+- A mode executor is transparent to the user.
+  It gets indirectly selected and activated through the owning mode, and thus the mode should be named accordingly.
+:::
+
+##### When would you use a mode executor?
+
+The most common use-case is the execution of an autonomous 'mission', such as a survey or a payload delivery: the executor triggers a take-off, then switches to a custom mode which performs the survey, and when completed, triggers an RTL.
```
hamishwillee, line 49: As above, I think this would be better integrated into the topic https://github.com/PX4/PX4-user_guide/pull/2740/files#r1332395891

en/ros2/px4_sdk.md
```diff
@@ -0,0 +1,336 @@
...
+
+#### Configuration Overrides
+
+Both modes and executors can define configuration overrides, allowing to customize certain behaviors while the mode or executor is active.
```
hamishwillee, line 53: Link to where we show how this is done

en/ros2/px4_sdk.md
```diff
@@ -0,0 +1,336 @@
+# PX4 SDK
+
+The [PX4 SDK](https://github.com/PX4/px4_sdk) is a C++ library to simplify controlling PX4 from ROS 2.
+It allows to write external modes that are dynamically registered with PX4 and behave the same way as internal ones.
+A mode can send different types of setpoints, ranging from high-level navigation tasks all the way down to direct actuator controls.
+
+## Overview
```
hamishwillee, line 7: You could consider omitting this heading and moving all others up a level. If you do this you can't add too much here, but you could turn "### Comparison to Offboard Control" into a few sentences and put that here too. So "These modes are first class citizens blah blah. This approach offers many benefits over conventional offboard control because ...".

Just a thought.

en/ros2/px4_sdk.md
```diff
@@ -0,0 +1,336 @@
...
+
+### Comparison to Offboard Control
+
+The above concepts provide a number of advantages over traditional [offboard control](../ros/offboard_control.md):
+
+- Multiple nodes or applications can coexist and even run at the same time. It is clearly defined which node can control the vehicle at any given time.
+- Exposing a selectable mode to the GCS.
+- Integrated with the failsafe state machine and arming checks.
+- It is well-defined which setpoint types can be sent.
+- It is possible to replace an FMU-internal mode (such as RTL).
```
hamishwillee, line 69: Is FMU better than flight controller internal?

en/ros2/px4_sdk.md
```diff
@@ -0,0 +1,336 @@
...
+
+## Getting Started
```
hamishwillee, line 71: Getting started is OK. However you might also split this into "Installing the SDK" and "Testing the Installation". I kind of prefer that.
One the other hand, if you were to pull this into its own doc and call it a QuickStart I might feel differently.

en/ros2/px4_sdk.md
```diff
@@ -0,0 +1,336 @@
...
+   MicroXRCEAgent udp4 -p 8888
+   ```
+
+6. Start QGroundControl.
+
+   :::note
+   Use QGroundControl Daily, which supports dynamically updating the list of modes.
+   :::
+
+7. Back in the ROS 2 terminal, run one of the example modes:
```
hamishwillee, line 115: This is the reason I was thinking of the split  "Installing the SDK" and "Testing the Installation". This is something you do once to verify the installation - it's not instructions you do every time. On the other hand, it does help you understand how you might test installation.

What debug do you get by default and what have you injected in your ROS node? I.e.

en/ros2/px4_sdk.md
```diff
@@ -0,0 +1,336 @@
...
+
+   ```shell
+   ros2 run example_mode_manual_cpp example_mode_manual
+   ```
+
+   You should get an output like this:
```
hamishwillee, line 121:
```suggestion
   You should get an output like this showing 'My Manual Mode' mode being registered:
```

en/ros2/px4_sdk.md
```diff
@@ -0,0 +1,336 @@
...
+  }
+
+private:
+  std::shared_ptr<px4_sdk::ManualControlInput> _manual_control_input;
+  std::shared_ptr<px4_sdk::RatesSetpointType> _rates_setpoint;
+};
+```
+
+- **[1]**: First we create a class that inherits from `px4_sdk::ModeBase`
+- **[2]**: In the constructor, we pass the mode name. This also allows to configure some other things, like replacing an FMU-internal mode.
```
hamishwillee, line 209: Can we link to where you show a list of all the things you can configure, or (not quite as good) the instructions on how to override an internal mode.
```suggestion
- **[2]**: In the constructor, we pass the mode name. This also allows us to configure some other things, like replacing an FMU-internal mode.
```

en/ros2/px4_sdk.md
```diff
@@ -0,0 +1,336 @@
...
+    ```
+    [DEBUG] [example_mode_manual]: Mode 'My Manual Mode' activated
+    ```
+
+11. Now you are ready to create your own mode.
+
+## Mode Tutorial
+
+This section steps through an example for a custom mode class.
+For a complete application, check out the [examples in the repository](https://github.com/PX4/px4_sdk/tree/main/examples/cpp).
```
hamishwillee, line 166: Might be worth providing "such as" links https://github.com/PX4/px4_sdk/blob/main/examples/cpp/modes/rtl_replacement/include/mode.hpp

en/ros2/px4_sdk.md
```diff
@@ -0,0 +1,336 @@
...
+10. Select the mode, make sure you have a manual control source (physical or virtual joystick), and arm the vehicle.
+    The mode will then activate, and it should print the following output:
+
+    ```
+    [DEBUG] [example_mode_manual]: Mode 'My Manual Mode' activated
+    ```
+
+11. Now you are ready to create your own mode.
+
+## Mode Tutorial
```
hamishwillee, line 163: You could call this a tutorial, but I'd called it a partially worked example. I think this is very useful, I'm just not sure that there isn't "more to it". For example, how do we associate this with a node? Why is it in an hpp file/does it matter.

This would be better as part of "How to use the SDK" - say as "Mode Class Definition", and separately have a full tutorial that works through the real example more thoroughly.

hamishwillee: PS There is obviously some magic that happens to tell PX4 it has a new class. It might be worth identifying that part - could be as simple as "Deriving from XX registers the mode using internal APIs" or similar. That would give hints to someone who wants to understand the underlying code, or if a class didn't register properly.

en/ros2/px4_sdk.md
```diff
@@ -0,0 +1,336 @@
+# PX4 SDK
+
+The [PX4 SDK](https://github.com/PX4/px4_sdk) is a C++ library to simplify controlling PX4 from ROS 2.
+It allows to write external modes that are dynamically registered with PX4 and behave the same way as internal ones.
```
hamishwillee, line 4: "allows to" omits a subject.
```suggestion
It allows developers to write external modes that are dynamically registered with PX4 and behave the same way as internal ones.
```

en/ros2/px4_sdk.md
```diff
@@ -0,0 +1,336 @@
...
+2. Clone the SDK into the workspace:
+
+   ```shell
+   cd $ros_workspace/src
+   git clone --recursive https://github.com/PX4/px4_sdk.git
+   ```
+
+   :::note
+   To ensure compatibility, use the latest _main_ branches for PX4, _px4_msgs_ and the SDK.
+   See also [here](https://github.com/PX4/px4_sdk#compatibility-with-px4).
```
hamishwillee, line 85: This makes me uncomfortable - it basically says we ensure compatibility of the SDK by forcing you to use an unstable version of PX4. The linked doc is better in that it says we might change this in future. We should come up with a plan now and let people know.

The linked doc mentions this, which should also appear in this doc (I haven't checked) but it is cool:

> The SDK checks for message compatibility on startup when registering a mode. ALL_PX4_SDK_MESSAGES defines the set of checked messages. If you use other messages, you can check them using:
```
if (!px4_sdk::messageCompatibilityCheck(node, {{"/fmu/in/vehicle_rates_setpoint"}})) {
  throw std::runtime_error("Messages incompatible");
}
```

en/ros2/px4_sdk.md
```diff
@@ -0,0 +1,336 @@
...
+
+### Setpoint Types
+
+A mode can choose its setpoint type(s) it wants to use to control the vehicle.
+The used types also define the compatibility with different vehicle types.
+
+The following sections provide a list of commonly used setpoint types.
+You can also add your own type by adding a class that inherits from `px4_sdk::SetpointBase`, sets the configuration flags according to what the setpoint requires, and then publishes any topic containing a setpoint.
+
+#### Smooth Position
```
hamishwillee, line 232: Nothing wrong with this, but might be worth providing context for all cases:

```suggestion
#### Smooth Position Setpoints
```

en/ros2/px4_sdk.md
```diff
@@ -0,0 +1,336 @@
...
+
+:::note
+If you want to control an actuator that does not control the vehicle's motion, but for example a payload servo, see [below](#controlling-an-independent-actuator-servo).
+:::
+
+### Controlling an Independent Actuator/Servo
+
+If you want to control an independent actuator (a servo), follow these steps:
+
+1. [Configure the output](../payloads/#generic-actuator-control-with-mavlink)
```
hamishwillee, line 250: FYI Some idiot removed MAVlink generic actuator control and it hasn't been added back yet - https://github.com/PX4/PX4-Autopilot/issues/21966

I'm sure this still works under the hood, but if that doesn't get fixed I will have to hide this linked doc or add a big warning. That might break this.

en/ros2/px4_sdk.md
```diff
@@ -0,0 +1,336 @@
...
+2. Create an instance of [px4_sdk::OffboardActuatorControls](https://github.com/PX4/px4_sdk/blob/main/px4_sdk_cpp/include/px4_sdk/control/offboard_actuators.hpp) in the constructor of your mode
+3. Call the `set` method to control the actuator(s). This can be done independently of any active setpoints.
```
hamishwillee, line 252: Code fragment or link to code showing last two points would be awesome.

en/ros2/px4_sdk.md
```diff
@@ -0,0 +1,336 @@
...
+
+### Telemetry
+
+Telemetry, such as local or global position estimates can be found under [px4_sdk/odometry](https://github.com/PX4/px4_sdk/tree/main/px4_sdk_cpp/include/px4_sdk/odometry).
```
hamishwillee, line 256: What is special about this? I mean there are other ways to get telemetry in ROS, so 
- why did you choose to create this
- can you give a list of the main telemetry items covered - reduce amount of digging
- An example might be nice "This is used by modes that need positioning to plan future movement blah blah"

en/ros2/px4_sdk.md
```diff
@@ -0,0 +1,336 @@
...
+- arming is not allowed, while the mode is selected
+- when already armed, the mode cannot be selected
+- when armed and the mode is selected, the relevant failsafe is triggered (e.g. RC loss for the manual control requirement).
+  Check the [safety page](../config/safety.md) for how to configure failsafe behavior.
+  A failsafe is also triggered when the mode crashes or becomes unresponsive while it is selected.
+
+It is possible to manually update any mode requirement after the mode is registered. For example to add home position as requirement:
+
+```cpp
+modeRequirements().home_position = true;
```
hamishwillee, line 278: Can we list all the requirements you can set, or link to how you can find them?

The example above of manual mode and what gets set is awesome. Would be good to expand out the other modes and their implicit requirements. Why I'm interested here is that it's good to make it easy for people to understand what they can do, and when they might be able to do it.

The trade off is time to document and time to maintain. So you might alternatively link to source that shows the options, or source code that shows how you can easily list the requirements (all of them) that the mode currently has.

en/ros2/px4_sdk.md
```diff
@@ -0,0 +1,336 @@
...
+```
+
+#### Deferring Failsafes
+
+A mode or mode executor can temporarily defer non-essential failsafes.
+To do so, use the method `deferFailsafesSync`. And to get notified when a failsafe would be triggered, override the method `void onFailsafeDeferred()`.
+Check the [integration test](https://github.com/PX4/px4_sdk/blob/main/px4_sdk_cpp/test/integration/overrides.cpp#L167) for an example.
```
hamishwillee, line 285: 1.  Copy in an example. These links change ^^^ - right now I could not do this without the link.
2. This should be linked from where you mention the config change previously. 
3. I'd be tempted to have config things all listed under a config section so that someone can see what their config options are "all together". This works too, but next person who comes along might not be so sure where to update/add a new config option.

en/ros2/px4_sdk.md
```diff
@@ -0,0 +1,336 @@
...
+
+11. Now you are ready to create your own mode.
+
+## Mode Tutorial
+
+This section steps through an example for a custom mode class.
+For a complete application, check out the [examples in the repository](https://github.com/PX4/px4_sdk/tree/main/examples/cpp).
+
+```cpp{1,5,7-9,24-31}
+class MyMode : public px4_sdk::ModeBase // [1]
```
hamishwillee, line 169: How do I set that my new mode is hidden in the ground station? 

How do I set that my new mode is advanced, and should only be displayed in an advanced section of the UI?

en/ros2/px4_sdk.md
```diff
@@ -0,0 +1,336 @@
...
+
+#### Deferring Failsafes
+
+A mode or mode executor can temporarily defer non-essential failsafes.
+To do so, use the method `deferFailsafesSync`. And to get notified when a failsafe would be triggered, override the method `void onFailsafeDeferred()`.
+Check the [integration test](https://github.com/PX4/px4_sdk/blob/main/px4_sdk_cpp/test/integration/overrides.cpp#L167) for an example.
+
+### Assigning a Mode to an RC Switch or Joystick Action
+
+External modes can be assigned to RC switches or joystick actions.
```
hamishwillee, line 289: Link to RC switches and Joysticks sections

en/ros2/px4_sdk.md
```diff
@@ -0,0 +1,336 @@
...
+When assigning a mode to an RC switch, you need to know the index.
```
hamishwillee, line 290: Why? Don't you just need the name?
To put it another way, the mode assignment shows a distinct name, so when/why do you need the index?

en/ros2/px4_sdk.md
```diff
@@ -0,0 +1,336 @@
...
+
+```plain
+   INFO  [commander] External Mode 1: nav_state: 23, name: My Manual Mode
+```
+
+means you would select **External Mode 1** in QGC:
+![QGC Mode Assignment](../../assets/middleware/ros2/px4_sdk/qgc_mode_assignment.png)
+
+:::note
+PX4 ensures a given mode is always assignment to the same index by storing a hash of the mode name.
```
hamishwillee, line 302:
```suggestion
PX4 ensures a given mode is always assigned to the same index by storing a hash of the mode name.
```

en/ros2/px4_sdk.md
```diff
@@ -0,0 +1,336 @@
...
+This makes it independent of startup ordering in case of multiple external modes.
+:::
+
+### Replacing an Internal Mode
+
+An external mode can replace an existing internal mode, such as RTL. By doing so, whenever RTL gets selected (through the user or a failsafe situation), the external mode is used instead of the internal one.
```
hamishwillee, line 308:
```suggestion
An external mode can replace an existing internal mode, such as [Return](../flight_modes/return.md) mode (RTL).
By doing so, whenever RTL gets selected (through the user or a failsafe situation), the external mode is used instead of the internal one.
```

en/ros2/px4_sdk.md
```diff
@@ -0,0 +1,336 @@
...
+The internal one is only used as a fallback when the external one becomes unresponsive or crashes.
+
+The replacement mode can be set in the settings of the `ModeBase` constructor:
+
+```cpp
+Settings{kName, false, ModeBase::kModeIDRtl}
```
hamishwillee, line 314: Perhaps show code with the surrounding constructor? I'm also interested in what the values are for modes, such as `kModeIDRtl` - list them?

en/ros2/px4_sdk.md
```diff
@@ -0,0 +1,336 @@
...
+```
+
+## CI: Integration Tests
```
hamishwillee, line 317: Great you added this. Any guidance needed for writing tests?

en/ros2/px4_sdk.md
bkueng: I've tested DO_SET_ACUTUATOR cmd within the sdk, it all works. I'm not sure what @junwoo091400 thinks is wrong.

bkueng:
Ok you were very quick! I'll need to go through it in detail.

> I hate the name: "PX4 SDK" - that's an SDK containing tools and libraries for developing PX4 modules. This is the "PX4 ROS SDK" or "PX4 ROS2 SDK" - a library for developing PX4 ROS2 applications.

I'm glad you mention it now. I'm prefectly fine with renaming it. Really doesn't matter to me.

hamishwillee:
> I'm glad you mention it now. I'm prefectly fine with renaming it. Really doesn't matter to me.

That's cool. Perhaps run it by the next dev call just to make sure no one else is as opinionated as me.

> I'll need to go through it in detail.

I'm very much in favour of reducing developer workload - so listing things that they could in theory look up. It might well be you are expecting churn, in which case the current approach is more pragmatic. I do think that someone could work out most things from this - even me - so this is all kind of "optimisation" thinking.

en/ros2/px4_sdk.md
hamishwillee: Oh, I thought that SDK used ROS2 - I thought the MAVLink interface had been removed. @julianoes do you have an existing test case for the do_set_actuator in MAVSDK that could easily verify it?

julianoes: https://github.com/mavlink/MAVSDK/tree/v1.4/examples/set_actuator

en/ros2/px4_sdk.md
bkueng: A few reasons:
- for convenience
- changing the message does not mean you have to change your code
- using them automatically sets the mode requirement flags

I expect more things to be added in near future.

en/ros2/px4_sdk.md
bkueng: The parameter metadata does not contain the dynamic mode name

bkueng:
@hamishwillee I took a pass through your comments. As I expect more code changes, I don't want to go too much into detail yet.
I'm still waiting for @tstastny to go through it.

hamishwillee:
> @hamishwillee I took a pass through your comments. As I expect more code changes, I don't want to go too much into detail yet.
I'm still waiting for @tstastny to go through it.

OK, I'll wait until you are pretty happy with it before reviewing again. I like the name.

What I would quite like is a more "real world" but actually no more difficult example. What I was thinking was a mission emulation using takeoff, fly to point, very simple survey or drop of payload, and RTL. For extra points, done with a fixed-wing. For extra extra points, explain how someone can have multiple missions on the vehicle and switch between them.

I haven't re-read this, but the two things that most left me wondering on the PX4 dev day were:
1. How the mode executor has to look - i.e. how structured is it. How to cope with things like "someone starts my mode that has a takeoff when I have already taken off.
2. Comparison to offboard mode 
3. How compatibility is to be maintained with ROS and with this library.

I'm not suggesting you change or address these, just braindumping.

en/ros2/px4_sdk.md
hamishwillee: It might be worth saying "why" in the docs, unless you already did so.

bkueng:
It looks like Thomas does not have time to go through this anytime soon. So I suggest to bring it in now.
@hamishwillee do you want to take another pass?

en/ros2/px4_ros2_interface_lib.md
```diff
@@ -0,0 +1,365 @@
+# PX4 ROS 2 Interface Library
+
+The [PX4 ROS 2 Interface Library](https://github.com/Auterion/px4-ros2-interface-lib) is a C++ library to simplify controlling PX4 from ROS 2.
```
hamishwillee, line 3: FMI, is plan to move library into PX4 org?

en/ros2/px4_ros2_interface_lib.md
```diff
@@ -0,0 +1,365 @@
...
+
+Actuators can be directly controlled using the [px4_ros2::DirectActuatorsSetpointType](https://github.com/Auterion/px4-ros2-interface-lib/blob/main/px4_ros2_cpp/include/px4_ros2/control/setpoint_types/direct_actuators.hpp) setpoint type.
+Motors and servos can be set independently.
+Be aware that the assignment is vehicle and setup-specific.
+For example to control a quadrotor, you need to set the first 4 motors according to its [output configuration](../concept/control_allocation.md).
+
+:::note
+If you want to control an actuator that does not control the vehicle's motion, but for example a payload servo, see [below](#controlling-an-independent-actuator-servo).
+:::
+
```
hamishwillee, line 260: This is supposed to list the setpoints right? But we only show `DirectActuatorsSetpointType`. Is that the only one that works? I ask because we above talk about ManualControlInput and RatesSetpointType so I would have expected those.

It would be good to list above in the parent all the types that you expect to have, and link down to the ones that are defined. If they are not implemented yet, don't link them and suffix with `(not yet implemented)`

en/ros2/px4_ros2_interface_lib.md
```diff
@@ -0,0 +1,365 @@
...
+The replacement mode can be set in the settings of the `ModeBase` constructor:
+
+```cpp
+Settings{kName, false, ModeBase::kModeIDRtl}
+```
+
+## CI: Integration Tests
+
+When opening a pull request to PX4, CI runs the integration tests of the library.
```
hamishwillee, line 348: What do the CI tests do for you?

Specifically I think this is interesting, but I assume these run ROS2 and the SDK together with PX4. That is very cool, so we should say that, and perhaps provide pointers to how people can write their own?

en/ros2/px4_ros2_interface_lib.md
```diff
@@ -0,0 +1,365 @@
...
+If you want to control an independent actuator (a servo), follow these steps:
+
+1. [Configure the output](../payloads/#generic-actuator-control-with-mavlink)
+2. Create an instance of [px4_ros2::PeripheralActuatorControls](https://github.com/Auterion/px4-ros2-interface-lib/blob/main/px4_ros2_cpp/include/px4_ros2/control/peripheral_actuators.hpp) in the constructor of your mode.
+3. Call the `set` method to control the actuator(s).
+   This can be done independently of any active setpoints.
+
+### Telemetry
+
+Telemetry, such as local or global position estimates can be found under [px4_ros2/odometry](https://github.com/Auterion/px4-ros2-interface-lib/tree/main/px4_ros2_cpp/include/px4_ros2/odometry).
```
hamishwillee, line 272: Are these topics the same as the normal PX4 ones? Would be good to explain the origin etc.

en/ros2/px4_ros2_interface_lib.md
```diff
@@ -0,0 +1,365 @@
...
+    _rates_setpoint->update(rates_sp, thrust_sp);
+  }
+
+private:
+  std::shared_ptr<px4_ros2::ManualControlInput> _manual_control_input;
+  std::shared_ptr<px4_ros2::RatesSetpointType> _rates_setpoint;
+};
+```
+
+- `[1]`: First we create a class that inherits from `px4_ros2::ModeBase`
```
hamishwillee, line 226: Do you think we could create a "reference" of the public API - just names and a description for the main classes at the end. Starting from a reference tells you what documentation is available/missing.

en/ros2/px4_ros2_interface_lib.md
```diff
@@ -0,0 +1,365 @@
...
+A mode defined using the interface library has the following properties:
+
+- A mode is a component that can send setpoints to the vehicle in order to control its motion (such as velocity or direct actuator commands).
+- A mode selects a setpoint type and sends it while it is active.
+  It can switch between multiple setpoint types.
+- A mode can't activate other modes, and must be activated by the user (through RC/GCS), the flight controller in a failsafe situation, a _mode executor_, or some other external system.
+- Has a name displayed by the GCS.
+- Can configure its mode requirements (for example that it requires a valid position estimate).
+- A mode can perform different tasks, such as flying to a target, lowering a winch, releasing a payload and then fly back.
+- A mode can replace a mode defined in PX4.
```
hamishwillee, line 30: FYI I added this point.

en/ros2/px4_ros2_interface_lib.md
```diff
@@ -0,0 +1,365 @@
...
+    ```
+
+11. Now you are ready to create your own mode.
+
+## How to Use the Library
+
+The following sections describe specific functionality provided by this library.
+In addition, any other PX4 topic can be subscribed or published.
+
+### Mode Class Definition
```
hamishwillee, line 180: We have no Mode Executor definition. Is that planned or just delayed, or what? It would be good to provide a realistic example.

en/ros2/px4_ros2_interface_lib.md
```diff
@@ -0,0 +1,365 @@
...
+
+### Setpoint Types
+
+A mode can choose its setpoint type(s) it wants to use to control the vehicle.
+The used types also define the compatibility with different vehicle types.
+
+The following sections provide a list of commonly used setpoint types.
+You can also add your own type by adding a class that inherits from `px4_ros2::SetpointBase`, sets the configuration flags according to what the setpoint requires, and then publishes any topic containing a setpoint.
+
+<!--
```
hamishwillee, line 244: Todo

hamishwillee:
@bkueng I added some comments. I'm not sure how far this has to go before we decide it is "right". Some things I'd like considered before we merge:

- Can you add a small section in https://docs.px4.io/main/en/releases/main.html for this. Essentially I'm trying to generate our release notes in advance of the final release. At least including the main features and autopilot PRs.
- Can we add the video from the dev summit
- This seems vehicle agnostic. Is it? I.e. can I expect this to work just as well on plane, and in similar ways? For example if my mode executor specifies takeoff will the system use the takeoff for plane on plane and mc on mc? Or to put it another way - can we have a section "Support for different vehicle types" that summarises what you have to think about. Ultimately we'd like examples for each.
- It would be good to have API reference at top level for the public API - classes, topics, setpoint types etc. That way even if we don't document them all, people can infer their purpose and we can expand out over time.

You might not want to do any of this right now. That's OK if you think this is useful as is and have a cunning plan going forward?

en/ros2/px4_ros2_interface_lib.md
bkueng: I don't know, it's not up to me.

en/ros2/px4_ros2_interface_lib.md
bkueng: Yes the other ones are still experimental, under a subdirectory: https://github.com/Auterion/px4-ros2-interface-lib/tree/main/px4_ros2_cpp/include/px4_ros2/control/setpoint_types/experimental.
We can mention that as well.

en/ros2/px4_ros2_interface_lib.md
bkueng: For now it makes sure the mode registration works, and a few other things like failsafes and mode replacement.
We can go much further with this and also test setpoints or just PX4 in general, since with DDS we can easily access PX4 internal topics. But I'm not sure about that yet, at the minimum it will require lockstep and work robustly.

en/ros2/px4_ros2_interface_lib.md
bkueng: Yes this will need a lot more documentation.

en/ros2/px4_ros2_interface_lib.md
bkueng: Yes we could use doxygen

en/ros2/px4_ros2_interface_lib.md
bkueng: I can add that.

bkueng:
> Can you add a small section in https://docs.px4.io/main/en/releases/main.html for this. Essentially I'm trying to generate our release notes in advance of the final release. At least including the main features and autopilot PRs.

Will do.

> Can we add the video from the dev summit

Yes later. It's not up yet and I wanted to add the link to the docs to the video first.

> This seems vehicle agnostic. Is it? I.e. can I expect this to work just as well on plane, and in similar ways? For example if my mode executor specifies takeoff will the system use the takeoff for plane on plane and mc on mc? Or to put it another way - can we have a section "Support for different vehicle types" that summarises what you have to think about. Ultimately we'd like examples for each.

Yes agnostic, but we'll have to see how much we can abstract a mode to be generic. I don't think they will be completely generic.

> You might not want to do any of this right now. That's OK if you think this is useful as is and have a cunning plan going forward?

Cunning certainly goes too far, but yes, there's a lot of things to be done.

en/releases/main.md
```diff
@@ -54,3 +54,9 @@ TBD ...
 ### Fixed-wing
 
 - TBD
+
+### ROS 2
+
+- [Experimental] [PX4 ROS 2 Interface Library](../ros2/px4_ros2_interface_lib.md): A new C++ library that simplifies controlling PX4 from ROS 2.
+  Supports adding flight modes in ROS 2 that are peers of the PX4 modes running on the flight controller.
+  Added to PX4 in [PX4-Autopilot#20707](https://github.com/PX4/PX4-Autopilot/pull/20707) (initial support).
```
hamishwillee, line 62: FYI the important thing here is to set expectation. So I have put a prefix experimental and added a link to the initial PR. 
As you enhance this you can add new PR links. 
If you consider significant parts of this stable, we can remove "Experimental".

en/ros2/px4_ros2_interface_lib.md
```diff
@@ -0,0 +1,465 @@
...
+At time of writing parts of the PX4 ROS 2 Interface Library is experimental, and hence subject to change:
+
+- The architecture and core interfaces for defining modes in ROS 2 modes are largely stable.
+  The library offers significant benefits over using offboard mode in its current state.
+- Only a few setpoint types have settled (the others are still under development).
+  You may need to use internal PX4 topics which may not remain backwards-compatible over time.
+- The API is not fully documented.
+- Testing in CI is still limited.
+
+:::
```
hamishwillee, line 13: FYI I added section "Experimental" up the top. What I'm trying to do here is explain that the mode stuff works now and already gives you a benefit over using offboard mode, but that many of the abstractions we provide for working with ROS 2 do not exist yet. So you can use it but likely things will change.

I've done it in bullets so you can extend easily if you want. I'm open to any text which sets the level of expectation appropriately (i.e. this is "for discussion").

en/ros2/px4_ros2_interface_lib.md
hamishwillee: I'll add links to the ones we know exist now.

en/ros2/px4_ros2_interface_lib.md
```diff
@@ -0,0 +1,478 @@
...
+The replacement mode can be set in the settings of the `ModeBase` constructor:
+
+```cpp
+Settings{kName, false, ModeBase::kModeIDRtl}
+```
+
+## CI: Integration Tests
+
+When opening a pull request to PX4, CI runs the library integration tests.
+These test that mode registration, failsafes, and mode replacement, work as expected.
```
hamishwillee, line 460: FYI just added a comment about what the tests do now. I know this means possible future maintenance of docs when they do more, but I think it is worth explaining.

en/ros2/px4_ros2_interface_lib.md
```diff
@@ -0,0 +1,478 @@
...
+
+Telemetry topics include:
+
+- [OdometryGlobalPosition](https://github.com/Auterion/px4-ros2-interface-lib/blob/main/px4_ros2_cpp/include/px4_ros2/odometry/global_position.hpp): Global position
+- [OdometryLocalPosition](https://github.com/Auterion/px4-ros2-interface-lib/blob/main/px4_ros2_cpp/include/px4_ros2/odometry/local_position.hpp): Local position, velocity and acceleration.
+
+:::note
+These topics provide a wrapper around the internal PX4 topics, allowing the library to maintain compatibility if the internal topics change.
+Check [px4_ros2/odometry](https://github.com/Auterion/px4-ros2-interface-lib/tree/main/px4_ros2_cpp/include/px4_ros2/odometry) for new topics, and of course you can use any ROS 2 topic published from PX4.
+:::
```
hamishwillee, line 382: This needs checking. I wanted to add a little information on what we support now at this level, and a reason why we bother.

en/ros2/px4_ros2_interface_lib.md
```diff
@@ -0,0 +1,478 @@
...
+```
+
+- `[1]`: First we create a class that inherits from `px4_ros2::ModeExecutorBase`.
+- `[2]`: The constructor takes our custom mode that is associated with the executor and passes it to the constructor of `ModeExecutorBase`.
+- `[3]`: We define an enum for the states we want to run through.
+- `[4]`: `onActivate` gets called when the executor becomes active. At this point we can start to run through our states.
+  How you do this is up to you, in this example a method `runState` is used to execute the next state.
+- `[5]`: According to the state, we call a method from `ModeExecutorBase` that asynchronously runs the requested mode.
+  A callback is passed which is called upon completion.
+  This provides a `Result` argument which tells you whether the operation succeeded or not.
```
hamishwillee, line 329: My C++ is a bit rusty, but I think that  `run`, `takeoff`, `rtl`, `WaitUntilDisarmed` are functions that take a callback when they compete, and you are using these to drive the state machine - right?

Do we have a list of the available functions and explanation of their behaviour? Because I expect these have mappings to existing modes.

Further, we need to mention  `scheduleMode()` because I think that is how you run an arbitrary mode. Right?

hamishwillee:
@bkueng I've addressed most of my own concerns with this - setting expectation - by adding an experimental header. Can you please check that is OK. Perhaps there is a better way of explaining the state of the SDK.

The only one thing missing that I think might block people from using this is this question on the mode executor: https://github.com/PX4/PX4-user_guide/pull/2740/files#r1402837201

I'll merge tomorrow if you think this is reasonable.

en/ros2/px4_ros2_interface_lib.md
bkueng: Yes and yes. We can mention it, but in general I think we should have generated API docs for these.

en/ros2/px4_ros2_interface_lib.md
```diff
@@ -1,5 +1,17 @@
...
 
+:::warning Experimental
+At time of writing parts of the PX4 ROS 2 Interface Library is experimental, and hence subject to change:
+
+- The architecture and core interfaces for defining modes in ROS 2 modes are largely stable.
+  The library offers significant benefits over using offboard mode in its current state.
+- Only a few setpoint types have settled (the others are still under development).
+  You may need to use internal PX4 topics which may not remain backwards-compatible over time.
+- The API is not fully documented.
+- Testing in CI is still limited.
```
bkueng, line 11: This is ok, but just fyi most of the things that are there are tested. Mostly the setpoint types (with their expected behavior) that are not.

bkueng:
Thanks Hamish. Looks good to me.

en/ros2/px4_ros2_interface_lib.md
hamishwillee: Then let's say that :-)

en/ros2/px4_ros2_interface_lib.md
```diff
@@ -0,0 +1,478 @@
...
+
+:::warning Experimental
+At time of writing parts of the PX4 ROS 2 Interface Library is experimental, and hence subject to change:
+
+- The architecture and core interfaces for defining modes in ROS 2 modes are largely stable.
+  The library offers significant benefits over using offboard mode in its current state.
+- Only a few setpoint types have settled (the others are still under development).
+  You may need to use internal PX4 topics which may not remain backwards-compatible over time.
+- The API is not fully documented.
+- Testing in CI is still limited.
```
hamishwillee, line 11: Minor variation. Clarifies that the core stuff is tested. Doesn't mention testing on the setpoint types, but you'd probably assume not fully tested as they are experimental.

```suggestion
- The architecture and core interfaces for defining modes in ROS 2 modes are largely stable, and are tested in CI.
  The library offers significant benefits over using offboard mode in its current state.
- Only a few setpoint types have settled (the others are still under development).
  You may need to use internal PX4 topics which may not remain backwards-compatible over time.
- The API is not fully documented.
```

en/ros2/px4_ros2_interface_lib.md
hamishwillee: @bkueng OK I spelled it out.

```suggestion
- `[1]`: First we create a class that inherits from `px4_ros2::ModeExecutorBase`.
- `[2]`: The constructor takes our custom mode that is associated with the executor and passes it to the constructor of `ModeExecutorBase`.
- `[3]`: We define an enum for the states we want to run through.
- `[4]`: `onActivate` gets called when the executor becomes active. At this point we can start to run through our states.
  How you do this is up to you, in this example a method `runState` is used to execute the next state.
- `[5]`: On switching to a state we call an asynchronous method from `ModeExecutorBase` to start the desired mode: `run`, `takeoff`, `rtl`, and so on.
  These methods are passed a function that is called on completion; the callback provides a `Result` argument that tells you whether the operation succeeded or not.
  The callback runs the next state on success.
- `[6]`: In the state handler for the mode executors "owned mode" we use the `scheduleMode()` method to start the mode.
  This follows the same callback pattern as the other methods/states.
```

en/ros2/px4_ros2_interface_lib.md
```diff
@@ -0,0 +1,477 @@
...
+
+    switch (state) { // [5]
+      case State::Reset:
+        break;
+
+      case State::TakingOff:
+        takeoff([this](px4_ros2::Result result) {runState(State::MyMode, result);});
+        break;
+
+      case State::MyMode:
```
hamishwillee, line 297:
```suggestion
      case State::MyMode: // [6]
```

en/ros2/px4_ros2_interface_lib.md
hamishwillee: I am going to assume you have this on your list :-)

hamishwillee (approved):
OK @bkueng made a few minor changes to address your responses to my feedback. I'm going to merge. I wish we could do a lot more, but good to get this out there and visible.

PR #2751, opened 2023-09-22
Tachometer: remove from main sensor list and change description

sfuhrer:
See commit messages for more detail.

### DIFF
--- a/en/getting_started/sensor_selection.md
+++ b/en/getting_started/sensor_selection.md
@@ -1,13 +1,16 @@
 # Sensors
 
-PX4-based systems use sensors to determine vehicle state (needed for stabilization and to enable autonomous control). The vehicle states include: position/altitude, heading, speed, airspeed, orientation (attitude), rates of rotation in different directions, battery level, etc.
+PX4-based systems use sensors to determine vehicle state (needed for stabilization and to enable autonomous control).
+The vehicle states include: position/altitude, heading, speed, airspeed, orientation (attitude), rates of rotation in different directions, battery level, and so on.
 
-The system *minimally requires* a gyroscope, accelerometer, magnetometer (compass) and barometer. A GPS or other positioning system is needed to enable all automatic [modes](../getting_started/flight_modes.md#categories), and some assisted modes. Fixed wing and VTOL-vehicles should additionally include an airspeed sensor (very highly recommended).
+PX4 *minimally requires* a gyroscope, accelerometer, magnetometer (compass) and barometer.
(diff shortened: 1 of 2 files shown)


en/sensor/tachometers.md
```diff
@@ -1,6 +1,6 @@
 # Tachometers (Revolution Counters)
 
-Tachometers (also known as [revolution-counter sensors](https://en.wikipedia.org/wiki/Tachometer#In_automobiles,_trucks,_tractors_and_aircraft)) are *highly recommended* for rotor-wing frames because they allow the autopilot to detect stall or another rotor failure (for rotor-flight it is the rotation of blades that guarantees lift not airspeed or ground speed).
+Tachometers (also known as [revolution-counter sensors](https://en.wikipedia.org/wiki/Tachometer#In_automobiles,_trucks,_tractors_and_aircraft)) can be used to measure the rate of rotation turning vehicle parts like rotors or wheels. Example applications of tachometers are the measuring of the rate of rotation of autogyro rotors or gasoline engines.
```
hamishwillee, line 3: So the point here is that in theory they can be useful but PX4 does not use them "specifically"? Should we make that clear?

MaEtUgR: Yes, I think right now it only logs the data in a generic rpm message. But that could be used in the future or custom implementation.

en/getting_started/sensor_selection.md
```diff
@@ -40,15 +40,6 @@ They are so important because the autopilot does not have other means to detect
 
 For more information and recommended hardware see: [Airspeed Sensors](../sensor/airspeed.md).
 
-## Tachometer
```
MaEtUgR, line 43: Same comment like here: https://github.com/PX4/PX4-user_guide/pull/2753#discussion_r1335013837

sfuhrer: But do we need to keep them in the sensor overview? I didn't remove the tachometer docs, just from the overview (I don't find them important enough for this list, especially as we nowhere in PX4 make use of the RPM feedback). If we want to cover all sensors that have a driver in PX4 then I find that we need to include others first (external IMU, baros, temperature sensors, current sensors etc).

hamishwillee: I agree "more" with @sfuhrer . I think we should have a list at the end of other peripherals that might be of interest, or even just a link and statement they exist. Let me think.

hamishwillee:
Thanks - this is good. I added a note that PX4 logs the RPM does not use it for state estimation or control.

PR #2760, opened 2023-09-27
Update README.md Gazebo Simulation instructions to include how to sol…

idorobotics:
…ve a common error.

### DIFF
--- a/en/sim_gazebo_gz/README.md
+++ b/en/sim_gazebo_gz/README.md
@@ -40,6 +40,10 @@ make px4_sitl gz_x500
 This will run both the PX4 SITL instance and the Gazebo client.
 Note that all gazebo make targets have the prefix `gz_`.
 
+:::note
+If `make px4_sitl gz_x500` gives the error `ninja: error: unknown target 'gz_x500'` then run `make distclean` to start from a clean slate, and try running `make px4_sitl gz_x500` again.
+:::
+
 The supported vehicles and `make` commands are listed below.
 
 Vehicle | Command | `PX4_SYS_AUTOSTART`


en/sim_gazebo_gz/README.md
```diff
@@ -40,6 +40,10 @@ make px4_sitl gz_x500
 This will run both the PX4 SITL instance and the Gazebo client.
 Note that all gazebo make targets have the prefix `gz_`.
 
+:::note
```
hamishwillee, line 43: Thank you. I've run the simulation thousands of time and never run into this error if following the instructions as written. Are you playing with multiple branches etc?

My concern is that the error has nothing to do with gazebo and everything to do with how you manage your code. In that case cluttering up this particular place with a tip like this doesn't make sense. We'd be better off having a troubleshooting guide for the error and link to all similar errors.

idorobotics: Not using multiple branches. To reproduce, see the steps here: https://github.com/PX4/PX4-Autopilot/issues/22119.

I agree that this is a make error but it comes about depending on how the docs are followed. The [Ubuntu development guide](https://docs.px4.io/main/en/dev_setup/dev_env_linux_ubuntu.html#simulation-and-nuttx-pixhawk-targets) says to skip to the ROS 2 installation before setting up the simulation. If you follow the [ROS 2 installation instructions](https://docs.px4.io/main/en/ros/ros2_comm.html#humble) and thereafter try to run the [Gazebo simulation instructions](https://docs.px4.io/main/en/sim_gazebo_gz/) you get this error. It is not obvious to a new user how to proceed. 

Adding this note to the docs will help users to proceed quickly. 
Other users previously encountered this same issue and reported it:
- https://github.com/PX4/PX4-Autopilot/issues/21163 
- https://github.com/PX4/PX4-Autopilot/issues/21599
- https://discuss.px4.io/t/unable-to-run-gazebo-classic-simulation/31540/2 
which shows that some users are struggling to resolve this. Therefore noting this somewhere would be helpful.

hamishwillee: Thank you for the explanation. I will look at this early next week.

hamishwillee: Hmmm. You're working on Ubuntu 20.04 right? So you [install ROS 2/PX4](https://docs.px4.io/main/en/ros/ros2_comm.html#install-px4 instructions) which adds Gazebo Classic. You then run `make px4_sitl` which builds PX4 but no simulator.

Then you're installing [Gazebo](http://docs.px4.io/main/en/sim_gazebo_gz/#installation-ubuntu-linux) and running `make px4_sitl gz_x500` - that should work because you now have Gazebo installed, but fails, most likely because you have build artifacts that haven't been cleaned up properly. 

So you're doing a `make distclean` to clean up the artifacts.
Had you installed onto Ubuntu 22.04 then gazebo would already be installed as part of installing PX4 and you should just be able to do `make px4_sitl gz_x500` at any point after that.

That all said, I think you're right - knowing when you're starting that you have to do a distclean is too much to ask. Thanks!

idorobotics: I'm working on Ubuntu 22.04 LTS with ROS Humble.

hamishwillee: OH, then you shouldn't have done this https://docs.px4.io/main/en/sim_gazebo_gz/#installation-ubuntu-linux

> Gazebo is installed by default on Ubuntu 22.04 as part of the development environment setup: [Ubuntu Dev Environment Setup > Simulation and NuttX (Pixhawk) Targets](https://docs.px4.io/main/en/dev_setup/dev_env_linux_ubuntu.html#simulation-and-nuttx-pixhawk-targets)

idorobotics: Didn't install Gazebo using PX4 instructions.

hamishwillee: Possibly! The `make px4_sitl` is like a step for `px4_sitl gz_x500` but perhaps if you run it without a gazebo target it leaves some configuration in place that says gazebo is not present.

I was thinking perhaps we should just add `make distclean # Optionally clean up any old build artifacts` to the main steps here - it costs very little and then the warning wouldn't be needed.

idorobotics: That's a good idea.

PR #2763, opened 2023-09-28
Rebrand contribution to community

hamishwillee:
This rebrands contribution as community as per @sfuhrer proposed restructure. I have updated the page and the SUMMARY.md.
I also prettified some affected docs - so it is hard to review. I'll tag the specific parts worth looking at.

@sfuhrer @MaEtUgR One question though - should Source Code management be here or in development?
Of course you could argue that is also true for licences, but I think the argument is stronger for source code.

### DIFF
--- a/en/SUMMARY.md
+++ b/en/SUMMARY.md
@@ -683,7 +683,7 @@
       * [ROS/MAVROS Installation on RPi](ros/raspberrypi_installation.md)
       * [External Position Estimation (Vision/Motion based)](ros/external_position_estimation.md)
   * [DroneKit](robotics/dronekit.md)
-* [Contribution (&Dev Call)](contribute/README.md)
+* [Community](contribute/README.md)
   * [Dev Call](contribute/dev_call.md)
   * [Support](contribute/support.md)
   * [Source Code Management](contribute/code.md)
--- a/en/contribute/README.md
+++ b/en/contribute/README.md
@@ -1,23 +1,24 @@
-# Contributing
+# Community
 
 <div v-if="$themeConfig.px4_version != 'main'">
   <div class="custom-block danger"><p class="custom-block-title">This page may be out of date</p>. <p>The latest version <a href="https://docs.px4.io/main/en/contribute/">can be found here</a>.</p>
   </div>
 </div>
 
+Welcome to the PX4 Community!
+
 :::tip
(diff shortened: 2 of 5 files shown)


en/SUMMARY.md
```diff
@@ -683,7 +683,7 @@
       * [ROS/MAVROS Installation on RPi](ros/raspberrypi_installation.md)
       * [External Position Estimation (Vision/Motion based)](ros/external_position_estimation.md)
   * [DroneKit](robotics/dronekit.md)
-* [Contribution (&Dev Call)](contribute/README.md)
+* [Community](contribute/README.md)
```
hamishwillee, line 686: Note, just a name change. I'm OK with losing "dev call" because where else would it be.

en/contribute/README.md
```diff
@@ -1,23 +1,24 @@
-# Contributing
+# Community
```
hamishwillee, line 1: This is updated page to review. All the other changes are mostly prettier updates.

en/contribute/README.md
```diff
@@ -1,23 +1,24 @@
...
-* [Dev Call](../contribute/dev_call.md) - Discuss architecture, pull requests, impacting issues with the dev team
-* [Codeline Management](../contribute/code.md) - Work with PX4 code
-* [Documentation](../contribute/docs.md) - Improve the docs
-* [Translation](../contribute/translation.md) - Translate using Crowdin
-* [Terminology/Notation](../contribute/notation.md)
-* [Licenses](../contribute/licenses.md)
 
-Contributors will also find these sections/topics useful:
-* [Support](../contribute/support.md) - Get help and raise issues
-* [Test Flights](../test_and_ci/test_flights.md) - Get your pull requests flight-tested.
```
hamishwillee, line 23: Test flights made sense here when test flights were a community service.

sfuhrer (approved):
Makes sense!

sfuhrer:
>Source Code management be here or in development

I feel here is better. As any random developer shouldn't have to care how "we" do source code management, it only starts to matter when they start contributing to the community. Makes sense?

hamishwillee:
>> Source Code management be here or in development

> I feel here is better. As any random developer shouldn't have to care how "we" do source code management, it only starts to matter when they start contributing to the community. Makes sense?

Yes!

hamishwillee:
PS Note this is still in current position - not nearer top of the list at top level. As per slack discussions, I think this is better.

PR #2770, opened 2023-09-29
Replay: add dynamic parameters and update working & structure

DanMesh:
- Add instructions for using dynamic parameters in replays.
   - See https://github.com/PX4/PX4-Autopilot/pull/22071.
- General improvements of the replay guide.

### DIFF
--- a/en/debug/system_wide_replay.md
+++ b/en/debug/system_wide_replay.md
@@ -14,7 +14,7 @@ All identified topics need to be logged at full rate (see [logging](../dev_log/l
 For `ekf2` this is already the case with the default set of logged topics.
 
 It is important that all replayed topics contain only a single absolute timestamp, which is the automatically generated field `timestamp`.
-Should there be more timestamps, they must be relative with respect to the main timestamp.
+Should there be more timestamps, they must be relative to the main timestamp.
 For an example, see [SensorCombined.msg](https://github.com/PX4/PX4-Autopilot/blob/main/msg/SensorCombined.msg).
 Reasons for this are given below.
 
@@ -27,11 +27,12 @@ Reasons for this are given below.
   export replay=<absolute_path_to_log_file.ulg>
   make px4_sitl_default
   ```
-  This will create the output in a separate build directory `build/px4_sitl_default_replay` (so that the parameters don't interfere with normal builds).
+  This will create the build/make output in a separate build directory `build/px4_sitl_default_replay` (so that the parameters don't interfere with normal builds).
   It's possible to choose any posix SITL build target for replay, since the build system knows through the `replay` environment variable that it's in replay mode.
 - Add ORB publisher rules in the file `build/px4_sitl_default_replay/rootfs/orb_publisher.rules`.
   This file defines which module is allowed to publish which messages.
   It has the following format:
+  
   ```
   restrict_topics: <topic1>, <topic2>, ..., <topicN>
   module: <module>
@@ -75,7 +76,7 @@ Reasons for this are given below.
   unset replay
   ```
 
-### Overriding parameters in the original log
+### Overriding Parameters in the Original Log
 
 By default, all parameters from the original log file are applied during a replay.
 If a parameter changes during recording, it will be changed at the right time during the replay.
--- a/en/debug/system_wide_replay.md
+++ b/en/debug/system_wide_replay.md
@@ -30,7 +30,7 @@ Reasons for this are given below.
   This will create the build/make output in a separate build directory `build/px4_sitl_default_replay` (so that the parameters don't interfere with normal builds).
   It's possible to choose any posix SITL build target for replay, since the build system knows through the `replay` environment variable that it's in replay mode.
 - Add ORB publisher rules in the file `build/px4_sitl_default_replay/rootfs/orb_publisher.rules`.
-  This file defines which module is allowed to publish which messages.
+  This file defines the modules that are allowed to publish particular messages.
   It has the following format:
   
   ```
@@ -53,7 +53,7 @@ Reasons for this are given below.
   
   With this, the modules that usually publish these topics don't need to be disabled for the replay.
 
-- _(Optional)_ Setup parameter overrides (see instructions below).
+- _(Optional)_ Setup parameter overrides (see [instructions below](#overriding-parameters-in-the-original-log)).
 - _(Optional)_ Copy a `dataman` mission file from the SD card to the build directory.
   This is only necessary if a mission should be replayed.
 - Start the replay:
--- a/en/debug/system_wide_replay.md
+++ b/en/debug/system_wide_replay.md
@@ -1,12 +1,12 @@
 # System-wide Replay
 
-Based on ORB messages, it's possible to record and replay arbitrary parts of the system.
+It is possible to record and replay arbitrary parts of the system based on ORB messages.
 
 Replay is useful to test the effect of different parameter values based on real data, compare different estimators, etc.
 
 ## Prerequisites
 
-The first thing that needs to be done is to identify the module or modules that should be replayed.
+The first step is to identify the module or modules that should be replayed.
 Then, identify all the inputs to these modules, i.e. subscribed ORB topics.
 For system-wide replay, this consists of all hardware input: sensors, RC input, MAVLink commands and file system.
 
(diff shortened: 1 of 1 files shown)


en/debug/system_wide_replay.md
```diff
@@ -1,93 +1,112 @@
...
+The first step is to identify the module or modules that should be replayed.
 Then, identify all the inputs to these modules, i.e. subscribed ORB topics.
 For system-wide replay, this consists of all hardware input: sensors, RC input, MAVLink commands and file system.
 
 All identified topics need to be logged at full rate (see [logging](../dev_log/logging.md)).
 For `ekf2` this is already the case with the default set of logged topics.
 
 It is important that all replayed topics contain only a single absolute timestamp, which is the automatically generated field `timestamp`.
-Should there be more timestamps, then they must be relative with respect to the main timestamp.
+Should there be more timestamps, they must be relative with respect to the main timestamp.
```
hamishwillee, line 17:
```suggestion
Should there be more timestamps, they must be relative to the main timestamp.
```

en/debug/system_wide_replay.md
```diff
@@ -1,93 +1,112 @@
...
 ## Usage
 
-- First, choose the file to replay, and build the target (from within the PX4-Autopilot directory):
+- First, choose the file to replay and build the target (from within the PX4-Autopilot directory):
   
   ```sh
   export replay=<absolute_path_to_log_file.ulg>
   make px4_sitl_default
   ```
   This will create the output in a separate build directory `build/px4_sitl_default_replay` (so that the parameters don't interfere with normal builds).
```
hamishwillee, line 30:
```suggestion
  This will create the build/make output in a separate build directory `build/px4_sitl_default_replay` (so that the parameters don't interfere with normal builds).
```

en/debug/system_wide_replay.md
```diff
@@ -1,93 +1,112 @@
...
-  It's possible to choose any posix SITL build target for replay, the build system knows through the `replay` environment variable that it's in replay mode.
-- Add ORB publisher rules file in `build/px4_sitl_default_replay/rootfs/orb_publisher.rules`.
+  It's possible to choose any posix SITL build target for replay, since the build system knows through the `replay` environment variable that it's in replay mode.
+- Add ORB publisher rules in the file `build/px4_sitl_default_replay/rootfs/orb_publisher.rules`.
   This file defines which module is allowed to publish which messages.
   It has the following format:
```
hamishwillee, line 34:
```suggestion
  It has the following format:
  
```

en/debug/system_wide_replay.md
```diff
@@ -1,93 +1,112 @@
...
+  Note that the above command will show the simulator as well, but - depending on what is being replayed - it will not show what's actually going on.
+  It is still possible to connect via QGC and, for example, view the changing attitude during replay.
 
 - Finally, unset the environment variable, so that the normal build targets are used again:
 
   ```sh
   unset replay
   ```
 
+### Overriding parameters in the original log
```
hamishwillee, line 78:
```suggestion
### Overriding Parameters in the Original Log
```

en/debug/system_wide_replay.md
```diff
@@ -100,50 +120,56 @@ The recording and replay of flight logs with [multiple EKF instances](../advance
...
-* Record the log
-* To replay:
+To perform an EKF2 replay:
 
-```
-export replay_mode=ekf2
-export replay=<abs_path_to_log.ulg>
-make px4_sitl none
-```
+- Record the original log. Optionally set `SDLOG_MODE` to `1` to log from boot.
```
hamishwillee, line 127:
```suggestion
- Record the original log.
  Optionally set `SDLOG_MODE` to `1` to log from boot.
```

en/debug/system_wide_replay.md
```diff
@@ -100,50 +120,56 @@ The recording and replay of flight logs with [multiple EKF instances](../advance
...
-Then edit the parameters in the file as needed and restart the replay process with `make px4_sitl none`. 
-This will create a new log file.
+- Once finished, unset both `replay` and `replay_mode`.
 
-The location of the generated log is printed with a message like this:
+  ```sh
+  unset replay; unset replay_mode
+  ```
+
+### Adjusting EKF2-specific parameters for the replay
```
hamishwillee, line 148: FYI We use first letter of each word capitalization, with normal exceptions.

```suggestion
### Adjusting EKF2-specific Parameters for the Replay
```

en/debug/system_wide_replay.md
```diff
@@ -1,93 +1,113 @@
...
   export replay=<absolute_path_to_log_file.ulg>
   make px4_sitl_default
   ```
-  This will create the output in a separate build directory `build/px4_sitl_default_replay` (so that the parameters don't interfere with normal builds).
-  It's possible to choose any posix SITL build target for replay, the build system knows through the `replay` environment variable that it's in replay mode.
-- Add ORB publisher rules file in `build/px4_sitl_default_replay/rootfs/orb_publisher.rules`.
+  This will create the build/make output in a separate build directory `build/px4_sitl_default_replay` (so that the parameters don't interfere with normal builds).
+  It's possible to choose any posix SITL build target for replay, since the build system knows through the `replay` environment variable that it's in replay mode.
+- Add ORB publisher rules in the file `build/px4_sitl_default_replay/rootfs/orb_publisher.rules`.
   This file defines which module is allowed to publish which messages.
```
hamishwillee, line 33:
```suggestion
  This file defines the modules that are allowed to publish particular messages.
```

en/debug/system_wide_replay.md
```diff
@@ -1,93 +1,113 @@
...
-  ```
-  EKF2_GB_NOISE 0.001
-  ```
-  
-  By default, all parameters from the log file are applied.
-  When a parameter changed during recording, it will be changed as well at the right time during replay.
-  A parameter in the `replay_params.txt` will override the value and changes to it from the log file will not be applied.
-- Optional: copy `dataman` missions file from the SD card to the build directory.
-  Only necessary if a mission should be replayed.
+- _(Optional)_ Setup parameter overrides (see instructions below).
```
hamishwillee, line 56:
```suggestion
- _(Optional)_ Setup parameter overrides (see [instructions below](#overriding-parameters-in-the-original-log)).
```

en/debug/system_wide_replay.md
```diff
@@ -1,93 +1,113 @@
...
-  A parameter in the `replay_params.txt` will override the value and changes to it from the log file will not be applied.
-- Optional: copy `dataman` missions file from the SD card to the build directory.
-  Only necessary if a mission should be replayed.
+- _(Optional)_ Setup parameter overrides (see [instructions below](#overriding-parameters-in-the-original-log)).
+- _(Optional)_ Copy a `dataman` mission file from the SD card to the build directory.
+  This is only necessary if a mission should be replayed.
 - Start the replay:
 
   ```sh
   make px4_sitl_default jmavsim
```
hamishwillee, line 62: I'm wondering if we should change to `make px4_sitl gz_x500`. jMAVSim is not getting much love currently. Something to think about for next iteration.

en/debug/system_wide_replay.md
```diff
@@ -1,93 +1,113 @@
...
   unset replay
   ```
 
+### Overriding Parameters in the Original Log
+
+By default, all parameters from the original log file are applied during a replay.
+If a parameter changes during recording, it will be changed at the right time during the replay.
+
+Parameters can be overridden during a replay in two ways: _fixed_ and _dynamic_.
+If an overridden parameter changes in the log, this change will not be applied in the replay.
```
hamishwillee, line 85:
```suggestion
When parameters are overridden, corresponding parameter changes in the log are not applied during replay.
```

en/debug/system_wide_replay.md
```diff
@@ -1,93 +1,113 @@
...
+  It's possible to choose any posix SITL build target for replay, since the build system knows through the `replay` environment variable that it's in replay mode.
+- Add ORB publisher rules in the file `build/px4_sitl_default_replay/rootfs/orb_publisher.rules`.
+  This file defines the modules that are allowed to publish particular messages.
   It has the following format:
+  
   ```
   restrict_topics: <topic1>, <topic2>, ..., <topicN>
   module: <module>
   ignore_others: <true/false>
   ```
```
hamishwillee, line 40: 1. Can you only define one module? I.e what if you want to define several modules and restrictions?

2. My understanding is that these publisher rules are optional - you could just play the log file your recorded. You add publisher rules because (from below) "the modules that usually publish these topics don't need to be disabled for the replay." . I'm not sure I get this - why do you want to disable publications from one place and not the other?

hamishwillee (approved):
Thanks @DanMesh - much appreciated. 

The changes you made are excellent - normally I have quite a few questions following edits.
I'm going to merge, but if you have answers to https://github.com/PX4/PX4-user_guide/pull/2770/files#r1344897577 we might adjust as a post process.

en/debug/system_wide_replay.md
DanMesh: 1. That is something I am not actually sure of - I have not tried it.

2. The rules probably are optional. That said, in the case of the EKF2 replay at least, the `replay` module republishes certain topics directly from the log. In this case, we want the module that originally published them to _not_ publish them such that there is only one source of those messages. These other modules can keep running, but if they publish on those topics we ignore it.

DanMesh:
Thanks for the feedback and edits @hamishwillee!

I'll follow up on those queries and propose updates if necessary.

en/debug/system_wide_replay.md
hamishwillee: Ah. That makes sense. Looking forward to an update, when and if you are able.

PR #2771, opened 2023-09-29
Add Land mode docs for none-Mission FW landing

sfuhrer:
As promised in https://github.com/PX4/PX4-Autopilot/pull/21036#issuecomment-1698538960.

Should get ported to the v1.14 docs.


hamishwillee (approved):
THanks very much @sfuhrer!  I subedited a little to my taste - mostly just made the warning near the top.

Yes, I will be splitting this out into separate modes. However I'm thinking maybe after taking the PX4 v1.14 cut tomorrow. Mostly because this will have most of a translation and I don't want to lose that.

PR #2775, opened 2023-10-04
pmw3901.md - fix typo pin mapping

hamishwillee:
pmw3901 [SPI wiring here](https://docs.px4.io/main/en/sensor/pmw3901.html#bitcraze-flow-breakout) says 

<html>
<body>

5 (CS1) | Do not connect
-- | --
6 (CS2) | CS

</body>
</html>

However that disagrees with the wiring diagram in same section and the text, which has them the other way around. 
IN addition the other way around was tested by Karthic here and worked https://discord.com/channels/1022170275984457759/1022185721450213396/1158374616268554300

So I am inverting the table to match all other sources of information.

### DIFF
--- a/en/sensor/pmw3901.md
+++ b/en/sensor/pmw3901.md
@@ -103,8 +103,8 @@ Pixhawk SPI Port (from left to right) | Bitcraze flow board
 2 (SCK) | CLK
 3 (MISO) | MISO
 4 (MOSI) | MOSI
-5 (CS1) | Do not connect
-6 (CS2) | CS
+5 (CS1) | CS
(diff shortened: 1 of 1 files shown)


hamishwillee:
FYI @bkueng I hope this makes sense. If there is a way to check this in code? I found an init line `initSPIDevice(DRV_FLOW_DEVTYPE_PMW3901, SPI::CS{GPIO::PortB, GPIO::Pin12})` but couldn't work out how to map the Port B/pin 12 GPI to a particular SPI pin.

Merging anyway - I think three indicators that this is correct are probably enough.

PR #2778, opened 2023-10-05
Flightmodes - split HOLD mode into FC and MC

hamishwillee:
Note that this means we lose the VTOL section. We'll have to address that separately anyway.

I have fixed up links to MC or FW locations based on use - i.e. for collision protection/obstacle avoidance only work on MC so that is where they link. The redirect is to the MC version. If both are relevant I have put both links in (sometimes awkward). I put cross links to equivalent version in the See Also.


sfuhrer:
Thanks, looks good!

PR #2780, opened 2023-10-05
Split up land mode into mc/fw/vtol variants

hamishwillee:
This splits up land mode into vehicle frame-specific variants. 

Generally VTOL is just MC in MC mode and FW in FW mode, however for land mode there is specific behaviour that configures whether an VTOL in FW mode switches to MC at the end of landing. As there is nowhere to put that (yet). I have temporarily made the original land mode page into a land-mode VTOL page. I will revisit VTOL modes and may remove this later.

I've also made some other general fixes in passing to remove content that has been split, and to fix up a few places where we only mention MC. 

I'm going to merge this, but I'll take Silvan with just one question inline.


en/flight_modes_fw/land.md
```diff
@@ -0,0 +1,49 @@
...
+Where possible, instead use [Return mode](../flight_modes/return.md) with a predefined [Fixed-wing mission landing](../flight_modes/mission.md#fw-mission-landing).
+:::
+
+:::note
+
+- This mode requires a valid global position estimate (from GPS or inferred from a [local position](../ros/external_position_estimation.md#enabling-auto-modes-with-a-local-position)).
+- In a failsafe the mode only requires altitude (typically a barometer is built into the flight controller).
+- This mode is automatic - no user intervention is _required_ to control the vehicle.
+- RC control switches can be used to change flight modes on any vehicle.
+- RC stick movement is ignored.
```
hamishwillee, line 21: @sfuhrer Note, it is implied but not stated in the old version that FW land mode ignores stick input. Is that correct? I have added that it is here.

PR #2798, opened 2023-10-12
Fixing airframe code for VTOL and Cessna

vudala:
Error while performing multi-vehicle simulation using Gazebo Garden: VTOL and Cessna airframe codes were inverted


hamishwillee (approved):
Thanks very much!

PR #2799, opened 2023-10-12
Fixing cessna aiframe code

vudala:
Error while performing multi-vehicle simulation using Gazebo Garden: Cessna airframe code was wrong


hamishwillee (approved):
Thanks very much!

PR #2801, opened 2023-10-13
update default interaction to use name alias

frede791:
With the old default example, vehicles were set half in the ground. Using `PX4_SIM_MODEL` rather than `PX4_GZ_MODEL` the model takes the default pose defined in its sdf file, rather than the default pose 0,0,0,0,0,0. 

Related to issue: https://github.com/PX4/PX4-Autopilot/issues/22214


Jaeyoung-Lim (approved):

hamishwillee:
Thanks very much!

PR #2804, opened 2023-10-16
Reversing simulation instructions 

frede791:
Reversing the simulation instructions for the second and third gazebo instruction in the README. This was incorrectly amended in #2801, commit 7a80d380af750ad489724663240e8b09b06661d0.

### DIFF
--- a/en/sim_gazebo_gz/README.md
+++ b/en/sim_gazebo_gz/README.md
@@ -138,7 +138,7 @@ Here are some examples of the different scenarios covered above.
 2. **Start simulator + default world + spawn vehicle at custom pose (y=2m)**
 
    ```sh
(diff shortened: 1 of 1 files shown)


hamishwillee:
Thanks. Happy to merge but FMI why is it different in the first case from the second two ?

frede791:
@hamishwillee Because the provided parameters are different: 
The first case used to use just `PX4_GZ_MODEL`. 
The second case also provided `PX4_GZ_MODEL_POSE` on top of that. This inclusion leads to a different path been selected due to a different if-condition being fulfilled. 
The third case uses just `PX4_GZ_MODEL_NAME` which again is a different path entirely. 
Thus setting the second and third case as the first one would mess with the provided logic. 

The relevant section of code can be found here: 
https://github.com/PX4/PX4-Autopilot/blob/f120ebcdc0f2110ccfaed1bb5552365f8e679be6/ROMFS/px4fmu_common/init.d-posix/px4-rc.simulator#L94C1-L188C4

Hope this clears it up.

hamishwillee:
Thanks very much.

PR #2805, opened 2023-10-17
Remote ID standard vs broadcast rules differentiation, Cube ID notes, & module testing

dirksavage88:
Basic updates to the remote ID docs that distinguish the standard remote ID rules from broadcast rules. Also some cube ID notes since it is not an ArduRemoteID firmware device and is not able to be updated via PX4 right now. 

Rationale: There is a **ton** of confusion in the community where we have people are doing YT videos on the cube ID but come from aircraft modeler backgrounds (they don't have autopilots so cube ID will def. not work). This confusion is exacerbated by the broadcast module shortage so people are looking to standard remote ID modules as alternatives despite requiring an autopilot and px4/ardu. 

Cube ID has some caveats and these differences are enough to warrant some notes on limitations with the arm status message flags as well as lack of firmware update capability on anything other than an autopilot flashed with ardupilot. 

Finally, we need to give users the ability to test their module before heading out to the flying field as a final check, hence linking drone scanner.

### DIFF
--- a/en/peripherals/remote_id.md
+++ b/en/peripherals/remote_id.md
@@ -7,6 +7,10 @@ Remote ID support is experimental in PX4 v1.14.
 Remote ID is a government mandated technology for UAVs in Japan, the United States of America and the European Union, designed to enable safe sharing of airspace between UAVs and other aircraft.
 The specification requires that UAVs broadcast data such as: real-time location/altitude, serial number, operator ID/location, status, etc.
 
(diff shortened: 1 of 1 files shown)


en/peripherals/remote_id.md
```diff
@@ -49,6 +53,10 @@ The TX and RX on the flight controller must be connected to the RX and TX on the
 | 3 (blk) | RX (IN)  |      |
 | 4 (blk) | GND      | 0    |
 
+Note that Cube ID does not support firmware updates in PX4 (only Ardupilot).
```
hamishwillee, line 56: I think you mean that you have to use ArduPilot/Mission Planner to update cube id firmware. I will replace this with a link to the Cube ID docs.

hamishwillee:
THanks so much for this. I added a little structure and worded in a way that seemed to capture your intent a little more clearly. I'm going to merge, but would happily correct/update further if you think I have it wrong.

Hoping to remove all the hand wavy compliance stuff soon. Please keep me posted on any updates to the behaviour.

PR #2809, opened 2023-10-22
Adding Pixhawk 6X-RT

dk7xe:
adding first version of documentation for Pixhawk 6X-RT based on NXP i.mx rt1176

PR #2810, opened 2023-10-22
pixhawk6-rt

hamishwillee:
Replaces #2809


dk7xe:
@hamishwillee when will this be available on px4.io ?

hamishwillee:
Already published. Usually takes about an hour after a merge but does depend on CI and how overloaded it is.

PR #2817, opened 2023-10-31
Adding explanation on how to specify a world in gazebo

frede791:
A short paragraph explaining how to select a world in a gazebo and the need to differentiate between the default world and any other world.

### DIFF
--- a/en/sim_gazebo_gz/README.md
+++ b/en/sim_gazebo_gz/README.md
@@ -57,24 +57,34 @@ Vehicle | Command | `PX4_SYS_AUTOSTART`
 The commands above launch a single vehicle with the full UI.
 *QGroundControl* should be able to automatically connect to the simulated vehicle.
 
-In order to run the simulation without running the Gazebo gui, use the `HEADLESS=1` flag:
+### Headless Mode
+
+The simulation can be run without the Gazebo GUI using the `HEADLESS=1` flag:
 
 ```bash
 HEADLESS=1 make px4_sitl gz_x500
 ```
 
-It is also possible to use the above `make` command to specify a desired world. To do so, simply concatenate the desired world to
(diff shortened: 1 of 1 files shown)


en/sim_gazebo_gz/README.md
```diff
@@ -63,6 +63,20 @@ In order to run the simulation without running the Gazebo gui, use the `HEADLESS
 HEADLESS=1 make px4_sitl gz_x500
 ```
 
+It is also possible to use the above `make` command to specify a desired world. To do so, simply concatenate the desired world to
+the desired vehicle. For example: If you want to run the windy world with the `x500` vehicle you can specify: `make px4_sitl gz_x500_windy`.
+If no world is specified, PX4 will use the default world. The supported worlds are listed below.
```
hamishwillee, line 68:
```suggestion
It is also possible to use the above `make` command to specify a desired world. To do so, simply concatenate the desired world to
the desired vehicle. For example: If you want to run the windy world with the `x500` vehicle you can specify: `make px4_sitl gz_x500_windy`.
If no world is specified, PX4 will use the default world. The supported worlds are listed below.
```

hamishwillee:
@frede791 This is useful - particularly making it clear that the default must not be added. I've subedited to my taste, mostly to start adding some structure to the running the simulation topic.

In addition, next time you make an edit, can you please create a branch rather than committing to your main branch. That will allow me to edit and test in git rather than having to modify using the UI.

I'll merge this now, but I have a question. The following section on usage assumes that you are calling the prebuilt binary with an environment variable. Do you happen to know you also do all of these with `make` in the same way you can with `HEADLESS` ? If so that means there are two ways you might set the world using `make`.

frede791:
@hamishwillee Concerning the main, I only realized I was on my own main branch after I had already made a merge request, sorry. Concerning your question, running `ARGS` with `make` is also possible. Do you want to add a note for this?

hamishwillee:
@frede791 Thanks very much. I'll add a note and ping you as a reviewer.

PR #2818, opened 2023-10-31
Add Vertiq Documentation to PX4 User Guide

vertiq-jordan:
This pull request proposes to add Vertiq module documentation to the PX4 User Guide as part of the effort to improve PX4 and Vertiq integration. I have added a new Vertiq page reachable from the side bar as well as the list of DroneCAN ESCs, similarly to VESC's documentation. There is one new image added to the project's assets.

This documentation has been through Vertiq's internal PR process.


mrpollo:
Awesome great to see @vertiq-jordan!

hamishwillee:
@vertiq-jordan Thanks very much for this! I couldn't push my subedits to your branch, so I have replaced this with #2819 and merged.

Next time you submit something would be great if you would branch off your main (i.e. rather than adding your commits direct to your fork main branch), and ideally open access to allow us to edit.

The main changes made in that PR were
- Shrink image using lossless compressor
- We're not "polite" so instructions tend to be "do X" not "please do X".
- Links go to enough information to understand the target (not "see here"). This is good for screenreaders.
- Prettier run to tidy layout.
- Recommended DShot rather than presenting as both DShot and PWM. Sure you support both, but PX4 does too, and DShot should always be used as PWM protocol if supported over any of the other PWM-like protocols. 

FWIW these look like an excellent product :-). Thanks again.

PR #2836, opened 2023-11-02
Gazebo make command is configured by env variables

hamishwillee:
Gazebo `make` can be configured using the same environment variables as used to directly start PX4 sitl from its binary. This adds a note to that effect. Follows on from #2817

@frede791 Can you please sanity check this? I added a question inline too.


en/sim_gazebo_gz/README.md
```diff
@@ -98,11 +107,23 @@ In particular, it is possible to:
 
 These scenarios are managed by setting the appropriate environment variables.
 
+:::note
+The environment variables described below can be used as a prefix to the `make px4_sitl gz*` command as well as for starting PX4 SITL from a prebuilt binary.
+
+These will override any settings defined by the make command itself.
+For example, this following command will start the `gz_rc_cessna` model rather than the `gz_x500` model.
```
hamishwillee, line 114: @frede791 This is an assumption - is it correct that the environment variables override make settings as I have indicated? If not, what's the rule?

frede791: This is correct: `PX4_GZ_MODEL` overrides make settings. However, the model itself does not actually load into Gazebo. It throws: `Error finding file [gz_rc_cessna/model.sdf].`. This even happens when you provide the same model for the make settings and `PX4_GZ_MODEL`. 
Why this happens I don't know but I'll investigate it further. Based on this, however, I wouldn't advertise this as a way of starting Gazebo.

frede791 (changes requested):
I would take out the override command provided as it currently doesn't work and breaks Gazebo.

hamishwillee:
OK, I removed the bit below. We can add it back if you are able to fix this. Thanks for the review, I'll merge it now.

>:::note
The environment variables described below can be used as a prefix to the `make px4_sitl gz*` command as well as for starting PX4 SITL from a prebuilt binary.

> These will override any settings defined by the make command itself.
For example, this following command will start the `gz_rc_cessna` model rather than the `gz_x500` model.

> ```sh
> PX4_GZ_MODEL=gz_rc_cessna make px4_sitl gz_x500
> ```

:::

PR #2840, opened 2023-11-02
Hiding older or less used tools

hamishwillee:
@dagar This is first bit of work following on from our discussion about showing more important tools first. I propose we put dev environments, IDEs, simulators under **Community Supported Developer Tools** which has a clear statement of "no support from dev team". Individual pages then have a header just saying they are community supported. Both cases shown at end of this note.

Note:
- Right now the supported stuff is Gazebo, Gazebo Classic, JMAVSim SITL.
- Once we get Gazebo working on macOS we could retire JMAVSim to the community supported bucket.
- Unsupported is HITL, SIH, JSBSim, Airsim, FlightGear. 

1. Is that OK - i.e. are things in right sets above?
2. Should any of this community supported stuff be deleted - I'm thinking Cygwin setup, and Qt Creator.
3. Is the approach OK. If you show broad agreement then I can proceed. 

I will proceed with this if you show broad agreement and respond to the questions 

## Community page, hidden away

image

## Banner on "hidden pages"

image


hamishwillee:
Well, you can't say I didn't try to engage. Merging since I already merged some of this in #2844 by accident

PR #2851, opened 2023-11-07
fix wrong code in mavros_offboard_python.md

Rainlv:


hamishwillee:
Thanks very much @Rainlv 

PX4 translations start as English and round-trip through a tool called Crowdin where the translations were done. We can't take fixes to the zh tree in the source because this comes out of Crowdin, and if we fix it here the version will get destroyed next time there is an update form Crowdin. I have fixed this in Crowdin. 

For more information see https://docs.px4.io/main/en/contribute/translation.html#translation

PR #2812, opened 2023-10-23
Add link to Pixhawk 6X-RT doc page

dk7xe:
Add Pixhawk 6X-RT doc link to the list of Flight controllers.
@hamishwillee have forgotten adding the link in previous pull request.

PR #2853, opened 2023-11-08
Fix up image and build instruction 6x-rt

hamishwillee:
Replaces #2812 which I am unable to edit.

Note that it doesn't update summary because that wasn't needed - this already appears in sidebar. It does fix up the images and the make command.

FYI @dk7xe


dk7xe:
Thank you!

PR #2861, opened 2023-11-09
WSL2 fixes

hamishwillee:
Fixes to WSL2 docs, in particular removing restrictions on Windows 10 and explaining how to work with multiple versions of Ubuntu. 

FYI @MaEtUgR


MaEtUgR:
Thanks @hamishwillee for the improvements! They were put to use when I just installed on a new machine 👍 
Is the `kisak-mesa` part to get gz running out of the box?

hamishwillee:
@MaEtUgR The `kisak-mesa` isn't "required" to get things running out of the box, which is why I put it in troubleshooting. Until I installed those latest drivers I was getting segfaults on closing the simulation. I'd rather not recommend this by default because at some point the default drivers should be good enough. But if you feel differently I will take direction from you!

There was a very long discussion that lead to all of this in https://github.com/gazebosim/gz-sim/issues/920

PR #2864, opened 2023-11-09
Update land.md - remove unhelpful reference to failsafe

sfuhrer:
The failsafe mode for any Auto mode is Descend, where the vehicle descends vertically without holding position. It's though not "Land" mode anymore. 
COM_POSCTL_NAVL allows you to set the failsafe action.

This was raised in https://discuss.px4.io/t/do-really-all-autonomous-modes-require-global-position/34890/10. We better keep all the failsafe docs in one place.

### DIFF
--- a/en/flight_modes_mc/land.md
+++ b/en/flight_modes_mc/land.md
@@ -8,7 +8,6 @@ The vehicle will disarm shortly after landing (by default).
 :::note
 
 - This mode requires a valid global position estimate (from GPS or inferred from a [local position](../ros/external_position_estimation.md#enabling-auto-modes-with-a-local-position)).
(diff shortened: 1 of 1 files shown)


hamishwillee:
@sfuhrer I'll merge this as it is confusing.  I'd like to capture that you've effectively changed mode in failsafe somehow, because it isn't obvious that "LAND" is descend. So perhaps a note in the failsafe docs here: https://docs.px4.io/main/en/config/safety.html#failsafe-actions - i.e. we note this is Descend?

sfuhrer:
True, also https://docs.px4.io/main/en/config/safety.html#position-gps-loss-failsafe needs to have "Land" replaced by "Descend"

PR #2871, opened 2023-11-13
Update uxrce_dds.md about multi publishing dds subscriptions

GuillaumeLaine:
Provides documentation concerning a new feature to add multi-publication for topics subscribed over DDS.  
Relates to: PX4/PX4-Autopilot#22266

### DIFF
--- a/en/middleware/uxrce_dds.md
+++ b/en/middleware/uxrce_dds.md
@@ -445,7 +445,7 @@ subscriptions_multi:
 
 Each (`topic`,`type`) pairs defines:
 
-1. A new subscription or publication depending on the list to which it is added.
+1. A new publication, subscription, or subscriptions_multi, depending on the list to which it is added.
 2. The topic _base name_, which **must** coincide with the desired uORB topic name that you want to publish/subscribe.
    It is identified by the last token in `topic:` that starts with `/` and does not contains any `/` in it.
    `vehicle_odometry`, `vehicle_status` and `offboard_control_mode` are examples of base names.
--- a/en/middleware/uxrce_dds.md
+++ b/en/middleware/uxrce_dds.md
@@ -455,9 +455,19 @@ Each (`topic`,`type`) pairs defines:
    - `/fmu/in/` for topics that are _subscribed_ by PX4.
 4. The message type (`VehicleOdometry`, `VehicleStatus`, `OffboardControlMode`, etc.) and the ROS 2 package (`px4_msgs`) that is expected to provide the message definition.
 
-We distinguish `subscriptions` from `subscriptions_multi` because PX4 uORB allows multiple publishers to a single uORB topic:
-  - Each field in `subscriptions` defines a ROS2 subscription (e.g. over `/fmu/in/vehicle_optical_flow_vel`) along with a uORB publication (e.g. over `vehicle_optical_flow_vel`) of those messages, under the assumption that only a **single publisher exists to the specified uORB topic**. 
-  - Each field in `subscriptions_multi` also defines such a ROS2 subscription and uORB republication, however **other publishers to the same uORB topic may coexist** (e.g. an internal PX4 module can also publishing over `vehicle_optical_flow_vel`).
+:::note
+We distinguish `subscriptions` from `subscriptions_multi` because PX4 uORB allows multiple publishers to a single uORB topic using the concept of topic instances.
+:::
+
+Add a topic to the `subscriptions` section to:
+- Expose messages from a ROS2 topic (e.g. subscribes over `/fmu/in/vehicle_odometry`) to instance 0 (the first instance) of a uORB topic (e.g. publishes over `vehicle_odometry`).
+- If another (internal) PX4 module is already publishing on the same uORB topic, its subscribers will receive both streams of messages.
+- This is the desired behavior when the specified uORB topic has no other publishers, or you want to replace an internal publisher on this topic (for example during offboard control).
+
+Add a topic to the `subscriptions_multi` section to:
+- Expose messages from a ROS2 topic to a newly allocated instance of a uORB topic (e.g. publishes over new instance of `vehicle_odometry`).
+- By doing so the subscribers can distinguish between the different publishers on a common topic by using only certain instances.
+- This is the desired behavior, for example, when you want PX4 to log the readings of two equal sensors; they will both publish on the same topic, but one will use instance 0 and the other will use instance 1.
 
 You can arbitrarily change the configuration.
 For example, you could use different default namespaces or use a custom package to store the message definitions.
--- a/en/middleware/uxrce_dds.md
+++ b/en/middleware/uxrce_dds.md
@@ -458,7 +458,7 @@ Each (`topic`,`type`) pairs defines:
 `subscriptions` and `subscriptions_multi` allow us to choose the uORB topic instance that ROS 2 topics are routed to: either a shared instance that may also be getting updates from internal PX4 uORB publishers, or a separate instance that is reserved for ROS2 publications, respectively.
 Without this mechanism all ROS 2 messages would be routed to the _same_ uORB topic instance (because ROS 2 does not have the concept of [multiple topic instances](../middleware/uorb.md#multi-instance)), and it would not be possible for PX4 subscribers to differentiate between streams from ROS 2 or PX4 publishers.
 
-Add a topic to the `subscriptions` section the uXRCE-DDS module to:
+Add a topic to the `subscriptions` section to:
 
 - Create a unidirectional route going from the ROS2 topic to the _default_ instance (instance 0) of the associated uORB topic.
   For example, it creates a ROS2 subscriber of `/fmu/in/vehicle_odometry` and a uORB publisher of `vehicle_odometry`.
@@ -466,9 +466,9 @@ Add a topic to the `subscriptions` section the uXRCE-DDS module to:
   The uORB subscriber will not be able to determine if an incoming message was published by PX4 or by ROS2.
 - This is the desired behavior when the ROS2 publisher is expected to be the sole publisher on the topic instance (for example, replacing an internal publisher to the topic during offboard control), or when the source of multiple publishing streams does not matter.
 
-Add a topic to the `subscriptions_multi` section the uXRCE-DDS module to:
+Add a topic to the `subscriptions_multi` section to:
 
-- Creates a unidirectional route going from the ROS2 topic to a _new_ instance of the associated uORB topic.
+- Create a unidirectional route going from the ROS2 topic to a _new_ instance of the associated uORB topic.
   For example, if `vehicle_odometry` has already `2` instances, it creates a ROS2 subscriber of `/fmu/in/vehicle_odometry` and a uORB publisher on instance `3` of `vehicle_odometry`.
 - This ensures that no other internal PX4 module will publish on the same instance used by uXRCE-DDS.
   The subscribers will be able to subscribe to the desired instance and distinguish between publishers.
--- a/en/middleware/uxrce_dds.md
+++ b/en/middleware/uxrce_dds.md
@@ -460,14 +460,14 @@ While in general ROS2 does not allow subscribers to distinguish between multiple
 For this reason, we distinguish between `subscriptions` and `subscriptions_multi`.
 :::
 
-When you add a topic to the  `subscriptions` section the uXRCE-DDS module:
-- Creates a unidirectional route going from the ROS2 topic to the _default_ instance (instance 0) of the associated uORB topic.
-For example, it creates a ROS2 subscriber of `/fmu/in/vehicle_odometry` and a uORB publisher of `vehicle_odometry`.
+Add a topic to the  `subscriptions` section the uXRCE-DDS module to:
+- Create a unidirectional route going from the ROS2 topic to the _default_ instance (instance 0) of the associated uORB topic.
+  For example, it creates a ROS2 subscriber of `/fmu/in/vehicle_odometry` and a uORB publisher of `vehicle_odometry`.
 - If other (internal) PX4 modules are already publishing on the same uORB topic instance as the ROS2 publisher, the instance's subscribers will receive all streams of messages.
   The uORB subscriber will not be able to determine if an incoming message was published by PX4 or by ROS2.
 - This is the desired behavior when the ROS2 publisher is expected to be the sole publisher on the topic instance (for example, replacing an internal publisher to the topic during offboard control), or when the source of multiple publishing streams does not matter.
 
-When you add a topic to the `subscriptions_multi` section the uXRCE-DDS module:
+Add a topic to the `subscriptions_multi` section the uXRCE-DDS module to:
 - Creates a unidirectional route going from the ROS2 topic to a _new_ instance of the associated uORB topic.
   For example, if `vehicle_odometry` has already `2` instances, it creates a ROS2 subscriber of `/fmu/in/vehicle_odometry` and a uORB publisher on instance `3` of `vehicle_odometry`.
 - This ensures that no other internal PX4 module will publish on the same instance used by uXRCE-DDS.
--- a/en/middleware/uxrce_dds.md
+++ b/en/middleware/uxrce_dds.md
@@ -456,7 +456,7 @@ Each (`topic`,`type`) pairs defines:
 4. The message type (`VehicleOdometry`, `VehicleStatus`, `OffboardControlMode`, etc.) and the ROS 2 package (`px4_msgs`) that is expected to provide the message definition.
 
 `subscriptions` and `subscriptions_multi` allow us to choose the uORB topic instance that ROS 2 topics are routed to: either a shared instance that may also be getting updates from internal PX4 uORB publishers, or a separate instance that is reserved for ROS2 publications, respectively.
-Without this mechanism all ROS 2 messages would be routed to the _same_ uORB topic instance (because ROS 2 does not have the concept of [multiple topic instances](../middleware/uorb.md#multi-instance)), and it would not be possible PX4 subscribers to differentiate between streams from ROS 2 or PX4 publishers.
+Without this mechanism all ROS 2 messages would be routed to the _same_ uORB topic instance (because ROS 2 does not have the concept of [multiple topic instances](../middleware/uorb.md#multi-instance)), and it would not be possible for PX4 subscribers to differentiate between streams from ROS 2 or PX4 publishers.
 
 Add a topic to the `subscriptions` section the uXRCE-DDS module to:
 
--- a/en/middleware/uxrce_dds.md
+++ b/en/middleware/uxrce_dds.md
@@ -463,13 +463,12 @@ For this reason, we distinguish between `subscriptions` and `subscriptions_multi
 When you add a topic to the  `subscriptions` section the uXRCE-DDS module:
 - Creates a unidirectional route going from the ROS2 topic to the _default_ instance (instance 0) of the associated uORB topic.
 For example, it creates a ROS2 subscriber of `/fmu/in/vehicle_odometry` and a uORB publisher of `vehicle_odometry`.
-- If other (internal) PX4 modules are already publishing on the same uORB topic, its subscribers will receive all streams of messages.
-The uORB subscriber will not be able to determine if an incoming message was published by PX4 or by ROS2.
-- This is the desired behavior when the specified uORB topic has no other publishers, or when you want to replace an internal publisher on this topic (for example during offboard control).
+- If other (internal) PX4 modules are already publishing on the same uORB topic instance as the ROS2 publisher, the instance's subscribers will receive all streams of messages.
+  The uORB subscriber will not be able to determine if an incoming message was published by PX4 or by ROS2.
+- This is the desired behavior when the ROS2 publisher is expected to be the sole publisher on the topic instance (for example, replacing an internal publisher to the topic during offboard control), or when the source of multiple publishing streams does not matter.
 
 When you add a topic to the `subscriptions_multi` section the uXRCE-DDS module:
-- Queries the uORB manager and get the number `n` of instances currently in use for the desired topic.
-- Creates a unidirectional route going from the ROS2 topic to the a _new_ instance (instance `n+1`) of the associated uORB topic.
+- Creates a unidirectional route going from the ROS2 topic to a _new_ instance of the associated uORB topic.
   For example, if `vehicle_odometry` has already `2` instances, it creates a ROS2 subscriber of `/fmu/in/vehicle_odometry` and a uORB publisher on instance `3` of `vehicle_odometry`.
 - This ensures that no other internal PX4 module will publish on the same instance used by uXRCE-DDS.
   The subscribers will be able to subscribe to the desired instance and distinguish between publishers.
--- a/en/middleware/uxrce_dds.md
+++ b/en/middleware/uxrce_dds.md
@@ -457,7 +457,7 @@ Each (`topic`,`type`) pairs defines:
 
 :::note
 While in general ROS2 does not allow subscribers to distinguish between multiple nodes publishing on the same topic, PX4 does implement this feature allowing uORB topics to have [multiple instances](../middleware/uorb.md#multi-instance).
-For this reason, we distinguish between `subscriptions` and `subscriptions_multi`.
+`subscriptions` and `subscriptions_multi` allows us to choose the uORB instance that ROS 2 topics are routed to: either a shared instance that may also be getting updates from PX4 published, or a separate instance that is reserved for ROS2 publications, respectively.
 :::
 
 Add a topic to the  `subscriptions` section the uXRCE-DDS module to:
--- a/en/middleware/uxrce_dds.md
+++ b/en/middleware/uxrce_dds.md
@@ -456,17 +456,25 @@ Each (`topic`,`type`) pairs defines:
 4. The message type (`VehicleOdometry`, `VehicleStatus`, `OffboardControlMode`, etc.) and the ROS 2 package (`px4_msgs`) that is expected to provide the message definition.
 
 :::note
-We distinguish `subscriptions` from `subscriptions_multi` because PX4 uORB allows multiple publishers to a single uORB topic using the concept of topic instances.
+While in general ROS2 does not allow subscribers to distinguish between multiple nodes publishing on the same topic, PX4 does implement this feature allowing uORB topics to have [multiple instances](uorb.md#multi-instance).
+For this reason, we distinguish between `subscriptions` and `subscriptions_multi`.
 :::
 
-Add a topic to the `subscriptions` section to:
-- Expose messages from a ROS2 topic (e.g. subscribes over `/fmu/in/vehicle_odometry`) to instance 0 (the first instance) of a uORB topic (e.g. publishes over `vehicle_odometry`).
-- If another (internal) PX4 module is already publishing on the same uORB topic, its subscribers will receive both streams of messages.
-- This is the desired behavior when the specified uORB topic has no other publishers, or you want to replace an internal publisher on this topic (for example during offboard control).
+When you add a topic to the  `subscriptions` section the uXRCE-DDS module:
+- Creates a unidirectional route going from the ROS2 topic to the _default_ instance (instance 0) of the associated uORB topic.
+For example, it creates a ROS2 subscriber of `/fmu/in/vehicle_odometry` and a uORB publisher of `vehicle_odometry`.
+- If other (internal) PX4 modules are already publishing on the same uORB topic, its subscribers will receive all streams of messages.
+The uORB subscriber will not be able to determine if an incoming message was published by PX4 or by ROS2.
+- This is the desired behavior when the specified uORB topic has no other publishers, or when you want to replace an internal publisher on this topic (for example during offboard control).
 
-Add a topic to the `subscriptions_multi` section to:
-- Expose messages from a ROS2 topic to a newly allocated instance of a uORB topic (e.g. publishes over new instance of `vehicle_odometry`).
-- By doing so the subscribers can distinguish between the different publishers on a common topic by using only certain instances.
+When you add a topic to the  `subscriptions_multi` section the uXRCE-DDS module:
+- Queries the uORB manager and get the number `n` of instances currently in use for the desired topic.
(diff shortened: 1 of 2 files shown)


en/middleware/uxrce_dds.md
```diff
@@ -447,6 +455,10 @@ Each (`topic`,`type`) pairs defines:
    - `/fmu/in/` for topics that are _subscribed_ by PX4.
 4. The message type (`VehicleOdometry`, `VehicleStatus`, `OffboardControlMode`, etc.) and the ROS 2 package (`px4_msgs`) that is expected to provide the message definition.
 
+We distinguish `subscriptions` from `subscriptions_multi` because PX4 uORB allows multiple publishers to a single uORB topic:
+  - Each field in `subscriptions` defines a ROS2 subscription (e.g. over `/fmu/in/vehicle_optical_flow_vel`) along with a uORB publication (e.g. over `vehicle_optical_flow_vel`) of those messages, under the assumption that only a **single publisher exists to the specified uORB topic**. 
+  - Each field in `subscriptions_multi` also defines such a ROS2 subscription and uORB republication, however **other publishers to the same uORB topic may coexist** (e.g. an internal PX4 module can also publishing over `vehicle_optical_flow_vel`).
+
```
beniaminopozzan, line 461: The true difference here is not about the number of publishers on the same tobic, but rather the fact that topics can have multiple instances:
```suggestion
We distinguish `subscriptions` from `subscriptions_multi` because PX4 uORB allows multiple publishers to a single uORB topic using the concept of topic instances:
  - Each field in `subscriptions` defines a ROS2 subscription (e.g. over `/fmu/in/vehicle_optical_flow_vel`) along with a uORB publication (e.g. over `vehicle_optical_flow_vel`) of that topic. Instance zero (the first instance) is used. Therefore, if other (internal) PX4 modules are already publishing on the same topic the subscribers will receive both streams of messages. This is the desired behavior when you want to replace an internal publisher (for example during offboard control).
  - Each field in `subscriptions_multi` also defines such a ROS2 subscription and uORB republication, However, upon the publisher creation, a new instance will be allocated. By doing so the subscribers can distinguish between the different publisers on a common topic by using only certain instances. This is the desired behaviour, for example, when you want PX4 to log the readings of two equal sensors. They will both publish on the same topic, but one will use instance 0 and the other will use instance 1.

```

beniaminopozzan (commented):
Thanks @GuillaumeLaine , I edited a bit the explanation. Let me know what you think about it.

en/middleware/uxrce_dds.md
hamishwillee: FYI added

en/middleware/uxrce_dds.md
```diff
@@ -433,11 +433,19 @@ subscriptions:
...
+    type: px4_msgs::msg::VehicleOpticalFlowVel
+
+  ...
+
 ```
 
 Each (`topic`,`type`) pairs defines:
 
-1. A new subscription or publication depending on the list to which it is added.
+1. A new publication, subscription, or subscriptions_multi, depending on the list to which it is added.
```
hamishwillee, line 448: FYI Added this

en/middleware/uxrce_dds.md
hamishwillee: @beniaminopozzan @GuillaumeLaine My gut feeling is it would be better to modify this a little such that rather than starting with 

> We distinguish `subscriptions` from `subscriptions_multi` because PX4 uORB allows multiple publishers to a single uORB topic using the concept of topic instances:"

We instead have two sections

> Add a topic to the `subscriptions_multi` section when:

and 

> Add a topic to the `subscriptions` section when:

Usually it is better to provide clear direction. If needed you could add what is there now in a note.

I am not sure I get this either. But I THINK that this is saying that a uorb topic such as a battery might have multiple instances, but the subscription yaml for ROS2 doesn't understand instances, only topics. This allows creation of a separate topic that can be subscribed to by ROS2 for each PX4 uorb instance?

hamishwillee:
Can we also have a line added to https://docs.px4.io/main/en/releases/main.html explaining what the feature allows and cross linking. YOu can see the other release notes to work out an appropriate format.

en/middleware/uxrce_dds.md
GuillaumeLaine: @beniaminopozzan Thanks I like your explanation much better, especially the introduction of the notion of instances. Also taking into account @hamishwillee 's feedback, how about:
```suggestion
:::note
We distinguish `subscriptions` from `subscriptions_multi` because PX4 uORB allows multiple publishers to a single uORB topic using the concept of topic instances.
:::

Add a topic to the `subscriptions` section to:
- Expose messages from a ROS2 topic (e.g. subscribes over `/fmu/in/vehicle_odometry`) to instance 0 (the first instance) of a uORB topic (e.g. publishes over `vehicle_odometry`).
- If another (internal) PX4 module is already publishing on the same uORB topic, its subscribers will receive both streams of messages.
- This is the desired behavior when the specified uORB topic has no other publishers, or you want to replace an internal publisher on this topic (for example during offboard control).
  
Add a topic to the `subscriptions_multi` section to:
- Expose messages from a ROS2 topic to a newly allocated instance of a uORB topic (e.g. publishes over new instance of `vehicle_odometry`).
- By doing so the subscribers can distinguish between the different publishers on a common topic by using only certain instances.
- This is the desired behavior, for example, when you want PX4 to log the readings of two equal sensors; they will both publish on the same topic, but one will use instance 0 and the other will use instance 1.
```

GuillaumeLaine: @hamishwillee What this is trying to say is that a `subscriptions` field defines a subscription to a ROS2 topic, and a publication to a uORB topic (using `uORB::Publication`) which is necessarily over instance 0 of the uORB topic. A `subscriptions_multi` field defines the same subscription, but a publication over a new instance of the same uORB topic (using `uORB::PublicationMulti`).
So the term "multi" refers to the publication on the PX4 side and the presence of multiple uORB topic instances, but still a single subscription to the ROS2 side, which is certainly a bit confusing. I'm not sure if that's what you understood, maybe this requires further clarification.

hamishwillee: @GuillaumeLaine Thanks for taking the time to update this. Personally I think it is better, and I'd like it to go into this PR so I can add suggestions as comments. However let's wait and see what @beniaminopozzan says. 

I _think_ I get it now.

Essentially ROS2 publishes/subscribes to _topics_ and does not understand instances? So you have a ROS2 node publishing Battery information say, then a subscriber would get that stream. If you had two ROS2 nodes publishing battery information then a subscriber would get both streams? There is no way to say in ROS2 I have two batteries, other than to create two topics? 

Whereas PX4 uORB can have one topic and multiple instances.

So what I think we're doing here is providing a way to publish from ROS2 to separate uORB instances.
How does this work ROS2 side? Does it mean that each _new_ ROS2 node that publishes to a topic that is marked as subscription_multi will result in a new instance being created PX4-side? I.e. how does the ROS2 mapping to PX4 uOrb topic instances happen? What if the ROS2 node dies - does the instance get removed? 

In other words you do subscription if you want all streams to go to just a single topic instance on PX4 and subscription multi if you want them to go to separate instances so they can be differentiated by PX4. 

Assuming I do get it, I might rewrite your suggestion a litte.

GuillaumeLaine: @hamishwillee Correct, ROS2 side two (battery information) publishers to the same topic cannot inherently be distinguished by a subscriber. But as you said, PX4 uORB side they could be distinguished using one topic and multiple instances.

However, what we are doing here doesn't go as far as as to enable multiple publishers to the same ROS2 topic:

What we're doing is providing a way to publish from a ROS2 topic (`/fmu/in/base_name`) to a **single** additional instance of the associated uORB topic (`base_name`), i.e. if we have internal PX4 uORB publishers to `n` instances of a uORB topic, we now enable a ROS2 topic to publish to a `n+1`-th instance of that topic. However, we're not providing any way to distinguish multiple ROS2 publishers of one ROS2 topic; having multiple publishers and adding the ROS2 topic name to subscriptions_multi will just relay all the messages from different ROS2 publishers to the `n+1`-th uORB topic instance, the messages would remain indistinguishable on that topic instance.

If you already had PX4 publishers to a uORB topic say `base_name` with multiple instances and were to add `/fmu/in/base_name` to the `subscriptions` field, the ROS2 topic messages would always go to the first instance of `base_name` which would either overwrite or get streamed in with a preexisting PX4 publisher's messages.

So `subscriptions` vs `subscriptions_mutli` is only a PX4 distinction, and depends on whether the uORB topic in question has no other publishers (former), or already has PX4 publishers (latter). The DDS mapping is still just from one ROS2 topic to one uORB topic (first instance in the former case, `n+1`th instance in the latter case).

I hope this clears it up, but I'm happy to go over this as many times as necessary. Thanks so much.



Side note: tbh this is all a bit confusing, and the distinction is here because `uORB::Publication` vs `uORB::PublicationMulti` exist separately. Ideally only `uORB::Publication` would exist with a way to automatically increment the topic instance count. I've heard this mentioned and this might already be in the works.

GuillaumeLaine:
> Can we also have a line added to https://docs.px4.io/main/en/releases/main.html explaining what the feature allows and cross linking. YOu can see the other release notes to work out an appropriate format.

@hamishwillee 
I'm not sure under which header this should go under, "Common" seems the most appropriate but is also very general?
When you say cross linking, does this refer to adding the relevant PR link?

hamishwillee:
>> Can we also have a line added to https://docs.px4.io/main/en/releases/main.html explaining what the feature allows and cross linking. YOu can see the other release notes to work out an appropriate format.

> I'm not sure under which header this should go under, "Common" seems the most appropriate but is also very general?
When you say cross linking, does this refer to adding the relevant PR link?

@GuillaumeLaine 

1. You are welcome to add your own heading. I'd say either "ROS2" or "XRCE-DDS" or "Middleware". Once it is there it might get moved around, but any of those are a good start.
2. With respect to cross linking, yes - a link to the new docs, and a link to the PR that did the work, and some connecting text. Some something like:

   > [ROS2] Support for ROS2 to publish to multiple uORB instances in PX4, allowing streams form different sources to be differentiated. Added to PX4 in [PX4-Autopilot#22266](https://github.com/PX4/PX4-Autopilot/pull/22266)

en/middleware/uxrce_dds.md
hamishwillee: @GuillaumeLaine Well done, I do get this now!!!

Essentially this is entirely about allowing PX4 to differentiate between publications from all internal instances and one external ROS2 publication. If you have multiple ROS2 publications then you have no way to differentiate them from each other. 

I am not sure how to capture that yet, and I'm off for a week. I am sure that this was not clear from the original text. I think the bit to capture is this "routing" of the ROS2 publication to either 0 or n+1 instance from ROS2, depending on the subscription multi. That makes it clear that there is just one external publication, and no still no differentiation in ROS2. Probably in the note that currently says 

> "We distinguish `subscriptions` from `subscriptions_multi` because PX4 uORB allows multiple publishers to a single uORB topic using the concept of topic instances."

Perhaps @beniaminopozzan will have an opinion, but if not I am pretty sure I can come up with a good compromise text next week if you guys havent done so before hand.

beniaminopozzan: Thanks @GuillaumeLaine for taking the time and explaining everything!

en/middleware/uxrce_dds.md
```diff
@@ -447,6 +455,28 @@ Each (`topic`,`type`) pairs defines:
...
 4. The message type (`VehicleOdometry`, `VehicleStatus`, `OffboardControlMode`, etc.) and the ROS 2 package (`px4_msgs`) that is expected to provide the message definition.
 
+:::note
+While in general ROS2 does not allow subscribers to distinguish between multiple nodes publishing on the same topic, PX4 does implement this feature allowing uORB topics to have [multiple instances](uorb.md#multi-instance).
+For this reason, we distinguish between `subscriptions` and `subscriptions_multi`.
+:::
+
+When you add a topic to the  `subscriptions` section the uXRCE-DDS module:
+- Creates a unidirectional route going from the ROS2 topic to the _default_ instance (instance 0) of the associated uORB topic.
+For example, it creates a ROS2 subscriber of `/fmu/in/vehicle_odometry` and a uORB publisher of `vehicle_odometry`.
```
hamishwillee, line 465:
```suggestion
  For example, it creates a ROS2 publisher of `/fmu/in/vehicle_odometry` and a uORB subscriber of `vehicle_odometry`.
```

en/middleware/uxrce_dds.md
```diff
@@ -447,6 +455,28 @@ Each (`topic`,`type`) pairs defines:
...
+- If other (internal) PX4 modules are already publishing on the same uORB topic, its subscribers will receive all streams of messages.
+The uORB subscriber will not be able to determine if an incoming message was published by PX4 or by ROS2.
```
hamishwillee, line 467:
```suggestion
- If other (internal) PX4 modules are already publishing on the same uORB topic instance as the ROS2 publisher, the instance's subscribers will receive all streams of messages.
  The uORB subscriber will not be able to determine if an incoming message was published by PX4 or by ROS2.
```

en/middleware/uxrce_dds.md
```diff
@@ -447,6 +455,28 @@ Each (`topic`,`type`) pairs defines:
...
+- This is the desired behavior when the specified uORB topic has no other publishers, or when you want to replace an internal publisher on this topic (for example during offboard control).
+
+When you add a topic to the  `subscriptions_multi` section the uXRCE-DDS module:
```
hamishwillee, line 470:
```suggestion
When you add a topic to the `subscriptions_multi` section the uXRCE-DDS module:
```

en/middleware/uxrce_dds.md
```diff
@@ -447,6 +455,28 @@ Each (`topic`,`type`) pairs defines:
...
+
+When you add a topic to the `subscriptions_multi` section the uXRCE-DDS module:
+- Queries the uORB manager and get the number `n` of instances currently in use for the desired topic.
+- Creates a unidirectional route going from the ROS2 topic to the a _new_ instance (instance `n+1`) of the associated uORB topic.
+For example, if `vehicle_odometry` has already `2` instances, it creates a ROS2 subscriber of `/fmu/in/vehicle_odometry` and a uORB publisher on instance `3` of `vehicle_odometry`.
+- This ensures that no other internal PX4 module will publish on the same instance used by uXRCE-DDS.
+The subscribers will be able to subscribe to the desired instance and distinguish between publishers.
+- Note, however, that this guarantees separation between PX4 and ROS2 publishers, not among multiple ROS2 publishers.
+In that scenario, their messages will still be routed to the same instance.
+- This is the desired behavior, for example, when you want PX4 to log the readings of two equal sensors; they will both publish on the same topic, but one will use instance 0 and the other will use instance 1.
```
hamishwillee, line 478:
```suggestion
- Queries the uORB manager and get the number `n` of instances currently in use for the desired topic.
- Creates a unidirectional route going from the ROS2 topic to the a _new_ instance (instance `n+1`) of the associated uORB topic.
  For example, if `vehicle_odometry` has already `2` instances, it creates a ROS2 subscriber of `/fmu/in/vehicle_odometry` and a uORB publisher on instance `3` of `vehicle_odometry`.
- This ensures that no other internal PX4 module will publish on the same instance used by uXRCE-DDS.
  The subscribers will be able to subscribe to the desired instance and distinguish between publishers.
- Note, however, that this guarantees separation between PX4 and ROS2 publishers, not among multiple ROS2 publishers.
  In that scenario, their messages will still be routed to the same instance.
- This is the desired behavior, for example, when you want PX4 to log the readings of two equal sensors; they will both publish on the same topic, but one will use instance 0 and the other will use instance 1.
```

en/middleware/uxrce_dds.md
```diff
@@ -447,6 +455,28 @@ Each (`topic`,`type`) pairs defines:
...
+- Creates a unidirectional route going from the ROS2 topic to the _default_ instance (instance 0) of the associated uORB topic.
+For example, it creates a ROS2 subscriber of `/fmu/in/vehicle_odometry` and a uORB publisher of `vehicle_odometry`.
+- If other (internal) PX4 modules are already publishing on the same uORB topic, its subscribers will receive all streams of messages.
+The uORB subscriber will not be able to determine if an incoming message was published by PX4 or by ROS2.
+- This is the desired behavior when the specified uORB topic has no other publishers, or when you want to replace an internal publisher on this topic (for example during offboard control).
+
+When you add a topic to the `subscriptions_multi` section the uXRCE-DDS module:
+- Queries the uORB manager and get the number `n` of instances currently in use for the desired topic.
+- Creates a unidirectional route going from the ROS2 topic to the a _new_ instance (instance `n+1`) of the associated uORB topic.
+  For example, if `vehicle_odometry` has already `2` instances, it creates a ROS2 subscriber of `/fmu/in/vehicle_odometry` and a uORB publisher on instance `3` of `vehicle_odometry`.
```
hamishwillee, line 473:
```suggestion
- Creates a unidirectional route going from the ROS2 topic to a _new_ instance of the associated uORB topic.
  For example, if `vehicle_odometry` has already `2` instances, it creates a ROS2 subscriber of `/fmu/in/vehicle_odometry` and a uORB publisher on instance `3` of `vehicle_odometry`.
```

en/middleware/uxrce_dds.md
beniaminopozzan: Hi @hamishwillee , no I did't.
The message flow goes from ROS2 to PX4, therefore uXRCE-DDS creates a subscriber on the ROS2 side and a publisher on the PX4 side.
The pipeline is
user created ROS2 node publishes topic -> uXRCE-DDS  agent subscribes to that topics -> agent/client communication -> uXRCE-DDS client advertise/publish that topic on PX4 -> PX4 modules subscribe to the topic and use the message coming from ROS2

beniaminopozzan (commented):
Thanks @hamishwillee !

en/middleware/uxrce_dds.md
hamishwillee: Duh. Sorry.

en/releases/main.md
```diff
@@ -39,6 +39,12 @@ TBD ...
 - [Gazebo] Support for [Advanced Plane](../sim_gazebo_gz/vehicles.md#advanced-plane), a simulated fixed-wing vehicle that provides better aerodynamic simulation than the regular plane.
   Added to PX4 in [PX4-Autopilot#22167](https://github.com/PX4/PX4-Autopilot/pull/22167) and [gz-sim#2185](https://github.com/gazebosim/gz-sim/pull/2185) (advanced lift drag plugin).
 
+### uXRCE-DDS / ROS2
+
+- [uXRCE-DDS] [Topics selector](../middleware/uxrce_dds.md#dds-topics-yaml) supports multi-instance ROS2 subscribers.
+  ROS2 users can now decide if the messages that they are sending to PX4 will overlap with the existing uORB ones or if the want to keep them in separate instances.
+  Added in PX4 in [PX4-Autopilot#22266](https://github.com/PX4/PX4-Autopilot/pull/22266).
```
hamishwillee, line 46:
```suggestion
- [uXRCE-DDS] [DDS Topics YAML](../middleware/uxrce_dds.md#dds-topics-yaml) allows use of `subscription_multi` to specify that ROS 2 messages that they are sending to PX4 will overlap with the existing uORB ones or if the want to keep them in separate instances.
  Added in PX4 in [PX4-Autopilot#22266](https://github.com/PX4/PX4-Autopilot/pull/22266).
```

```suggestion
- [uXRCE-DDS] [DDS Topics YAML](../middleware/uxrce_dds.md#dds-topics-yaml) now allows the use of `subscription_multi` to specify that indicated ROS 2 topics are sent to a separate uORB topic instance reserved for ROS 2.
  This allows PX4 to differentiate between updates from ROS and those from PX4 uORB publishers.
  With this change ROS2 users can now decide if the messages that they are sending to PX4 will overlap with the existing uORB ones or be kept in separate instances.
  Added in PX4 in [PX4-Autopilot#22266](https://github.com/PX4/PX4-Autopilot/pull/22266).
```

en/middleware/uxrce_dds.md
hamishwillee, line 448:
```suggestion
1. A new `publication`, `subscription`, or `subscriptions_multi`, depending on the list to which it is added.
```

en/middleware/uxrce_dds.md
```diff
@@ -447,6 +455,27 @@ Each (`topic`,`type`) pairs defines:
    - `/fmu/in/` for topics that are _subscribed_ by PX4.
 4. The message type (`VehicleOdometry`, `VehicleStatus`, `OffboardControlMode`, etc.) and the ROS 2 package (`px4_msgs`) that is expected to provide the message definition.
 
+:::note
+While in general ROS2 does not allow subscribers to distinguish between multiple nodes publishing on the same topic, PX4 does implement this feature allowing uORB topics to have [multiple instances](uorb.md#multi-instance).
```
hamishwillee, line 459: Generally I fully qualify links.

```suggestion
While in general ROS2 does not allow subscribers to distinguish between multiple nodes publishing on the same topic, PX4 does implement this feature allowing uORB topics to have [multiple instances](../middleware/uorb.md#multi-instance).
```

en/middleware/uxrce_dds.md
```diff
@@ -447,6 +455,27 @@ Each (`topic`,`type`) pairs defines:
    - `/fmu/in/` for topics that are _subscribed_ by PX4.
 4. The message type (`VehicleOdometry`, `VehicleStatus`, `OffboardControlMode`, etc.) and the ROS 2 package (`px4_msgs`) that is expected to provide the message definition.
 
+:::note
+While in general ROS2 does not allow subscribers to distinguish between multiple nodes publishing on the same topic, PX4 does implement this feature allowing uORB topics to have [multiple instances](../middleware/uorb.md#multi-instance).
+For this reason, we distinguish between `subscriptions` and `subscriptions_multi`.
```
hamishwillee, line 460:
```suggestion
`subscriptions` and `subscriptions_multi` allows us to choose the uORB instance that ROS 2 topics are routed to: either a shared instance that may also be getting updates from PX4 published, or a separate instance that is reserved for ROS2 publications, respectively.
```

en/middleware/uxrce_dds.md
```diff
@@ -447,6 +455,27 @@ Each (`topic`,`type`) pairs defines:
...
+
+When you add a topic to the `subscriptions_multi` section the uXRCE-DDS module:
+- Creates a unidirectional route going from the ROS2 topic to a _new_ instance of the associated uORB topic.
+  For example, if `vehicle_odometry` has already `2` instances, it creates a ROS2 subscriber of `/fmu/in/vehicle_odometry` and a uORB publisher on instance `3` of `vehicle_odometry`.
+- This ensures that no other internal PX4 module will publish on the same instance used by uXRCE-DDS.
+  The subscribers will be able to subscribe to the desired instance and distinguish between publishers.
+- Note, however, that this guarantees separation between PX4 and ROS2 publishers, not among multiple ROS2 publishers.
+  In that scenario, their messages will still be routed to the same instance.
+- This is the desired behavior, for example, when you want PX4 to log the readings of two equal sensors; they will both publish on the same topic, but one will use instance 0 and the other will use instance 1.
```
hamishwillee, line 477:
```suggestion
Add a topic to the  `subscriptions` section the uXRCE-DDS module to:
- Create a unidirectional route going from the ROS2 topic to the _default_ instance (instance 0) of the associated uORB topic.
  For example, it creates a ROS2 subscriber of `/fmu/in/vehicle_odometry` and a uORB publisher of `vehicle_odometry`.
- If other (internal) PX4 modules are already publishing on the same uORB topic instance as the ROS2 publisher, the instance's subscribers will receive all streams of messages.
  The uORB subscriber will not be able to determine if an incoming message was published by PX4 or by ROS2.
- This is the desired behavior when the ROS2 publisher is expected to be the sole publisher on the topic instance (for example, replacing an internal publisher to the topic during offboard control), or when the source of multiple publishing streams does not matter.

Add a topic to the `subscriptions_multi` section the uXRCE-DDS module to:
- Creates a unidirectional route going from the ROS2 topic to a _new_ instance of the associated uORB topic.
  For example, if `vehicle_odometry` has already `2` instances, it creates a ROS2 subscriber of `/fmu/in/vehicle_odometry` and a uORB publisher on instance `3` of `vehicle_odometry`.
- This ensures that no other internal PX4 module will publish on the same instance used by uXRCE-DDS.
  The subscribers will be able to subscribe to the desired instance and distinguish between publishers.
- Note, however, that this guarantees separation between PX4 and ROS2 publishers, not among multiple ROS2 publishers.
  In that scenario, their messages will still be routed to the same instance.
- This is the desired behavior, for example, when you want PX4 to log the readings of two equal sensors; they will both publish on the same topic, but one will use instance 0 and the other will use instance 1.
```

en/middleware/uxrce_dds.md
```diff
@@ -447,6 +455,29 @@ Each (`topic`,`type`) pairs defines:
    - `/fmu/in/` for topics that are _subscribed_ by PX4.
 4. The message type (`VehicleOdometry`, `VehicleStatus`, `OffboardControlMode`, etc.) and the ROS 2 package (`px4_msgs`) that is expected to provide the message definition.
 
+ROS2 does not allow subscribers to distinguish between multiple nodes publishing on the same topic, while PX4 can, using [multiple uORB topic instances](../middleware/uorb.md#multi-instance).
+
+`subscriptions` and `subscriptions_multi` allow us to choose the uORB topic instance that ROS 2 topics are routed to: either a shared instance that may also be getting updates from internal PX4 uORB publishers, or a separate instance that is reserved for ROS2 publications, respectively.
```
hamishwillee, line 460:
```suggestion
`subscriptions` and `subscriptions_multi` allow us to choose the uORB topic instance that ROS 2 topics are routed to: either a shared instance that may also be getting updates from internal PX4 uORB publishers, or a separate instance that is reserved for ROS2 publications, respectively.
```

en/middleware/uxrce_dds.md
```diff
@@ -447,6 +455,29 @@ Each (`topic`,`type`) pairs defines:
...
+Without this mechanism all ROS 2 messages would be routed to the _same_ uORB topic instance (because ROS 2 does not have the concept of [multiple topic instances](../middleware/uorb.md#multi-instance)), and it would not be possible PX4 subscribers to differentiate between streams from ROS 2 or PX4 publishers.
+
+Add a topic to the `subscriptions` section the uXRCE-DDS module to:
```
GuillaumeLaine, line 463: Did we mean:
```suggestion
Add a topic to the `subscriptions` section of the uXRCE-DDS module to:
```

GuillaumeLaine:
@hamishwillee @beniaminopozzan I was a bit out of the loop the last days, thanks for revising this further! I'm giving this another read now.

en/middleware/uxrce_dds.md
```diff
@@ -447,6 +455,29 @@ Each (`topic`,`type`) pairs defines:
    - `/fmu/in/` for topics that are _subscribed_ by PX4.
 4. The message type (`VehicleOdometry`, `VehicleStatus`, `OffboardControlMode`, etc.) and the ROS 2 package (`px4_msgs`) that is expected to provide the message definition.
 
+ROS2 does not allow subscribers to distinguish between multiple nodes publishing on the same topic, while PX4 can, using [multiple uORB topic instances](../middleware/uorb.md#multi-instance).
+
+`subscriptions` and `subscriptions_multi` allow us to choose the uORB topic instance that ROS 2 topics are routed to: either a shared instance that may also be getting updates from internal PX4 uORB publishers, or a separate instance that is reserved for ROS2 publications, respectively.
+Without this mechanism all ROS 2 messages would be routed to the _same_ uORB topic instance (because ROS 2 does not have the concept of [multiple topic instances](../middleware/uorb.md#multi-instance)), and it would not be possible PX4 subscribers to differentiate between streams from ROS 2 or PX4 publishers.
```
GuillaumeLaine, line 461:
```suggestion
Without this mechanism all ROS 2 messages would be routed to the _same_ uORB topic instance (because ROS 2 does not have the concept of [multiple topic instances](../middleware/uorb.md#multi-instance)), and it would not be possible for PX4 subscribers to differentiate between streams from ROS 2 or PX4 publishers.
```

en/middleware/uxrce_dds.md
hamishwillee: No, I mean't remove everything after "section". But had a power failure mid edit. Won't be able to look at this until tomorrow.

hamishwillee: No, I mean't remove everything after "section". But had a power failure mid edit. Won't be able to look at this until tomorrow.

hamishwillee:
```suggestion
Add a topic to the `subscriptions` section to:
```

en/middleware/uxrce_dds.md
```diff
@@ -447,6 +455,27 @@ Each (`topic`,`type`) pairs defines:
...
+
+- Create a unidirectional route going from the ROS2 topic to the _default_ instance (instance 0) of the associated uORB topic.
+  For example, it creates a ROS2 subscriber of `/fmu/in/vehicle_odometry` and a uORB publisher of `vehicle_odometry`.
+- If other (internal) PX4 modules are already publishing on the same uORB topic instance as the ROS2 publisher, the instance's subscribers will receive all streams of messages.
+  The uORB subscriber will not be able to determine if an incoming message was published by PX4 or by ROS2.
+- This is the desired behavior when the ROS2 publisher is expected to be the sole publisher on the topic instance (for example, replacing an internal publisher to the topic during offboard control), or when the source of multiple publishing streams does not matter.
+
+Add a topic to the `subscriptions_multi` section the uXRCE-DDS module to:
+
+- Creates a unidirectional route going from the ROS2 topic to a _new_ instance of the associated uORB topic.
```
hamishwillee, line 471:
```suggestion
Add a topic to the `subscriptions_multi` section to:

- Create a unidirectional route going from the ROS2 topic to a _new_ instance of the associated uORB topic.
```

hamishwillee (commented):
@GuillaumeLaine @beniaminopozzan OK, thanks for your edits. I left it a bit half-way before due to power failure. I think this is pretty clear now. 

I'll merge tomorrow if you are both happy. Or possibly anyway, since it may not be perfect, but I think it is OK

GuillaumeLaine:
Thanks @hamishwillee , LGTM too!

hamishwillee:
Cheers!

PR #2872, opened 2023-11-13
Local vs global position requirements

sfuhrer:
This was raised in https://discuss.px4.io/t/do-really-all-autonomous-modes-require-global-position/34890/1

### DIFF
--- a/en/flight_modes/mission.md
+++ b/en/flight_modes/mission.md
@@ -12,6 +12,7 @@ The mission is typically created and uploaded with a Ground Control Station (GCS
   - Flying vehicles can't switch to this mode without global position.
   - Flying vehicles will failsafe if they lose the position estimate.
   - Disarmed vehicles can switch to mode without valid position estimate but can't arm.
+- Mode requires wind and flight time are within specified limits.
 - RC control switches can be used to change flight modes on any vehicle.
 - RC stick movement in a multicopter (or VTOL in multicopter mode) will [by default](#COM_RC_OVERRIDE) change the vehicle to [Position mode](../flight_modes_mc/position.md) unless handling a critical battery failsafe.
 
--- a/en/flight_modes/mission.md
+++ b/en/flight_modes/mission.md
@@ -12,7 +12,7 @@ The mission is typically created and uploaded with a Ground Control Station (GCS
   - Flying vehicles can't switch to this mode without global position.
   - Flying vehicles will failsafe if they lose the position estimate.
   - Disarmed vehicles can switch to mode without valid position estimate but can't arm.
-- Mode requires wind and flight time are within specified limits.
+- Mode requires wind and flight time are within allowed limits (specified via parameters).
 - RC control switches can be used to change flight modes on any vehicle.
 - RC stick movement in a multicopter (or VTOL in multicopter mode) will [by default](#COM_RC_OVERRIDE) change the vehicle to [Position mode](../flight_modes_mc/position.md) unless handling a critical battery failsafe.
 
--- a/en/flight_modes/mission.md
+++ b/en/flight_modes/mission.md
@@ -7,11 +7,15 @@ The mission is typically created and uploaded with a Ground Control Station (GCS
 
 :::note
 
-- This mode requires a global 3d position estimate (from GPS or inferred from a [local position](../ros/external_position_estimation.md#enabling-auto-modes-with-a-local-position)).
-- The vehicle must be armed before this mode can be engaged.
-- This mode is automatic - no user intervention is _required_ to control the vehicle.
+- Mode is automatic - no user intervention is _required_ to control the vehicle.
+- Mode requires a global 3d position estimate (from GPS or inferred from a [local position](../ros/external_position_estimation.md#enabling-auto-modes-with-a-local-position)).
+  - Flying vehicles can't switch to this mode without global position.
+  - Flying vehicles will failsafe if they lose the position estimate.
+  - Disarmed vehicles can switch to mode without valid position estimate but can't arm.
 - RC control switches can be used to change flight modes on any vehicle.
 - RC stick movement in a multicopter (or VTOL in multicopter mode) will [by default](#COM_RC_OVERRIDE) change the vehicle to [Position mode](../flight_modes_mc/position.md) unless handling a critical battery failsafe.
+
+<!-- https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/commander/ModeUtil/mode_requirements.cpp -->
 
 :::
 
--- a/en/flight_modes_fw/hold.md
+++ b/en/flight_modes_fw/hold.md
@@ -11,10 +11,16 @@ It is usually activated with a pre-programmed switch.
 
 :::note
 
-- This mode requires GPS.
-- This mode is automatic - no user intervention is _required_ to control the vehicle.
+- Mode is automatic - no user intervention is _required_ to control the vehicle.
+- Mode requires a global 3d position estimate (from GPS or inferred from a [local position](../ros/external_position_estimation.md#enabling-auto-modes-with-a-local-position)).
+  - Flying vehicles can't switch to this mode without global position.
+  - Flying vehicles will failsafe if they lose the position estimate.
+  - Disarmed vehicles can switch to mode without valid position estimate but can't arm.
+- Mode requires wind and flight time are within specified limits.
 - RC control switches can be used to change flight modes on any vehicle.
 - RC stick movement is ignored.
+
+<!-- https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/commander/ModeUtil/mode_requirements.cpp -->
 
 :::
 
--- a/en/flight_modes_fw/hold.md
+++ b/en/flight_modes_fw/hold.md
@@ -16,7 +16,7 @@ It is usually activated with a pre-programmed switch.
   - Flying vehicles can't switch to this mode without global position.
   - Flying vehicles will failsafe if they lose the position estimate.
   - Disarmed vehicles can switch to mode without valid position estimate but can't arm.
-- Mode requires wind and flight time are within specified limits.
+- Mode requires wind and flight time are within allowed limits (specified via parameters).
 - RC control switches can be used to change flight modes on any vehicle.
 - RC stick movement is ignored.
 
--- a/en/flight_modes_fw/land.md
+++ b/en/flight_modes_fw/land.md
@@ -14,7 +14,7 @@ Where possible, instead use [Return mode](../flight_modes/return.md) with a pred
 
 :::note
 
-- This mode requires a valid global position estimate (from GPS or inferred from a [local position](../ros/external_position_estimation.md#enabling-auto-modes-with-a-local-position)).
+- This mode requires a valid local position estimate.
 - This mode is automatic - no user intervention is _required_ to control the vehicle.
 - RC control switches can be used to change flight modes on any vehicle.
 - RC stick movement is ignored.
--- a/en/flight_modes_fw/land.md
+++ b/en/flight_modes_fw/land.md
@@ -14,12 +14,16 @@ Where possible, instead use [Return mode](../flight_modes/return.md) with a pred
 
 :::note
 
-- This mode requires a valid local position estimate.
-- This mode is automatic - no user intervention is _required_ to control the vehicle.
+- Mode is automatic - no user intervention is _required_ to control the vehicle.
+- Mode requires at least a valid local position estimate (does not require a global position).
+  - Flying vehicles can't switch to this mode without valid local position.
+  - Flying vehicles will failsafe if they lose the position estimate.
+- Mode prevents arming (vehicle must be armed when switching to this mode).
 - RC control switches can be used to change flight modes on any vehicle.
 - RC stick movement is ignored.
 - The mode can be triggered using the [MAV_CMD_NAV_LAND](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_LAND) MAVLink command, or by explicitly switching to Land mode.
 
+<!-- https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/commander/ModeUtil/mode_requirements.cpp -->
 :::
 
 ## Technical Summary
--- a/en/flight_modes_fw/takeoff.md
+++ b/en/flight_modes_fw/takeoff.md
@@ -8,12 +8,16 @@ Vehicles are [hand or catapult launched](#catapult-hand-launch) by default, but 
 
 :::note
 
-- This mode requires at least a valid local position estimate (does not require a global position).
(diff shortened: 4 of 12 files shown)


en/flight_modes_fw/land.md
```diff
@@ -14,7 +14,7 @@ Where possible, instead use [Return mode](../flight_modes/return.md) with a pred
 
 :::note
 
-- This mode requires a valid global position estimate (from GPS or inferred from a [local position](../ros/external_position_estimation.md#enabling-auto-modes-with-a-local-position)).
+- This mode requires a valid local position estimate.
```
hamishwillee, line 17: @sfuhrer As per https://discuss.px4.io/t/do-really-all-autonomous-modes-require-global-position/34890/20?u=hamishwillee I suspect land requires a position estimate that can be either local or global.

As it reads now, you could have a global position and as a reader I would not think it would work. My assumption here is that it requires a valid position estimate, either local or global. 

So what modes do you really require a global position for? Mission and return I presume because that specifies locations in terms of GPS. But what about orbit mode? That requires a GPS position, so how can it work with only a local position?

Lastly, consider https://docs.px4.io/main/en/config/safety.html#position-gps-loss-failsafe - I believe this to be correct - it reflects position as either local or global and that is fine. Right?

sfuhrer: >be either local or global.

There is no global position without local position, so the text could read "requires at least local position estimate"

[Here](https://github.com/PX4/PX4-Autopilot/blob/44d1003f8e4d1c1ed1d8c7a3d6f8f623b43a8fe4/src/modules/commander/ModeUtil/mode_requirements.cpp#L137-L142) is where the mode requirements are set in PX4.
Those modes do only require local position:
NAVIGATION_STATE_AUTO_TAKEOFF
NAVIGATION_STATE_AUTO_LAND
NAVIGATION_STATE_AUTO_FOLLOW_TARGET
NAVIGATION_STATE_AUTO_PRECLAND
NAVIGATION_STATE_ORBIT
NAVIGATION_STATE_AUTO_VTOL_TAKEOFF

And for those you need global:
NAVIGATION_STATE_AUTO_MISSION
NAVIGATION_STATE_AUTO_LOITER
NAVIGATION_STATE_AUTO_RTL

>Lastly, consider https://docs.px4.io/main/en/config/safety.html#position-gps-loss-failsafe - I believe this to be correct - it reflects position as either local or global and that is fine. Right?

Yes, that's fine, as in the failsafe case you usually lose local and global position estimate at the same time. Local vs global mode requirements only really matter for mode acceptance (which modes you are able to fly indoors for example).

hamishwillee:
@sfuhrer Very helpful.  I'll do some fixes but can you give me a precise understanding of what these requirements mean in terms of behaviour. Specifically. 

@sfuhrer From [mode_requirements](https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/commander/ModeUtil/mode_requirements.cpp):

```
	// NAVIGATION_STATE_AUTO_TAKEOFF
	setRequirement(vehicle_status_s::NAVIGATION_STATE_AUTO_TAKEOFF, flags.mode_req_angular_velocity);
	setRequirement(vehicle_status_s::NAVIGATION_STATE_AUTO_TAKEOFF, flags.mode_req_attitude);
	setRequirement(vehicle_status_s::NAVIGATION_STATE_AUTO_TAKEOFF, flags.mode_req_local_position);
	setRequirement(vehicle_status_s::NAVIGATION_STATE_AUTO_TAKEOFF, flags.mode_req_local_alt);
```

This doesn't have `flags.mode_req_prevent_arming`
- I assume flag means "all other reqs must be met to arm"?
- Since the flag is not present, presumably you can ARM even if those requirements are not met. Does that mean:
  - you can switch to this mode even if the requirements are not met and then takeoff, but then you will failsafe, or
  - You can't switch to the mode.
  - You might switch to the mode if the requirements are met, but once you're in the mode you can't arm. Would you failsafe on the ground in this case?
  - In other words "other than arming, what does not meeting a requirement affect w.r.t. failsafes and modes"?

For example, what does `flags.mode_req_wind_and_flight_time_compliance` prevent if not met?

What is `mode_req_local_position_relaxed` ? Specifically, for position control it looks like you need global position for fixed wing, but this relaxed condition for MC?

Imagine 

So what does it mean if you don't have a local position? Can you switch to this mode? i.e. it could be that 


You do require at least a local position/alt. What does that requirement mean? Given `flags.mode_req_prevent_arming` is not present my assumption is that you can switch to this mode

en/flight_modes_fw/takeoff.md
```diff
@@ -8,8 +8,8 @@ Vehicles are [hand or catapult launched](#catapult-hand-launch) by default, but
 
 :::note
 
-- This mode requires a good position estimate (e.g. from GPS).
-- The vehicle must be armed before this mode will activate.
+- This mode requires at least a valid local position estimate (does not require a global position).
+  <!-- https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/commander/ModeUtil/mode_requirements.cpp -->
```
hamishwillee, line 12: > - This mode requires at least a valid local position estimate (does not require a global position).

I want to make sure that it is obvious that "needing at least a valid local position estimate" means "you can fly indoors with a valid local position". I think " (does not require a global position)."  probably captures that,right?

hamishwillee:
@sfuhrer OK, I have updated the set of automatic modes you listed except for precland and VTOL_TAKEOFF, which do not have flight mode docs.

1. Can you please review?

2. Do you think I should create a doc for vtol takeoff? I.e. is there enough special to say about it - minimally we probably want to talk about how you invoke it from MAVLink and any differences from normal MC takeoff.

3. What about [precland](https://docs.px4.io/main/en/advanced_features/precland.html)? It's a bit odd in that you can't invoke it as a mode in a normal way - it's invoked via return settings or a mission. On the other hand it might be useful to also have a mode doc for Descend, that could say nothing more than it exists, it has this high level characteristic, and see X for more info. Thoughts?

bkueng:
> Very helpful. I'll done some fixes on the assumption I understand the [requirements](https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/commander/ModeUtil/mode_requirements.cpp), such as

@hamishwillee you are correct, it's the same as https://github.com/PX4/PX4-user_guide/pull/2740/files#diff-9ebcb7aa2dd9d514f99d8f5a4a5ed34cc6197777853ebfe43098e72f1358e200R270 (we can move it and link from there).

> Does flags.mode_req_prevent_arming? mean you can't arm from the mode? You have to arm in another mode THEN switch to the mode to use it.

Yes

> What is mode_req_local_position_relaxed ? Specifically, for position control it looks like you need global position for fixed wing, but this relaxed condition for MC? I have not updated manual modes.

relaxed is for manual modes with less strict requirements (e.g. for optical flow where you can manually correct some inaccuracy).

> What does flags.mode_req_wind_and_flight_time_compliance check?

This is a regulatory-thing: when flying in high wind or exceeding a certain flight time, you cannot switch into certain modes anymore. Both thresholds are set by parameters (disabled by default if I recall correctly).

en/flight_modes_fw/takeoff.md
sfuhrer: Requires local position, does not require global position, yes.

sfuhrer:
Thanks Beat, hope that answers most of your questions @hamishwillee .

>  create a doc for vtol takeoff?

That would be nice indeed.

> What about [precland](https://docs.px4.io/main/en/advanced_features/precland.html)?

I don't know more than you about that. Never touched/used it.

en/flight_modes/mission.md
```diff
@@ -7,12 +7,17 @@ The mission is typically created and uploaded with a Ground Control Station (GCS
...
 
-- This mode requires a global 3d position estimate (from GPS or inferred from a [local position](../ros/external_position_estimation.md#enabling-auto-modes-with-a-local-position)).
-- The vehicle must be armed before this mode can be engaged.
-- This mode is automatic - no user intervention is _required_ to control the vehicle.
+- Mode is automatic - no user intervention is _required_ to control the vehicle.
+- Mode requires a global 3d position estimate (from GPS or inferred from a [local position](../ros/external_position_estimation.md#enabling-auto-modes-with-a-local-position)).
+  - Flying vehicles can't switch to this mode without global position.
+  - Flying vehicles will failsafe if they lose the position estimate.
+  - Disarmed vehicles can switch to mode without valid position estimate but can't arm.
+- Mode requires wind and flight time are within specified limits.
```
hamishwillee, line 15:
```suggestion
- Mode requires wind and flight time are within allowed limits (specified via parameters).
```

en/flight_modes_fw/hold.md
```diff
@@ -11,11 +11,17 @@ It is usually activated with a pre-programmed switch.
...
 :::note
 
-- This mode requires GPS.
-- This mode is automatic - no user intervention is _required_ to control the vehicle.
+- Mode is automatic - no user intervention is _required_ to control the vehicle.
+- Mode requires a global 3d position estimate (from GPS or inferred from a [local position](../ros/external_position_estimation.md#enabling-auto-modes-with-a-local-position)).
+  - Flying vehicles can't switch to this mode without global position.
+  - Flying vehicles will failsafe if they lose the position estimate.
+  - Disarmed vehicles can switch to mode without valid position estimate but can't arm.
+- Mode requires wind and flight time are within specified limits.
```
hamishwillee, line 19:
```suggestion
- Mode requires wind and flight time are within allowed limits (specified via parameters).
```

en/flight_modes_mc/follow_me.md
```diff
@@ -4,6 +4,24 @@
...
 _Follow Me_ mode allows a multicopter to autonomously hold position and altitude relative to another system that is broadcasting its position (and optionally velocity) using the [FOLLOW_TARGET](https://mavlink.io/en/messages/common.html#FOLLOW_TARGET) MAVLink message.
 
+:::note
+
+- Mode is automatic - no user intervention is _required_ to control the vehicle.
+- Mode requires at least a valid local position estimate (does not require a global position).
+  - Flying vehicles can't switch to this mode without valid local position.
+  - Flying vehicles will failsafe if they lose the position estimate.
+- Mode prevents arming (vehicle must be armed when switching to this mode).
+- Mode requires wind and flight time are within specified limits.
```
hamishwillee, line 14:
```suggestion
- Mode requires wind and flight time are within allowed limits (specified via parameters).
```

en/flight_modes_mc/hold.md
```diff
@@ -11,11 +11,17 @@ It is usually activated with a pre-programmed switch.
...
 :::note
 
-- This mode requires GPS.
-- This mode is automatic - no user intervention is _required_ to control the vehicle.
+- Mode is automatic - no user intervention is _required_ to control the vehicle.
+- Mode requires a global 3d position estimate (from GPS or inferred from a [local position](../ros/external_position_estimation.md#enabling-auto-modes-with-a-local-position)).
+  - Flying vehicles can't switch to this mode without global position.
+  - Flying vehicles will failsafe if they lose the position estimate.
+  - Disarmed vehicles can switch to mode without valid position estimate but can't arm.
+- Mode requires wind and flight time are within specified limits.
```
hamishwillee, line 19:
```suggestion
- Mode requires wind and flight time are within allowed limits (specified via parameters).
```

en/flight_modes_mc/orbit.md
```diff
@@ -6,11 +6,19 @@ The _Orbit_ guided flight mode allows you to command a multicopter (or VTOL in m
...
 :::note
 
-- This mode requires a valid global position estimate (from GPS or inferred from a [local position](../ros/external_position_estimation.md#enabling-auto-modes-with-a-local-position)).
-- This mode is automatic - no user intervention is _required_ to control the vehicle.
+- Mode is automatic - no user intervention is _required_ to control the vehicle.
+- Mode requires at least a valid local position estimate (does not require a global position).
+  - Flying vehicles can't switch to this mode without valid local position.
+  - Flying vehicles will failsafe if they lose the position estimate.
+- Mode prevents arming (vehicle must be armed when switching to this mode).
+- Mode requires wind and flight time are within specified limits.
```
hamishwillee, line 14:
```suggestion
- Mode requires wind and flight time are within allowed limits (specified via parameters).
```

hamishwillee:
@sfuhrer @bkueng Thanks guys. Since you looked at this last I updated the compliance check info to note that this is set by parameters. We should update to list those parameters "at some point". Otherwise this is accurate for now as far as I can tell and I'm going to merge it. 

Wr.t. VTOL takeoff mode, @sfuhrer  Can you outline how this works? I will add as a post process.

hamishwillee:
> you are correct, it's the same as https://github.com/PX4/PX4-user_guide/pull/2740/files#diff-9ebcb7aa2dd9d514f99d8f5a4a5ed34cc6197777853ebfe43098e72f1358e200R270 (we can move it and link from there).

@bkueng I like that, particularly the diagram. 

Short term, what has been done in terms of the notes (as duplicated below) works for me because it is clear and human readable text about what the requirements are. The Mode requires a global ... and the following sub bullets ... capture the behaviour - "requirements must be met to arm when disarmed or to switch to a mode when flying. If not met when flying you will failsafe. You can switch to a mode without the requirements being met but not arm"

> - Mode is automatic - no user intervention is _required_ to control the vehicle.
> - Mode requires a global 3d position estimate (from GPS or inferred from a [local position](../ros/external_position_estimation.md#enabling-auto-modes-with-a-local-position)).
>   - Flying vehicles can't switch to this mode without global position.
>   - Flying vehicles will failsafe if they lose the position estimate.
>   - Disarmed vehicles can switch to mode without valid position estimate but can't arm.
> - Mode requires wind and flight time are within allowed limits (specified via parameters).
> - RC control switches can be used to change flight modes on any vehicle.
> - RC stick movement in a multicopter (or VTOL in multicopter mode) will [by default](#COM_RC_OVERRIDE) change the vehicle to [Position mode](../flight_modes_mc/position.md) unless handling a critical battery failsafe.
```

What I don't like about the approach:
- There are more requirements than just "local/global" position. Those have to be caught separately - e.g. the wind limits, arming requirement. 
- This isn't particularly scalable - you couldn't auto-extract this for example.
- Note though, you couldn't anyway - it would be nice if there was a requirement or flag to indicate "RC control switches can be used to change flight modes on any vehicle." and the one after.
- Someone coming along to this can work out what the requirements are quite easily, but they don't know anything about how this happens in code. 

So longer term I think it would be good to think about:
- A new page PX4 Mode Requirements that covers how this works and what the requirements are - that you can link to in ROS. The ROS docs could then just list the specific requirements associated with the ROS2 manual flag (say).
- Then I update the box above to list the requirements and link to the explanations of what a requirement means. This would be slightly less readable first time you check a mode, but would become very familiar quite quickly.

Thoughts?

(note, this is merged, but happy to continue discussion here).

bkueng:
Sounds good. The mode requirements also appear in the failsafe state machine simulation: https://docs.px4.io/main/en/config/safety_simulation.html

PR #2873, opened 2023-11-13
Land (FW): this mode only works with valid local position, and we shouldn't mix it with failsafe modes

sfuhrer:


hamishwillee:
Fair enough!

PR #2875, opened 2023-11-13
Update ark_flow.md

afwilkin:
Some parameters on this  list are not available until you 
 Enable UAVCAN_SUB_FLOW.
Enable UAVCAN_SUB_RNG.

I wouldl like to move them up the list so no one is confused on why these other parameters arent there


afwilkin:
@AlexKlimaj

hamishwillee (approved):
Thank you. 

This is good - similar approach to what I used in https://docs.px4.io/main/en/dronecan/#optical-flow-sensor - though I might update that to be a little more explicit and state that some parameters aren't present if these are not enabled.

PR #2888, opened 2023-11-15
Update pixhawk6x-rt.md

dk7xe:
Support for Pixhawk 6X-RT is in PX4 main since today. Corrected the build target name.


hamishwillee (approved):
Thanks @dk7xe !

hamishwillee:
Is support going to be backported to PX4 v1.14? I.e. should we note that currently only mainline firmware is available via QGC?

hamishwillee:
PS NOrmally px4_fmu-v6xrt_default and px4_fmu-v6xrt are synonyms - i.e. the _default is optional.

PR #2897, opened 2023-11-17
Added documentation for the throw-launch feature.

zeroos:
This is the documentation of the recently added throw-launch feature. 

The PR introducing the feature:
https://github.com/PX4/PX4-Autopilot/pull/21170

@MaEtUgR, as you've helped in the development, can you please have a look and provide the technical review?


hamishwillee:
Thanks @zeroos - this is pretty cool! I've subedited. As part of that I've made some assumptions for which I will add a few notes. Can you confirm those assumptions are correct?

en/SUMMARY.md
```diff
@@ -68,6 +68,9 @@
       * [Omnicopter](frames_multicopter/omnicopter.md)
       * [DJI F450 (CUAV v5+)](frames_multicopter/dji_f450_cuav_5plus.md)
       * [DJI F450 (CUAV v5 nano)](frames_multicopter/dji_f450_cuav_5nano.md)
+    * [Features](features_mc/README.md)
```
hamishwillee, line 71: FYI we're moving towards more vehicle-type centric docs. This is still a work in progress, but for that reason I have moved this, along with precision landing, into a section under multicopters.

For now this is just MC features, but probably will also have to link other interesting features that apply to all platforms.

en/features_mc/throw_launch.md
```diff
@@ -0,0 +1,87 @@
...
+- It has not yet been broadly tested on different vehicle configurations or scenarios. various platforms and in different scenarios.
+- The majority of testing has been done in position mode.
+  Other modes should also work.
+
+:::
+
+This feature allows a multicopter to be started by arming it from a fixed position and then throwing it into the air.
+The vehicle then turns on the motors and operates according to its current mode.
+
+The vehicle detects that it has been thrown based on reaching a certain speed and altitude (5m/s and 2m by default, respectively) and then starts the motors at the apex of the throw, once it determines that it has started to fall.
```
zeroos, line 15: Actually, altitude is not checked, only the speed. I've mentioned that the drone should be thrown "about 2m away and 2m up" just as an example of a throw that results in a speed of 5m/s. Do you think it's not clear that was only an example and should be clarified?

en/features_mc/throw_launch.md
```diff
@@ -0,0 +1,87 @@
...
+Note that unlike in other configurations, when throw launch is enabled the propellers should not spin until the launch is detected ( even when the vehicle is armed).
+
+The vehicle will not arm while it is moving, and should not be used on a moving platform (this is to reduce the chance of accidental triggering if a vehicle is armed and then moved to a launch point).
+As usual the vehicle will disarm if the throw is not detected within a few seconds of arming.
```
hamishwillee, line 19: Please check this. My assumption is that your safety feature is "will not arm while moving".

zeroos, line 19: That's not true, when the throw mode is enabled, automatic disarming does not trigger. It happened once to us that the drone disarmed just before the drone and then we introduced this change.

en/features_mc/throw_launch.md
```diff
@@ -0,0 +1,87 @@
...
+
+:::
+
+This feature allows a multicopter to be started by arming it from a fixed position and then throwing it into the air.
+The vehicle then turns on the motors and operates according to its current mode.
+
+The vehicle detects that it has been thrown based on reaching a certain speed and altitude (5m/s and 2m by default, respectively) and then starts the motors at the apex of the throw, once it determines that it has started to fall.
+Note that unlike in other configurations, when throw launch is enabled the propellers should not spin until the launch is detected ( even when the vehicle is armed).
+
+The vehicle will not arm while it is moving, and should not be used on a moving platform (this is to reduce the chance of accidental triggering if a vehicle is armed and then moved to a launch point).
```
zeroos, line 18: The first part of the sentence is probably true (I did not modify the arming conditions in any way), but in the initial comment about "moving platform" I was trying to highlight something else. I propose to change this sentence in the following way, please let me know if it's more clear for you now:

> The vehicle should not be transported after being armed and before the throw. In particular, the throw should not be executed from a moving platform. The reason behind this is that the condition to start the motors depends on absolute speed of the multicopter and does not account for any additional movement. Trying to throw the drone from a moving platform might result in the motors being started prematurely.

en/SUMMARY.md
zeroos: Nice, I think it's a good change! Before, I've struggled a bit to find where this feature would "fit" in the documentation, now it's clear.

en/features_mc/throw_launch.md
```diff
@@ -0,0 +1,87 @@
...
+- The majority of testing has been done in position mode.
+  Other modes should also work.
+
+:::
+
+This feature allows a multicopter to be started by arming it from a fixed position and then throwing it into the air.
+The vehicle then turns on the motors and operates according to its current mode.
+
+The vehicle detects that it has been thrown based on reaching a certain speed and altitude (5m/s and 2m by default, respectively) and then starts the motors at the apex of the throw, once it determines that it has started to fall.
+Note that unlike in other configurations, when throw launch is enabled the propellers should not spin until the launch is detected ( even when the vehicle is armed).
```
zeroos, line 16: Should we also mention that with throw mode enabled, it repeats the arming sound until the vehicle is thrown (or disarmed)?

zeroos (commented):
Thank you for the changes, I like the new organization!

I left a couple of comments, please let me know if you agree with them and I'll be happy to incorporate the proposed changes.

en/features_mc/throw_launch.md
zeroos: From what I understand, that's the default behavior (not related to throw mode). The problems might occur when the vehicle moves after arming (I've proposed a description of these problems here: https://github.com/PX4/PX4-user_guide/pull/2897#discussion_r1402955353 ). Do you think we should also mention that the vehicle does not arm while moving?

en/features_mc/throw_launch.md
hamishwillee: @zeroos Absolutely - because I read it as a hard limit.
The point though is that we want to throw fast enough to invoke the throw trigger, and high enough that the motors can start before the vehicle falls to the height of a person.
Is 2 m likely to be high enough in most cases?

zeroos: It depends on a platform, but I feel like for most multicopters small enough to be thrown, it should be fine, unless they are very limited in power. I'll try to make it clear in the modification.

en/SUMMARY.md
hamishwillee: @zeroos Understandable.

Thanks for your review. I have integrated your comments into the introduction, so you can have another look if you like. Much appreciated.

Note, I expected to be wrong. But by writing all the assumptions a normal user might make, it forces you to correct me :-)

en/features_mc/throw_launch.md
hamishwillee: No need, I have fixed it :-)

en/features_mc/throw_launch.md
```diff
@@ -0,0 +1,92 @@
...
+- The majority of testing has been done in position mode.
+  Other modes should also work.
+
+:::
+
+This feature allows a multicopter to be started by arming it from a fixed position and then throwing it into the air.
+The vehicle then turns on the motors and operates according to its current mode.
+
+The vehicle detects that it has been thrown based on reaching a certain speed (5m/s), and then starts the motors at the apex of the throw, once it determines that it has started to fall.
+You need to throw it high enough so that the vehicle can stabilize its height well before falling in range of people (2 or 3 metres is usually sufficient).
```
zeroos, line 16: This might suggest to some that it's okay to throw the drone 2m away from people, I try to keep at least 10-20m even when flying normally... Maybe we can just remove "in range of people"?

en/SUMMARY.md
zeroos: Thank you, nice work, looks good to me!

I left just one minor comment about mentioning the distance to people, but overall I'm very happy with the result, thanks! :)

en/features_mc/throw_launch.md
```diff
@@ -0,0 +1,92 @@
+# Throw Launch (Multicopter)
+
+:::warning Experimental
+This feature was introduced after PX4 v1.14.
+
+- It has not yet been broadly tested on different vehicle configurations or scenarios. various platforms and in different scenarios.
```
MaEtUgR, line 6:
```suggestion
- It has not yet been broadly tested on different vehicle configurations or scenarios.
```

en/features_mc/throw_launch.md
```diff
@@ -0,0 +1,92 @@
...
+- The majority of testing has been done in position mode.
+  Other modes should also work.
```
MaEtUgR, line 8:
```suggestion
- The majority of testing has been done in position mode. Other modes should also work.
```

en/features_mc/throw_launch.md
```diff
@@ -0,0 +1,92 @@
...
+
+:::
+
+This feature allows a multicopter to be started by arming it from a fixed position and then throwing it into the air.
+The vehicle then turns on the motors and operates according to its current mode.
+
+The vehicle detects that it has been thrown based on reaching a certain speed (5m/s), and then starts the motors at the apex of the throw, once it determines that it has started to fall.
```
MaEtUgR, line 15:
```suggestion
The vehicle detects that it has been thrown based on reaching a certain speed (default 5m/s), and then starts the motors at the apex of the throw, once it determines that it has started to fall.
```

en/features_mc/throw_launch.md
MaEtUgR: I agree maybe "in range of any obstacles" is less tempting to bring people close.

hamishwillee: Good catch. It really depends on your vehicle - if you had a small drone with prop guards it wouldn't matter. Let's do this.

```suggestion
You need to throw it high enough so that the vehicle can stabilize its height well before falling anywhere near people.
```

en/features_mc/throw_launch.md
```diff
@@ -0,0 +1,92 @@
...
+
+The vehicle should not be transported after being armed and before the throw.
+In particular, the throw should not be executed from a moving platform.
+The reason for this is that the condition to start the motors depends on absolute speed of the multicopter and does not account for any additional movement.
+Trying to throw the drone from a moving platform might result in the motors being started prematurely.
+
+## Safety
+
+:::warning
+Throw launch is potentially dangerous as it requires the operator to hold an armed multicopter and be in proximity when it is flying.
```
MaEtUgR, line 29:
```suggestion
Throw launch is dangerous as it requires the operator to hold an armed multicopter and be in proximity when it is flying.
```
Let's be honest.

en/SUMMARY.md
hamishwillee: Great. I've addressed that. @MaEtUgR says he is going to look at this very soon. With some luck today.

MaEtUgR (commented):
Nice, thanks for the very useful, safety aware documentation!
The only thing really missing in my eyes is mentioning the fact that once the feature is enabled you cannot take off "normally". So the throw takeoff is mutually exclusive with a takeoff from the ground and you have to choose using the parameter before arming.

en/SUMMARY.md
```diff
@@ -68,6 +68,9 @@
...
+      * [Throw Launch](features_mc/throw_launch.md)
+      * [Precision Landing](advanced_features/precland.md)
```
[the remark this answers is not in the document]
hamishwillee: I'd be happy for them to be modes when they are - indeed I tried to push precland into being a mode recently. But it's hard because it can happen inside a return or mission or forced through shell - but not standalone. Just odd. 

Are there any other features that we might consider MC only that don't depend on Computer vision - other than collision prevention of course?

en/features_mc/throw_launch.md
hamishwillee: Whoops. Thanks.

en/features_mc/throw_launch.md
hamishwillee: I like things on separate lines.

en/features_mc/throw_launch.md
hamishwillee: I've played with a throwable caged rotor copter at about 200g that wasn't scary. But I take your point.

en/features_mc/throw_launch.md
hamishwillee: I put default because that speed is a default. I guess it doesn't matter since the fact it can be changed is clear lower down.

en/advanced_features/README.md
```diff
@@ -13,3 +13,4 @@ This section contains topics related to some of the more advanced features of th
 - [Iridium/RockBlock Satellite Communication System](../advanced_features/satcom_roadblock.md)
 - [Precision Landing](../advanced_features/precland.md)
 - [RTK GPS](../gps_compass/rtk_gps.md)
+* [Throw Launch](../advanced_features/throw_launch.md)
```
hamishwillee, line 16:
```suggestion
```

en/SUMMARY.md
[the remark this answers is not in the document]
hamishwillee: PS I moved the document for throw launch into flight_modes_mc folder, which will affect the URL. However it will still be displayed as an mc feature until it is really a flight mode. This means it won't break later when it is a mode. 
Note that we will be grouping the modes together under MC (except this one) soon. I just have to finish splitting Return mode and mission mode.

en/flight_modes_mc/throw_launch.md
```diff
@@ -0,0 +1,97 @@
...
+You need to throw the vehicle high enough so that it can stabilize its height well before falling anywhere near people or obstacles.
+
+Notes:
+
+- The mode is disabled by default, and must be enabled using a [parameter](#parameters) before arming.
+- When enabled you cannot take off from the ground using the normal modes.
+- The vehicle should not be transported after being armed and before the throw.
+  In particular, the throw should not be executed from a moving platform.
+  The reason for this is that the condition to start the motors depends on absolute speed of the multicopter and does not account for any additional movement.
+  Trying to throw the drone from a moving platform might result in the motors being started prematurely.
```
hamishwillee, line 29: @MaEtUgR OK, I rewrote this a little to ensure it mentions the arming tone, and that the vehicle must be enabled using a parameter, and that you can't take off from the ground using normal modes. 

This should all be about right now. If no comments will merge tomorrow.

hamishwillee:
Merged! Thanks very much @zeroos !

zeroos:
Yay! :rocket: Thank you! :)

en/features_mc/throw_launch.md
MaEtUgR: I destroyed the keyboard of my laptop using a 35g micro quad. I agree it's less dangerous but if my finger (or eye) would have taken that hit that would have lead to injury...

hamishwillee: Ouch!

PR #2932, opened 2023-11-23
Geofence update

sfuhrer:
Most important docs update after https://github.com/PX4/PX4-Autopilot/pull/22394.

Should we actually start calling it "geo-fence" instead of "geofence" or "geoFence"?


de/flying/geofence.md
```diff
@@ -41,5 +41,9 @@ In summary:
 1. Once finished, click on the **Upload** button (top right) to send the fence (along with rally points and mission) to the vehicle.
 1. Set the breach action in the [GeoFence Failsafe](../config/safety.md#geofence-failsafe).
 
+:::note Geofences that do not include the Home position will be rejected by the flight controller.
```
hamishwillee, line 44: This is a change to a german doc. Needs to be removed from here and moved into the /en tree.

hamishwillee:
I have no problem with "geofence" - and it is is very widely used - it appears more prevalent than geo-fence or variants. If you feel strongly though it doesn't matter as long as we are consistent across the documentation.

de/flying/geofence.md
sfuhrer: thanks, done!

sfuhrer:
> I have no problem with "geofence" - and it is is very widely used - it appears more prevalent than geo-fence or variants. If you feel strongly though it doesn't matter as long as we are consistent across the documentation.

See last commit - let's go with geofence consistently.

hamishwillee (approved):
Excellent! Thanks

PR #2950, opened 2023-11-29
Return mode - split mc, fw, vtol

hamishwillee:
This splits the return mode docs into vehicle specific versions:

- [x] MC return
- [x] FW return
- [x] VTOL return
- [x] Generic return

This is an 80:20 split - the vehicle specific docs simple outline of the default behaviour of the return type plus configuration of the type. This makes them quite simple, and useful "out of the box" if you want to use the recommended behaviour.

All possible return types are detail information in the "generic" return mode doc, which is cross linked. 
The generic version includes the RTL types without talking about specific details. So  "could apply to any vehicle".

There are some open questions against VTOL, and also FW. For example, it is not clear if FW ignores the cone setting (always) or just sets it to zero by default and it is not expected to be touched.


en/flight_modes_fw/return.md
```diff
@@ -0,0 +1,229 @@
...
+- Flies via direct path to the home position or a rally point (whichever is closest)
+- [Land or waits](#arrival) at descent altitude (depending on landing parameters).
+
+:::note
+If no rally points are defined, this is the same as a _Return to Launch_ (RTL)/_Return to Home_ (RTH).
+:::
+
+<a id="mission_landing_return"></a>
+
+### Mission Landing/Rally Point Return Type (RTL_TYPE=1)
```
sfuhrer, line 60: Rally points landings are not possible yet in FW. They really only can be used to "rally" a/multiple vehicles at certain points on the map (they then just loiter and wait there).

sfuhrer: Proper rally point landing support is coming though

hamishwillee:
I'm going to get back to this as soon as I can. Promise.

en/flight_modes/return.md
```diff
@@ -59,26 +78,20 @@ If no rally points are defined, this is the same as a _Return to Launch_ (RTL)/_
...
-<a id="mission_landing_pattern"></a>
-
-:::note
-A mission landing pattern consists of a [MAV_CMD_DO_LAND_START](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_LAND_START), one or more position waypoints, and a [MAV_CMD_NAV_LAND](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_LAND).
-:::
+- If the destination is a rally point or home it will [land or wait](#hover-landing-at-destination) at descent altitude (depending on landing parameters).
 
 :::warning
-When this type is set PX4 will reject any mission without a valid landing pattern.
+When this type is set PX4 will reject a mission without a valid landing pattern.
```
hamishwillee, line 92: @sfuhrer I wanted to clarify the VTOL Return mode landing behaviour for RTL_TYPE=1 (the default), and specifically this point "When this type is set PX4 will reject a mission without a valid landing pattern.".

For a plane this makes sense. You probably won't use a rally point because there is no way to specify the flight path to land. So you define a mission and a mission landing pattern, and you always land on it whether in a mission or return mode.

But what about VTOL? I just tried this and I was forced to define a "fixed wing" landing pattern. But a VTOL can land vertically - it doesn't need to define the flight path. Now I could define rally points and all should work well, but I am STILL forced to define the mission landing pattern, and if I actually want to have a mission, I'm going to have to land using this pattern as though I was fixed wing. What's up?

en/flight_modes/return.md
```diff
@@ -45,11 +60,15 @@ More detailed explanations for each of the types are provided in the following s
...
+
 In this return type the vehicle:
 
-- Ascends to a safe [return altitude](#return_altitude) (above any expected obstacles).
+- Ascends to a safe [minimum return altitude](#minimum-return-altitude) (above any expected obstacles).
 - Flies via direct path to the home position or a rally point (whichever is closest)
-- [Land or waits](#arrival) at descent altitude (depending on landing parameters).
+- On [arrival](#hover-landing-at-destination) descends to "descent altitude" and waits for a configurable time.
+  This time may be used to deploy landing gear.
+- Lands or waits (this depends on landing parameters: a fixed-wing vehicle circles at the descent altitude).
```
hamishwillee, line 71: What does a VTOL do for RTL_TYPE=0. Is it similar to VTOL land - by default transition to MC mode at the descent altitude, based on `NAV_FORCE_VT`, and then wait based on the descent delay? If that isn't set, it lands as FW?

en/flight_modes/return.md
```diff
@@ -117,29 +130,28 @@ If no mission is defined PX4 will fly directly to home location and land (rally
...
 :::note
-This is true for most [return types](#return_types).
-The exception is when executing a [mission path return](#mission_path_return) from within a mission, where the vehicle follows mission waypoints (we can assume these avoid any obstacles).
+The exception is when executing a [mission path return](#mission-path-return-type-rtl-type-2) from _within a mission_.
+In this case the vehicle follows mission waypoints, which we assume are planned to avoid any obstacles.
 :::
 
 The return altitude for a fixed-wing vehicle is configured using the parameter [RTL_RETURN_ALT](#RTL_RETURN_ALT).
+
 The return altitude for multicopter and VTOL vehicles is configured using the parameters [RTL_RETURN_ALT](#RTL_RETURN_ALT) and [RTL_CONE_ANG](#RTL_CONE_ANG), which define a half cone centered around the destination (home location or safety point).
```
hamishwillee, line 155: @sfuhrer Here we imply that the fixed wing return altitude does not use the code. Is it that it "cannot" or that "it does not by default". If it is the later, is it that it "should not"?

en/flight_modes_vtol/return.md
```diff
@@ -0,0 +1,67 @@
+# Return Mode (VTOL)
```
hamishwillee, line 1: @sfuhrer I have VTOL questions for RTL_TYPE=1 (the default)

1. There is a note that says "When this type is set PX4 will reject a mission without a valid landing pattern."
   - I tried this in simulation and it appears to be correct.
   - Doesn't that mean you have to have a landing pattern for VTOL? How does that make sense? A VTOL might want to use a mission with a normal VTOL_LAND - doesn't this preclude it?
2. The original doc defined the behaviour for VTOL RTL_TYPE=1 as:
   > - If a mission landing is defined, fly direct to the mission landing start point and then land.
   > - Otherwise fly directly to the home position, transition to multicopter mode, and land as a multicopter.

   This differs from the general instructions that say "closest of mission landing or any rally point gets run", home gets used if no mission landing or rally point is defined" - my assumption is that the "general instructions" are correct (i.e. rally points are used by VTOL too, in preference to home)?

3. The docs state that:
   >   (for a rallypoint/home location) If not in a mission landing, a VTOL in FW mode will _always_ transition back to MC just before landing (ignoring [NAV_FORCE_VT](../advanced_config/parameter_reference.md#NAV_FORCE_VT)).

   - If we're flying as an FW when return mode is triggered, do we use the FW settings for return altitude - i.e ignore the cone setting? 
   - If we're flying as an MC when return mode is triggered, do we switch to FW to fly back? 
   - If the answer is "no" then do we obey the minimum return altitude of MC, taking account of the cone?
   - When does the transition happen? Does it happen below the descent altitude (i.e. on final landing) or at the descent altitude (i.e. when waiting)?
   
   My guess here is that we use the MC settings in MC mode, FW settings in FW mode. 
   - So if we're in FW when the return mode occurs we fly back at RTL_ALTITUDE (ignoring the cone), spiral down to the descent altitude, start landing, transition to MC, then land. The only difference from FW in return behaviour being no indefinite loiter at the wait point (since we can land) and the transition.
   - If we're in MC mode we'd fly back at the minimum return altitude as an MC, taking account of the cone parameter (which seems to default to 45 deg in SITL), and simply land as normal as an MC. No switching back to FW.
   
   Can you confirm?

hamishwillee:
@sfuhrer I've updated this and I am pretty happy with it. There are some open questions in here that need an answer, and of course it needs a review. In particular the VTOL case. Once you pull the main use cases apart like this it becomes more obvious "what we don't say".

en/flight_modes/return.md
sfuhrer: The cone it only respects in a hovering mode.

TYPE=0 for VTOL: climb to RTL_RETURN_ALT, fly over Home, start descending to RTL_DESCEND_ALT while loitering with RTL_LOITER_RAD, once that altitude is reached aligns heading to Home (center of loiter) and starts transitioning.

en/flight_modes/return.md
sfuhrer: It cannot use it.

en/flight_modes_vtol/return.md
sfuhrer: >There is a note that says "When this type is set PX4 will reject a mission without a valid landing pattern."

This comment is outdated, the param `MIS_TKO_LAND_REQ` governs the requirements now.

> Doesn't that mean you have to have a landing pattern for VTOL? How does that make sense? A VTOL might want to use a mission with a normal VTOL_LAND - doesn't this preclude it?

VTOL_LAND also counts as a landing (now). And if you don't have any landing planned but the type is 1, then it just goes to Home.


> my assumption is that the "general instructions" are correct (i.e. rally points are used by VTOL too, in preference to home)?

Yes rally points are used as well, they are like multiple Home points, and the closest is taken. The parameter description of RTL_TYPE should be correct.


>If we're flying as an FW when return mode is triggered, do we use the FW settings for return altitude - i.e ignore the cone setting?

Don't fully understand what you mean, there is no specific "FW setting for return altitude". Cone is ignored. Vehicle returns as if it would be a fixed-wing vehicle, and then transitions to MC before landing.

I confirm your other points, especially the last two.

en/flight_modes_fw/return.md
```diff
@@ -0,0 +1,69 @@
...
+  Note that return altitude cannot be configured using the "cone" parameter in fixed-wing vehicles.
+- Flies via direct constant-altitude path to the destination, which will be the closest of the start of a _mission landing pattern_ and any rally point, or the home location if no mission landing pattern or rally points are defined.
+- If the destination is a _mission landing pattern_ it will follow the pattern to land.
+- If the destination is a rally point or home it will descend to the descent altitude, and then loiter or land (depending on landing parameters).
+
+A fixed-wing vehicle is expected to use a landing pattern defined in a mission as the return destination, as this is the safest way to land automatically.
+A mission landing pattern consists of a [MAV_CMD_DO_LAND_START](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_LAND_START), one or more position waypoints, and a [MAV_CMD_NAV_LAND](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_LAND).
+
+:::warning
+When this return type is set PX4 will reject a mission without a valid landing pattern.
```
sfuhrer, line 44: Outdated.

en/flight_modes_fw/return.md
```diff
@@ -0,0 +1,69 @@
...
+:::
+
+When the destination is a rally point or the home location, on arrival the vehicle will rapidly descend to the altitude defined by [RTL_DESCEND_ALT](#RTL_DESCEND_ALT), and by default circle above the destination indefinitely at radius [NAV_LOITER_RAD](#NAV_LOITER_RAD).
```
sfuhrer, line 47:
```suggestion
When the destination is a rally point or the home location, on arrival the vehicle will rapidly descend to the altitude defined by [RTL_DESCEND_ALT](#RTL_DESCEND_ALT), and by default circle above the destination indefinitely at radius [RTL_LOITER_RAD](#RTL_LOITER_RAD).
```

en/flight_modes_fw/return.md
```diff
@@ -0,0 +1,69 @@
...
+| <a id="RTL_LAND_DELAY"></a>[RTL_LAND_DELAY](../advanced_config/parameter_reference.md#RTL_LAND_DELAY)    | Time to hover at `RTL_DESCEND_ALT` before landing (default: 0.5s) -by default this period is short so that the vehicle will simply slow and then land immediately. If set to -1 the system will loiter at `RTL_DESCEND_ALT` rather than landing. The delay is provided to allow you to configure time for landing gear to be deployed (triggered automatically). |
+| <a id="NAV_LOITER_RAD"></a>[NAV_LOITER_RAD](../advanced_config/parameter_reference.md#NAV_LOITER_RAD)    | [Fixed-wing Only] The radius of the loiter circle (at [RTL_LAND_DELAY](#RTL_LAND_DELAY).                                                                                                                                                                                                                                                                         |
+
```
sfuhrer, line 64: RTL_LOITER_RAD should be added.

en/flight_modes_vtol/return.md
```diff
@@ -0,0 +1,67 @@
...
+  <!-- Note that return altitude cannot be configured using the "cone" parameter in fixed-wing vehicles. -->
+- Flies via direct constant-altitude path to the destination, which will be the closest of the start of a _mission landing pattern_ and any rally point, or the home location if no mission landing pattern or rally points are defined.
+- If the destination is a mission landing pattern it will follow the pattern to land.
+- If the destination is a rally point or home it will transition to a multicopter at the descent altitude and then land.
+  Note that [NAV_FORCE_VT](../advanced_config/parameter_reference.md#NAV_FORCE_VT) is igored: the vehicle will always land as a multicopter for these destinations.
+
+A mission landing pattern consists of a [MAV_CMD_DO_LAND_START](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_LAND_START), one or more position waypoints, and a [MAV_CMD_NAV_LAND](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_LAND).
+
+:::warning
+When this type is set PX4 will reject a mission without a valid landing pattern.
```
sfuhrer, line 46: outdated

en/flight_modes_vtol/return.md
```diff
@@ -0,0 +1,67 @@
...
+| <a id="RTL_LAND_DELAY"></a>[RTL_LAND_DELAY](../advanced_config/parameter_reference.md#RTL_LAND_DELAY)    | Time to hover at `RTL_DESCEND_ALT` before landing (default: 0.5s) -by default this period is short so that the vehicle will simply slow and then land immediately. If set to -1 the system will loiter at `RTL_DESCEND_ALT` rather than landing. The delay is provided to allow you to configure time for landing gear to be deployed (triggered automatically). |
+| <a id="NAV_LOITER_RAD"></a>[NAV_LOITER_RAD](../advanced_config/parameter_reference.md#NAV_LOITER_RAD)    | [Fixed-wing Only] The radius of the loiter circle (at [RTL_LAND_DELAY](#RTL_LAND_DELAY).                                                                                                                                                                                                                                                                         |
+
```
sfuhrer, line 62: add RTL_LOITER_RAD

en/flight_modes_vtol/return.md
```diff
@@ -0,0 +1,67 @@
...
+<!-- https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/commander/ModeUtil/mode_requirements.cpp -->
+
+:::
+
+## Technical Summary
+
+VTOL vehicles use the [Mission Landing/Rally Point](../flight_modes/return.md#mission-landing-rally-point-return-type-rtl-type-1) return type by default.
+In this return type the vehicle:
+
```
sfuhrer, line 34: I would separate it into two parts: RTL sequence when triggered while in hovering state, and while in forward flying state. And for the hovering it maybe is enough to reference to the MC docs, as it's 1:1 the same.

en/flight_modes_fw/return.md
hamishwillee: Fixed

en/flight_modes_fw/return.md
hamishwillee: Done - replaced the NAV_LOITER_DELAY above in the table.

en/flight_modes_vtol/return.md
hamishwillee: Thanks. For the landing pattern part, I have taken this to mean that a landing pattern is defined as:

> This consists of a [MAV_CMD_DO_LAND_START](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_LAND_START), one or more position waypoints, and a [MAV_CMD_NAV_LAND](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_LAND) (or [MAV_CMD_NAV_VTOL_LAND](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_VTOL_LAND) for a VTOL Vehicle).

So you still need the LAND_START. Presumably an MC that used this would just LAND on the final point.

hamishwillee: Otherwise, all good I think. Thanks.

en/flight_modes/return.md
hamishwillee: So that's the FW behaviour. The MC behaviour would just be to fly back as an MC.

en/flight_modes_vtol/return.md
hamishwillee: Done.

hamishwillee:
@sfuhrer Thanks very much for the review - really helpful and I am pretty sure I "get it". I have integrated the changes.

Since I'm fairly happy and I'm going on holiday I'm going to merge this. If you see anything glaring, feel free to patch the docs.

PR #2966, opened 2023-12-08
removes occurrences of PX4_GZ_MODEL env variable

frede791:
PX4_GZ_MODEL has been deprecated but there were still instructions in the user guide which used it.


hamishwillee (approved):
I should have caught this when you made the previous change. Thanks very much for sorting this out.

I do wonder if in the root simulation page we should have  flat list of these variables and a description - since they are SITL and not simulator dependent?

PR #2968, opened 2023-12-09
correct serial adrees line

avcuenes:


en/middleware/uxrce_dds.md
```diff
@@ -199,7 +199,7 @@ When working with real hardware, the setup depends on the hardware, OS, and chan
 For example, if you're using the RasPi `UART0` serial port, you might connect using this command (based on the information in [Raspberry Pi Documentation > Configuring UARTS](https://www.raspberrypi.com/documentation/computers/configuration.html#configuring-uarts)):
 
 ```sh
-sudo MicroXRCEAgent serial --dev /dev/AMA0 -b 921600
+sudo MicroXRCEAgent serial --dev /dev/ACM0 -b 921600
```
hamishwillee, line 202: I don't think this is correct. My understanding is that ACM refers to a USB device while AMA refers to a UART. The text before says this is an example of a UART, which would mean this doesn't match the rest of the page. Right?

avcuenes: Thanks, I understand difference.

PR #2979, opened 2023-12-13
[WIP][DONT_MERGE] Vitepress tests

hamishwillee:
Needs new merge after: 
- [ ] https://github.com/PX4/PX4-Autopilot/pull/22943
  - [ ] Remove the vuepress items after this all working
- [ ] Deployment
  - [ ] Will look a bit like this https://github.com/mavlink/qgroundcontrol/blob/master/.github/workflows/docs_deploy.yml and my deployment test. 
  - [ ] Need to disable jenkins-based deployment

This is a test version of PX4 user guide running on [Vitepress](https://vitepress.dev/). PX4 user guide runs on Vuepress 1 - whereas Vitepress is effectively "Vuepress 3". It has a nicer default theme, and much better build characteristics. 
I'm not ready to update yet, but this project allows me to track what needs to happen for that to occur.

- [ ] public folder in root 
  -  The public folder contains site stuff like logos, and also uorb graph and failsafe sim code. Needs to move from `/.vuepress` to root of the docs. 
  - CI needs to update the build to copy files to same.
  - [ ] need to fix up pages so this renders still. 
- [x] Package.json - make sure it matches original properly
- [x] .gitignore - needs to exclude build and temp files.
- [x] Fix `The language '5' is not loaded, falling back to 'txt' for syntax highlighting. (x2)`
- [x] index files in root and language roots
- [x] `::: note` to `:::info NOTE`
- [ ] Special cases to test/fix
   - [ ] Algolia
   - [x] Zoom images 
   - [x] uorb graph - yes, but automated stuff happening
   - [x] failsafe
   - [x] tabs - caac4ffe70653b92cf411cae33454f209fb70e9e - needed plugin.
   - [x] favicon
   - [x] version switchers
   - [ ] Only export translated docs, not empty translations.
   - [x] notes to info blocks conversion or just support somehow
   - [ ] Support, dev call etc where we have a page that needs to redirect - because it won't be latest.
     - Done in two items - english and also "all other langs". 
     - [ ] Before merging we'd remove that PR for all other langs and let crowdin sort it out.  
     - [ ] For translated pages., don't have Edit this link. Instead if possible point to crowdin
- [x] Convert SUMMARY.md to use `-`rather than `*`
- [ ] Disabled checks using `ignoreDeadLinks: true,` in config.mjs. Need to re-enable this and fix up remaining cases for last part.
- [x] Convert README to index.html
  - [x] Done for English in 8439e8abfe391865915bf2d7317dd548dd23e5ac but may need to revert if doesn't rebase cleanly
  - [x] Summary would need to be updated along with all other files.
  - [x] Going to need to think about best way to import in crowdin.

   
- [ ] Can fix up anchors `<a id="reference_design_generations"></a>` to be in the title `## Reference Design Generations {reference_design_generations}`. Consider if needed.
- [x] Google analytics - needs Ramon to provide new key
- [x] Root redirector pages
- [ ] Root redirect vue component
- [ ] LinkBadge component

Other things to consider
- [ ] Part builds for extra languages?


.vitepress/get_sidebar.js
```diff
@@ -0,0 +1,268 @@
+// .vuepress/get_sidebar.js
```
hamishwillee, line 1: Builds sidebar from readme. Need to remove some of  the debug.

PR #2980, opened 2023-12-13
add cuav gps

cuhome:


hamishwillee:
@cuhome I'm setting this to Draft since this isn't finished. Please undraft it when ready for review. Minimally there should be no errors reported in https://github.com/PX4/PX4-user_guide/pull/2980#issuecomment-1853504342 by the flaw checker - this says you are missing some files.

PR #2995, opened 2023-12-20
ethernet: update default IP to 10.41.10.x

bkueng:
- [x] https://github.com/PX4/PX4-Autopilot/pull/22517

### DIFF
--- a/en/advanced_config/ethernet_setup.md
+++ b/en/advanced_config/ethernet_setup.md
@@ -35,7 +35,7 @@ To connect systems over Ethernet you need to configure them to run on the same I
 This might be done using a DHCP server to allocate addresses, or by manually configuring the addresses of each system on the network.
 
 There is no single "out of the box configuration" that we can provide that will necessarily work in your local network.
-Therefore as an example of the kind of configuration you might do, below we show how to set up the systems on an IP network with static addresses in the range `192.168.0.Xxx`, where PX4 has a statically allocated address `192.168.0.4` and the computer has address `192.168.0.1`.
+Therefore as an example of the kind of configuration you might do, below we show how to set up the systems on an IP network with static addresses in the range `10.41.10.Xxx`, where PX4 has a statically allocated address `10.41.10.2` and the computer has address `10.41.10.1`.
 If you wanted to connect a companion computer or other system to the network you could use a similar approach to allocate a static address.
 
 ::: info
@@ -56,10 +56,10 @@ A configuration file might look like this:
 ```
 DEVICE=eth0
 BOOTPROTO=fallback
-IPADDR=192.168.0.4
+IPADDR=10.41.10.2
 NETMASK=255.255.255.0
-ROUTER=192.168.0.254
-DNS=192.168.0.254
+ROUTER=10.41.10.254
+DNS=10.41.10.254
 ```
 
 Where the values are:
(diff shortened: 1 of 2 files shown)


en/advanced_config/ethernet_setup.md
```diff
@@ -35,7 +35,7 @@ To connect systems over Ethernet you need to configure them to run on the same I
 This might be done using a DHCP server to allocate addresses, or by manually configuring the addresses of each system on the network.
 
 There is no single "out of the box configuration" that we can provide that will necessarily work in your local network.
-Therefore as an example of the kind of configuration you might do, below we show how to set up the systems on an IP network with static addresses in the range `192.168.0.Xxx`, where PX4 has a statically allocated address `192.168.0.4` and the computer has address `192.168.0.1`.
+Therefore as an example of the kind of configuration you might do, below we show how to set up the systems on an IP network with static addresses in the range `10.41.10.Xxx`, where PX4 has a statically allocated address `10.41.10.2` and the computer has address `10.41.10.1`.
```
hamishwillee, line 38: @bkueng Just to be clear, this default IP address is set on boards, meaning that there is a default `net.cfg` somewhere, and below we're adding a value to overwrite this? The value is then used for setting things like where XRCE-DDS expects the default port and so on?

The reason I ask is that here and everywhere else in the doc the IP address is presented as a random example you might use. If it is a default then we probably should say that, because then presumably you don't have to do anything in the "PX4 Ethernet Network Setup" section?

So I was thinking you might say something like:

```suggestion
Therefore as an example of the kind of configuration you might do, below we show how to set up the systems on an IP network with static addresses in the range `10.41.10.Xxx`, where PX4 has a statically allocated address `10.41.10.2` (the default) and the computer has address `10.41.10.1`.
```

And then below in the "PX4 Ethernet Network Setup" section say something like - if you are happy to use this IP address you don't need to modify the configuration on PX4?

bkueng: Yes exactly, I was wondering if we should actually mention this is already the default.
The default behavior is also set to FALLBACK, which means it tries DHCP first. If you anyway want to use a static IP, you can directly set that and it gets faster initialized on startup.

hamishwillee: So PX4 will try read the IP from whatever DHCP server is configured. If it doesn't get an IP assigned it uses the default static IP. If you want to drop through that first step you have to explicitly define a config file with the default static IP (or some other value). Right?

bkueng: Correct

hamishwillee: Cool. Note added

hamishwillee (approved):

en/releases/main.md
```diff
@@ -48,6 +48,12 @@ TBD ...
   Gazebo additional supports drag-and-drop via the resource spawner in Gazebo GUI.
   Added to PX4 in [PX4-Autopilot#22467](https://github.com/PX4/PX4-Autopilot/pull/22467).
 
+### Ethernet
+
+- [Ethernet](../advanced_config/ethernet_setup.md): The default static IP address (used as a fallback for DHCP) has changed from `192.168.0.3` to `10.41.10.2`.
```
hamishwillee, line 53: @bkueng I added this release note. I'm trying to do this for all changes that touch docs FYI, because I think it is useful, but also so that when we get to the next release we're not scabbling to remember what we did and why. Appreciate it if you could check this (and of course start to make adding these a habit).

hamishwillee:
@bkueng I think this is good. Feel free to merge this along side your source code PR.

hamishwillee:
@bkueng Is this going to happen? I ask mostly to understand if this is a v1.15 thing or a one day thing.

bkueng:
It's merged

hamishwillee:
Thanks @bkueng

PR #2999, opened 2023-12-21
Mavlink custom messages update

hamishwillee:
This is a significant update to the MAVLink custom message doc. This did not match the current way of doing things, which is to have a separate file for the streaming class and include it. Also somewhat confusing for readers who don't understand they have to create their own CA_TRAJECTORY message and Uorb topic.

There are several main parts to the change
1. Explains the streaming class file and how it is included
2. Changes correct way for streaming message to be updating the groups in mavlink_main.cpp (not using startup file, though that is mentioned).
3. Explains how you can use MAVLink to set streaming rates. The old "set streaming rates" uses the shell, and I have made that part of testing
4. Adds a section on updating MAVSDK/QGC. Cross links to this from testing.

I've added a few notes, where I'd quite like help.

@julianoes @bkueng Would you be able to review?

### DIFF
--- a/en/middleware/mavlink.md
+++ b/en/middleware/mavlink.md
@@ -9,7 +9,7 @@ The protocol defines a number of standard [messages](https://mavlink.io/en/messa
 This tutorial explains how you can add PX4 support for your own new "custom" messages.
 
 :::note
-The tutorial assumes you have a [custom uORB](../middleware/uorb.md) `ca_trajectory` message in `msg/ca_trajectory.msg` and a custom MAVLink `ca_trajectory` message in `mavlink/include/mavlink/v2.0/custom_messages/mavlink_msg_ca_trajectory.h`.
+The tutorial assumes you have created your own [custom uORB](../middleware/uorb.md) message named `CaTrajectory` in `msg/CaTrajectory.msg` and a corresponding custom MAVLink `CA_TRAJECTORY` message that is defined in `\src\modules\mavlink\mavlink\message_definitions\v1.0\custom_messages.xml` (code for this will be generated as part of the build to `/build/px4_sitl_default/mavlink/custom_messages/mavlink_msg_ca_trajectory.h`).
 :::
 
 ## Defining Custom MAVLink Messages
@@ -38,22 +38,34 @@ The [MAVLink Developer guide](https://mavlink.io/en/getting_started/) has more i
 
 ## Sending Custom MAVLink Messages
 
-This section explains how to use a custom uORB message and send it as a MAVLink message.
+This section explains how to stream the content of a custom uORB message as a MAVLink message.
 
-Add the headers of the MAVLink and uORB messages to
-[mavlink_messages.cpp](https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/mavlink/mavlink_messages.cpp)
+### Define the Streaming Class
+
+First create a file named `CA_TRAJECTORY.hpp` for your streaming class inside the [/src/modules/mavlink/streams](https://github.com/PX4/PX4-Autopilot/tree/main/src/modules/mavlink/streams) directory (named after the corresponding MAVLink message).
+
+Add the headers for the MAVLink and uORB messages to the top of the file:
 
 ```C
 #include <uORB/topics/ca_trajectory.h>
 #include <v2.0/custom_messages/mavlink.h>
 ```
 
-Create a new class in [mavlink_messages.cpp](https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/mavlink/mavlink_messages.cpp#L2193)
+:::note
+The uORB topic's snake-case header file is generated from the CamelCase uORB filename at build time.
+The `custom_messages/mavlink.h` header is also generated at build time.
+:::
+
+Then copy the streaming class definition below into the file:
 
 ```C
 class MavlinkStreamCaTrajectory : public MavlinkStream
 {
 public:
+    static MavlinkStream *new_instance(Mavlink *mavlink)
+    {
+        return new MavlinkStreamCaTrajectory(mavlink);
+    }
     const char *get_name() const
     {
         return MavlinkStreamCaTrajectory::get_name_static();
@@ -69,10 +81,6 @@ public:
     uint16_t get_id()
     {
         return get_id_static();
-    }
-    static MavlinkStream *new_instance(Mavlink *mavlink)
-    {
-        return new MavlinkStreamCaTrajectory(mavlink);
     }
     unsigned get_size()
     {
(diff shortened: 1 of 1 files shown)


en/middleware/mavlink.md
```diff
@@ -105,7 +112,7 @@ protected:
             _msg_ca_trajectory.seq_id = _ca_trajectory.seq_id;
```
hamishwillee, line 112: This line and above define the mapping from uorb to MAVLInk message CA_TRAJECTORY. I think we might help people by providing the definitions of both of those in the doc for people to copy. They could then run the tutorial end to end and know it makes sense. 

For the uorb topic, would this make sense based on the above values? Advice on coefficients?

```
uint64 timestamp				# time since system start (microseconds)
uint64 time_start_usec			# time for trajectory start (microseconds)
uint64 time_stop_usec			# time for trajectory stop (microseconds)
float32 coefficients				# NOT SURE ABOUT THIS - it matches the type, but is singular, so does not match name
uint32 seq_id					# sequence id
```

What do you think of the idea, and if so, is this message OK or we could just pretend that BATTERY_STATUS say is a new message?

@julianoes Opinions?

bkueng: I think using an existing one would be a bit simpler. Then we can also directly point to the messages.

en/middleware/mavlink.md
```diff
@@ -114,28 +121,83 @@ protected:
...
+- `MAVLINK_MODE_NORMAL`: Streamed to a GCS.
+- `MAVLINK_MODE_ONBOARD`: Streamed to a companion computer on a fast link, such as Ethernet
+- `MAVLINK_MODE_ONBOARD_LOW_BANDWIDTH`: Streamed to a companion computer on slower link
+- `MAVLINK_MODE_GIMBAL`: Streamed to a gimbal
+- `MAVLINK_MODE_EXTVISION`: Streamed to an external vision system
+- `MAVLINK_MODE_EXTVISIONMIN`: Streamed to an external vision system on a slower link
+- `MAVLINK_MODE_OSD`: Streamed to an OSD, such as an FPV headset.
+- `MAVLINK_MODE_CUSTOM`: Stream nothing by default. Used when configuring streaming using MAVLink.
+- `MAVLINK_MODE_MAGIC`: Same as `MAVLINK_MODE_CUSTOM`
+- `MAVLINK_MODE_CONFIG`: ?
```
bkueng, line 172: This is used over USB by default, with a bit higher rates than NORMAL

en/middleware/mavlink.md
```diff
@@ -114,28 +121,83 @@ protected:
...
+
+### Streaming by Default
 
+The easiest way to stream your messages by default (as part of a build) is to add them to [mavlink_main.cpp](https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/mavlink/mavlink_main.cpp) in the appropriate message group.
+
+If you search in the file you'll find groups of messages defined in a switch statement:
+
+- `MAVLINK_MODE_NORMAL`: Streamed to a GCS.
+- `MAVLINK_MODE_ONBOARD`: Streamed to a companion computer on a fast link, such as Ethernet
+- `MAVLINK_MODE_ONBOARD_LOW_BANDWIDTH`: Streamed to a companion computer on slower link
```
bkueng, line 165: We use this on skynode, not because of slower link but to reduce traffic afterwards (routing to the GCS)

en/middleware/mavlink.md
```diff
@@ -216,33 +279,59 @@ It should be used for development only!
...
+
+### Updating QGroundControl
+
+You will need to [Build QGroundControl](https://docs.qgroundcontrol.com/master/en/qgc-dev-guide/getting_started/index.html) including a pre-built C library that contains your custom messages.
+
+QGC uses a pre-built C library that must be located at [/qgroundcontrol/libs/mavlink/include/mavlink](https://github.com/mavlink/qgroundcontrol/tree/master/libs/mavlink/include/mavlink) in the QGC source.
+
+By default this is pre-included as a submodule from <https://github.com/mavlink/c_library_v2> but you can [generate your own MAVLink Libraries](https://mavlink.io/en/getting_started/generate_libraries.html).
+
+QGC uses the ArduPilotMega.xml dialect by default, which includes **common.xml**.
```
bkueng, line 331:
```suggestion
QGC uses the all.xml dialect by default, which includes **common.xml**.
```

en/middleware/mavlink.md
hamishwillee: Thank you!

en/middleware/mavlink.md
hamishwillee: Cheers!

```suggestion
- `MAVLINK_MODE_ONBOARD_LOW_BANDWIDTH`: Streamed to a companion computer for re-routing to a reduced-traffic link, such as a GCS.
```

en/middleware/mavlink.md
hamishwillee: Thanks
```suggestion
- `MAVLINK_MODE_CONFIG`: Streaming over USB with higher rates than `MAVLINK_MODE_NORMAL`.
```

en/middleware/mavlink.md
```diff
@@ -2,200 +2,419 @@
...
+
+:::note
+Not all "standard" messages, commands and microservices are _implemented_.
+:::
 
 PX4 includes the [mavlink/mavlink](https://github.com/mavlink/mavlink) repo as a submodule under [/src/modules/mavlink](https://github.com/PX4/PX4-Autopilot/tree/main/src/modules/mavlink), and generates the MAVLink 2 C header files at build time.
 
 There are are number of XML dialect files in [/mavlink/messages/1.0/](https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/).
-The dialect that is built is specified using the variable `MAVLINK_DIALECT` in [/src/modules/mavlink/CMakeLists.txt](https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/mavlink/CMakeLists.txt#L34); by default this is [development.xml](https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/development.xml).
+The dialect that is built is defined in the variable `CONFIG_MAVLINK_DIALECT` used in [/src/modules/mavlink/CMakeLists.txt](https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/mavlink/CMakeLists.txt).
```
hamishwillee, line 65: The value is set in `CONFIG_MAVLINK_DIALECT` rather than `MAVLINK_DIALECT` as indicated, and appears to be set in `default.px4board` for SITL. This seems odd - is the setting there used for ALL boards?

en/middleware/mavlink.md
```diff
@@ -2,200 +2,419 @@
...
+	}
 
-        return false;
-    }
 };
 ```
 
+Most streaming classes are very similar (see examples in [/src/modules/mavlink/streams](https://github.com/PX4/PX4-Autopilot/tree/main/src/modules/mavlink/streams)):
+
+- The streaming class derives from [`MavlinkStream`](https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/mavlink/mavlink_stream.h).
```
hamishwillee, line 228:
```suggestion
- The streaming class derives from [`MavlinkStream`](https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/mavlink/mavlink_stream.h) and is named using the pattern `MavlinkStream``CamelCaseMessageName>`.
```

en/middleware/mavlink.md
```diff
@@ -2,200 +2,419 @@
...
+	}
 
-        return false;
-    }
 };
 ```
 
+Most streaming classes are very similar (see examples in [/src/modules/mavlink/streams](https://github.com/PX4/PX4-Autopilot/tree/main/src/modules/mavlink/streams)):
+
+- The streaming class derives from [`MavlinkStream`](https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/mavlink/mavlink_stream.h) and is named using the pattern `MavlinkStream``CamelCaseMessageName>`.
```
hamishwillee, line 228:
```suggestion
- The streaming class derives from [`MavlinkStream`](https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/mavlink/mavlink_stream.h) and is named using the pattern `MavlinkStream``<CamelCaseMessageName>`.
```

en/middleware/mavlink.md
```diff
@@ -2,200 +2,419 @@
...
+
+- The streaming class derives from [`MavlinkStream`](https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/mavlink/mavlink_stream.h) and is named using the pattern `MavlinkStream``<CamelCaseMessageName>`.
+- The `public` definitions are "near-boilerplate", allowing PX4 to get an instance of the class (`new_instance()`), and then to use it to fetch the name, id, and size of the message from the MAVLink headers (`get_name()`, `get_name_static()`, `get_id_static()`, `get_id()`, `get_size()`).
+  For your own streaming classes these can just be copied and modified to match the values for your MAVLink message.
+- The `private` definitions subscribe to the uORB topics that need to be published.
+  Here we also define constructors to prevent the definition being copied.
+- The `protected` section is where the important work takes place!
+  Here we override the `send()` method, copying values from the subscribed uORB topic(s) into appropriate fields in the MAVLink message, and then send the message.
+
+  In this case the uORB topic has multiple instances: one for each battery.
```
hamishwillee, line 236: Downside of battery is that it support multiple instances so more complicated processing. Upside is that we get to discuss. I wasn't aware for example difference between subscription `update()` and `copy()`. Section below needs to be checked carefully.

en/middleware/mavlink.md
```diff
@@ -2,200 +2,419 @@
...
-
-    f.timestamp = hrt_absolute_time();
-    f.seq_id = traj.seq_id;
-    f.time_start_usec = traj.time_start_usec;
-    f.time_stop_usec = traj.time_stop_usec;
-    for(int i=0;i<28;i++)
-        f.coefficients[i] = traj.coefficients[i];
-        
-    _ca_traj_msg_pub.publish(f);
+	if ((msg->sysid != mavlink_system.sysid) || (msg->compid == mavlink_system.compid)) {
```
hamishwillee, line 380: So what instance are we writing to? It looks like maybe "the default instance"? What if there are two mavlink batteries and one "internal" one? How do we sort that out? I don't plan to say too much but we should comment on what is happening.

en/middleware/mavlink.md
```diff
@@ -2,200 +2,419 @@
...
+This is the kind of implementation that you would provide to support a MAVLink battery integration with PX4.
 
-```C
-#include <uORB/topics/ca_trajectory.h>
-#include <v2.0/custom_messages/mavlink_msg_ca_trajectory.h>
+Add the headers for the incoming MAVLink message and the uORB topic to publish to in [mavlink_receiver.h](https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/mavlink/mavlink_receiver.h#L77):
+
+```cpp
+#include <uORB/topics/battery_status.h>
+#include <v2.0/development/mavlink.h>
```
hamishwillee, line 356: I presume that we don't need the mavlink import even for custom messages? Presumably this is done automatically based on config value?

en/middleware/mavlink.md
```diff
@@ -2,200 +2,419 @@
...
+	bool send() override
+	{
+		bool updated = false;
 
-        if (_sub.update(&_ca_trajectory)) {
-            mavlink_ca_trajectory_t _msg_ca_trajectory;  //make sure mavlink_ca_trajectory_t is the definition of your custom MAVLink message
+		// Loop through _battery_status_subs (subscription to array of BatteryStatus instances)
+		for (auto &battery_sub : _battery_status_subs) {
+            // battery_status_s is a struct that can hold the battery object topic
+			battery_status_s battery_status;
```
hamishwillee, line 197: Original version imported this as a struct - ie. `struct battery_status_s battery_status;` or similar. Doesn't seem to be necessary to declare that type.

en/middleware/mavlink.md
```diff
@@ -2,200 +2,419 @@
...
+
+### Define the Streaming Class
+
+First create a file named `BATTERY_STATUS_DEMO.hpp` for your streaming class (named after the message to stream) inside the [/src/modules/mavlink/streams](https://github.com/PX4/PX4-Autopilot/tree/main/src/modules/mavlink/streams) directory.
+
+Add the headers for the MAVLink and uORB messages to the top of the file:
+
+```cpp
+#include <uORB/topics/battery_status.h>
+#include <v2.0/development/mavlink.h>
```
hamishwillee, line 137: You shouldn't need to import mavlink here right? I.e. headers are imported at PX4 level based on the config variable, and by this point you should get the correct selection?

bkueng: Yes exactly

en/middleware/mavlink.md
bkueng: `uORB::Publication<battery_status_s>` means to use single-instance. For multiple it really depends on what you want to achieve. The simplest (and automatic) is to use an array of `uORB::PublicationMulti`.

en/middleware/mavlink.md
bkueng: The other boards use `common` by default: https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/mavlink/Kconfig#L10

hamishwillee: Ah. So you'd change `default "common"` there to your XML file if you wanted something custom?

en/middleware/mavlink.md
hamishwillee: @bkueng I don't think that quite answers my question. For the incoming battery status message (the current/real one) we have this single publication, so what we have here is a single topic.

```cpp
uORB::Publication<battery_status_s>	_battery_pub{ORB_ID(battery_status)};
```

But when we're streaming we have multiple instances we iterate through and publish a MAVLink message for each.

My inference is that there are multiple battery instance definitions possible in PX4 but we only write into to one of them - if that is correct (?) then which one - the first item, which I expect is the battery with mavlink instance id `1`?

Further doesn't this imply that while we can emit MAVLink messages for any number of connected batteries, we can only ever have one "incoming" battery connected to PX4 via MAVLink?

Lastly, given your answer, I think that this means that if we wanted to have multiple batteries "incoming" and connected via MAVLink we'd need to use a multiple publication, and take the instance number to work out which one to write to. Right?

I'm not going to block on this but if you answer I'll put in an update on my return.

----

As an aside, I think we/I need to go through uORB docs again and document this both as a guide and as a reference - explaining all these methods for subscription. The current docs are ... insufficient.

hamishwillee:
Merging as "much better and more accurate than before". There is an open question here https://github.com/PX4/PX4-user_guide/pull/2999/files#r1446771518 that I hope to address (following a response) as a post process.

en/middleware/mavlink.md
bkueng: You could, but need to be aware if a board has set `CONFIG_MAVLINK_DIALECT`, it will override it.

en/middleware/mavlink.md
bkueng: > My inference is that there are multiple battery instance definitions possible in PX4 but we only write into to one of them - if that is correct (?) then which one - the first item, which I expect is the battery with mavlink instance id 1?

Yes in that case it would publish to the first uorb instance. The battery case is a bit complex for that reason. The message also has an `id` field that's used for mavlink.

> Lastly, given your answer, I think that this means that if we wanted to have multiple batteries "incoming" and connected via MAVLink we'd need to use a multiple publication, and take the instance number to work out which one to write to. Right?

Yes that's where `PublicationMulti` comes in.

en/middleware/mavlink.md
hamishwillee: Thanks. The interesting bit here is that common is really the default on real hardware - a bit of a worry since testing is mostly in SITL so if you don't do real hardware testing you could well use definitions that aren't available in the real world.

I made all this a bit more clear in https://github.com/PX4/PX4-user_guide/pull/3027

en/middleware/mavlink.md
hamishwillee: Added explanation in https://github.com/PX4/PX4-user_guide/pull/3027

PR #3000, opened 2023-12-21
VTOL airspeedless: update parameter names

sfuhrer:
Required after https://github.com/PX4/PX4-Autopilot/pull/22510

### DIFF
--- a/en/config_vtol/vtol_without_airspeed_sensor.md
+++ b/en/config_vtol/vtol_without_airspeed_sensor.md
@@ -1,23 +1,25 @@
 # VTOL Without an Airspeed Sensor
 
+<Badge type="warning" text="Experimental" />
+
 :::warning
-The use of an airspeed sensor is advised.
 Support for VTOLs without an airspeed sensor is considered experimental and should only be attempted by experienced pilots.
+The use of an airspeed sensor is recommended.
 :::
 
 Fixed-wing vehicles use airspeed sensors to determine the speed at which the airplane is moving through the air.
-Depending on wind this could vary from groundspeed. 
-Every airplane has a minimum airspeed below which the airplane will stall. 
-In mild weather conditions and with settings significantly above stall speed a VTOL can operate without the use of an airspeed sensor. 
+Depending on wind this could vary from groundspeed.
+Every airplane has a minimum airspeed below which the airplane will stall.
+In mild weather conditions and with settings significantly above stall speed a VTOL can operate without the use of an airspeed sensor.
 The settings should also be applicable to non-VTOL fixed-wings but this is currently untested.
 
 This guide will outline the parameter settings required to bypass the airspeed sensor for VTOL planes.
 
 ## Preparation
 
-Before attempting to eliminate an airspeed sensor you should first determine a safe throttle level. 
+Before attempting to eliminate an airspeed sensor you should first determine a safe throttle level.
 Also the duration for a front transition needs to be known.
-To do this you can either perform a reference flight with an airspeed sensor or fly the vehicle manually. 
+To do this you can either perform a reference flight with an airspeed sensor or fly the vehicle manually.
 In both cases the reference flight should be performed in very low wind conditions.
 
 The flight should be performed at a speed that would be sufficient to fly in high wind conditions and should consist of:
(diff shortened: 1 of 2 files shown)


en/config_vtol/vtol_without_airspeed_sensor.md
```diff
@@ -1,23 +1,25 @@
 # VTOL Without an Airspeed Sensor
 
+<Badge type="warning" text="Experimental" />
```
hamishwillee, line 3: FYI only, I'm experimenting with badges to indicate this kind of thing "as well". I hope to link to a definition at some point.

en/config_vtol/vtol_without_airspeed_sensor.md
```diff
@@ -32,49 +34,45 @@ The flight should be performed at a speed that would be sufficient to fly in hig
 After the reference flight download the log and use [FlightPlot](../log/flight_log_analysis.md#flightplot) (or another analysis tool) to examine the log.
 Plot the altitude (`GPOS.Alt`), thrust (`ATC1.Thrust`), groundspeed (Expression: `sqrt(GPS.VelN\^2 + GPS.VelE\^2)`), pitch (`ATT.Pitch`) and roll (`AT.Roll`).
 
-Examine the throttle level (thrust) when the vehicle is level (no or little pitch and roll), 
-during the ascend (increasing altitude) and when the vehicle is banking (more roll). 
-The initial value to use as cruise speed should be the highest thrust applied during a roll or ascend, 
-the thrust during level flight should be considered the minimum value if you decide to further tune down your speed.
+Examine the throttle level (thrust) when the vehicle is level (no or little pitch and roll), during the ascend (increasing altitude) and when the vehicle is banking (more roll).
```
hamishwillee, line 37: FYI only I tend to break on sentences. It's a preference that helps me review the flow of docs when editing, but I also think most comments apply to a sentence, so it's not a bad point to break.

hamishwillee (changes requested):
This is is awesome, but can you add a release note to https://docs.px4.io/main/en/releases/main.html#vtol ?

sfuhrer:
> This is is awesome, but can you add a release note to https://docs.px4.io/main/en/releases/main.html#vtol ?

Nice, I wasn't aware of that page, put it on it now. 
It's though not clear if it should fall into the VTOL or FW category, as VTOL and FW share this feature. The same struggle we will have with many MC features, as they also are used on VTOLs. 
Maybe we should put all shared FW/VTOL features into the FW category, and specify that VTOL only covers the custom VTOL features and that also most of the MC and FW changes are important for it.

en/releases/main.md
```diff
@@ -72,7 +72,11 @@ TBD ...
 
 ### Fixed-wing
 
-- TBD
+- [Simplified airspeed sensor configuration](../config_vtol/vtol_without_airspeed_sensor.md): 
+Replace parameter CBRK_AIRSPD_CHK by [SYS_HAS_NUM_ASPD](../advanced_config/parameter_reference.md#SYS_HAS_NUM_ASPD) and rename parameter FW_ARSP_MODE to [FW_USE_AIRSPD](../advanced_config/parameter_reference.md#FW_USE_AIRSPD).
+To be able to arm without an airspeed sensor set SYS_HAS_NUM_ASPD to 0.
+To not use the airspeed sensor data in the controller, set FW_USE_AIRSPD to 0.
+  Added to PX4 in [PX4-Autopilot#22510](https://github.com/PX4/PX4-Autopilot/pull/22510).
```
hamishwillee, line 79:
```suggestion
- [Simplified airspeed sensor configuration](../config_vtol/vtol_without_airspeed_sensor.md): 
  Replacef parameter `CBRK_AIRSPD_CHK` with [SYS_HAS_NUM_ASPD](../advanced_config/parameter_reference.md#SYS_HAS_NUM_ASPD) and renamed parameter `FW_ARSP_MODE` to [FW_USE_AIRSPD](../advanced_config/parameter_reference.md#FW_USE_AIRSPD).
  To be able to arm without an airspeed sensor set `SYS_HAS_NUM_ASPD` to 0.
  To not use the airspeed sensor data in the controller, set `FW_USE_AIRSPD` to 0.
  Added to PX4 in [PX4-Autopilot#22510](https://github.com/PX4/PX4-Autopilot/pull/22510).
```

hamishwillee:
> > This is is awesome, but can you add a release note to https://docs.px4.io/main/en/releases/main.html#vtol ?
> 
> Nice, I wasn't aware of that page, put it on it now. It's though not clear if it should fall into the VTOL or FW category, as VTOL and FW share this feature. The same struggle we will have with many MC features, as they also are used on VTOLs. Maybe we should put all shared FW/VTOL features into the FW category, and specify that VTOL only covers the custom VTOL features and that also most of the MC and FW changes are important for it.

Good points. First thing to remember is that the release note structure is not set in stone. If we can generally find a way that captures the broad changes better we can do that. For example, we might always assume everything is vehicle specific, but have a heading "shared features" or similar and use badges or tags to indicate specific vehicles if it is not all of them.

Or we stick with the current structure and add a badge to indicate the supported vehicle types - in this case both vtol and fw?

Second thing, this is FW and VTOL but the document says VTOL and I would not have assumed as a reader it was for FW too. So does this need to be retitled? Is anything else VTOL or FW specific in it? I can work on duplicating in some way if it is truly generic and we have retitled. And again, we can use badges to indicate vehicle types it applies to.

I'm going to merge. Let's continue to discuss here and update this later.

sfuhrer:
> Good points. First thing to remember is that the release note structure is not set in stone. If we can generally find a way that captures the broad changes better we can do that. For example, we might always assume everything is vehicle specific, but have a heading "shared features" or similar and use badges or tags to indicate specific vehicles if it is not all of them.

We could do:
shared MC/VTOL
shared FW/VTOL
MC specific
FW specific
VTOL specific

> Or we stick with the current structure and add a badge to indicate the supported vehicle types - in this case both vtol and fw?

I'm atm rather in favor of the above, so without badges but with "shared" categories.

> Second thing, this is FW and VTOL but the document says VTOL and I would not have assumed as a reader it was for FW too. So does this need to be retitled? Is anything else VTOL or FW specific in it? I can work on duplicating in some way if it is truly generic and we have retitled. And again, we can use badges to indicate vehicle types it applies to.

Most of that page is as applicable to FW as it is to VTOL, yes. There's a couple of things like transition tuning and quad-chute that ofc is VTOL specific.

hamishwillee:
Thanks. I've added a note in the airspeedless doc about FW https://github.com/PX4/PX4-user_guide/pull/3014

You're welcome to update the shared FW/VTOL sections when and if you want! I probably won't for now.

PR #3027, opened 2024-01-31
MAVLink custom messages - follow up about publication multi

hamishwillee:
This addresses issues from #2999

- Clarity on `CONFIG_MAVLINK_DIALECT` 
  - essentially you should use this to set the dialect on a board specific basis.
  - MORE important, this means that common.xml is used by default on real hardware and development on SITL. I was thinking that development.xml was used by default everywhere.
- Point out that our battery code has limitations, and how we might work around them.
- A little moving of stuff around.

FYI @bkueng This falls out of your comments. Feedback welcome but I think this is better.

### DIFF
--- a/en/middleware/mavlink.md
+++ b/en/middleware/mavlink.md
@@ -243,18 +243,22 @@ protected:
 
 Most streaming classes are very similar (see examples in [/src/modules/mavlink/streams](https://github.com/PX4/PX4-Autopilot/tree/main/src/modules/mavlink/streams)):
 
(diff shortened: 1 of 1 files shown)


en/middleware/mavlink.md
```diff
@@ -375,6 +392,13 @@ In this case the [BatteryStatus](../msg_docs/BatteryStatus.md) uORB topic alread
 uORB::Publication<battery_status_s> _battery_pub{ORB_ID(battery_status)};
 ```
 
+This creates a publication to a single uORB topic instance, which by default will be the _first_ instance.
+
+:::note
+This implementation won't work on multi-battery systems, because several batteries might be publishing data to the first instance of the topic, and there is no way to differentiate them.
+To support multiple batteries we'd need to use `PublicationMulti` and map the MAVLink message instance IDs to specific uORB topic instances.
+:::
```
hamishwillee, line 400: Explanation of why this implementation is flawed. I think worth saying, and pointing out PublicationMulti

PR #2842, opened 2023-11-02
[WIP] Fixed-wing performance model docs

RomanBapst:
Docs for 

- [x] https://github.com/PX4/PX4-Autopilot/pull/22091

- [x] mention that some compensations apply only for propellers (e..g maximum climbrate would be different for jet engine)
- [x] document procedure when user does initial tuning in conditions very different from sea level standard atmospheric conditions

PR #3029, opened 2024-02-01
Fixed-wing performance model docs - update

hamishwillee:
This follows on from #2842 which I merged too quickly. I'm going to merge this to fix some mostly minor things. FYI @RomanBapst @sfuhrer

### DIFF
--- a/en/config_fw/weight_and_altitude_tuning.md
+++ b/en/config_fw/weight_and_altitude_tuning.md
@@ -1,52 +1,63 @@
 # Advanced TECS Tuning (Weight and Altitude)
 
-Before following the guidance provided by this section you should have completed the [basic TECS tuning](../config_fw/position_tuning_guide_fixedwing.md#tecs-tuning-altitude-and-airspeed) which established the key performance limitations of the vehicle that are very important for the altitude and airspeed controller to function properly.
-While those limitations have been defined using constant parameters, in reality vehicle performance is not constant and is affected by various factors.
-This section specifically looks at how vehicle weight and air density affect performance and what you can do do compensate for those effects.
+This topic shows how you can compensate for changes to the [weight of the vehicle](#specify-the-weight-of-the-vehicle) and the [air density](#compensation-for-air-density), along with information about the [algorithms](#weight-and-density-compensation-algorithms) that are used.
 
-Both the vehicle weight and the air density, which depends on altitude and air temperature, affect vehicle performance.
-For example, when a vehicle climbs to a higher altitude the air density decreases, and usually the maximum climb rate reduces.
-Similarly, the maximum climb rate of a vehicle will reduce with the weight of the vehicle.
+:::warning
+This topic requires that you have already performed [basic TECS tuning](../config_fw/position_tuning_guide_fixedwing.md#tecs-tuning-altitude-and-airspeed).
+:::
 
+[Basic TECS tuning](../config_fw/position_tuning_guide_fixedwing.md#tecs-tuning-altitude-and-airspeed) established the key performance limitations of the vehicle that are required for the altitude and airspeed controller to function properly.
+
+While those limitations are specified using constant parameters, in reality vehicle performance is not constant and is affected by various factors.
 If changes in weight and air density are not taken into account, altitude and airspeed tracking will likely deteriorate in the case where the configuration (air density and weight) deviate significantly from the configuration at which the vehicle was tuned.
 
-In section [Specify the Weight of the Vehicle](../config_fw/weight_and_altitude_tuning.md#specify-the-weight-of-the-vehicle) and [Compensation for Air Density](../config_fw/weight_and_altitude_tuning.md#compensation-for-air-density) you will find the parameters you need to adjust for weight and density compensation to take place.
-In the [Notes](../config_fw/weight_and_altitude_tuning.md#notes-to-the-derivation-of-weight-and-density-compensation) section the curious reader can find more details on theory of each compensation.
+## Vehicle Weight Compensation
 
-## Specify the Weight of the Vehicle
+Set (both) the following parameters to scale the maximum climb rate, minimum sink rate, and adjust airspeed limits for weight:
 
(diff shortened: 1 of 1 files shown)


en/config_fw/weight_and_altitude_tuning.md
```diff
@@ -55,28 +66,28 @@ By calibrated we mean the value of that variable measured at sea level in standa
...
+The minimum sink rate ([FW_T_SINK_MIN](../advanced_config/parameter_reference.md#FW_T_SINK_MIN)) is scaled as a function of weight ratio
 
 The minimum sink rate can be written as:
 
 $$\dot{h}_{min} = \sqrt{2mg\over{\rho F}} f(c_A, c_W)$$
 
 where $\rho$ is the air density, F is the wing area and $f(c_A, c_W)$ is a function of the polars.
-It's easy to see that the minimum sink rate scales with the square root of the weight ratio.
+
+From this equation we see that the minimum sink rate scales with the square root of the weight ratio.
```
hamishwillee, line 90: FYI It is "easy to see", but what some see as easy is not as obvious to others. My training tells me to avoid suggesting things are easy.

en/config_fw/weight_and_altitude_tuning.md
```diff
@@ -1,52 +1,63 @@
...
+
+While those limitations are specified using constant parameters, in reality vehicle performance is not constant and is affected by various factors.
 If changes in weight and air density are not taken into account, altitude and airspeed tracking will likely deteriorate in the case where the configuration (air density and weight) deviate significantly from the configuration at which the vehicle was tuned.
 
-In section [Specify the Weight of the Vehicle](../config_fw/weight_and_altitude_tuning.md#specify-the-weight-of-the-vehicle) and [Compensation for Air Density](../config_fw/weight_and_altitude_tuning.md#compensation-for-air-density) you will find the parameters you need to adjust for weight and density compensation to take place.
-In the [Notes](../config_fw/weight_and_altitude_tuning.md#notes-to-the-derivation-of-weight-and-density-compensation) section the curious reader can find more details on theory of each compensation.
+## Vehicle Weight Compensation
```
hamishwillee, line 14: FYI This section is how I would (ideally) like all instructional sections - this is what you need to change and how.

en/config_fw/weight_and_altitude_tuning.md
```diff
@@ -1,52 +1,63 @@
...
+## Air Density Compensation
 
 ### Specify a Service Ceiling
+
 In PX4 the service ceiling [FW_SERVICE_CEIL](../advanced_config/parameter_reference.md#FW_SERVICE_CEIL) specifies the altitude in standard atmospheric conditions at which the vehicle is still able to achieve a maximum climb rate of 0.5 m/s at maximum throttle and weight equal to [WEIGHT_BASE](../advanced_config/parameter_reference.md#WEIGHT_BASE).
 By default this parameter is disabled and no compensation will take place.
-This parameter needs to be estimated experimentally. It is always better to set a conservative value (lower value) than an optimistic value.
+
+This parameter needs to be determined experimentally.
+It is always better to set a conservative value (lower value) than an optimistic value.
```
hamishwillee, line 35: From "Effect of Density on Maximum Climb Rate"

> where  is the air density at sea level in the standard atmosphere and K is a scaling factor which determines the slope of the function. Rather than trying to identify this constants, the usual practice in aviation is to specify a service ceiling altitude at which the vehicle is still able to achieve a minimum specified climb rate

So my understanding is that while this scales by density ratio the scaling factor is hard to work out. So instead we set a service ceiling in which we know that the plane will fly OK. Is that right?

So what does the service ceiling do? I mean if I set this to 200m does this mean the vehicle is capped to that height?

Is there a rule of thumb about how you go about determining this? Can you suggest one? I.e. can you do some single maneuver and work it out from your logs?

en/config_fw/weight_and_altitude_tuning.md
```diff
@@ -1,52 +1,63 @@
...
 
 ### Apply Density Correction to Minimum Sink Rate
-During the initial tuning the minimum sink rate is set using [FW_T_SINK_MIN](../advanced_config/parameter_reference.md#FW_T_SINK_MIN).
 
-If the initial tuning is not done in standard sea level conditions then [FW_T_SINK_MIN](../advanced_config/parameter_reference.md#FW_T_SINK_MIN) needs to be multiplied with correction factor P:
+The minimum sink rate is set in [FW_T_SINK_MIN](../advanced_config/parameter_reference.md#FW_T_SINK_MIN).
+
+If the [Basic TECS tuning](../config_fw/position_tuning_guide_fixedwing.md#tecs-tuning-altitude-and-airspeed) was not done in standard sea level conditions then the [FW_T_SINK_MIN](../advanced_config/parameter_reference.md#FW_T_SINK_MIN) parameter must be modified by multiplying with correction factor $P$ (where $\rho$ is the air density during tuning):
 
 $$P = \sqrt{\rho\over{\rho_{sealevel}}}$$
```
hamishwillee, line 43: So the density has to be known at tuning time right? Doesn't that mean you'll have to re-tune to set this because you probably didn't measure it?

Should we suggest us 1.225 kg/m^3 for the sea level value. How do you go about measuring the density when you're tuning? Can we get this from any of our on-board sensors? 

How sensitive is vehicle behaviour to getting this wrong?

en/config_fw/weight_and_altitude_tuning.md
RomanBapst: @hamishwillee I think I have read too many papers which suggest that something is "easy to see". I assume it's a way to try and make people feel dumb. I don't have a problem removing it. To me this section is for someone who is into the PX4 code and wants to double check calculations.

hamishwillee: That's precisely it - we don't want to make people feel dumb. I do think this is fairly easy to see, but it isn't a good habit to get into :-)

en/config_fw/weight_and_altitude_tuning.md
hamishwillee: @RomanBapst I'm going to stop chasing you on these.

PR #3031, opened 2024-02-01
[WIP] Offboard mode topic restructure

hamishwillee:
This attempts to improve the offboard mode topic so it is more clear how things work and to ease splitting it in future.

Still thinking about it.


en/flight_modes/offboard.md
```diff
@@ -2,12 +2,17 @@
...
+
+The offboard controller supplies a stream of position, velocity, acceleration, attitude, attitude rate, or thrust/torque setpoints, which are injected into the [PX4 control architecture](../flight_stack/controller_diagrams.md) as though they had been generated by the equivalent internal controller — the corresponding internal controllers that would otherwise have generated the setpoint (in another mode) are turned off.
 
+The setpoints may be provided using MAVLink (or a MAVLink API such as [MAVSDK](https://mavsdk.mavlink.io/)) or by [ROS 2](../ros/ros2.md).
 PX4 requires that the external controller provides a continuous 2Hz "proof of life" signal, by streaming any of the supported MAVLink setpoint messages or the ROS 2 [OffboardControlMode](../msg_docs/OffboardControlMode.md) message.
 PX4 enables offboard control only after receiving the signal for more than a second, and will regain control if the signal stops.
 
+Different vehicle frame types implement different [control architectures](../flight_stack/controller_diagrams.md).
+As a result the setpoints and setpoint combinations that can be set depend on the vehicle.
```
hamishwillee, line 14: @beniaminopozzan How you going? This is my understanding of how offboard mode works (irrespective of ROS2 or MAVLink). Is it correct?

Specifically, the bit about how we inject in the control architecture, and the bit about different vehicle frames implementing different architectures being the reason we can't use the same setpoints/setpoint combinations in all cases.

There might also be a subtlty that some combinations just aren't tested.

beniaminopozzan: Hi @hamishwillee . Nicely busy :) How about you?

Yes, this is how offboard control works.
The setpoints are injected in different ways between MAVLink and ROS 2, in particular MAVLink adds a level of abstraction as MAVLink messages are converted in uorb. The ROS 2 approach instead goes straight at the uorb level. This is why MAVLink offboard control does not require the  "prof of life" while ROS 2 does.
There is activity in trying to simplify this though: https://github.com/PX4/PX4-Autopilot/pull/22530

hamishwillee: Thanks @beniaminopozzan - busy too. A bit refreshed from my holiday though which is good. 

So what we have now in docs for ROS2 is that:
1. you call `OffboardControlMode` to turn on the mode and configure which controllers to turn off (or to put it another way, where you want to inject your external setpoints). 
2. YOu call the setpoint messages for the external setpoints you want. You've turned off the preceding controllers so you no longer have to worry about the values being swamped.

That's correct right - the docs don't actually make it quite as clear that you're turning off the controllers, more that you're setting the setpoints that you use (if you see the distinction).

But what is missing here is that presumably calling  `OffboardControlMode` can fail or you might slip out by doing a manual mode change? How are you expected to handle this case? I don't think the docs cover this?

As I understand the linked PR everything _can_ work exactly as above.
What is added is some "proxy" offboard topics for _some_ of the topics. If you're in offboard mode these are used to set the real offboard topic, and if not, then they don't. So you would write to those topics instead.

It's going to be a bit more messy to describe, but not much more.

en/flight_modes/offboard.md
```diff
@@ -2,12 +2,17 @@
 
 [<img src="../../assets/site/position_fixed.svg" title="Position fix required (e.g. GPS)" width="30px" />](../getting_started/flight_modes.md#key_position_fixed)
 
-The vehicle obeys position, velocity, acceleration, attitude, attitude rates or thrust/torque setpoints provided by some source that is external to the flight stack, such as a companion computer.
-The setpoints may be provided using MAVLink (or a MAVLink API such as [MAVSDK](https://mavsdk.mavlink.io/)) or by [ROS 2](../ros/ros2.md).
+The _Offboard_ mode allows PX4 to be controlled by some source that is external to the flight stack, such as a companion computer.
```
hamishwillee, line 5: Note to self. Intent is to push the ROS 2 specifics and MAVLink specifics as far as possible into those instructions. The MAVLink part more or less done but the ROS 2 not yet done, as is pending response to https://github.com/PX4/PX4-user_guide/pull/3031#discussion_r1480614076

When that is done probably merge the content in the top of page and description into the description (as far as possible) as there is duplication now.

hamishwillee:
Closing as stale. This is a year old and no longer reflects the doc structure / info. If we were to do this we'd start a design from scratch.

PR #3032, opened 2024-02-01
Kakute H7 Mini - update pinouts and serial ports

iftahnaf:
Hi, 

I updated the Kakute H7 Mini documentation page.

* Adding pinout image.
* Adding `TELEM 2` to the `Serial Port Mapping` and the `Connectors and Pins` sections.


hamishwillee:
Thanks very much!

PR #3038, opened 2024-02-08
Update flight_modes.md

bradenwagstaff:


hamishwillee (approved):
Thanks very much. FYI I am restructuring flight modes stuff as we speak so this might perhaps have been touched anyway.

PR #3039, opened 2024-02-08
Flight modes restructure

hamishwillee:
This restructures flight modes under the associated vehicle type. So for example:

image

This gives us a soft intro to the flight modes that is specific to each vehicle, with links to the detail.
- The top level readme for MC, FW, etc is a simple introduction created from the overview getting started doc on flight modes. 
- The original overview summary doc with the big HTML table is hard to parse and maintain. I have moved the content into the sidebar topics as a "technical description". 
- Not sure what happens to the old top level overview and the getting started.
- There should be no _loss_ of information.

Progress:

- [x] Fixed Wing
- [x] MC
- [x] VTOL
- [x] Rover
- [x] General - make just links to vehicle topics
- [x] Getting Started - remove the stand alone topic (redirect) and move vehicle specific links into Basic concepts section on Flight modes.

Done. FYI @sfuhrer


en/flight_modes_fw/README.md
```diff
@@ -0,0 +1,173 @@
+# Flight Modes (Fixed Wing)
```
hamishwillee, line 1: Note, new top level Fixed wing doc. Created from the original getting started. The Overview is a list of the modes with just enough info for selection. Lower down is the high level explanation. Sidebar topic includes this plus the detailed information

en/flight_modes_fw/stabilized.md
```diff
@@ -1,32 +1,49 @@
...
+    Position of stick maps to the orientation of vehicle.
+  - Pitch: Assistance from autopilot to stabilize the attitude.
+    Position of stick maps to the orientation of vehicle.
+  - Yaw: Stick input is sent directly to control allocation.
+  - Throttle: Stick input is sent directly to control allocation.
 
 ## Parameters
 
 | Parameter | Description |
 | --------- | ----------- |
```
hamishwillee, line 46: @sfuhrer What params affect position mode? Is it the same ones as documented for altitude - https://docs.px4.io/main/en/flight_modes_fw/altitude.html#parameters

en/flight_modes_fw/position.md
```diff
@@ -1,25 +1,40 @@
...
+- Global position estimate is required.
+- Manual control input is required.
+- Roll, Pitch, Yaw, Throttle: Assistance from autopilot to hold position or altitude against wind.
 
 ## Parameters
 
 | Parameter | Description |
 | --------- | ----------- |
-| &nbsp;    |
+| &nbsp;    | &nbsp;      |
```
hamishwillee, line 40: @sfuhrer What params affect position mode? Is it the same ones as documented for altitude - https://docs.px4.io/main/en/flight_modes_fw/altitude.html#parameters

en/SUMMARY.md
```diff
@@ -57,6 +57,19 @@
       * [Setpoint Tuning (Trajectory Generator)](config_mc/mc_trajectory_tuning.md)
         * [Jerk-limited Type Trajectory](config_mc/mc_jerk_limited_type_trajectory.md)
       * [Multicopter Racer Setup](config_mc/racer_setup.md)
+    * [Flight Modes](flight_modes_mc/README.md)
```
hamishwillee, line 60: Note, doc does not yet exist.

en/SUMMARY.md
```diff
@@ -87,6 +111,10 @@
       * [Back-transition Tuning](config_vtol/vtol_back_transition_tuning.md)
       * [VTOL w/o Airspeed Sensor](config_vtol/vtol_without_airspeed_sensor.md)
       * [VTOL Weather Vane](config_vtol/vtol_weathervane.md)
+    * [Flight Modes](flight_modes_vtol/README.md)
```
hamishwillee, line 114: Note, doc does not yet exist. I also haven't worked out the "story". Probably see MC for MC mode and FW for FW mode.

en/SUMMARY.md
```diff
@@ -117,33 +145,8 @@
   * [Geofence](flying/geofence.md)
   * [Safety Point Planning](flying/plan_safety_points.md)
   * [Flight Modes](flight_modes/README.md)
```
hamishwillee, line 147: This is the "table summary". Probably just a link to the other docs. Not sure yet, we still have a few "general" docs undeneath.

en/getting_started/flight_modes.md
```diff
@@ -2,7 +2,7 @@
 
 Flight modes define how the autopilot responds to remote control input, and how it manages vehicle movement during fully autonomous flight.
 
-The modes provide different types/levels of autopilot assistance to the user (pilot), ranging from automation of common tasks like takeoff and landing, through to mechanisms that make it easier to regain level flight, hold the vehicle to a fixed path or position, etc.
+The modes provide different types/levels of autopilot assistance to the user (pilot). This assistance ranges from automation of common tasks like takeoff and landing, mechanisms that make it easier to regain level flight, hold the vehicle to a fixed path or position, etc.
```
hamishwillee, line 5: FYI This doc all still to be done. MOst of this material moving into the flight-specific versions so this will have to become a linking doc with only high level summary.

en/flight_modes_fw/altitude.md
```diff
@@ -28,11 +37,15 @@ RC/manual mode like Stabilized mode but with altitude stabilization (centered st
...
-:::note
-
-- Manual input is required (RC controller, or gamepad/thumbsticks through MAVLink).
-- The altitude is normally measured using a barometer, which may become inaccurate in extreme weather conditions.
+- Manual control input is required (such as RC control, joystick).
+  - Roll: Assistance from autopilot to hold altitude against wind.
+  - Pitch: Assistance from autopilot to stabilize the attitude.
+    Position of RC stick maps to the orientation of vehicle.
+  - Yaw: Input is sent directly to control allocation.
+  - Throttle: Assistance from autopilot to hold altitude against wind.
```
hamishwillee, line 45: FYI @sfuhrer I'm trying to get rid of the tables in https://docs.px4.io/main/en/flight_modes/#fixed-wing - its HTML and not all that easy to "parse" as a reader. 

So the entry points to each mode is now a high level topic like we previously used for the getting started flight modes topic.

But I didn't want to throw away the information in the table for what is controlled in each axis, so I have pulled it here into the technical summary. Hope this makes sense to you.

image

en/flight_modes_vtol/README.md
```diff
@@ -0,0 +1,20 @@
+# Flight Modes (VTOL)
```
hamishwillee, line 1: This introduction is somewhat better than before, since all we did was say that VTOL uses MC modes and FW modes based on its flying config. Now we say that, but point to the main exception docs.

However it is not as good if you're a VTOL user as say an MC, where you get a full list of the modes.

I think good enough for now though, since if you've started working in VTOL you probably already know about the other mode types.

en/SUMMARY.md
hamishwillee: I removed the big table, replacing with links to vehicle specific docs. The content of the table is in each of the detail pages.

hamishwillee:
Merging. This is pretty significant update. Questions inline still need to be answered, but this is "no worse" than before and more aligned to our intent structurally.

Note, this won't build. I will do another update once crowdin has done its thing.

en/flight_modes_fw/position.md
sfuhrer: addressed in https://github.com/PX4/PX4-user_guide/pull/3046

en/flight_modes_fw/stabilized.md
sfuhrer: addressed in https://github.com/PX4/PX4-user_guide/pull/3046

en/flight_modes_fw/README.md
sfuhrer: I would think about removing the middle one, and only have a rough summary and then the very detailed one.

en/flight_modes_fw/altitude.md
sfuhrer: For me it would be okay to drop that table, I never found it very readable. Words are better here in my eyes.

PR #3046, opened 2024-02-15
Flight mode reorg follow up (for FW)

sfuhrer:
Follow up on https://github.com/PX4/PX4-user_guide/pull/3039

Do we really need the mode description in three levels of details (README overveiw, README bit more detailed, and then separate section for them all). Maybe okay to remove the middle level? I've done my proposals here though mainly in there, so those things I would need to copy over to the detail sections then.

### DIFF
--- a/en/flight_modes_fw/README.md
+++ b/en/flight_modes_fw/README.md
@@ -51,7 +51,7 @@ A high level description of the modes is provided below (select the mode-specifi
 [Position mode](../flight_modes_fw/position.md) is the easiest and safest manual mode.
 It is supported on vehicles that have a position estimate (e.g. GPS).
 
-The roll stick controls left/right horizontal movement and the pitch stick is used to ascend/descend.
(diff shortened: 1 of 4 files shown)


hamishwillee (approved):
Thanks @sfuhrer - These are all excellent. Merging, because what we will have when this builds is wrong, and we can reduce that now. 

1. Yes, please copy the corrections down into the detail docs.
2. As discussed, if we're going to delete the the middle level we need to decide what the top level really needs. 
  For example, for Position mode, it might be worth having the text "The vehicle performs a [coordinated turn](https://en.wikipedia.org/wiki/Coordinated_flight) if the roll sticks are non-zero, while the pitch stick controls the rate of ascent/descent.
The throttle determines airspeed — at 50% throttle the aircraft will hold its current altitude with a preset cruise speed."

That makes the high level more like the middle level - but still compact. Thoughts?

PR #3051, opened 2024-02-16
Update README.md

KTechDev4213:
Added the 3dr Zero Quad, a drone kit compatible with the px4 autopilot.


hamishwillee (approved):
Looks good thanks. You might want to consider providing a build/assembly guide on the site at some point.

PR #3068, opened 2024-02-22
[WIP] Modify hardware selection to highlight components better

hamishwillee:
This modifies the hardware section as discussed with @sfuhrer. The current version is too nested.


en/SUMMARY.md
```diff
@@ -339,18 +292,67 @@
...
+      * [Servo Gripper](peripherals/gripper_servo.md)
+    * [Parachute](peripherals/parachute.md)
     * [I2C Peripherals](sensor_bus/i2c_general.md)
       * [I2C bus accelerators](sensor_bus/i2c_general.md#i2c-bus-accelerators)
       * [TFI2CADT01 I2C address translator](sensor_bus/translator_tfi2cadt.md)
     * [CAN Peripherals](can/README.md)
     * [DroneCAN Peripherals](dronecan/README.md)
       * [PX4 DroneCAN Firmware](dronecan/px4_cannode_fw.md)
       * [ARK CANnode](dronecan/ark_cannode.md)
+    * [Remote ID](peripherals/remote_id.md)
```
hamishwillee, line 355: @sfuhrer You had this in operations. From a pilot point of view it is invisible - you need to know about it when assembling a drone, so put it here.

en/SUMMARY.md
```diff
@@ -339,18 +292,67 @@
...
+    * [CUAV HV pm](power_module/cuav_hv_pm.md)
+    * [CUAV CAN PMU](dronecan/cuav_can_pmu.md)
+    * [Holybro PM02](power_module/holybro_pm02.md)
+    * [Holybro PM07](power_module/holybro_pm07_pixhawk4_power_module.md)
+    * [Holybro PM06 V2](power_module/holybro_pm06_pixhawk4mini_power_module.md)
+    * [Holybro PM02D (digital)](power_module/holybro_pm02d.md)
+    * [Holybro PM03D (digital)](power_module/holybro_pm03d.md)
+    * [Pomegranate Systems Power Module](dronecan/pomegranate_systems_pm.md)
+    * [Sky-Drones SmartAP PDB](power_module/sky-drones_smartap-pdb.md)
+  * [Smart Batteries](smart_batteries/README.md)
```
hamishwillee, line 337: Separate from power modules IMO

en/SUMMARY.md
```diff
@@ -339,18 +292,67 @@
...
+    * [Grippers](peripherals/gripper.md)
+      * [Servo Gripper](peripherals/gripper_servo.md)
+    * [Parachute](peripherals/parachute.md)
     * [I2C Peripherals](sensor_bus/i2c_general.md)
       * [I2C bus accelerators](sensor_bus/i2c_general.md#i2c-bus-accelerators)
       * [TFI2CADT01 I2C address translator](sensor_bus/translator_tfi2cadt.md)
     * [CAN Peripherals](can/README.md)
     * [DroneCAN Peripherals](dronecan/README.md)
       * [PX4 DroneCAN Firmware](dronecan/px4_cannode_fw.md)
       * [ARK CANnode](dronecan/ark_cannode.md)
```
hamishwillee, line 354: I wonder if we should move Ethernet Setup down here with the other "BUS" stuff. Inclined not to.

en/SUMMARY.md
```diff
@@ -339,18 +291,68 @@
...
+    * [Parachute](peripherals/parachute.md)
+    * [Remote ID](peripherals/remote_id.md)
+  * [I2C Peripherals](sensor_bus/i2c_general.md)
+    * [I2C bus accelerators](sensor_bus/i2c_general.md#i2c-bus-accelerators)
+    * [TFI2CADT01 I2C address translator](sensor_bus/translator_tfi2cadt.md)
+  * [CAN Peripherals](can/README.md)
+  * [DroneCAN Peripherals](dronecan/README.md)
+    * [PX4 DroneCAN Firmware](dronecan/px4_cannode_fw.md)
+    * [ARK CANnode](dronecan/ark_cannode.md)
+  * [Cable Wiring](assembly/cable_wiring.md)
```
hamishwillee, line 355: Is cable writing really something that should be in hardware? I'm "OK" with it here, but it kind of does belong in an assembly topic too. I guess perhaps good with proper interlinking.

en/SUMMARY.md
sfuhrer: Okay, putting it in either is probably fine, and interlink it.

sfuhrer: Or do you split up the "integration" from the "operational" parts?

hamishwillee: I am still getting my head around what operational means. Where would integration live - i.e. the basic vehicle assembly guide?

hamishwillee: @sfuhrer What I'd like to do is have a guide on the integration/assembly on a per-vehicle basis.

This would be like a wiring quickstart but unbranded and also vehicle specific. It would be centred around a diagram like the wiring chart in the assembly guide - like this one: https://docs.px4.io/main/en/assembly/quick_start_pixhawk6x.html#wiring-chart-overview and have similar instructions on writing up the bits an bobs.

So this would be the image for an MC, but we'd probably have more appropriate outputs for a Quadplane VTOL (say).
The instructions would be assuming Pixhawk standard outputs and point to the hardware section for options.

These existing specific assembly guides would go below the flight controller docs.

My reasoning is that even though there is a lot of common information for assembly there are lots of things that aren't well covered. For example, we talk very little about what outputs you might need for plane and VTOL anywhere other than to say you need a BEC.

I'd be OK for the guide to be common, but only if we can expand out the powering and wiring for the other frames. Either way though, I think it would help to present that as part of the instructions for each vehicle.

Ultimately I'd prefer that users face this almost entirely through vehicle-specific docs.

Make sense?

en/data_links/README.md
```diff
@@ -0,0 +1,3 @@
+# Data Links
```
hamishwillee, line 1: Note, to populate when content of section populated.

en/gps_compass/README.md
```diff
@@ -1,7 +1,7 @@
-# GPS & Compass
+# Global Navigation Satellite Systems (GNSS)
```
hamishwillee, line 1: Note, even though this will cover both, I'm working on making it less compass-centric, and will link.

en/assembly/mount_gps_compass.md
```diff
@@ -1,15 +1,15 @@
...
+The diagram below shows the heading marker on the Pixhawk 4 flight controller and compass.
 
 ![Connect compass/GPS to Pixhawk 4](../../assets/flight_controller/pixhawk4/pixhawk4_compass_gps.jpg)
 
+
 ## Compass Orientation
 
-The compass can be mounted in any of the standard MAVLink orientations defined in [MAV_SENSOR_ORIENTATION](https://mavlink.io/en/messages/common.html#MAV_SENSOR_ORIENTATION).
+If you can't mount it towards the front of the vehicle then it can be mounted in any of the standard MAVLink orientations defined in [MAV_SENSOR_ORIENTATION](https://mavlink.io/en/messages/common.html#MAV_SENSOR_ORIENTATION).
 The orientation follows the same frame convention as when [orienting the flight controller](../config/flight_controller_orientation.md#calculating-orientation).
```
hamishwillee, line 14: @sfuhrer It says here that the compass must be mounted in a standard orientation, and below you can see it says "You must mount the compass in a supported orientation!", and also that the orientations are _autodetected_.

BUT if you look at the `CAL_MAG1_ROT` there is option for a custom orientation. Further, if you look at the [advanced FC orientation](https://docs.px4.io/main/en/advanced_config/advanced_flight_controller_orientation_leveling.html) it looks like the FC allows more than these orientations, so it is possible that the mag does too - for example via `CAL_MAG1_PITCH`

So my assertion is that you don't have to use the defined angles - right? Is it still true though that if you don't use those angles the correct orientation would be auto detected?

en/SUMMARY.md
```diff
@@ -168,22 +158,22 @@
...
-  * [Bootloader Update](advanced_config/bootloader_update.md)
-    * [Bootloader Update FMUv6X-RT via USB](advanced_config/bootloader_update_v6xrt.md)
-* [Hardware (Drones&Parts)](hardware/drone_parts.md)
-  * [Complete Vehicles](complete_vehicles/README.md)
+
+* [Hardware Selection & Setup](hardware/drone_parts.md)
   * [Flight Controllers (Autopilots)](flight_controller/README.md)
+    * [Mounting the Flight Controller](assembly/mount_and_orient_controller.md)
+    * [Updating Firmware](config/firmware.md)
+    * [Flight Controller/Sensor Orientation](config/flight_controller_orientation.md)
```
hamishwillee, line 171: FYI, note orientation would ideally be after "mounting". However it is here, because it is config, which must be done after firmware update.

en/assembly/mount_and_orient_controller.md
```diff
@@ -10,7 +10,7 @@ The controller should be placed on the frame top-side up, oriented so that the a
 ![FC Orientation](../../assets/qgc/setup/sensor/fc_orientation_1.png)
```
hamishwillee, line 10: Does position matter, or just orientation? I.e. does it need to be near COG?

What about optical flow? That is relative to COG according to docs, but is that always correct, or does it depend on FC position?

en/assembly/mount_gps_compass.md
sfuhrer: The auto detection as well as the custom angle are fairly new features, and the docs are outdated here.
The flow is now that you only have to set the autopilot rotation, and then do a mag cal. The calibration then detects the mag orientation and sets it. AFAIK you could then override that either by setting the _ROT yourself to the correct value (e.g. "pitch 90"), or set CAL_MAG1_PITCH, CAL_MAG1_ROLL and CAL_MAG1_YAW, which then automatically sets CAL_MAG1_ROT to "Custom Euler Angle". @bresch correct?

en/assembly/mount_and_orient_controller.md
sfuhrer: It ideally is close to the CG, and otherwise the EKF_IMU_POS_X/Y/Z should be set.

en/SUMMARY.md
hamishwillee: Yes. Though I was thinking we might duplicate even more - and have a per vehicle type version of  https://docs.px4.io/main/en/assembly/quick_start_pixhawk6x.html#wiring-chart-overview with links to setup for the bits that are per-vehicle.

I won't know until we get to that point and I try some alternatives. But ideally we're trading off as vehicle specific as possible with some duplication.

en/assembly/mount_and_orient_controller.md
hamishwillee: This is now fixed - there is a "position" section below (I did in another PR and rebased).

en/assembly/mount_gps_compass.md
hamishwillee: I'm going to try fix this in https://github.com/PX4/PX4-user_guide/pull/3107 and then rebase. Don't want to block on this.

en/SUMMARY.md
```diff
@@ -339,18 +287,75 @@
...
+        * [3DR Telemetry Wifi (Discontinued)](telemetry/3dr_telemetry_wifi.md)
+      * [Microhard Serial Telemetry Radio](telemetry/microhard_serial.md)
+        * [ARK Electron Microhard Serial Telemetry Radio](telemetry/ark_microhard_serial.md)
+        * [Holybro Microhard P900 Telemetry Radio](telemetry/holybro_microhard_p900_radio.md)
+      * [CUAV P8 Telemetry Radio](telemetry/cuav_p8_radio.md)
+      * [HolyBro XBP9X - Discontinued](telemetry/holybro_xbp9x_radio.md)
+    * [FrSky Telemetry](peripherals/frsky_telemetry.md)
+    * [TBS Crossfire (CRSF) Telemetry](telemetry/crsf_telemetry.md)
+    * [Satellite Comms (Iridium/RockBlock)](advanced_features/satcom_roadblock.md)
+  * [Power Systems](power_systems/README.md)
```
hamishwillee, line 330: Note, still need to add the configuration stuff to power systems

en/SUMMARY.md
```diff
@@ -367,6 +369,10 @@
         * [Realsense T265 Tracking Camera (VIO)](peripherals/camera_t265_vio.md)
     * [Video Streaming](companion_computer/video_streaming.md)
       * [Video Streaming using WFB-ng Wifi (Long range)](companion_computer/video_streaming_wfb_ng_wifi.md)
+  * [Static Pressure Buildup](advanced_config/static_pressure_buildup.md)
+  * [Serial Port Configuration](peripherals/serial_configuration.md)
+  * [PX4 Ethernet Setup](advanced_config/ethernet_setup.md)
```
hamishwillee, line 374: @sfuhrer I can't find ethernet and static pressure build up in your modified guide - does this location at the end of Hardware Setup and Configuration make sense to you?

hamishwillee:
@sfuhrer This has become very large, and I think I need to merge it to progress. I don't mind if you don't review it "properly" but it would be good to familiarise yourself with what has been done.

This "almost" maps your guide. In some cases I haven't created top level docs and others I have. For example, accelerometer now has a parent page (that feels bare) but I didn't put Cable Wiring under wiring, because what other types are there :-).

This still omits RC/Joystick, which I am thinking about. 

Do you recall where you thought autotune and airframe setup should go?

Note that even though much has moved into Hardware, the overall structure is the same as before from an entry point point of view. That's because the old docs still exist and link to the moved docs. That is a good thing. We don't yet have the full vehicle specific stuff in place.

I hope to merge next week "come what may".

sfuhrer:
> @sfuhrer This has become very large, and I think I need to merge it to progress. I don't mind if you don't review it "properly" but it would be good to familiarise yourself with what has been done.

Sure, makes sense, it IS hard to keep the overview here!

> Do you recall where you thought autotune and airframe setup should go?

I have a feeling that I thought both should be not stand-alone sections but should belong within the MC part and the FW part seperately (so MC airframe selection, MC autotuning, FW airframe selection. FW autotuning). VTOL autotuning should simply link to the MC and FW ones.

PR #3070, opened 2024-02-22
ulog: a number of clarifications

bkueng:
- add valid characters for names
- remove unused non-uint64_t timestamp types
- https://github.com/foxglove/studio as supported ulog reader (https://docs.foxglove.dev/docs/connecting-to-data/local-data/#supported-formats)


hamishwillee (approved):
Thanks very much.

PR #3073, opened 2024-02-24
Update generic_dev_tools.md

callpri:
Codelobster editor is added


hamishwillee:
Thanks, but I'm not gong to accept this. Firstly because all changes must be made to the English topic and then propagate to other languages automatically (this change is to the Korean tree). Secondly because there is no evidence that code lobster is in any way a reasonable or popular choice for PX4 development - it is just a distraction. Unless there is a good reason not to, Visual Studio Code should be preferred.

PR #3084, opened 2024-03-04
fix  mavros_installation.md code block display

wp19991:
The format of the sh code block is incorrect, resulting in webpage display errors

fix


hamishwillee:
Thank you very much. I'm closing this, but I have fixed it in our translation service Crowdin. You can find more about the process in https://docs.px4.io/main/en/contribute/translation.html, but basically if we take fixes in Github on translations then there will be a clash with Crowdin.
You can join the project to make fixes or just do what you did here - A PR like this makes it easy for me to see the broken parts.

PR #3101, opened 2024-03-13
Add info about offsets to use for compass, flight stack

hamishwillee:
This adds info about the mounting offsets for GNSS and FC/IMU as discussed in https://discord.com/channels/1022170275984457759/1212732679011700748

### DIFF
--- a/en/assembly/mount_and_orient_controller.md
+++ b/en/assembly/mount_and_orient_controller.md
@@ -1,6 +1,5 @@
 # Mounting the Flight Controller
 
-
 The flight controller should be placed on the frame as close as possible to the centre-of-gravity (CoG), top-side up, and oriented so that the _heading mark arrow_ points towards the front of the vehicle.
 [Vibration isolation](#vibration-isolation) is often needed, and you should follow the manufacturer recommendations.
 If mounted in this way, no further PX4 configuration is required.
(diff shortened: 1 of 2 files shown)


en/assembly/mount_and_orient_controller.md
```diff
@@ -13,6 +18,14 @@ The controller should be placed on the frame top-side up, oriented so that the a
 If the controller cannot be mounted in the recommended/default orientation (e.g. due to physical constraints) you will need to configure the autopilot software with the orientation that you actually used: [Flight Controller Orientation](../config/flight_controller_orientation.md).
 :::
 
+## Position
+
+The flight controller should be placed on the frame as close as possible to the centre-of-gravity.
+
+If you can't mount the controller in this position, then you can [configure](../advanced_config/parameters.md) the following parameters to set offset relative to the CoG: [EKF2_IMU_POS_X](../advanced_config/parameter_reference.md#EKF2_IMU_POS_X), [EKF2_IMU_POS_Y](../advanced_config/parameter_reference.md#EKF2_IMU_POS_Y), [EKF2_IMU_POS_Z](../advanced_config/parameter_reference.md#EKF2_IMU_POS_Z) (for the default EKF2 estimator).
+
+Note that if you don't set these offsets then EKF2 position/velocity estimates will be at the IMU location rather that at the CoG.
```
hamishwillee, line 27: @bresch What would you "see" in this case - i.e how big a difference does it make? Are there conditions in which it would make a bigger difference?

en/assembly/mount_gps_compass.md
```diff
@@ -21,3 +21,7 @@ You must mount the compass in a supported orientation!
 If you mount the compass at an orientation that isn't supported, for example `Yaw 30`, PX4 will detect the closest supported value.
 This will result in errors/warnings, even if the calibration appeared to succeed.
 :::
+
+## Position
+
+In order to compensate for the relative motion between the receiver and the CoG, [configure](../advanced_config/parameters.md) the following parameters to set the offsets: [EKF2_GPS_POS_X](../advanced_config/parameter_reference.md#EKF2_GPS_POS_X), [EKF2_GPS_POS_Y](../advanced_config/parameter_reference.md#EKF2_GPS_POS_Y) and [EKF2_GPS_POS_Z](../advanced_config/parameter_reference.md#EKF2_GPS_POS_Z).
```
hamishwillee, line 27: @bresch same questions for GPS - What would you "see" given an offset - say of 10cm or 2m - i.e how big a difference does it make? What are the conditions in which it would make a bigger difference? 

My guess is that this means you're true position will be off by the amount in the offset. Whether that matters depends on how accurate you expect that position to be - I imagine it would normally not matter too much?

en/assembly/mount_and_orient_controller.md
bresch: You'll see the difference is position control mode for example. If the IMU is in front of the c.g., a pitching motion (around the c.g.) will result in a reduction of the altitude estimate (which is correct, the IMU goes down); as a reaction, the altitude controller will give more thrust to compensate for it. The amplitude depends on how far the IMU is located from the c.g. and can be negligible in some cases but it's still some unneeded control effort that is constantly applied.

If the offsets are specified, a pure pitch motion would not create any change in the altitude estimate so there will be less parasitic corrections.

en/assembly/mount_gps_compass.md
hamishwillee: That's awesome - thanks!

hamishwillee:
@bresch I've slightly reworded your comments and added them as details boxes (boxes that expand when you click them). So now you get a high level explanation "you might get oscillations" and then you can open the box for the reasons. I THINK I didnt change your meaning, but if you could quickly scan the new version would be good. 

I'm going to merge on the assumption this is correct.

PR #3107, opened 2024-03-14
Remove note that you must use supported compass orientation

hamishwillee:
This removes a note from the compass mounting docs stating that you have to use the supported orientations.

The orientations can now be tuned by setting `CAL_MAGn_ROT` to the "Custom Euler Angle" option and then setting the orientation using `CAL_MAGn_ROLL`, `CAL_MAGn_PITCH`, `CAL_MAGn_YAW` to set the actual orientation.

This is draft because it is not clear WHEN you would need to do this. Specifically compass tuning will automatically select the right  "standard" `CAL_MAGn_ROT` during calibration, but it isn't clear whether it will recognise that the standard setting is wrong, and try set  `CAL_MAGn_ROT` to the "Custom Euler Angle" and the other values for you?

It might be that if you know the values are not right you have to set them yourself. Or that you set to the closest normal rotation and then tune. 

Can you confirm the right process @bresch  or @sfuhrer ?

EDIT NOte, I see that @sfuhrer has partially answered in https://github.com/PX4/PX4-user_guide/pull/3068#discussion_r1515759817 :

> The auto detection as well as the custom angle are fairly new features, and the docs are outdated here.
The flow is now that you only have to set the autopilot rotation, and then do a mag cal. The calibration then detects the mag orientation and sets it. AFAIK you could then override that either by setting the _ROT yourself to the correct value (e.g. "pitch 90"), or set CAL_MAG1_PITCH, CAL_MAG1_ROLL and CAL_MAG1_YAW, which then automatically sets CAL_MAG1_ROT to "Custom Euler Angle". @bresch correct?

So that indicates, but does not "confirm" that if you want a custom euler angle you need to set it explicitly - right?

### DIFF
--- a/en/assembly/mount_gps_compass.md
+++ b/en/assembly/mount_gps_compass.md
@@ -15,13 +15,6 @@ The orientation follows the same frame convention as when [orienting the flight 
 If you're using the normal [Compass Calibration](../config/compass.md) process (with parameter [SENS_MAG_AUTOROT](../advanced_config/parameter_reference.md#SENS_MAG_AUTOROT) enabled), the orientation should be detected automatically.
 Otherwise you can directly select the appropriate value in [CAL_MAGn_ROT](../advanced_config/parameter_reference.md#CAL_MAG1_ROT) for up to three compasses.
 
-:::warning
-You must mount the compass in a supported orientation!
-
-If you mount the compass at an orientation that isn't supported, for example `Yaw 30`, PX4 will detect the closest supported value.
-This will result in errors/warnings, even if the calibration appeared to succeed.
-:::
-
 ## Position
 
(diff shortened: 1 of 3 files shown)


sfuhrer (approved):

bresch:
During the mag calibration, the autopilot attempts to find the orientation of the external magnetometer using the internal one. The algorithm can only find rotations that are multiples of 45 degrees (e.g.: if the real orientation is yaw = 22.5deg, the algorithm will either converge on 0 or 45 degrees). The user should ideally verify the detected rotation.
If the rotation is not part of the "standard set", one can set `CAL_MAGx_ROLL/PITCH/YAW` to a custom value (using yaw-pitch-roll intrinsic Euler angles). When doing this, `CAL_MAGx_ROT` is automatically set to "custom euler angle".

hamishwillee:
Thanks @bresch @sfuhrer - this has been updated appropriately (and I am happy with it). But happy for a follow on review from you guys :-)

Note that I checked that setting a custom rotation does NOT disable the auto mag detection. It probably should.

PR #3139, opened 2024-03-27
Hardware: ModalAI added VOXL 2

lululauren101:
I added a link to VOXL 2 under one of the sections.
cc: @mrpollo

### DIFF
--- a/en/companion_computer/README.md
+++ b/en/companion_computer/README.md
@@ -28,6 +28,7 @@ The following integrated companion computer/flight controller systems use manage
 They are listed here as they can be updated with "vanilla" PX4 firmware for testing/rapid development.
 
 - [Auterion Skynode](../companion_computer/auterion_skynode.md)
+- [ModalAI VOXL 2](https://docs.modalai.com/voxl-2/)
 
 ## Companion Computer Options
 


en/companion_computer/README.md
```diff
@@ -28,6 +28,7 @@ The following integrated companion computer/flight controller systems use manage
 They are listed here as they can be updated with "vanilla" PX4 firmware for testing/rapid development.
 
 - [Auterion Skynode](../companion_computer/auterion_skynode.md)
+- [ModalAI VOXL 2](https://docs.modalai.com/voxl-2/)
```
hamishwillee, line 31: @lululauren101 This is listed below under `## Companion Computer Options`, while this section is for systems where the flight controller and general computing parts are pre-integrated. 
Which group should VOXL2 be in? If it is this one, then we should remove from below.

lululauren101: Hi Hamish, we believe that VOXL 2 qualifies as hardware options for both sections, so would be good if we can list it under both companion computers and managed integrated systems.

Thanks! 
cc: @modalai-sweetc @mrpollo

mrpollo: @hamishwillee it's possible to use the VOXL 2 as a companion computer and as an integrated FC+CC the links for each are also different pointing to relevant sections hope this helps clarify

hamishwillee: You're right - it fits here if PX4 is running on DSP, and as a stand-alone companion computers if it is externally attached to a flight controller. Both cases documented in your link.

Thanks for the clarification/education. Merging.

hamishwillee: Ah @mrpollo - thanks. Didn't see your note, was checking at the same time.

PR #3200, opened 2024-05-01
MAVLink camera updates

hamishwillee:
This splits out MAVLink cameras and FC-connected cameras into their own documents. Their setup is quite different, and the current story is unclear.

This fell out of the discussion here: https://discuss.px4.io/t/connect-siyi-camera-module-to-px4/37782/3


en/camera/mavlink_camera.md
```diff
@@ -0,0 +1,88 @@
+# MAVLink Cameras
```
hamishwillee, line 1: @julianoes Can you sanity check this document and the one above it.

This attempts to capture what you do for a MAVLink camera more clearly. I think it is much better, but I am biased :-) The thing is, with this you can just keep on adding more links to other camera managers and so on - its a scalable doc. 

Also please confirm you are OK with the way we refer to your camera manager. I have tried to capture that it is "community" not something you will necessarily fix, and that it is a good example of MAVSDK. It would be great to have a guide sometime on how you constructed it.

en/camera/mavlink_camera.md
```diff
@@ -0,0 +1,88 @@
...
+The camera manager runs on a companion computer and bridges between the MAVLink camera protocol interface and the camera's native interface.
+
+Generic/extensible camera managers:
+
+- [MAVLink Camera Manager](https://github.com/mavlink/mavlink-camera-manager) - Extensible cross-platform MAVLink Camera Server built on top of GStreamer and Rust-MAVLink.
+- [Dronecode Camera Manager](https://camera-manager.dronecode.org/en/) - Adds Camera Protocol interface for cameras connected to Linux computer.
+
+Camera-specfic camera managers:
+
+- [SIYI A8 mini camera manager](https://github.com/julianoes/siyi-a8-mini-camera-manager) - Community-supported MAVSDK-plugin based camera manager for the [SIYI A8 mini](https://shop.siyi.biz/products/siyi-a8-mini) (includes tutorial).
```
julianoes, line 83: I probably wouldn't even write "community-supported" because that would imply it's supported when it might not be.

julianoes (commented):
I left some comments. All in all it's good but cameras with PX4 are just very confusing and likely won't work as expected, just because all setups are ever so slightly different. Best would be to have a few concrete examples for a trigger based camera as well.

en/camera/mavlink_camera.md
```diff
@@ -0,0 +1,88 @@
...
+First configure the PX4 camera driver to enable the MAVLink camera backend, and the triggering mode to capture on command in survey missions.
+
+You can do this using _QGroundControl_:
+
+- Open [Vehicle Setup > Camera](https://docs.qgroundcontrol.com/master/en/qgc-user-guide/setup_view/camera.html#px4-camera-setup).
+- Set the values as shown:
+
+  ![Camera Setup Screen - Trigger mode and interface for MAVLink](../../assets/camera/mavlink_camera_settings.png)
+
+- Select **Apply and Reboot** to apply the parameters.
```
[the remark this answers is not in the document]
hamishwillee: Are you sure? I think that PX4 requires you to reboot if you modify the parameter - it doesn't care what value you set it to.

en/camera/mavlink_camera.md
```diff
@@ -0,0 +1,88 @@
...
+
+::: info
+You can also [set the parameters directly](../advanced_config/parameters.md) and then reboot the flight controller:
+
+- [TRIG_MODE](../advanced_config/parameter_reference.md#TRIG_MODE) — `4`: Distance based, on command (Survey mode)
+- [TRIG_INTERFACE](../advanced_config/parameter_reference.md#TRIG_INTERFACE) — `3`: MAVLink (forward via MAV_CMD_IMAGE_START_CAPTURE)
+
+:::
+
+This configuration ensures that PX4 properly handles the [camera trigger commands](/camera/index.md#commands-supported-by-all-cameras) found in missions by re-emitting them as MAVLink commands to the camera, and forwards the same commands if received from a ground station.
```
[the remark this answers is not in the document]
hamishwillee: Yes! Essentially the camera driver just handles how PX4 handled the camera trigger commands in missions or if received from a ground station. Everything else is done by forwarding. So in fact if you had a mission processor outside of the FC you wouldnt need to do this for a mavlink camera.

en/camera/mavlink_camera.md
hamishwillee: OK. I guess you can be explicit about the support you provide on the site.

en/camera/mavlink_camera.md
```diff
@@ -0,0 +1,88 @@
...
+
+  ![Camera Setup Screen - Trigger mode and interface for MAVLink](../../assets/camera/mavlink_camera_settings.png)
+
+- Select **Apply and Reboot** to apply the parameters.
+
+::: info
+You can also [set the parameters directly](../advanced_config/parameters.md) and then reboot the flight controller:
+
+- [TRIG_MODE](../advanced_config/parameter_reference.md#TRIG_MODE) — `4`: Distance based, on command (Survey mode)
+- [TRIG_INTERFACE](../advanced_config/parameter_reference.md#TRIG_INTERFACE) — `3`: MAVLink (forward via MAV_CMD_IMAGE_START_CAPTURE)
```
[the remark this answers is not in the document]
hamishwillee: Good catch. I copied that from old docs.
My assumption is that if this is in a mission the mission would be rejected. Really PX4 should process this as well and re-emit.

en/camera/index.md
```diff
@@ -1 +1,53 @@
...
+
+PX4 can be used with with two classes of cameras:
+
+- [MAVLink cameras](../camera/mavlink_camera.md), which are controlled using the [MAVLink camera protocol](https://mavlink.io/en/services/camera.html).
+
+  These are suitable for most use cases, because they allow access to any feature of the camera that has been exposed via the protocol, such as image and video capture and control over where the data is saved, video streaming to a ground station, control over zoom and focus, selecting between infrared and visible light feeds, and so on.
+  
+  However, since most cameras do not support the MAVLink camera protocol, you'll generally need a camera manager running on a companion computer to interface between MAVLink and the camera's native protocol.
+
+- Cameras that are attached directly to the flight controller outputs.
```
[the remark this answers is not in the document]
hamishwillee: It does. It's a bullet attached to the parent statement, but that isn't obvious unless you look at the rendered version.

So it is:

---

PX4 can be used with with two classes of cameras:

- Cameras that are attached directly to the flight controller outputs.

---

en/camera/mavlink_camera.md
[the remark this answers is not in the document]
julianoes: Yes to reboot. I meant the whole paragraph.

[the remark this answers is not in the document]
hamishwillee: I don't understand. Would this be right if I just removed this bullet point?

en/camera/mavlink_camera.md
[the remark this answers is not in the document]
hamishwillee: > TRIG_INTERFACE and the command MAV_CMD_IMAGE_START_CAPTURE are pretty much separate. I think the command is usually used by the camera directly, and not via PX4.

Note though, this is a "bug" in PX4. We basically have multiple messages to start capture, but we've only implemented a few of them in the driver. We could implement all of them, and then start deprecating the old ones.

en/camera/index.md
```diff
@@ -1 +1,99 @@
 # Cameras
```
hamishwillee, line 1: @julianoes I did a big update today, but I've now had yet another epiphany: **you should not need the camera driver** for a MAVLink camera integration. 

PX4 basically handles mavlink camera protocol commands in missions (and the older commands) by emitting the mission item as a command addressed to component id 100. It handles incoming commands by forwarding them to the camera.

So what does the driver do in this scenario? Essentially for MAVLink camera backend it just emits [CAMERA_TRIGGER](https://mavlink.io/en/messages/common.html#CAMERA_TRIGGER) for some, but not all camera trigger events.
That should be done by the camera. 

So what we should do is make sure that the camera protocol requires [CAMERA_TRIGGER](https://mavlink.io/en/messages/common.html#CAMERA_TRIGGER) to be emitted on image capture, and then you wouldn't need the driver at all.

Am I right?

en/simulation/index.md
```diff
@@ -269,7 +269,7 @@ PX4 connects/integrates with this camera in _exactly the same way_ as it would w
 1. [TRIG_INTERFACE](../advanced_config/parameter_reference.md#TRIG_INTERFACE) must be set to `3` to configure the camera trigger driver for use with a MAVLink camera
    :::tip
    In this mode the driver just sends a [CAMERA_TRIGGER](https://mavlink.io/en/messages/common.html#CAMERA_TRIGGER) message whenever an image capture is requested.
-   For more information see [Camera](../peripherals/camera.md).
+   For more information see [Cameras Connected to Flight Controller Outputs](../camera/fc_connected_camera.md).
```
hamishwillee, line 272: TBD. This might need to be completely removed if we're not using the trigger interface/don't need it. I don't think we should.

en/camera/index.md
```diff
@@ -1 +1,99 @@
...
+  Photo capture can be triggered from a flight controller pin, and in some cases an additional pin can toggle camera modes.
+
+Both approaches allow image capture in missions and when commanded by a ground station, and are suitable for mapping and survey applications.
+MAVlink cameras support many more functions, such as streaming and video capture, and can be used for a broader range of applications.
+
+## Overview
+
+### Mission Camera Commands
+
+PX4 _allows_ the following camera commands in missions (i.e. does not reject the mission if they are included):
```
hamishwillee, line 27: Note to me. If I'm right that Mavlink camera doesnt need the camera driver, then this should be modified. Part of that would be to explain that PX4 just emits all camera commands, both old and new, and that they are different and used by different cameras. Lets see.

en/camera/fc_connected_camera.md
```diff
@@ -1,24 +1,82 @@
...
-This timestamp can be used for several applications, including: timestamping photos for aerial surveying and reconstruction, synchronising a multi-camera system or visual-inertial navigation.
+Whenever a camera is triggered, the MAVLink [CAMERA_TRIGGER](https://mavlink.io/en/messages/common.html#CAMERA_TRIGGER) message is published containing the current session's image _sequence number_ and the corresponding timestamp.
+This timestamp can be used for several purposes, including: timestamping photos for aerial surveying and reconstruction, synchronising a multi-camera system or visual-inertial navigation.
 
-Cameras can also (optionally) signal PX4 at the exact moment that a photo/frame is taken using a camera capture pin.
+Cameras can also (optionally) signal PX4 at the exact moment that a photo/frame is taken using a [camera capture pin](#camera-capture).
 This allows more precise mapping of images to GPS position for geotagging, or the right IMU sample for VIO synchronization, etc.
```
hamishwillee, line 14: Below there is a section on geotagging - using QGC to apply the camera data based on timestamps. Does this work for the new camera API as well?

en/camera/mavlink_camera.md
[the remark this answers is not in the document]
hamishwillee: Closing this. IMO the driver is only used for the "old" mavlink interface. The new camera interface doesn't care about the driver. Asked for confirmation elsewhere.

en/camera/index.md
julianoes: > Are there old MAVLink cameras that just support the old API for which using the backend makes sense? 

I think most simple triggering cameras are like that. I'm blanking on the name of an example that I have in mind.

julianoes: 1. Depends on the definition of a "MAVLink camera". It just comes down to what the camera supports. I'd assume most implement the ArduPilot camera messages like DIGICAM_...
2. Don't know.
3. No idea. It's probably just incomplete.
4. I don't think it's deliberate. Was just always an incomplete/tricky topic.

hamishwillee:
@julianoes I've finished this round of update and I plan to merge. Would love a follow on sanity update if you have time. 

This is now split into four camera docs - the two mavlink protocols and the FC connected have hardware setup/usage guides. There is also a "camera architecture/integration" doc. The first ones are pretty much - if you have this camera this is what you do, while the integration one explains how it works, and points to some code (the stuff you helped me find).

We don't perhaps need the integration one, but given how long it took to work out how it all works, this might same some time next time.

Some interesting points:
- MAVLink camera protocol 2 trigger events are not logged in PX4, so you can't use QGC geotagging. PX4 should log incoming CAMERA_CAPTURE_INFO messages if we want to support this using the same mechanism.
- Camera logging is a little broken, in that every capture is logged, not just the "most relevant": https://github.com/PX4/PX4-Autopilot/pull/23103
- The MAVLink 1 camera integration doesn't strictly need camera trigger etc - that essentially just provides logging for this particular type of camera. You could send commands direct to the camera.
  It was this way almost certainly because then you can just always send commands to the autopilot for both FC and old-Mavlink cameras.

PR #3201, opened 2024-05-01
Broken links in docker.md

federicociresola:
I found some broken links pointing to the PX4 docker repository https://github.com/PX4/PX4-containers/tree/master?tab=readme-ov-file#container-hierarchy


hamishwillee (approved):
Thanks very much @federicociresola - good catch

PR #3209, opened 2024-05-13
Updated the VOXL 2 flight controller page to include latest information

katzfey:
Added update about mainline support and newer VOXL 2 based developer kits


hamishwillee (approved):
Thanks @katzfey

PR #3245, opened 2024-06-12
Deprecate jMAVSim

hamishwillee:
JMAVSim doesn't work on the lasted mainline on Ubuntu or WSL on Windows, and presumably won't work on v1.15. There is no one making any particular effort to maintain it, and I know there is a strong argument to drop all support and focus on gazebo.

What this does is "bite the bullet" and deprecate JMAVSim in the docs

- removes it being mentioned in the main flow of developer setup instructions.
- Replaces where it is mentioned with Gazebo/Gazebo classic
- pushes the documentation into the Community Supported Simulators.
- Adds a release note about the removal.
- While here I simplified some of the instructions around setup on ubuntu to push the RaPi and ROS2 instructions into their own docs - leaving this doc just about setup of Gazebo/Pixhawk. Decentralizing is a little bit harder if you work on some other FC type, but is much better for the 80% working on Pixhawk-series and simulators.

@mrpollo @dagar Any disagreements/suggestions? In this release should we actually remove the installation of jmavsim from the ubuntu.sh installer script too?

Are we any closer to also deprecating Ubuntu 18 and ROS (1)?


hamishwillee:
@romain-chiap Howdy! This PR moves JMAVSim from part of the PX4 core supported simulators to "community supported", leaving Gazebo/Gazebo-Classic the only devteam-maintained simulator. In the short term nothing will actually change.

I'm contacting you because it is possible in future that JMAVSim will break or be retired, and I don't want to also break [SIH](https://docs.px4.io/main/en/sim_sih/).

Is it possible to update SIH so that it works using Gazebo on Ubuntu 22+ as the visualizer? Or possibly without any visualizer at all?

hamishwillee:
Merging, but open to further discussion.

romain-chiap:
Hey, thanks I appreciate.

Short answer: the SIH can already work without visualization.

To make it work with Gazebo we need to run Gazebo as a visualizer only (i.e. disable its physics engine). The SIH sends periodic mavlink message [HIL_STATE_QUATERNION](https://mavlink.io/en/messages/common.html#HIL_STATE_QUATERNION). If gazebo already uses this mavlink message to display the vehicle that can be pretty straight forward.

I've never opened the code for gazebo. Can you ping in someone knowledgable to this conversation?

hamishwillee:
Awesome @Jaeyoung-Lim is the best person to talk to about Gazebo-Classic - and should be able to recommend the best people for Gazebo (?). 

Jae, can you comment on ^^^ - essentially how can we run Gazebo(s) as a visualizer for SIH?

romain-chiap:
Thanks Jae!

How can I do that given it will be my first time coding in gazebo?
Can you point me to an example plugin? and to the structure/function for the pose?

Jaeyoung-Lim:
@romain-chiap I was more implying that we should not implement the visualizer in gazebo...

 but if you want to you can go through some gazebo tutorials and check the plugins we have integrated into gazebo: https://github.com/PX4/PX4-SITL_gazebo-classic/tree/main/src

romain-chiap:
Aha, thanks for the clarification.

In this case I'm happy to leave it as it is. 
@hamishwillee, I will update the docs

hamishwillee:
@Jaeyoung-Lim Can you provide pointers for working around Gazebo - GZ, or point to someone else who might be able to help with this? Docs updated by @romain-chiap to note that a visualizer is optional.

Jaeyoung-Lim:
@hamishwillee @romain-chiap I am not sure if I message is getting through.

I think we would want some sort of visualizer to visualize SIH, but I personally think gz is the wrong software to do so.
@romain-chiap Would you be able to look into somekind of 3D visualizer tool that is light weight enough that we can use for SIH?

romain-chiap:
For some projects, I use Xplane for visualization from the SIH. Using the plugin from nasa [Xplane connect](https://github.com/nasa/XPlaneConnect), this is not leightweight, but the free version allows for 10min of flights, and the people really love the graphics.

QGC used to be able to connect to Xplane in version 3. Can we imagine putting that back in place in the current version of QGC?

I feel I'm missing a piece of information here, why going for another visualizer when jmavsim is already working. Can we imagine supporting jmavsim as a visualizer for SIH only?

PR #3246, opened 2024-06-12
highlight the need for daily build of QGC

ayhamalharbat:
Emphasis that the Daily Build of QGC is needed.

### DIFF
--- a/en/ros2/px4_ros2_control_interface.md
+++ b/en/ros2/px4_ros2_control_interface.md
@@ -100,7 +100,7 @@ The above concepts provide a number of advantages over traditional [offboard con
 
 The following steps are required to get started:
 
-1. Make sure you have a working [ROS 2 setup](../ros/ros2_comm.md), with [`px4_msgs`](https://github.com/PX4/px4_msgs) in the ROS 2 workspace.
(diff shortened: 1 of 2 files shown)


hamishwillee:
1. WHY do you need the daily build? 
2. Care is needed here because the plan is to release PX4 v1.15 soon, and that will presumably come with a new QGC stable release.
3. You have made these changes to the `/tr` tree, which is a translation. Can you make them to the `en` tree? (they will then automatically sync to all other branches)

ayhamalharbat:
Thanks @hamishwillee 
1. As you mentioned, this is until we have a stable release that allows for dynamic control modes
2. I agree, but until then I thought it is worth it to be highlighted
3. Done!

I think that when we have a stable release of QGC it is also important to mention in these docs that the older versions will not work.

hamishwillee:
Thanks @ayhamalharbat , but I am going to close this.

My reason is that step 6 in https://docs.px4.io/main/en/ros2/px4_ros2_control_interface.html#px4-ros-2-control-interface already says to use QGC daily. When a new release comes out that supports this we'll need to update that.

PR #3247, opened 2024-06-13
Move ros2 topics

hamishwillee:
This moves all the ROS2 topics out from the /ros subfolder into the /ros2 folder. This makes organisation easier, and later retirement of ROS.

Note, I still plan to add redirects (next PR). Also in a following PR will actually split the built tree.

FYI @beniaminopozzan !


beniaminopozzan:
Thanks @hamishwillee!

PR #3260, opened 2024-06-19
Update index.md

romain-chiap:
update of the SIH documentation, jMAVSim getting deprecated


hamishwillee (approved):
Thanks @romain-chiap . I've done a minor subedit to make it clear that the visualizer is optional - you can fly with QGC. I removed the comment about JMAVSim being deprecated. Its stability is of course a worry, but the important thing here is know that SIH is useful without it.

Though I do hope we can visualize with Gazebo in future!

PR #3265, opened 2024-06-26
Add user guide for new ackermann module

chfriedrich98:
This PR adds a new section to the Rovers tab on the new ackermann rover module
- [x] https://github.com/PX4/PX4-Autopilot/pull/23024 
- [x] https://github.com/PX4/PX4-Autopilot/pull/23310
- [x] https://github.com/PX4/PX4-Autopilot/pull/23383
- [x] https://github.com/PX4/PX4-gazebo-models/pull/46

This is the start of an update on the whole rover (and later USV) documentation in PX4.


en/frames_rover/ackermann_rover.md
```diff
@@ -0,0 +1,201 @@
...
+:::
+
+To deal with the edge case that the line segment is outside the look ahead radius around the rover there are 2 steps:
+
+1. The look ahead is scaled to the crosstrack error, which creates an intersection point for the rover to drive towards.
+2. Scaling the look ahead too much can have undesirable effects such as the rover driving a huge circle to return to the path, rather then taking a straight line. For this purpose the maximum look ahead radius can be capped with the parameter [RA_LOOKAHEAD_MAX](#RA_LOOKAHEAD_MAX), after which the rover will drive directly towards the current waypoint until it gets close enough to the line segment again.
+
+To summarize, the following parameters can be used to tune the controller:
+
+| Parameter                                                                                                   | Description                             | Unit |
```
hamishwillee, line 93: FYI, note that generally we use a table to summarise the params, just as you have done. This is prefixed with an anchor, with the same name and links to the actual parameter.
Then in the rest of the doc we link to the table entry - you can see that above for `RA_LOOKAHEAD_MAX`. You don't have to link everything to the table if you want, but you should link the first instance, and other instances should be code style to highlight them.

This ensures that people can get the info they want without heading out of the context, but at most they will have one jump.

en/frames_rover/ackermann_rover.md
```diff
@@ -0,0 +1,201 @@
...
+| ----------------------------------------------------------------------------------------------------------- | --------------------------------------- | ---- |
+| <a id="RA_LOOKAHEAD_TUN"></a>[RA_LOOKAHEAD_TUN](../advanced_config/parameter_reference.md#RA_LOOKAHEAD_TUN) | Main tuning parameter                   | -    |
+| <a id="RA_LOOKAHEAD_MAX"></a>[RA_LOOKAHEAD_MAX](../advanced_config/parameter_reference.md#RA_LOOKAHEAD_MAX) | Maximum value for the look ahead radius | m    |
+| <a id="RA_LOOKAHEAD_MIN"></a>[RA_LOOKAHEAD_MIN](../advanced_config/parameter_reference.md#RA_LOOKAHEAD_MIN) | Minimum value for the look ahead radius | m    |
+
+#### Cornering Parameters
```
hamishwillee, line 99: Note, first letter of all words are capitalised. Generally "related" is unnecessary in the heading - they are "Cornering paramters".

en/frames_rover/ackermann_rover.md
```diff
@@ -0,0 +1,201 @@
...
+
+The module employs a special cornering logic causing the rover to "cut corners" to achieve a smooth trajectory.
+This is done by scaling the acceptance radius based on the corner the rover has to drive (for geometric explanation see [Cornering logic](#cornering-logic)).
+
+![image](https://github.com/PX4/PX4-Autopilot/assets/125505139/541935df-c10b-4ca9-a9cb-594312d1afac)
```
hamishwillee, line 104: These images should be pulled into the /assets/ path in this tree. Also a meaningful description - other than `image`.

hamishwillee:
@chfriedrich98 Appreciate that you're still draft, but I've done a subedit to the standard docs conventions, not all of which are documented in https://docs.px4.io/main/en/contribute/docs.html#style-guide . I also ran prettier.

When you move out of draft I'll give it a more thorough review. But on quick scan I think this looks excellent. About time we had some good docs for Rover.

A few things in passing - you cover flight modes here - we need to make this work with https://docs.px4.io/main/en/flight_modes_rover/

For mission mode it might be worth adding a dedicated page, but either way, would be good to list the commands that can be used in a mission, and how the system works with respect to when others are missing or present. For example, takeoff is a meaningless rover mission item, but it is also harmless. We should still make it clear whether that command would be ignored or fail a feasibility test.

We should also clarify what happens when the mission is paused (presumably vehicle just stops, but does it go into some kind of "hold mode" (undocumented) or is it in a paused mode "within mission mode".

When we get to boat we'll need to cover hold behaviour too - vehicles might follow a radius like a fixed wing, or they might head to centre of a hold circle and drift, returning to centre on exiting the radius, or something else.

Thanks again.

hamishwillee:
PS I assume this is in v1.15? In that case we will need a release note in https://docs.px4.io/main/en/releases/1.15.html

hamishwillee:
(That could go in before the rest of this).

chfriedrich98:
@hamishwillee Thank you for the review!
I'll start implementing your suggestions and re-request a review once I move out of draft.

> PS I assume this is in v1.15? In that case we will need a release note in https://docs.px4.io/main/en/releases/1.15.html

Yes, I'll have a look and open a seperate PR for the release notes entry.

hamishwillee:
> Nevermind, it is planned vor 1.16.

Great, then can you please, when you get to it, add a release note to the doc in `en/releases/main.md`

-

chfriedrich98:
@hamishwillee I added the realease notes and made some minor improvements. 
In regards to the flight modes: With the restrucutre of the USV support the different types of rovers no longer share the same module so the supported flight modes are different and each would have their own definitions. 
The goal with the restructure is to eventually remove support for the old rover module (which was used for all USV's) and with it most of the current rover related doc entries, which would include https://docs.px4.io/main/en/flight_modes_rover/.
Until then there is currently some conflicting information I tried to solve by adding this warning https://github.com/PX4/PX4-user_guide/pull/3265/files#diff-68836498fff5b72612f91d31bfa9e8c70a34bbed4bb0f9696e5816afe563f2c5R31-R34.

chfriedrich98:
This is the structure I have in mind:

USV (Experimental) (some basic introduction to USV support in PX4)
  - Rover (similar to https://github.com/PX4/PX4-user_guide/blob/main/en/frames_rover/index.md?plain=1#L1-L23)
	  - Ackermann (This PR)
		  - Specific hardware example (something like https://github.com/PX4/PX4-user_guide/blob/main/en/frames_rover/traxxas_stampede.md)
	  - Differential (https://github.com/PX4/PX4-user_guide/pull/3133)
		  - Specific hardware example
  - Boats
	  - ...
  - ...

And the USV tab could be on the same hieararchy level as Multicopter/planes and vtol.

hamishwillee:
@chfriedrich98 Thanks very much for the heads up. I'm now on another job until Wednesday, so I will look at this "properly" then. 

- I understand the situation is that we now have two modules for USVs - ackermann and the old module that is still used for differential rovers and boats. So right now we have certain behaviours for ackermann and the docs for other vehicles aren't right.
- I understand that you intend to have separate modules/modes. 

What if anything is planned to be shared, if anything? What I'm heading towards here is that I would expect that USVs would share modes such as manual and mission - since the behaviours of all are likely to be the same in most cases. However there might be some exceptions between UGVs and boats - for example hold behaviour for a boat can vary quite a lot. Are you "really" planning on having completely separate implementations even for missions?

If the intent is that these will be very separate and continue to be very separate then I very much like your proposal above https://github.com/PX4/PX4-user_guide/pull/3265#issuecomment-2222725717 - and I think we should work to move there earlier rather than later.

So essentially, my plan would be to move to your proposed structure now, but initially keep all of that down in experimental section until one of the USVs becomes "non-experimental".
Does that make sense, or do you feel it is too early?

chfriedrich98:
@hamishwillee 
> What if anything is planned to be shared, if anything? What I'm heading towards here is that I would expect that USVs would share modes such as manual and mission - since the behaviours of all are likely to be the same in most cases. However there might be some exceptions between UGVs and boats - for example hold behaviour for a boat can vary quite a lot. Are you "really" planning on having completely separate implementations even for missions?

Manuall/Mission are going to be very similar for all USVs , so I think it could make sense to have a USV Flight Modes section and mention there which USVs support which flight modes (and of course also in the respective USV docs).  @PerFrivik wrote a very nice flight modes documentation in #3133 and I would propose to use a simplified version of this that applies to all USVs in the general USV Flight modes tab and then expand on it in the individual modules. I think that would make sense since the very detailed explanation of the individual guidance algorithms is not _necessary_ for the user to understand to use it.

> So essentially, my plan would be to move to your proposed structure now, but initially keep all of that down in experimental section until one of the USVs becomes "non-experimental".
Does that make sense, or do you feel it is too early?

I think this PR could be merged on its own for now, since it is esentially a standalone file. If we do the same for differential rover  we would then have all the "pieces" ready for the restructure in a subsequent PR (this could be sinchronized with the PX4 PR that removes the old rover module).
During this transition we just need to clearly indicate that there is currently 2 different ways to run rovers.
Keeping the USVs in the experimental tab until they are more established makes sense to me.

en/frames_rover/ackermann_rover.md
```diff
@@ -0,0 +1,246 @@
...
+
+::: info
+The module does not require the ackermann geometry and will work with any front-steering rover.
+:::
+
+## Airframe Setup
+
+This section will go through the most basic setup that is necessary to start using the rover.
+
+1. To enable the module flash the PX4 rover build onto your flight controller.
```
hamishwillee, line 18: How? QGC does offer a master "alpha" build, but not one for Rover? How do I get it and how do I install it.

If the answer is "for now you must build it yourself, we should provide build instructions for creating this firmware.

en/frames_rover/ackermann_rover.md
```diff
@@ -0,0 +1,246 @@
...
+
+   ::: info
```
hamishwillee, line 20: I have pushed this info into firmware. It's clearly not recommended so we should make it clearly "a less preferred resort".

en/frames_rover/ackermann_rover.md
```diff
@@ -0,0 +1,246 @@
...
+   :::
+
+2. In the [Airframe](../config/airframe.md) configuration select the _Generic ackermann rover_:
+
+   ![QGC screenshot showing selection of the airframe 'Generic ackermann rover'](../../assets/config/airframe/airframe_ackermann_rover.png)
+
+   Select the **Apply and Restart** button.
+
+   ::: info
+   If this airframe does not show up in the UI, it can alternatively be selected by setting the [SYS_AUTOSTART](../advanced_config/parameter_reference.md#SYS_AUTOSTART) parameter to `50010`.
```
hamishwillee, line 38: We could do this, but normally we explain the QGC minimum build you need to get the ability to install the appropriate airframe - do you know what that is (I may run out of time before I try out latest daily).

en/frames_rover/ackermann_rover.md
```diff
@@ -0,0 +1,246 @@
...
+| -------------------------------------------------------------------------------------------------------- | -------------------------------------- | ----- |
+| <a id="RA_MAX_ACCEL"></a>[RA_MAX_ACCEL](../advanced_config/parameter_reference.md#RA_MAX_ACCEL)          | Limit on the acceleration of the rover | m/s^2 |
+| <a id="RA_MAX_STR_RATE"></a>[RA_MAX_STR_RATE](../advanced_config/parameter_reference.md#RA_MAX_STR_RATE) | Limit on the steering rate             | deg/s |
+
+:::warning
+The slew rates are not based on measurements but on assumed linear relation between the throttle input and [RA_MAX_SPEED](#RA_MAX_SPEED) or steering input and [RA_MAX_STR_ANG](#RA_MAX_STR_ANG) respectively.
+Therefore these two parameters have to be set for the slew rates to work!
+:::
+
+### Mission Parameters
```
hamishwillee, line 98: IN my opinion there is enough information here for us to create a mission mode topic for rover and pull all this information into it. Are you OK if I do that?

How much of this would be in common with differential rovers in future?

en/frames_rover/ackermann_rover.md
hamishwillee: ... and we _must_ update QGC to allow selection of firmwares. Really we need a custom build server and to pull the other options from there.

en/frames_rover/index.md
```diff
@@ -28,6 +28,10 @@ The supported frames can be seen in [Airframes Reference > Rover](../airframes/a
 
 ### Ackermann Steering Configuration
 
+:::warning
```
hamishwillee, line 31: How would you characterise the old ackermann vs new module in "v1.16"? Will the old module still exist in v1.16, or will it be removed, or will it be deprecated?

How about right now if someone is working on main? What should they use?

The reason I ask is that if the "old" ackermann module is going to be gone then I would work towards removing it from the docs now. We'd structure as differential and ackermann. I can do this, but I need to know what the intent is and how "firm" it is.

en/frames_rover/index.md
```diff
@@ -19,7 +19,7 @@ PX4 supports rovers with:
 
 - **Differential steering**: direction is controlled by moving the left- and right-side wheels at different speeds.
   This kind of steering commonly used on bulldozers, tanks, and other tracked vehicles.
-- **Ackermann steering**: direction is controlled by pointing wheels in the direction of travel ([ackermann geometry](https://en.wikipedia.org/wiki/Ackermann_steering_geometry) compensates for the fact that wheels on the inside and outside of the turn move at different rates).
+- [**Ackermann steering**](../frames_rover/ackermann_rover.md): direction is controlled by pointing wheels in the direction of travel.
```
hamishwillee, line 22: Shouldn't we be "Ackermann" everywhere, not this mix of "ackermann"?

en/flight_modes_rover/index.md
```diff
@@ -1,5 +1,10 @@
 # Flight Modes (Rover)
 
+::: warning
+This page contains flight mode information for differential rovers and the legacy ackermann rover implementation.
```
hamishwillee, line 4: I've called it legacy here, but I need to understand how long it will live. If it is likely gone in v1.16, then we could make this page `Drive Modes (Differential Rovers)`

hamishwillee:
Some of the documented params don't exist - probably changed names after you wrote this. YOu can see them listed by the flaw checker https://github.com/PX4/PX4-user_guide/pull/3265#issuecomment-2235743284

I fixed the ones I could be sure of, but you'll have to update the docs for the others appropriately.

en/frames_rover/index.md
hamishwillee: @chfriedrich98 This comment above relates to your comment:

> I think this PR could be merged on its own for now, since it is esentially a standalone file. I

I don't think it is standalone, though you're quite right we don't need to fully split everything up now. But if we know that this module is going to be the only thing we would recommend for Ackermann right now, we can very much restructure this to hide the old stuff.

If that is the case, I would also perhaps suggest removing the old airframe so people can't select it.

en/frames_rover/ackermann_rover.md
chfriedrich98: Yes, for now the rover build (will probably renamed to USV build)  has to be built by the user. I'll add a section on how to do this.
As a side note: All the new USV modules have that same problem because we can't enable them by default on most boards due to flash constraints.

> ... and we must update QGC to allow selection of firmwares. Really we need a custom build server and to pull the other options from there.

Definitely, that would make this much easier for the end user.

en/frames_rover/ackermann_rover.md
chfriedrich98: Makes sense to me!

en/frames_rover/ackermann_rover.md
chfriedrich98: The pure pursuit part is shared, all new USV modules that are currently being worked on use that algorithm.
This also includes the [Pure Pursuit Algorithm](https://github.com/PX4/PX4-user_guide/pull/3265/files#diff-6ad9e9c11f97f5ef09c780ba38fb6c3242db85a2fcdca8e77ed1f28ab387760eR199-R220) section in this PR.
The cornering part is ackermann specific (at least for now).

> IN my opinion there is enough information here for us to create a mission mode topic for rover and pull all this information into it. Are you OK if I do that?

That works for me.

en/frames_rover/index.md
chfriedrich98: > How would you characterise the old ackermann vs new module in "v1.16"? Will the old module still exist in v1.16, or will it be removed, or will it be deprecated?
How about right now if someone is working on main? What should they use?

The old module uses a guidance algorithm that was designed for fixed-wing and the module was used for all types of USV's but didn't do a great job with any of them. This is why we started creating dedicated modules for the different USV's. The feedback I have received so far is that the new module works significantly better for ackermann rovers than the old one, so I would definitely recommend using the new one.

> The reason I ask is that if the "old" ackermann module is going to be gone then I would work towards removing it from the docs now. We'd structure as differential and ackermann. I can do this, but I need to know what the intent is and how "firm" it is.

I want to remove the old module once the differential and boat modules are properly done, but there might be some pushback on that, so I am not sure yet.

> I don't think it is standalone - it is confusing as is. Though you're quite right we don't need to fully split everything up now. But if we know that this module is going to be the only thing we would recommend for Ackermann right now, we can very much restructure this to hide the old stuff.

I agree it is quite confusing like this, hiding the old stuff sounds like a good idea.

> If that is the case, I would also perhaps suggest removing the old airframe so people can't select it. You could even just name it with "deprecated" in the title.

Good idea, I'll look into that!

en/frames_rover/index.md
chfriedrich98: Yes, I'll go through and make it consistent.

chfriedrich98:
@hamishwillee Thanks for the detailed review!
I'll start implementing your suggestions and ping you again once I'm through.

> Some of the documented params don't exist - probably changed names after you wrote this. YOu can see them listed by the flaw checker https://github.com/PX4/PX4-user_guide/pull/3265#issuecomment-2235743284
I fixed the ones I could be sure of, but you'll have to update the docs for the others appropriately.

Yes I think i renamed some of them because the names were too long, I'll get right on that!

en/frames_rover/index.md
chfriedrich98: Implemented in [9db2bef](https://github.com/PX4/PX4-user_guide/pull/3265/commits/9db2bef6c17098cb78f9137468e3970ad920071c)

en/frames_rover/ackermann_rover.md
chfriedrich98: Implemented in [dffb334](https://github.com/PX4/PX4-user_guide/pull/3265/commits/dffb334f2d64eabea527a27f0d1d863b39cb7927)

hamishwillee:
@chfriedrich98 Thanks very much for all the updates.

OK, so I have "positioned" this module now. Essentially renamed the old and new modules to "Ackermann Rover (v1)" and "Ackermann Rover (v2)". I've split out the docs for the "v1" and "differential steering" so that there are three very clear choices.
I moved the flight mode stuff into the v1 and diff steering docs because they are not the same as the ackermann v2.

I think we're good with this because it is pretty clear what differs and what your choices are. I decided not to split out the mission mode for now though it is a bit messy, we can do so easily enough as this progresses.

I haven't merged, because I'd appreciate you checking that what I have done makes sense. We can merge tomorrow if it is.

BTW, will the new module accept offboard commands yet?

hamishwillee:
Excellent. So pleased you're taking over the differential stuff too! Merging - we can iterate.

hamishwillee:
@chfriedrich98 FYI have posted about this this on #rover channel in discord - https://discord.com/channels/1022170275984457759/1026905519560077372/threads/1265812295804522557

I couldn't tag you because I don't know your ID.

hamishwillee:
PS How are we planning on deploying this "eventually"? Is it that we'll dump the old module and we'll have space in the standard build? Or is it that we need some mechanism to allow firmware variants to be selected in QGC (my preference)? Do you know who is sorting this out?

chfriedrich98:
> PS How are we planning on deploying this "eventually"? Is it that we'll dump the old module and we'll have space in the standard build? Or is it that we need some mechanism to allow firmware variants to be selected in QGC (my preference)?

Getting rid of the old module is unfortunately not enough since it is not even enabled on the default build for all boards. The second option would be ideal, but I'm not familiar with that topic so I don't know how difficult that change is. 

>  Do you know who is sorting this out?

I don't know of anyone that is working on this.

chfriedrich98:
@hamishwillee Just FYI I started working on the restructure for the rover docs as we discussed here in https://github.com/PX4/PX4-user_guide/pull/3311. It is still a draft (some upstream PR's have to be merged first too) so there is no need to review it yet, but I set up the overall structure which I'd appreciate your input on.

PR #3270, opened 2024-07-03
Fixes up the main entry point for getting started

hamishwillee:
To compensate for the restructure.

Also adds some other non-vital infrastructure changes


en/index.md
```diff
@@ -14,62 +14,60 @@ This guide contains everything you need to assemble, configure, and safely fly a
 
 :::warning
 This guide is for the _development_ version of PX4 (`main` branch).
```
hamishwillee, line 16: FYI @sfuhrer @MaEtUgR 

If you happen to be around, would appreciate your review of just this entry point doc. Is it "not wrong". Specifically I say things like "if you add a payload or new hardware you should retune". But should I be more specific about what kind of tuning (e.g. is autotune sufficient) or should I also recommend running the whole tune cycle again? 

I'll be doing a little more tomorrow and merging.

MaEtUgR: I still stand by my experience that with normal payload swaps usually you don't need to retune the vehicle. Of course if you add 2kg far away from the center of gravity then you are changing how it flies but chances are that you then also do that out of specs and can only reach a stay in the air state with an adapted tuning.

en/index.md
```diff
@@ -14,62 +14,60 @@ This guide contains everything you need to assemble, configure, and safely fly a
...
+Use the **Version** selector to find the current _stable_ version.
 
 Documented changes since the stable release are captured in the evolving [release note](releases/main.md).
 :::
 
 ## How Do I Get Started?
 
-[Getting Started](getting_started/index.md) should be read by all users!
+[Basic Concepts](../getting_started/px4_basic_concepts.md) should be read by all users!
 It provides an overview of PX4, including features provided by the flight stack (flight modes and safety features) and the supported hardware (flight controller, vehicles, airframes, telemetry systems, RC control systems).
```
MaEtUgR, line 25: Just caught my attention but vehicle vs airframe or rover airframe seem weird. We might want to consider renaming to limit confusion, just a thought.

en/index.md
```diff
@@ -14,62 +14,60 @@ This guide contains everything you need to assemble, configure, and safely fly a
...
 
-Depending on what you want to achieve, the following tips will help you navigate through this guide:
+If you're just getting started with flying drones then we recommend [multicopters](frames_multicopter/index.md), because they are easier to fly than planes and less complicated to build than VTOLs.
+Once you're familiar with multicopters, then moving onto planes and VTOL is much easier.
```
MaEtUgR, line 28: Having flown a multicopter and a plane definitely makes VTOLs easier. I agree planes are harder because they cannot stop in air and the support in terms of PX4 getting started is still slightly worse probably because of also less demand but in terms of basic stability and tuning they are easier no? All perspective dependent, I wonder what @sfuhrer thinks.

en/index.md
```diff
@@ -14,62 +14,60 @@ This guide contains everything you need to assemble, configure, and safely fly a
...
 
-**I already have a drone and I just want to fly:**
+Depending on what you want to achieve, the following tips will help you navigate through this guide:
 
-If you have a Ready To Fly (RTF) vehicle that supports PX4:
+### How do get a vehicle that works with PX4?
```
MaEtUgR, line 32:
```suggestion
### How to get a vehicle that works with PX4?
```

en/index.md
```diff
@@ -14,62 +14,60 @@ This guide contains everything you need to assemble, configure, and safely fly a
...
 
-- [Basic Configuration](config/index.md) explains how to update your firmware to the latest version, calibrate the main sensors (compass, gyro/IMU, airspeed etc.), and setup your remote control and safety features.
-- [Flying](flying/index.md) teaches flight essentials, including where and how to fly safely, and how to debug arming and flight issues. It also provides detailed information about flight modes.
+If you don't have already have Multicopter vehicle then [Complete Vehicles](complete_vehicles_mc/index.md) lists "Ready to Fly" (RTF) pre-build vehicles that you can buy, [Kits](frames_multicopter/kits.md) lists drones that you have to build yourself from a package, and [DIY Builds](frames_multicopter/diy_builds.md) shows some examples of drones that have been build using parts that have been sourced individually.
```
MaEtUgR, line 34:
```suggestion
If you don't already have a Multicopter vehicle then [Complete Vehicles](complete_vehicles_mc/index.md) lists "Ready to Fly" (RTF) pre-built vehicles that you can buy, [Kits](frames_multicopter/kits.md) lists drones that you have to build yourself from a package, and [DIY Builds](frames_multicopter/diy_builds.md) shows some examples of drones that have been built using parts that were sourced individually.
```
You write better than me so might be garbage.

en/index.md
```diff
@@ -14,62 +14,60 @@ This guide contains everything you need to assemble, configure, and safely fly a
...
+The [VTOL](frames_vtol/index.md) and [Planes (Fixed-Wing)](frames_plane/index.md) areas have similar topics.
 
-**I want to build a drone with PX4 from scratch:**
+Both kits and complete vehicles usually include everything you need except for a battery and RC System.
+Kits are usually not hard to build, provide a good introduction to how drones fit together, and are relatively inexpensive.
+We provide generic instructions for assembly, such as [Assembling a Multicopter](../assembly/assembly_mc.md), and most kits come with specific instructions too.
 
-:::tip
+If the kits and complete drones aren't quite right then you can build one from scratch, but this requires more knowledge.
```
MaEtUgR, line 41:
```suggestion
If the kits and complete drones aren't quite right for you then you can build a vehicle from scratch, but this requires more knowledge.

```

en/index.md
```diff
@@ -14,62 +14,60 @@ This guide contains everything you need to assemble, configure, and safely fly a
...
-- [Basic Configuration](config/index.md) shows how to update your firmware and configure it with settings appropriate for your airframe.
-  This section also explains how to calibrate the main sensors (compass, gyro/IMU, airspeed etc.), and setup your remote control and safety features.
+Before you fly you should read [Operations](config/operations.md) to understand how to set up the safety features of your vehicle and the common behaviours of all frame types.
+Once you've done that you're ready to fly.
 
-Once you are ready to fly your vehicle, visit the [Flying](flying/index.md) section.
+Basic instructions for flying each vehicle type are provided in the respective sections, such as [Basic Flying (Multicopter)](flying/basic_flying_mc.md).
 
-**I want to add payload or a camera:**
+### I want to add payload or a camera
```
MaEtUgR, line 54:
```suggestion
### I want to add a payload e.g. a camera

```

en/index.md
```diff
@@ -14,62 +14,60 @@ This guide contains everything you need to assemble, configure, and safely fly a
...
 
-The payloads section describes how to add a camera or how to configure PX4 to enable you to deliver packages.
+The [Payloads](payloads/index.md) section describes how to add a camera or how to configure PX4 to enable you to deliver packages.
 
-- [Payloads](payloads/index.md) describes how to integrate payloads
+After adding a payload you should retune the vehicle, which for multicopters you can do with [Autotune](config/autotune.md).
```
MaEtUgR, line 58:
```suggestion
After adding a heavy payload that's mounted far away from the center of gravity you might need to retune the vehicle, which for multicopters you can do with [Autotune](config/autotune.md).
```

en/index.md
```diff
@@ -14,62 +14,60 @@ This guide contains everything you need to assemble, configure, and safely fly a
...
+You should also consider the other tuning and calibration listed in your vehicle configuration guide (e.g. [Multicopter Configuration/Tuning](config_mc/index.md)).
```
MaEtUgR, line 59:
```suggestion
You should also consider the other tuning and calibration listed in your vehicle configuration guide (e.g. [Multicopter Configuration/Tuning](config_mc/index.md), [Autotune](config/autotune.md)).
```

en/index.md
```diff
@@ -14,62 +14,60 @@ This guide contains everything you need to assemble, configure, and safely fly a
...
 
-Modifications of the flight controller and basic sensors are covered above.
-In order to use new sensors, or if you have made changes that significantly affect flight characteristics:
+The [Hardware Selection & Setup](hardware/drone_parts.md) section provides both high level and product-specific information about hardware that you might use with PX4 and its configuration.
+This is the first place you should look if you want to modify a drone and add new components.
 
-- [Peripheral Hardware](peripherals/index.md) provides additional information about using external sensors.
-- [Basic Configuration](config/index.md) explains how to calibrate the main sensors.
-- [Advanced Configuration](advanced_config/index.md) should be used to re/fine-tune the airframe.
+Just as when you add a payload, you should re-tune the vehicle (see [Multicopter Configuration/Tuning](config_mc/index.md) for links to the relevant information).
```
MaEtUgR, line 66: I really wouldn't stress this so much. Probably it's rather a detail that needs to be in the steps of adding a payload e.g. how to mount it, does it have a big influence on the weight, airflow, distance or airspeed sensor?

Also I'd point to one tuning landing page and then explain there autotuning, manual tuning and a procedure to try them out.

en/index.md
sfuhrer: I guess it depends where you come from and what you aim for. If you have flown planes manually then yes, they're straight forward. But if you anyway want to end up operating MCs in auto modes then don't go through the struggle with planes first...

sfuhrer: I honestly challenge this section, there is no general advice on what to start with, MC vs FW. It's only really valid for VTOLs - I recommend that you fly a MC before a VTOL.

en/index.md
hamishwillee: Yeah, this isn't right. I usually prefer to use vehicle type or vehicle frame instead of "airframe", in particular when talking about a non-flying vehicle..

```suggestion
It provides an overview of PX4, including features provided by the flight stack (flight modes and safety features) and the supported hardware (flight controller, vehicle types, telemetry systems, RC control systems).

```

en/index.md
hamishwillee, line 28:
```suggestion
```

hamishwillee: I can't remember where I heard this opinion and if you both disagree with it there's no need for it - deleted. 

I suspect it is actually true that you're best off starting with MC because the documentation is much more comprehensive, and it is easier to get a developer kit etc. But we don't need to say that, and this docs effort is designed to address that problem.

en/index.md
hamishwillee: Nope, sadly in this case your writing is much better than mine :-(. Rushed to get it out without proper self-review.

en/index.md
hamishwillee: Thank you. That's the kind of thing I wasn't sure about.

en/index.md
hamishwillee: Thanks. I had this line as weasel words in case you should, say, redo compass calibration after changing your payload. But if autotune is the main thing you might want to do we should just mention that - which you did in the previous line.
```suggestion

```

en/index.md
hamishwillee: OK, so we'll have to add those notes to the payload page. I think I'll just remove this here. We're not trying to explain everything, just provide board pointers.

```suggestion

```

PR #3271, opened 2024-07-03
Add a warning in user_guide.md for ros_gz_bridge package

federicociresola:
Added a warning on how to install the correct version for ros_gz_bridge compatible with Gazebo Garden

### DIFF
--- a/en/ros2/user_guide.md
+++ b/en/ros2/user_guide.md
@@ -464,6 +464,18 @@ At this point, every ROS2 node must be instructed to use the newly bridged `/clo
 This concludes the modifications required on the ROS2 side. On the PX4 side, you are only required to stop the uXRCE-DDS time synchronization, setting the parameter [UXRCE_DDS_SYNCT](../advanced_config/parameter_reference.md#UXRCE_DDS_SYNCT) to `false`.
 By doing so, Gazebo will act as main and only time source for both ROS2 and PX4.
 
+:::warning
+If you notice that the ros_gz_bridge node doesn't publish anything on the /clock topic, it's possible that you have the wrong version of ros_gz_bridge installed. 
+The following commands install the correct version for **Gazebo Garden** with ROS2 **Humble**:
+```bash
+# Remove the wrong version for ignition fortress
+sudo apt remove ros-humble-ros-gz-bridge* 
+
+# Install the version for gazebo garden 
+sudo apt install ros-humble-gzgarden-bridge 
+```
+:::
+
 ## ROS 2 Example Applications
 
 ### ROS 2 Listener
--- a/en/ros2/user_guide.md
+++ b/en/ros2/user_guide.md
@@ -7,10 +7,12 @@ This topic provides an overview of the architecture and application pipeline, an
 ::: info
 From PX4 v1.14, ROS 2 uses [uXRCE-DDS](../middleware/uxrce_dds.md) middleware, replacing the _FastRTPS_ middleware that was used in version 1.13 (v1.13 does not support uXRCE-DDS).
 
-The [migration guide](../middleware/uxrce_dds.md#fast-rtps-to-uxrce-dds-migration-guidelines) explains what you need to do in order to migrate ROS 2 apps from PX4 v1.13 to PX4 v1.14. 
+The [migration guide](../middleware/uxrce_dds.md#fast-rtps-to-uxrce-dds-migration-guidelines) explains what you need to do in order to migrate ROS 2 apps from PX4 v1.13 to PX4 v1.14.
 
 If you're still working on PX4 v1.13, please follow the instructions in the [PX4 v1.13 Docs](https://docs.px4.io/v1.13/en/ros/ros2_comm.html).
+
 <!-- remove this when there are PX4 v1.14 docs for some months -->
+
 :::
 
 ## Overview
@@ -43,7 +45,6 @@ In PX4v1.13 and earlier, ROS 2 was dependent on definitions in [px4_ros_com](htt
 This repo is no longer needed, but does contain useful examples.
 :::
 
-
 ## Installation & Setup
 
 The supported ROS 2 platforms for PX4 development are ROS 2 "Humble" on Ubuntu 22.04, and ROS 2 "Foxy" on Ubuntu 20.04.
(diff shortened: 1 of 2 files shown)


en/ros2/user_guide.md
```diff
@@ -464,6 +464,18 @@ At this point, every ROS2 node must be instructed to use the newly bridged `/clo
 This concludes the modifications required on the ROS2 side. On the PX4 side, you are only required to stop the uXRCE-DDS time synchronization, setting the parameter [UXRCE_DDS_SYNCT](../advanced_config/parameter_reference.md#UXRCE_DDS_SYNCT) to `false`.
 By doing so, Gazebo will act as main and only time source for both ROS2 and PX4.
 
+:::warning
```
hamishwillee, line 467: Thanks @federicociresola . 

How likely is this? My thinking is that if you follow the instructions, "not very", so perhaps this should move this down into the troubleshooting section?

@beniaminopozzan Are you happy with this?

federicociresola: I think it is likely because when the official release of gazebo for ROS" humble is ignition not garden. You can look it [here ](https://gazebosim.org/docs/harmonic/ros_installation#summary-of-compatible-ros-and-gazebo-combinations)

beniaminopozzan: I agree that this might go in the troubleshooting section. There is already a (small) note about where to find the right version after all:
https://github.com/PX4/PX4-user_guide/blob/a79a24a30ad74150a3b7c27ca5f73e10b37a8cdf/en/ros2/user_guide.md?plain=1#L454

I think this is a good addition tot he doc as the existing note points to manual build installation instruction while users can actually install the right binaries.

@federicociresola , it seems to me that the package is `ros-humble-ros-gzgarden-bridge` though, could you double check? I'd also suggest to point to `ros-humble-ros-gzgarden`, so _all_ the ros/gz interface packages are installed, not just the bridge.

en/ros2/user_guide.md
```diff
@@ -451,9 +458,38 @@ No further action is required by the user.
...
+
+:::
+
+::::
+
+::: info
+The [repo](https://github.com/gazebosim/ros_gz#readme) and [package](https://github.com/gazebosim/ros_gz/tree/ros2/ros_gz_bridge#readme) READMEs show the package versions that need to be installed depending on your ROS2 and Gazebo versions.
+:::
+
+Once the packages are installed and sourced, the node `parameter_bridge` provides the bridging capabilities and can be used to create an unidirectional `/clock` bridge:
```
hamishwillee, line 492: @federicociresola @beniaminopozzan Looking at this, I think that there was an unnecessary risk in the instructions in pointing to the readmes to work out the correct versions of the bridge. We know the correct versions because we know the versions PX4 supports, so why not spell it out. 

So what I have done is added this section ^^^ which outlines the humble and foxy variants of binaries to install (@beniaminopozzan - specifying the interface package, not just the bridge, as you suggested). I added the old note as Info at the end so people can see that there are different versions that might work.

en/ros2/user_guide.md
```diff
@@ -768,43 +804,62 @@ average rate: 247.485
...
+
+  ```sh
+  sudo apt install ros-foxy-eigen3-cmake-module
+  ```
+
+  :::
 
+  ::::
+
+### ros_gz_bridge not publishing on the \clock topic
```
hamishwillee, line 849: This section below is the warning, turned into a section. Can you sanity check that too please. IN particular the instruction to remove the wrong version, and the line that starts "This might happen ...".

en/ros2/user_guide.md
federicociresola: @beniaminopozzan I didn't see the line but I think it is good to highlight it because a person can lose a lot of time trying to find the problem. Yes, you can suggest to install `ros-humble-ros-gzgarden-bridge` but I had a bad experience to install gazebo from ROS2 and gazebo standalone #23259. Then I think it is time to decide how to install gazebo (split in standalone or from ROS2 package) and should suggest a modification in `../setup/ubuntu.sh`. What do you think about it?

hamishwillee: @federicociresola Gazebo is installed via the script standalone because: 
- PX4 isn't validated with random versions, just Garden and Gazebo-Classic 11.
- Many (the majority) don't care about ROS.

Open to any solution that means you can still use PX4/Gazebo with and without ROS in a predictable way.

You haven't commented on my solution, which is to tell people exactly what they should install so there is no guesswork.

federicociresola: you are right we can't modify the script because there is already the option to run the installation with/out the simulation tool.

Sorry, I have not commented on your solution because I have not seen it

en/ros2/user_guide.md
federicociresola: For it is a good proposal, I would like to suggest to add a line with wrote a think like "if you have installed the gazebo from ubuntu.sh we suggest you to remove it and keep only the version installed from the ROS2 package so that there are no conflicts between versions"

What do @beniaminopozzan think about it?

en/ros2/user_guide.md
federicociresola: For me, It is good. I have checked and when someone installs `ros-humble-ros-gzgarden` in automatic way it also installs  the correct bridge

en/ros2/user_guide.md
hamishwillee: > What do @beniaminopozzan think about it?

Good question. Note, I am away until Wednesday, but I will respond then to whatever @beniaminopozzan says.

I think the suggestion to remove any conflicting version would be good, if it is possible to do.

hamishwillee: @beniaminopozzan You must have missed this request or been too busy. 

I am pretty sure that the rest of this is good. I'm going to merge, but if you could do a post-review that would be good.

@federicociresola Thanks for this. I have chosen not to add the note:

>  "if you have installed the gazebo from ubuntu.sh we suggest you to remove it and keep only the version installed from the ROS2 package so that there are no conflicts between versions"

Because there shouldn't be any conflict if you follow the instructions. But @beniaminopozzan we can add it he thinks it is needed.

federicociresola: Perfect, I am happy to contribute to this amazing project when I am able

beniaminopozzan: Thanks @hamishwillee , apologises for missing this (again).
It looks very good to me, nice work!

hamishwillee: NO worries! Thanks for the post submit check!

PR #3280, opened 2024-07-10
correct battery realease notes

chfriedrich98:
Slight correction: The internal resistance estimation is used to _calculate the open circuit voltage_ which is used instead of the _measured voltage_ to get a more accurate capacity estimate.


hamishwillee (approved):
Thank you very much - that makes more sense.

PR #3282, opened 2024-07-10
Bugfix - made the flighttask links visible

Claudio-Chies:
the old @[youtube] didnt work, just linked the titles to the videos

PR #3294, opened 2024-07-17
[TEST] Try markdown-it-block-embed for video embed

hamishwillee:
Attempting fix of https://github.com/PX4/PX4-user_guide/pull/3282 using https://github.com/rotorz/markdown-it-block-embed

As suggested in https://github.com/vuejs/vitepress/issues/4053

NOTE, does not display at all:

image

### DIFF
--- a/.vitepress/config.mjs
+++ b/.vitepress/config.mjs
@@ -2,7 +2,7 @@ import { defineConfig } from "vitepress";
 const getSidebar = require("./get_sidebar.js");
 //import { getSidebar } from "./get_sidebar";
 
-import markdownItVideo from "markdown-it-video";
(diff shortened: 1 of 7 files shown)


.vitepress/config.mjs
```diff
@@ -26,15 +26,10 @@ export default defineConfig({
...
-    //   container: {
-    //     tipLabel: "Note",
-    //   },
-
     config: (md) => {
       // use more markdown-it plugins!
-      md.use(markdownItVideo);
-
+      md.use(blockEmbedPlugin);
+      console.log("Screw");
```
hamishwillee, line 32:
```suggestion
```

PR #3296, opened 2024-07-17
added x500_lidar model

Claudio-Chies:
Updated documentation because of 
- [ ] https://github.com/PX4/PX4-Autopilot/pull/22418 
- [ ] https://github.com/PX4/PX4-gazebo-models/pull/41


en/sim_gazebo_gz/vehicles.md
```diff
@@ -52,6 +52,16 @@ The camera cannot yet be used to stream video or for image capture in QGroundCon
 [PX4-Autopilot#22563](https://github.com/PX4/PX4-Autopilot/issues/22563) can be used to track the additional work needed to fully enable these use cases.
 :::
 
+### X500 Quadrotor with 2D LIDAR
+
+This model have a 2D LIDAR attached, modelled on the Hokuyo TM-30LX.
```
hamishwillee, line 57: This all looks excellent. A few follow ons:

1. did you mean [UTM-30LX](https://hokuyo-usa.com/products/lidar-obstacle-detection/utm-30lx) - google didn't find anything at all on TM-30LX.
2. Is this a lidar that we support in PX4? I can't find it here: https://docs.px4.io/main/en/sensor/rangefinders.html#distance-sensors-rangefinders
3. Also in that doc we have Gazebo Classic simulation section - https://docs.px4.io/main/en/sensor/rangefinders.html#gazebo-classic-simulation - Would it be possible for you to add a few words about simulation with Gazebo in its own section?

hamishwillee: PS You'll note there are more images in this PR than you had. I shrank all images linked in this folder (losslessly) so they just happen to appear as part of this.

Claudio-Chies: 1. yea, youre right. thats a copypasta error. 
2. not that i can see. the only similar one we support is the [Lightware SF45/B](https://lightwarelidar.com/shop/sf45-b-50-m/) then it might make sense to remove the specific model. not sure of the best practices here.
3. currently the rotating Lightware lidar is also in this section, but we could add a subsection...
4. i agree that it would make sense to add a gazebo section to rangefinders, but i'm not sure if i'm the right person, as i have limited expierience creating my own models.
5. see (2.)

en/sensor/rangefinders.md
```diff
@@ -161,6 +161,17 @@ On the Simulator you can run the commands directly in the terminal.
...
+
+Lidar rangefinders can be used in the [Gazebo](../sim_gazebo_gz/index.md) simulator.
+To do this you must start the simulator using a vehicle model that includes the rangefinder.
+
+The x500 lidar model includes a 2D Lidar rangefinder:
+
+```sh
+make px4_sitl gz_x500_lidar
+```
+
```
hamishwillee, line 174: > 4. Also in that doc we have Gazebo Classic simulation section - https://docs.px4.io/main/en/sensor/rangefinders.html#gazebo-classic-simulation - Would it be possible for you to add a few words about simulation with Gazebo in its own section? (assuming this is the right place)?
>> i agree that it would make sense to add a gazebo section to rangefinders, but i'm not sure if i'm the right person, as i have limited expierience creating my own models.

So what I was really trying to get at here was "OK, cool, we have a simulated vehicle with a rangefinder, what can I do with it? how can I use it?

My assumption is that if I use this model PX4 will be getting rangefinder data as though there were a real rangefinder attached to the vehicle and already enabled.

On that assumption I have added this section. So basically someone else reading this doc this will assume that they can just use this lidar. Do you happen to know what range it is set up to use? Or how this can be configured?

If that is not true, then we need to know how to make it useful.

It would be good to explain how to add this same model to other simulated frames, but that is not be as important.

en/computer_vision/collision_prevention.md
```diff
@@ -204,3 +204,4 @@ See [PX4/PX4-Avoidance](https://github.com/PX4/PX4-Avoidance) for setup instruct
 
 <!-- PR companion collision prevention (initial): https://github.com/PX4/PX4-Autopilot/pull/10785 -->
 <!-- PR for FC sensor collision prevention: https://github.com/PX4/PX4-Autopilot/pull/12179 -->
+<!-- using rangefinder? -->
```
hamishwillee, line 207: > 5. We should certainly update https://docs.px4.io/main/en/computer_vision/collision_prevention.html#px4-distance-sensor right?
 >> 5. see (2.)

So I've tagged this section at the end of the Collision parameters page as an explanation of what I mean.

My assumption is that you added this for testing collision prevention. So it might be good to remove this section pointing to the PX4-Avoidance repo, which is broken, and replace it with a section that explains how you can test collision prevention using this new lidar.

Do you know. IF not, I consider this non-blocking, but would be great if you can tell me who might know? (or I can try it myself).

en/sim_gazebo_gz/vehicles.md
hamishwillee: Thanks @chfriedrich98 

> 2. not that i can see. the only similar one we support is the [Lightware SF45/B](https://lightwarelidar.com/shop/sf45-b-50-m/) then it might make sense to remove the specific model. not sure of the best practices here.
> 3. currently the rotating Lightware lidar is also in this section, but we could add a subsection...

For now we leave both of these as is - can't do any harm and I'm not sure we'd be improving things.

I've added comments for 4 and 5 in other places - mostly for you to help me understand "what can we do now we have this model".

en/computer_vision/collision_prevention.md
Claudio-Chies: i did add a small section, might be expanded once i'm done with collision prevention

en/sensor/rangefinders.md
Claudio-Chies: yes, the only thing there is that the x500_lidar model generates a gazebo laserscan which gets translated into a obstacle distance message. not sure if this is the most representative example for rangefinder simulation... its more a collision prevention example. do we want to differentiate there?
i did add some specs on the simulated sensor. but these are hardcoded in the gazebo model. so no way to change them via parameters.

hamishwillee (approved):
Thanks @Claudio-Chies ! Looking at the code, Gazebo does not support normal rangefinder use cases because it doesn't write to DistanceSensor (at all). It would be great if this model did so, because then you could use it for both collision prevention and other terrain following/position hold etc.

For now though I have made it explicit that the model is intended for collision prevention, and in the Rangefinder doc noted that currently these use cases are not supported. I kept the section though because I hope that you or someone else will update this or some other model to work as a rangefinder.

Thanks very much for this.

en/computer_vision/collision_prevention.md
```diff
@@ -197,10 +185,21 @@ For more information on hardware and software setup see: [PX4/PX4-Avoidance > Ru
...
+_Collision Prevention_ can be tested using [Gazebo](../sim_gazebo_gz/index.md) with the [x500_lidar](../sim_gazebo_gz/vehicles.md#x500-quadrotor-with-2d-lidar) model.
+To do this, start a simulation with the x500 lidar model by running the following command:
+
+```sh
+make px4_sitl gz_x500_lidar
+```
+
+Next, adjust the relevant parameters to the appropriate values and add arbitrary obstacles to your simulation world to test the collision prevention functionality.
+
+The diagram below shows how the simulation looks when viewed in RViz.
```
Claudio-Chies, line 199: Actually its Gazebo and not RViz 😅

hamishwillee: Thanks @Claudio-Chies, I'll fix - didn't know you could get the sensor view like that. BTW, do you know much about prearm conditions on collision prevention  - got a bug https://github.com/PX4/PX4-Autopilot/issues/22994 and from my comment to that you can see a linked (current) discussion board post.

Claudio-Chies: puh, for Collision Prevention, i think you just have to have a obstacle_distance message present. but in the issue he's talking about avoidance, and there its handeled differently from what i know. as it has to wait for the onbaord computer to be ready

hamishwillee: Ah, we should probably document that :-). So to be clear, the implication is that you need a component that is sending a HEARTBEAT and has type https://mavlink.io/en/messages/common.html#MAV_TYPE_ONBOARD_CONTROLLER right?

hamishwillee: Ignore that, I can see the code 
```
			case MAV_COMP_ID_OBSTACLE_AVOIDANCE:
				_heartbeat_component_obstacle_avoidance = now;
				_mavlink->telemetry_status().avoidance_system_healthy = (hb.system_status == MAV_STATE_ACTIVE);
				break;
```

Incorrect in that it relies on the companion computer component ID rather than mavlink type
-

PR #3304, opened 2024-07-21
Update mavlink.md - Example Bugs

rafaelmelomacieira:
First: The "mavlink_msg_battery_status_demo.h" is built in the "/build.../mavlink/development" and not at "common", since the message definition was set at  "...message_definitions/development.xml".
Second: the message field at "BATTERY_STATUS_DEMO.hpp", line 60, must be changed from battery_remaining to percent_remaining, since the message definition says so.


hamishwillee:
Thanks @rafaelmelomacieira - that all seems very likely. I will review this on Wednesday!

hamishwillee (approved):
Thank you @rafaelmelomacieira - appreciate it when someone using these guides  takes the time to fix the bugs they discover.

rafaelmelomacieira:
Thank you @hamishwillee. These tutorials are very important and useful! You can count on me!

PR #3308, opened 2024-07-25
PX4 v115 docs around safety/arming params

hamishwillee:
Fixes for changes in these docs: https://discuss.px4.io/t/px4-v1-15-public-changes-what-needs-docs/39850

- ‘COM_ARMABLE’, ‘COM_ARM_BAT_MIN’, ‘COM_DISARM_MAN’,

### DIFF
--- a/en/config/safety.md
+++ b/en/config/safety.md
@@ -65,6 +65,10 @@ The settings and underlying parameters are shown below.
 | <a id="BAT_CRIT_THR"></a>Battery Failsafe Level | [BAT_CRIT_THR](../advanced_config/parameter_reference.md#BAT_CRIT_THR)       | Percentage capacity for Return action (or other actions if a single action selected).                                                                                     |
 | Battery Emergency Level                         | [BAT_EMERGEN_THR](../advanced_config/parameter_reference.md#BAT_EMERGEN_THR) | Percentage capacity for triggering Land (immediately) action.                                                                                                             |
 
+::: tip
+There is also a configuration parameter [COM_ARM_BAT_MIN](#COM_ARM_BAT_MIN) that allows you to prevents arming in the first place if the battery level is too low.
+:::
+
 ### Manual Control Loss failsafe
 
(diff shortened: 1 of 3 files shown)


en/config/safety.md
```diff
@@ -219,6 +223,18 @@ The relevant parameters are shown below:
 | [COM_OF_LOSS_T](../advanced_config/parameter_reference.md#COM_OF_LOSS_T)   | Delay after loss of offboard connection before the failsafe is triggered.                                         |
 | [COM_OBL_RC_ACT](../advanced_config/parameter_reference.md#COM_OBL_RC_ACT) | Failsafe action if RC is available: Position mode, Altitude mode, Manual mode, Return mode, Land mode, Hold mode. |
 
+### Flight Time Failsafe
```
hamishwillee, line 226: FYI @sfuhrer Little odd to have two mechanisms for deciding on max flight time return - one estimated and one determined by you. Unless these are actually the trigger time and action that have been miss-documented?

hamishwillee:
Merging this. So can port into main

en/config/safety.md
sfuhrer: Those are two different things, one (COM_FLT_TIME_MAX) is to limit the flight time, and the other is a battery failsafe action to actually make sure you reach your landing point before you're out of juice.

sfuhrer: I would document COM_FLTT_LOW_ACT as part of the battery failsafes.

hamishwillee: @sfuhrer Thanks, done in https://github.com/PX4/PX4-user_guide/pull/3320

Note I chose not to do this originally because the separation in safety doc was between QGC failsafes and "Failsafes you have to set with parameters" and this crossed both. Also because the low level setting was quite distinct.

However as QGC safety view omits so much decided to make this feature centric instead, and have moved this as you suggested. It is an adjacent section to the old settings, both grouped under "Battery Failsafes". Works much better IMO.

PR #3310, opened 2024-07-29
pid_tuning_guide_multicopter: adapt to Acro default rates having changed for tuning purposes in PX4 1.15

MaEtUgR:
Follow-up to
-  https://github.com/PX4/PX4-Autopilot/pull/22098
after reading https://discuss.px4.io/t/px4-v1-15-public-changes-what-needs-docs/39850/3

- [ ] When this goes in, need to also push into PX4 v1.15

### DIFF
--- a/en/flight_modes_mc/acro.md
+++ b/en/flight_modes_mc/acro.md
@@ -27,7 +27,21 @@ Manual control input is required (such as RC control, joystick):
 
 ## Stick Input Mapping
 
-The default roll, pitch, and yaw input stick mapping for Acro mode is shown below. The curve enables a high turn rate at maximum stick input for performing acrobatic maneuvers, and a zone of lower sensitivity close to the stick center for small corrections.
+The default values for expo and rate [parameters](#parameters) are _beginner friendly_, reducing the chance that users will flip the vehicle when first trying this mode, or when using Acro mode for manual rate tuning.
+
+Unfortunately they are poor for Acro flying though!
+The following values are more reasonable for experienced users:
+
+- [MC_ACRO_R_MAX](#MC_ACRO_R_MAX): `720` (2 turns per second)
+- [MC_ACRO_P_MAX](#MC_ACRO_P_MAX): `720` (2 turns per second)
+- [MC_ACRO_Y_MAX](#MC_ACRO_Y_MAX): `540` (2 turns per second)
+- [MC_ACRO_EXPO](#MC_ACRO_R_MAX): `0.69`
+- [MC_ACRO_EXPO_Y](#MC_ACRO_EXPO_Y): `0.69`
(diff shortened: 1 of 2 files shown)


en/flight_modes_mc/acro.md
```diff
@@ -27,7 +27,21 @@ Manual control input is required (such as RC control, joystick):
```
hamishwillee, line 27: Thanks @MaEtUgR 

I've updated the parameter values in the Acro mode doc here as well. It seems to me that otherwise users will be very disappointed when they try acro mode and it isn't much fun. Because of the existing diagrams in stick input I've suggested that they use the previous default as experienced users. Is this OK?

Further, this really only matters for _manual_ tuning right? I.e. you don't need to worry about default values when autotuning?

hamishwillee:
@MaEtUgR This seems to have fallen off our radar, but still seems relevant. I have rebased.

Can you sanity check, and look at https://github.com/PX4/PX4-user_guide/pull/3310#discussion_r1697759964 ?

en/flight_modes_mc/acro.md
MaEtUgR: > I.e. you don't need to worry about default values when autotuning?

Correct

Yes, the reason I lowered the default parameters is that it's much more likely for someone inexperienced to use the defaults and also very likely only for tuning purposes in Acro mode. In that case they rather be the low ones I use for tuning (100°/s) than the high ones we'd usually use on a small FPV racer to be able to turn quickly and flip around. Someone experienced very likely wants to see the plot I explained here:
https://github.com/PX4/PX4-Autopilot/pull/8036#issuecomment-333774139
it essentially allows you to set maximum rate, expo and super expo and see the mathematical result of how input is then interpreted. This way you can dial it in to your liking and flying style:
https://www.desmos.com/calculator/zuqgbayd9g
I still hope we can somehow make that very accessible at least in the docs but ideally also in QGroundcontrol 👀

MaEtUgR:
Thank you @hamishwillee It is in Text all really nice now!
The only thing I saw is that the math part which actually already links to desmos is not rendered properly but we can solve that seprately. Is there any way to try this out locally in order to find the problem?
image

MaEtUgR:
Why does CI fail with `Could not resolve "./../../assets/airframes/rover/flight_modes/manual_inputs.png" from "ko/flight_modes_rover/index.md"`
That should be unchanged.

en/flight_modes_mc/acro.md
hamishwillee: @MaEtUgR Thanks! The calculation and graph are linked below, but there was a rendering issue which I have fixed in https://github.com/PX4/PX4-user_guide/pull/3459/files

So these should be accessible in docs.

For QGC it would probably be nice to include the graph and save the results from that into params automatically. A lot of work though.

PR #3311, opened 2024-07-29
Restructure rover docs

chfriedrich98:
Adds a section for the new differential rover module, restructures the rover docs and moves all the information on the old rover module to a deprecated section.

- [x] https://github.com/PX4/PX4-Autopilot/pull/22402
- [x] https://github.com/PX4/PX4-Autopilot/pull/23430
- [x] https://github.com/PX4/PX4-Autopilot/pull/23506


hamishwillee:
@chfriedrich98 I think this structure looks excellent. Its good to see common flight modes.

I adopted the v1 and v2 naming because it is rarely a good idea to refer to "old" and "new" in naming conventions - they date badly. I don't think it matters here calling the original module "old" because we're hoping to eventually remove it. At that point the prior existence is irrelevant. Ping m when you're good for a proper review - but I don't think it will need much based on my quick scan.

We are going to need to chase up and "nag" about deployment though. Suggesting a custom build will always make this a second class citizen. The first step is that someone needs to make sure that all the variants are build to the build server, not just the defaults. Second step is to modify QGC to allow selection of other variants. Final step would be to look at the relationship between airframes and firmware so that we can select configurations sensibly - the way it is now you can select an airframe that isn't supported by your firmware. Beat Kung would have ideas on that, but has become overloaded and would be hard to track down.

en/flight_modes_rover/index.md
```diff
@@ -45,6 +45,19 @@ Acro Mode is similar to Manual Mode, but with a closed-loop yaw rate control. In
 [Mission mode](../flight_modes/mission.md) is an automatic mode that causes the vehicle to execute a predefined autonomous [mission](../flying/missions.md) plan that has been uploaded to the flight controller.
 The mission is typically created and uploaded with a Ground Control Station (GCS) application, such as [QGroundControl](https://docs.qgroundcontrol.com/master/en/).
 
+Following is the list of currently implemented mission commands:
+
+| Command             | Description                                          |
+| ------------------- | ---------------------------------------------------- |
+| Mission start       | Starts the mission (necessary)                       |
```
hamishwillee, line 52: It is worth listing/linking to the MAVLink messages too. Why? Because that is ultimately what a mission is, and you can't know how the commands are being sent or named.

chfriedrich98:
rebased on main and squashed

en/flight_modes_rover/index.md
chfriedrich98: Is [this](https://github.com/PX4/PX4-user_guide/blob/9cc703a2e440e4c11284f82b20ce3bf440f6663e/en/flight_modes_rover/index.md#mission-commands) what you had in mind?

hamishwillee: Perfect!

en/frames_rover/ackermann_rover.md
```diff
@@ -88,7 +62,7 @@ To get an overview of all parameters that are related to the Ackermann rover mod
 
 ### General Parameters
 
-These parameters affect the general behaviour of the rover. This will influence both auto and manual modes.
+These parameters affect the general behavior of the rover. This will influence both auto and manual modes.
 
 | Parameter                                                                                       | Description                                | Unit |
 | ----------------------------------------------------------------------------------------------- | ------------------------------------------ | ---- |
```
hamishwillee, line 68: The following section is "Slew rates". This says "The slew rates are not based on measurements but on assumed linear relation between the throttle".

This tells me what they are, but not what I should set them to. How do I go about deciding what they should be. Say assuming my vehicle can go from 0 to 60km/hour in 5 seconds, and stay on the road if doing a circle with radius X, at Y m/s.

hamishwillee: Similarly for `RA_MAX_SPEED` - is the approach to just set this impossibly high, drive the vehicle as fast as it can go, and put in the top value?

Where possible we need to be instructional - "this is what you do". The parameter reference is informational "this is what it is"

en/frames_rover/ackermann_rover.md
```diff
@@ -112,46 +86,25 @@ Therefore these two parameters have to be set for the slew rates to work!
```
hamishwillee, line 86: Above it says "two parameters have to be set for the slew rates to work!". I assume you mean slew rate limiting to work. Again, worth saying here what happens if you get it wrong. I.e. "Too high slew rate settings mean that the vehicle will skid at higher speeds and faster turns", or to put it another way, assume I'm an idiot "why do I care about slew rates", "why am I setting these"

en/frames_rover/aion_r1.md
```diff
@@ -2,7 +2,7 @@
 
 <Badge type="tip" text="PX4 v1.15" />
 
-The [Aion R1](https://www.aionrobotics.com/)vehicle was chosen to test and improve the differential drive support for PX4, and to improve driver support for Roboclaw Motor Controllers, such as the [RoboClaw 2x15A](https://www.basicmicro.com/RoboClaw-2x15A-Motor-Controller_p_10.html).
+The [Aion R1](https://www.aionrobotics.com/) vehicle was chosen to test and improve the differential drive support for PX4, and to improve driver support for Roboclaw Motor Controllers, such as the [RoboClaw 2x15A](https://www.basicmicro.com/RoboClaw-2x15A-Motor-Controller_p_10.html).
```
hamishwillee, line 5: This is shown in Summary.md like this:
```
    - [Differential-steering Rover](frames_rover/differential_rover.md)
      - [Aion Robotics R1](frames_rover/aion_r1.md)
```

But the setup instructions are for the old module. So setup changes (below) need to be adjusted to point to new rover page. OR this has to move below the old position control topic.

en/frames_rover/differential_rover.md
```diff
@@ -0,0 +1,104 @@
...
+## Tuning (Mission)
+
+:::warning
+The parameters in [Tuning (Basic)](#tuning-basic) must also be set to drive missions!
+:::
+
+The module uses a control algorithm called pure pursuit, see [Mission Mode](../flight_modes_rover/index.md#mission-mode) for the basic tuning process.
+The additional parameters are separated into the following sections:
+
+### Mission Velocity
```
hamishwillee, line 67: So the mission params for the two rover types are different? We should link back here from the mission mode section.

en/frames_rover/ackermann_rover.md
```diff
@@ -255,3 +160,9 @@ $$
 | $r_{min}$      | Minimum turn radius                | m    |
 | $\delta_{max}$ | Maximum steer angle                | m    |
 | $r_{acc}$      | Acceptance radius                  | m    |
+
+<!--
```
hamishwillee, line 164: As a pattern for docs, normally we'd list ALL parameters affecting the frame that the user needs to know about in a table. Above we'd **list** or describe the items in each section, linking down to here. So for example, the mission parameters above would be a description or a list. 

The idea being that you have all the parameters in one place and link down to them from descriptive text. Something to think about. You might not want to do this if you thought moving the mission parameters to the drive mode docs made sense.

en/frames_rover/differential_rover.md
```diff
@@ -0,0 +1,112 @@
...
+### State Machine
+
+The module employs the following state machine to make full use of a differential-steering rovers ability to turn on the spot:
+![Differential state machine](../../assets/airframes/rover/rover_differential/differential_state_machine.png)
+
+These transition thresholds can be set with [RD_TRANS_DRV_TRN](#RD_TRANS_DRV_TRN) and [RD_TRANS_TRN_DRV](#RD_TRANS_TRN_DRV).
+
+### Parameters
+
+The following parameters affect the differential-steering rover in mission mode (overview):
```
hamishwillee, line 100: These are the mission params. Should we include ALL params here and reduce the heading level by one?

en/flight_modes_rover/index.md
```diff
@@ -1,16 +1,101 @@
...
+  - Pushed left/right: Rover rotates counter-clockwise/clockwise at the rate commanded by the stick input, using a closed-loop controller (e.g., PID) to try to ensure the vehicle yaw rate matches the given setpoint.
+
+See [Tuning(basic)](../frames_rover/differential_rover.md#tuning-basic) to go through the necessary setup to use acro mode for differential-steering rovers.
+
+## Mission Mode
+
+_Mission mode_ is an automatic mode that causes the vehicle to execute a predefined autonomous [mission](../flying/missions.md) plan that has been uploaded to the flight controller.
+The mission is typically created and uploaded with a Ground Control Station (GCS) application, such as [QGroundControl](https://docs.qgroundcontrol.com/master/en/).
+
+### Mission commands
```
hamishwillee, line 52: Add another section `### Mission Parameters` linking to the sections for each vehicle. Unless they are largely shared/common (doesn't appear so) in which case you might  go the other way - have the mission settings here.

en/frames_rover/ackermann_rover.md
hamishwillee: This is ^^^ an important point. When you come to this again, look at every section and ask "what am I asking the user to actually _do_.

en/frames_rover/ackermann_rover.md
```diff
@@ -88,7 +62,7 @@ To get an overview of all parameters that are related to the Ackermann rover mod
 
 ### General Parameters
 
-These parameters affect the general behaviour of the rover. This will influence both auto and manual modes.
+These parameters affect the general behavior of the rover. This will influence both auto and manual modes.
```
hamishwillee, line 65: We use UK spelling "generally".
```suggestion
These parameters affect the general behaviour of the rover. This will influence both auto and manual modes.
```

hamishwillee (approved):
Looks great. Certainly good enough for now. I've added some notes for you to address (or reject) in a follow on update. Thank you.

en/frames_rover/ackermann_rover.md
hamishwillee: So this comment is "something to think about"

en/frames_rover/differential_rover.md
```diff
@@ -0,0 +1,112 @@
...
+:::warning
+The parameters in [Tuning (Basic)](#tuning-basic) must also be set to drive missions!
+:::
+
+The module uses a control algorithm called pure pursuit, see [Mission Mode](../flight_modes_rover/index.md#mission-mode) for the basic tuning process.
+The additional parameters are separated into the following sections:
+
+### Mission Velocity
+
+These parameters tune velocity control in missions:
```
hamishwillee, line 77: This is what I thought they were for. Is it correct?

In line with the other comment, how do I work out how to set these values? I.e. instructional or rule of thumb guidance. This might be "just experiment - if the vehicle does X, Y, Z, then you need to lower this". Does that make sense?

PR #3312, opened 2024-07-29
dshot.md: make it clear that command ID refers to output pin number

sfuhrer:
And not to motor number.

### DIFF
--- a/en/peripherals/dshot.md
+++ b/en/peripherals/dshot.md
@@ -56,7 +56,7 @@ See [here](../modules/modules_driver.md#dshot) for a full reference of the suppo
 
 The most important ones are:
 
-- Make the first motor beep (helps with identifying motors):
+- Make the motor connected to the first output pin beep (helps with identifying motors):
 
   ```
   dshot beep1 -m 1
@@ -79,7 +79,7 @@ The most important ones are:
   INFO  [dshot] LED 3: unsupported
   ```
 
-  - Permanently reverse the spin direction of the first motor:
+  - Permanently reverse the spin direction of motor connected to the first output pin:
 
   ```
   dshot reverse -m 1
--- a/en/peripherals/dshot.md
+++ b/en/peripherals/dshot.md
@@ -56,7 +56,7 @@ See [here](../modules/modules_driver.md#dshot) for a full reference of the suppo
 
 The most important ones are:
 
-- Make the motor connected to the first output pin beep (helps with identifying motors):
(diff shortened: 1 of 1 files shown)


en/peripherals/dshot.md
```diff
@@ -56,7 +56,7 @@ See [here](../modules/modules_driver.md#dshot) for a full reference of the suppo
 
 The most important ones are:
 
-- Make the first motor beep (helps with identifying motors):
+- Make the motor connected to the first output pin beep (helps with identifying motors):
```
hamishwillee, line 59: FMI This is FMU pin 1 as numbered by the manufacturer right? Not "motor 1" as per the actuator mapping?

If so we might perhaps say that "connected to FMU PWM output pin 1" ?

MaEtUgR: > pin 1 as numbered by the manufacturer right? Not "motor 1" as per the actuator mapping?

Yes

> FMU PWM output pin 1

Is that the case for every board? 🤔 But maybe I'm thinking too far and you're right.

hamishwillee: Hmmm, well it has to be FMU, but the port doesn't have to be labelled that - but we mention the mapping back up the page. Further, as it is written now you might interpret it as "the first connected output pin from the right" (say) - I know it is a stretch.

I think if we have "first" we are definitely assuming that you have connected to pin 1 first. Of course you would, but it is ambiguous.

```suggestion
- Make a motor connected to to FMU PWM output pin 1 beep (helps with identifying motors):
```

sfuhrer: It could be a bit confusing talking about PWM pins when it's in this case here actually used for DSHOT, but I'm not strongly opinionated, we just should do the pin naming consistently.

sfuhrer:
```suggestion
- Make a motor connected to to FMU output pin 1 beep (helps with identifying motors)
```

en/peripherals/dshot.md
```diff
@@ -79,7 +79,7 @@ The most important ones are:
   INFO  [dshot] LED 3: unsupported
   ```
 
-  - Permanently reverse the spin direction of the first motor:
+  - Permanently reverse the spin direction of motor connected to the first output pin:
```
hamishwillee, line 82:
```suggestion
  - Permanently reverse the spin direction of a motor connected to FMU output pin 1:
```

hamishwillee (approved):
Thanks @sfuhrer @MaEtUgR !

PR #3313, opened 2024-07-29
Dshot: correct reversing a reversed motor again

MaEtUgR:


en/peripherals/dshot.md
```diff
@@ -79,26 +75,26 @@ The most important ones are:
...
-  after saving it with `dshot save -m 1` command, reversed direction will become new normal direction:
+  after saving it with `dshot save -m 1` command, the reversed direction will become the new normal direction:
 
-  ```
+  ```sh
   Rotation Direction: normal
   ```
 
-  To change direction again new `dshot reverse -m 1` command needs to be sent.
+  To change direction back to the "original" normal, you will need to set `dshot normal -m 1` and then the `dshot save -m 1` command.
```
hamishwillee, line 97: @MaEtUgR Are you sure? (the previous text made more sense - and either this change is wrong or the previous lines are wrong).

To put it another way, we say that saving the value makes it the new normal. Once it is the new normal the old normal become the new reversed. So either saving does not change what normal means and this is wrong, or it does, and the previous lines need to be tidied up.

MaEtUgR: To be honest I'm not sure. I suggested this change based on @markusachtelik 's feedback but haven't tried myself. I'll quickly try on a hardware setup.

hamishwillee: I'm going to be on it being wrong, because Beat is never wrong :-)

MaEtUgR: I thoroughly tested a 4-in-1 DShot ESC, which has a binary state: `normal` or `reverse`.

Running `dshot reverse -m 1` switches the ESC on FMU Pin 1 to reverse. If it was already in reverse, nothing changes. Using `dshot save -m 1` makes this change persistent over reboots; otherwise, the ESC resets to its previously saved state.

- ESCs usually come in `normal` state from the factory
- Running a motor test shows the motor spinning based on the order of the phase wire connections.
- Running `dshot reverse -m 1` reverses the motor's direction.
- Repeating `dshot reverse -m 1` has no further effect.
- After a reboot, the motor returns to its normal state unless `dshot save -m 1` was used.
- If saved, the motor remains in reverse until switched back with `dshot normal -m 1` and potentially saved.

Note: These commands have no effect while the motor is spinning.

THe idea of this pr was to correctly describe this behavior. What else do we need to change to accommodate? 🤔👀

markusachtelik: Thanks @MaEtUgR ! what you describe is exactly the behavior we observed as well. So with the fixed documentation, all should be fine.

hamishwillee: Thanks very much for the exhaustive testing - let's just hope that this is not ESC specific :-).

I've fixed it up now so that it is clear that the meaning of normal and reverse do not change after saving, that not saving means direction resets on reboot, and that calling commands while motors spinning or setting the direction to the current direction has no effect.

PR #3316, opened 2024-07-31
[v1.15] Kill gesture docs

hamishwillee:
This add docs and release note for the kill gesture added in https://github.com/PX4/PX4-Autopilot/pull/18509.

@MaEtUgR Can you please sanity check this. I am assuming, for example, that this applies to all frame types.

FWIW I think the difference between disarm, kill, and flight termination need to be clarified.  My understanding is that 
- arm - means powered and enabled.
- disarm - means motors unpowered (and by definition disabled)
- "kill" means turn off the motors, but you're still armed - so motors are powered but disabled.
- "flight termination" means motors are turned off but other outputs locked to certain values - so this is like kill but you also might deploy a parachute.

Is that about right? The reason I find it confusing is that there are so many things that seem to achieve "kind of" the same thing. As a user, should I disarm? Should I kill? Should I flight terminate? Why? When?

To me kill and disarm previously meant recoverable and unrecoverable motors off. Now you've added a kill gesture that isn't recoverable I don't know why you wouldn't just disarm? Then why not a flight termination switch so I can deploy a parachut.?

- [ ] When this goes in, push to main too.


hamishwillee:
@MaEtUgR Can you please sanity check this?

en/config/safety.md
```diff
@@ -321,9 +321,17 @@ This section lists the available emergency switches.
...
-The motors will restart if the switch is reverted within 5 seconds.
-After 5 seconds the vehicle will automatically disarm; you will need to arm it again in order to start the motors.
+A kill switch immediately stops all motor outputs — if flying, the vehicle will start to fall!
+
+[By default](#COM_KILL_DISARM) the motors will restart if the switch is reverted within 5 seconds, after which the vehicle will automatically disarm, and you will need to arm it again in order to start the motors.
+
+| Parameter                                                                                                | Description                                                                                |
+| -------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------ |
+| <a id="COM_KILL_DISARM"></a>[COM_KILL_DISARM](../advanced_config/parameter_reference.md#COM_KILL_DISARM) | Timeout value for disarming after kill switch is engaged. Default: `5` seconds (Disabled). |
```
MaEtUgR, line 330:
```suggestion
| <a id="COM_KILL_DISARM"></a>[COM_KILL_DISARM](../advanced_config/parameter_reference.md#COM_KILL_DISARM) | Timeout value for disarming after kill switch is engaged. Default: `5` seconds. |
```
I don't get it, what does "Disabled" mean here?

MaEtUgR (approved):
Sorry, totally fell through the cracks. Thanks for taking care of documenting this feature! 🙏
I think I only found a small copy paste error or a detail I don't understand.

en/config/safety.md
hamishwillee: Yes, copy paste error! Thanks.

PR #3333, opened 2024-08-07
Assembly - add distance sensor section

hamishwillee:
@sfuhrer This adds distance sensor and optical flow to assembly guide. The sections are rendered according to vehicle type. 

This is OK as far as it goes, but I'll add some questions/comments inline.

### DIFF
--- a/en/assembly/_assembly.md
+++ b/en/assembly/_assembly.md
@@ -163,6 +163,31 @@ You can use an I2C level converter to connect these devices to a Pixhawk flight 
 
 </div>
 
+## Distance Sensor
+
+<div v-if="(($frontmatter.frame === 'Multicopter') || ($frontmatter.frame === 'VTOL'))">
+
+[Distance sensors](../sensor/rangefinders.md) can significantly improve vehicle robustness and performance, and are required for some use cases:
+
+- Landing can be much improved with a distance sensor
+  - Land detector more robust
+  - Smoother landing it can help detect right point to slow down vehicle before touchdown
+  - Reduce risk of very hard touchdown due to bad altitude estimate or incorrectly set touch point altitude
+- Enables terrain following
+- Required for robust state estimation when flying with gnss-denied navigation (along with optical flow)
+
+</div>
+<div v-if="$frontmatter.frame === 'VTOL'">
+
+- Allows disabling of "pusher assist" when close to the ground.
+
+</div>
+<div v-if="$frontmatter.frame === 'Plane'">
+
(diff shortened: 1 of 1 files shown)


en/assembly/_assembly.md
```diff
@@ -163,6 +163,50 @@ You can use an I2C level converter to connect these devices to a Pixhawk flight
...
+<div v-if="$frontmatter.frame === 'Plane'">
+
+[Distance sensors](../sensor/rangefinders.md) are highly recommended as they allow for proper flaring during landing, without which smooth automated fixed-wing landings are near-impossible.
+
+</div>
+
+Unlike for some other components, there is no particular bus or port that is commonly used for distance sensors.
+Different rangefinders will connect via I2C, CAN, serial ports, and even PWM inputs!
+
+See [Distance sensors](../sensor/rangefinders.md) and manufacturer documentation for instructions on how to integrate a specific sensor with PX4.
```
hamishwillee, line 194: This is all I say about distance sensors, and I say similar things for optical flow sensors. However I have some questions that basically come down to "Is there general rule of thumb guidance for choosing good components of these?"

- What is the basis for selecting a distance sensor that is NOT to be used for Optical flow? 
- E.g. what distance on the rangefinder min/max would you be after? What would PX4 want for:
  - smooth landing MC? 
  - What about for FW? 
  - What about for pusher assist?
- Are there other criteria you need to look at? Are sonar ones any good? What about laser, or radar altimiter?
- At what point do max/min values not matter?
- What else should we mention? Weather proofing?

What about for optical flow systems? Is there any way to differentiate them in terms of technologies and ranges?
If you get one that comes without a distance sensor, is there some range on the sensor at which there is no point in getting more range?

sfuhrer: >What is the basis for selecting a distance sensor that is NOT to be used for Optical flow?

I would highlight range and robustness in the environment where it's used - e.g. a laster range finder is usable over glass.

>E.g. what distance on the rangefinder min/max would you be after? What would PX4 want for:
smooth landing MC? --> 10m+
What about for FW? --> 20m+
What about for pusher assist? --> 10m+

>Are there other criteria you need to look at? Are sonar ones any good? What about laser, or radar altimiter?

I personally only have experience with laser ones.

>At what point do max/min values not matter?

You mean when do you have to set it and when not? AFAIK it's of most importance when using the distance sensor together with optical flow. Not sure if you need to set it otherwise.

Maybe @haumarco has some more ideas how we can help with sensor selection.

sfuhrer (approved):
Looks correct to me, I would get this in independently of the "how to chose your distance sensor" selection.

hamishwillee:
Thanks @sfuhrer . Makes sense. Merging.

PR #3340, opened 2024-08-08
Path planning need an onboard computer with heartbeat streaming

hamishwillee:
This makes is to help people who have the `Avoidance system not ready` prearm message have an easy way to determine why that  is.

I have inferred the requirement from https://github.com/PX4/PX4-Avoidance/issues/568 and then looked at the source.

FYI, the code is not MAVLink compliant because it hard codes the check on the component id of the companion computer rather than on the type `MAV_TYPE_ONBOARD_CONTROLLER`

Fixes https://github.com/PX4/PX4-Autopilot/issues/22994

(since that is not a bug - it's a feature)

### DIFF
--- a/en/computer_vision/obstacle_avoidance.md
+++ b/en/computer_vision/obstacle_avoidance.md
@@ -60,6 +60,9 @@ If PX4 stops receiving setpoint updates for more than half a second it will swit
 
 Obstacle avoidance is enabled within PX4 by [setting](../advanced_config/parameters.md) the [COM_OBS_AVOID](../advanced_config/parameter_reference.md#COM_OBS_AVOID) to 1.
 
+Note that PX4 also requires that the companion computer meets the requirements of the [Path Planning Interface](../computer_vision/path_planning_interface.md).
+For example, the vehicle will fail the prearm check `Avoidance system not ready` if the companion computer isn't broadcasting a MAVLink `HEARTBEAT` with `type` set appropriately.
+
 ::: info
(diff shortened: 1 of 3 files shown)


Jaeyoung-Lim:
@hamishwillee Thanks for taking the initiative.

It is a bit unfortunate, but currently px4-avoidance is not being maintained and therefore I have been asking to remove the functionality completely from px4 until somebody makes sure everything is working. what would be your opinion on this?

hamishwillee:
> It is a bit unfortunate, but currently px4-avoidance is not being maintained and therefore I have been asking to remove the functionality completely from px4 until somebody makes sure everything is working. what would be your opinion on this?

I think we need to set expectation appropriately.

I don't know enough to make an informed opinion on whether we should also remove the code.

- How much harder is it for "someone" to get this working once it is removed from PX4? I mean then you'd need to re-add to PX4 before you could even start thinking about your companion side.
- Was the architecture "good", "suboptimal" - would we do it different if we started again?
- What does it cost us to have the code sitting around now?
- Can we get @mrpollo to see if any of the dronecode members want to sponsor getting this all working again ... and next time with the CI to keep it going?

So depending on your answer:

- if we think that this about the right architecture and that the cost of keeping the code is low, then we should mark all of it as experimental to set expectations appropriately.
- If we think the architecture is flawed, and someone would be better starting from scratch, and we have a good idea what that would look like, then we remove it as you suggest.

hamishwillee:
@Jaeyoung-Lim But I'm going to merge this anyway as "correct for now".

hamishwillee:
PS @Jaeyoung-Lim Very happy to add "experimental" tagging and a note on the docs for this, but I would like to understand your answers to the "how good is architecture" side of things before doing that.

PR #3348, opened 2024-08-15
Path planning/obstacle avoidance - deprecation/removal

hamishwillee:
PX4 is removing support for the features Obstacle Avoidance in Missions and Safe Landing, and the underling MAVLink Path Planning interface, as discussed in https://discord.com/channels/1022170275984457759/1270983027429736581

Specifically, the ROS-side interfaces are not supported or maintained, and haven't been for some years. The features can't be used out of the box (or even with significant effort) but that isn't obvious. So what we're doing is removing the documentation for the feature except for basic information about the removal. Note the actual code is still present, so could be resurrected.

This removes as many links to the removed features as possible. The Obstacle/Safe landing/Path Planning docs are removed from the sidebar and redirected to the Path Planning Interface. The path planning interface doc has been cut back to include information about what has changed so that we have a place to point people to, or that they will find by search. But it isn't in the sidebar.

There are still mentions in the PX4 Autonomy Vision Kit. I have put a clear warning there about this software not being used, but that it is still useful.

Also added a release note


mrpollo (changes requested):
Let's clarify that there is nothing inherently wrong with PX4 that makes building a collision prevention or avoidance solution. We should also point out that this feature was deprecated due to architectural constraints of the implementation, which has led to poor adoption from the community, making it hard to maintain.

hamishwillee:
> Let's clarify that there is nothing inherently wrong with PX4 that makes building a collision prevention or avoidance solution. We should also point out that this feature was deprecated due to architectural constraints of the implementation, which has led to poor adoption from the community, making it hard to maintain.

Yes to the first bit - there isn't anything wrong with PX4 - you'd have these same problems on any system. 

I don't believe the second part. No one adopted it because it was too hard. That's nothing to do with the architecture - it's to do with complexity of ROS2 and computer vision. For adoption this would have had to be turnkey, and then you would have had lots of adoption.

hamishwillee:
@mrpollo OK, I've added some appropriate words in https://github.com/PX4/PX4-user_guide/pull/3348/commits/80b81657e0ffa0a2de5e52262ac4c0c849414068

Along the lines of:

>  Note that this code was abandoned due to architectural constraints of the implementation making it hard to maintain, extend, and adopt.
  PX4 now advocates more generic and scalable approaches for integrating these kinds of features, such as using the [PX4 ROS 2 Interface Library](../ros2/px4_ros2_interface_lib.md) to replace the PX4 flight modes with enhanced versions written using ROS 2.

I don't love it, because I think this will be just as hard or harder than the previous approach. On the other hand, the text is true, and the approach is pretty cool.

FYI at some point Beat was talking about putting in place a CI solution that allows running a PX4 simulation/companion side simulation. That's a precondition to having contributed companion software as part of PX4, because it allows _ongoing_ maintenance.

mrpollo (approved):

PR #3352, opened 2024-08-17
Kakute H7 - update UART2 workaround

iftahnaf:
Hi,

After discussing with @vincentpoont2 - PX4/PX4-Autopilot#23553, there is a workaround to use UART2 on the KakuteH7.
Since the `TELEM2` serial connection is already configured in the `default.px4board`, it's worth mentioning this workaround


vincentpoont2:
Thank you @iftahnaf . 

@hamishwillee should be good to merge.

en/flight_controller/kakuteh7.md
```diff
@@ -120,11 +119,19 @@ In addition to the [basic configuration](../config/index.md), the following para
...
+| USART2 | /dev/ttyS1 | TELEM2                |
 | USART3 | /dev/ttyS2 | Debug Console         |
 | UART4  | /dev/ttyS3 | GPS1                  |
 | USART6 | /dev/ttyS4 | RC SBUS               |
 | UART7  | /dev/ttyS5 | ESC telemetry (DShot) |
 
+### UART2 to TELEM2 Workaround
+
+Since UART2 is used by Bluetooth telemetry (which doesn't work with PX4), there are no exposed solder pads on the board.
+The workaround is as follows, the two resistors marked with an X needs to be removed.
```
hamishwillee, line 131: It is better not to talk about this as a "workaround" and more as a "solution to a problem".
Is this correct? In particular, is any other configuration required in PX4?

```suggestion
### Using TELEM2 (USART2)

The `TELEM2` port (USART2) has no exposed solder pads as it is intended for use with Bluetooth telemetry (this does not work with PX4).

You can expose the solder pads and use the port by removing the two resistors marked with an X.
No other configuration should be required.
```

hamishwillee:
Thanks @iftahnaf and @vincentpoont2 . Much appreciated!

PR #3355, opened 2024-08-21
Tiltrotor format fix 

Claudio-Chies:
links were on the same line


hamishwillee (approved):
You're absolutely right. I moved the second link (and first one) though to new section, because there was no video for it.

PR #3356, opened 2024-08-22
MC mode is stabilized mode

hamishwillee:
We referred to it as Manual/Stabilized because Manual maps to stabilised on this frame if you select the mode.
However it is possible to have other types of manual modes with stabilisation, or with different stabilization.

This names the mode Stabilized everywhere. In the top level it explains that the mode manual is mapped.

This falls out of this discussion: https://github.com/PX4/PX4-Autopilot/pull/23544#issuecomment-2302163156

FYI only @sfuhrer


de/_sidebar.md
```diff
@@ -23,6 +23,7 @@
     - [Throw Launch](/flight_modes_mc/throw_launch.md)
   - [Assembly](/assembly/assembly_mc.md)
   - [Configuration/Tuning](/config_mc/index.md)
+    - [Auto-tune](/config/autotune_mc.md)
```
hamishwillee, line 26: FYI, all the _sidebar changes are just tacked on here. Ignore.

sfuhrer:
Neat thanks! Let's see if we at some point can remove the arbitrary mapping from Manual-->Stabilized and just don't support Manual in Hover mode.

PR #3357, opened 2024-08-22
Actuator outputs fixes / flaps

hamishwillee:
Firstly this cleans up a few minor issues with the actuator docs - adding gripper and landing gear outputs that were added in v1.14.

There is more work to be done on this to address issue with flaps testing raised in https://discord.com/channels/1022170275984457759/1255605125943398481/1275732677458399324


en/config/actuators.md
```diff
@@ -528,10 +529,15 @@ For each of the control surfaces:
 1. Set the `Disarmed` value so that the surfaces will stay at neutral position when disarmed.
    This is usually around `1500` for PWM servos.
 2. Move the slider for the surface upwards (positive command) and verify that it moves in the direction defined in the [Control Surface Convention](#control-surface-deflection-convention).
+
+   For example, an aileron would be expected to move up if the slider moves up, while a rudder would move right.
```
hamishwillee, line 533: @ryanjAA I've been trying to shoe-horn in the information about flaps testing here. I know it doesn't fit for now, but if captured I can pull it out better later. In the [linked discussion](https://discord.com/channels/1022170275984457759/1255605125943398481) you said:

> That actually answers this. You’re right. When you setup any plane you need to ensure correct servo orientation (aileron stick left means left aileron up, right aileron down, etc). Should be same here but since no flaps output on RC you have to do it actuators slider (which is totally fine). 
>
> **If flaps do not deflect downwards when raising the flaps channel slider(s), the Rev Range checkbox should be selected.**

However the sliders move control surfaces in a positive direction, where that is defined as upward for horizontal controls in [Control Surface Deflection Convention](https://docs.px4.io/main/en/config/actuators.html#control-surface-deflection-convention), which is the opposite of what you suggested.

So I **think** it would be correct to say (can you check):

> If the flap connected to the output does not move upwards with the corresponding slider, then it reverse the Rev range.
> Note that in some vehicle builds flaps may be configured such that both flaps are controlled from a single output. In this case, if using the `LFlap` output (say) then the left flap should deflect up, and the right flap should deflect down.

ryanjAA: But doesn't up on the slider go to 2000us? Mid is 1500 and down is 1000? (unless you have the rev checked or intentionally have min and max backwards?)

> If the flap connected to the output does not move upwards with the corresponding slider, then it reverse the Rev range.
Note that in some vehicle builds flaps may be configured such that both flaps are controlled from a single output. In this case, if using the LFlap output (say) then the left flap should deflect up, and the right flap should deflect down.

I think we need to say deployed and not deployed.
AKA

If the flap connected to the output does not extend/deploy (as you would expect on landing, not at cruise) with the corresponding slider, then reverse it via the Rev range check box.
Note that in some vehicle builds, flaps may be configured such that both flaps are controlled from a single output. In this case, you need to ensure that both flaps extend/deploy when raising the corresponding slider. If this is not the case and one servo deploys correctly and one does not, you need to change the servo direction with a third party servo programmer or move the servo (that is not deflecting in the correct orientation) to its own servo output channel and then then reverse its direction via the Rev range check box.


Thats a lot of words for somethings simple but I think encompasses it.

hamishwillee: Thanks!

> But doesn't up on the slider go to 2000us? Mid is 1500 and down is 1000? (unless you have the rev checked or intentionally have min and max backwards?)

Hmmm, the docs say positive is up. This makes sense instructionally because the instructions always will be the same whichever way you mount your servos. Is it not possible that the system works with a scaled up and down once you finish configuration and doesn't care about the actual PWM values? Or to put it another way, if you configure it following those instructions the system will always drive the flaps in the direction it is supposed to?

@sfuhrer So you know how the actuator sliders work? Is "up" abstracted, or is it always that moving the slider up increases the PWM value (unless reversed). The docs indicate that the actuators should follow the slider up and down to be "correct".

> > If the flap connected to the output does not move upwards with the corresponding slider, then it reverse the Rev range.
> > Note that in some vehicle builds flaps may be configured such that both flaps are controlled from a single output. In this case, if using the LFlap output (say) then the left flap should deflect up, and the right flap should deflect down.
> 
> I think we need to say deployed and not deployed. AKA
> 
> If the flap connected to the output does not extend/deploy (as you would expect on landing, not at cruise) with the corresponding slider, then reverse it via the Rev range check box. Note that in some vehicle builds, flaps may be configured such that both flaps are controlled from a single output. In this case, you need to ensure that both flaps extend/deploy when raising the corresponding slider. If this is not the case and one servo deploys correctly and one does not, you need to change the servo direction with a third party servo programmer or move the servo (that is not deflecting in the correct orientation) to its own servo output channel and then then reverse its direction via the Rev range check box.
> 
> Thats a lot of words for somethings simple but I think encompasses it.

If it is correct I'm happy with that, except what do you mean "not at cruise"?

Any comment @sfuhrer 


Note, I'm starting to think about splitting out vehicle specific docs/tutorial for plane/vtol config so we can talk a bit more carefully about things like Flaps.

sfuhrer: >@sfuhrer So you know how the actuator sliders work? Is "up" abstracted, or is it always that moving the slider up increases the PWM value (unless reversed). The docs indicate that the actuators should follow the slider up and down to be "correct".

It's abstracted. Slider up should move the surfaces up/right, otherwise the output needs to be inverted.


>except what do you mean "not at cruise"?

sfuhrer (approved):

en/config/actuators.md
hamishwillee: @sfuhrer @ryanjAA  I slightly improved the existing wording 

image


Sorry, but I'm still not sure I have this quite right.
The image at the top of the [Control Surface Setup](https://docs.px4.io/main/en/config/actuators.html#control-surface-setup) is like this:



That's good for a rudder or aileron that deflect in either direction from a neutral point.

Are flaps the same. My understanding was that they were primarily for lift and only have one direction: when flying they are retracted. When landing they are  deployed "out and down". Further, unlike ailerons which move in opposite directions on each side, the flaps on left and right move in the same direction. 

What that would mean is that the image above would be wrong for flaps, as the disarmed value would be (presumably) the minimum PWM.

More important, it would mean that flaps don't actually move up. In the neutral position you'd expect them to be retracted, and in the deployed position you'd expect them to move down. Is that right?
SO our instructions to test the surface won't work - moving the slider up can be done, but "positive" means nothing if the control can only move down.

So for flaps what kind of instructions make sense? 

Depending on the answer I'll re-jig the comment about dual flaps control from one output.

hamishwillee: @sfuhrer That is contrary to the [Control surface deflection convention](https://docs.px4.io/main/en/config/actuators.html#control-surface-deflection-convention) which defines positive as up or right, and negative as the reverse. I didn't come up with that, but I've been asserting it as true. Bloody annoying.

So to be clear, 
- these are the controls we say we support: LeftAileron, RightAileron, Elevator, Rudder, Left Elevon, Right Elevon, Left V-Tail, Right V-Tail, Left Flap, Right Flap, Airbrakes, Custom.
- Other than flaps, do all of these move either way around the neutral position and obey the convention? I know ailerons do, what about airbrakes, vtails, elevators?
- If not, which are exceptions, and how do they move?

> Just talking about **flaps** now ...

It sounds like flap servos are set up similar to aileron servos. The disarmed value is the center of the PWM range. So in theory you can apply a positive and negative deflection by applying more or less PWM than the centre value.
The difference for a flap is that increasing the slider above the centre should lower the flap, and below the centre should do nothing. Right?

For above we have min (1000), disarmed (1500), max (2000). 
1. My thinking is that you've thrown away half the range of your servo - i.e. it could obey 1000-1500 say, but it won't reverse. Does that matter?
2. Could you set this up as `min (1000), disarmed (1100), max (2000)` to get more of the range? In other words, my assumption is that the system will always split the range around the disarmed value? Or,
3. Could I do it like a motor `min (1000), disarmed (900), max (2000)` - what happens in this case?


KNowing the above makes this easier to write, because Flaps are an exception and have to be written as such.

hamishwillee: PS I've asked for access to your video from summit. That might answer some or all of this. Either way, this reinforces the need to split this doc up a bit.

sfuhrer: >contrary to the [Control surface deflection convention](https://docs.px4.io/main/en/config/actuators.html#control-surface-deflection-convention) which defines positive as up or right

True, I didn't have the flaps in mind when defining these. Surfaces that conceptually do only have one direction from "no effect" to "full effect" cannot be actuated with a negative input, and thus there is no need for a convention (as long as it is clear that they're not part of any, which atm is not the case).


> LeftAileron, RightAileron, Elevator, Rudder, Left Elevon, Right Elevon, Left V-Tail, Right V-Tail, Left Flap, Right Flap, Airbrakes, Custom.

Ailerons, Elevons, V-tails, A-tails: up positive, down negative
Rudder: right positive
Airbrakes, Spoiler, Flaps: 0: no effect, 1: full effect

>The difference for a flap is that increasing the slider above the centre should lower the flap, and below the centre should do nothing. Right?

Yes correct.

>My thinking is that you've thrown away half the range of your servo - i.e. it could obey 1000-1500 say, but it won't reverse. Does that matter?

Yes you throw away servo range, and with it you'll need a stronger servo to deflect the flap. I guess it does matter if you optimize your hardware design.

>Could you set this up as min (1000), disarmed (1100), max (2000) to get more of the range? In other words, my assumption is that the system will always split the range around the disarmed value? Or,

Having a different disarm than MIN or MAX for flaps probably doesn't make sense, but yes, if you have the hardware setup to exploit the full servo range then MIN should be around 1000 and MAX around 2000. 

>Could I do it like a motor min (1000), disarmed (900), max (2000) - what happens in this case?

I would not do that. You should set your MIN and MAX according to the flap range, and then set the disarm to either of them (whichever is the one to have the flap flush with the wing). Otherwise you damage the servo as it tries hard to move the flap into a position that it cannot go into.

en/config/actuators.md
```diff
@@ -227,6 +228,13 @@ In summary:
 - **Mixed Control Surfaces:** Upwards/rightwards movement is positive (as above).
   Includes V-Tail etc.
 
+::: tip
+Control surfaces that can only deflect in one direction from the neutral point include: Airbrakes, Spoiler, and Flaps.
+
+For these a positive input is always deflection from the neutral (0: no effect, 1: full effect), irrespective of the direction that the control itself moves.
+These do not respond to negative input.
+:::
+
```
hamishwillee, line 237: FYI @sfuhrer This is new text for the convention. This is intended to make it clear what controls the convention applies to, and that Airbrakes, spoilers, flaps are different.

en/config/actuators.md
```diff
@@ -523,15 +532,29 @@ If you're using PWM servos, PWM50 is far more common.
...
+   - Ailerons, elevons, V-tails, A-tails, and other horizontal surfaces should move up.
+   - Rudders and other "purely vertical" surfaces should move right.
+
+   ::: tip
+   It is important that the slider movement matches the control surface convention, in order to normalize control for different servo mountings (moving the slider up may actually decrease the output value sent to the servo).
+   :::
+
+   If the control surface moves in the opposite direction, click on the `Rev Range` checkbox to reverse the range.
+
+3. Move the slider again to the middle and check if the Control Surfaces are aligned in the neutral position of the wing.
```
hamishwillee, line 554: @sfuhrer Note, we only talk about the fact that they go to the middle. Do we have to worry about the servo having too great a range (i.e. to we need to configure max/min at all?

en/config/actuators.md
```diff
@@ -540,16 +563,33 @@ For each of the control surfaces:
...
+
+For these controls you should set the minimum and maximum PWM values according to the full range of the control.
+The `Disarmed` value should then match the value (of maximum or minimum) that corresponds to control being in the "neutral" position.
+For a flap, that is when the flap is fully retracted and flush with the wing.
+
+One approach for setting these up is:
+
+1. Set the `Disarmed` value so that the surfaces are at neutral position when disarmed.
+2. ...
+3. Trim?
```
hamishwillee, line 586: @sfuhrer I decided that the controls that deflect in only one direction from neutral are sufficiently different to warrant a new subsection. That way we can have a note for flaps and we don't have to assume the centered disarmed position.

But I can't write this, because I can't try it. I mean I have put your note above and the special note below, but I don't know the best practical steps for this?

My guess was that you might 
- set values disarmed/min/max of 1500, 1200, 1700. 
- move slider up and check it moves, and that it is extending. If not, reverse range.
- enable slider, them change `disarmed` PWM until control retracted/flush with wing. If this is closer to min the move min to same value. If close to max, them move max to same value.
- whichever max/min you didn't set to same value as disarmed represents the value when fully extended. So set slider to max, then change either max/min so that the control is fully extended when the slider is at top.

Does that make sense. There may be a safter and less confusing rule!

sfuhrer: This process is spot on. 

>Then what about trimming?

You don't need to trim anything anymore if you made MIN or MAX being flush in step 3)

en/config/actuators.md
sfuhrer: >Do we have to worry about the servo having too great a range (i.e. to we need to configure max/min at all?

Yes that is generally something you have to worry about for each servo configuration.

en/config/actuators.md
sfuhrer: makes sense, looks good

hamishwillee:
@sfuhrer @ryanjAA Thanks for all the help. This is now documented, and I am fairly confident that it is much better :-)

PR #3358, opened 2024-08-22
Robotics/Apps section remove intermediate ROS abstraction

hamishwillee:
This tidies up the Robotics sections:

- Removed DroneKit from the sidebar. It now appears as small section in top level doc only, just to say "don't use this"
- Add MAVSDK to sidebar so it is more visible
- Moved ROS2 and ROS1 to top level of the robotics section, rather than hiding under a top level ROS topic.

  From

   ```
   Ros
      Ros 2
      Ros 1
   DroneKit
   ```
  To
  
   ```
   MAVSDK
   ROS2
   ROS1 (Deprecated)
   ```
- Marked ROS 1 as deprecated (with reasons).
- Top level doc now compares MAVSDK and ROS2. Dronekit and ROS 1 are marked as "Deprecated".

@beniaminopozzan Would be good to get your feedback.


en/ros/offboard_control.md
```diff
@@ -1,6 +1,6 @@
-# Offboard Control
```
hamishwillee, line 1: FYI this doc is first doc in the  Robotics API section. It is supposed to explain how offboard control works/what it is.

The concept is useful but this is so out of date it probably should be deleted. This is added to make sure I look at this topic next week and decide what can be used or to discard it.

en/ros/ros1.md
```diff
@@ -1,14 +1,18 @@
-# ROS 1 with MAVROS
+# ROS 1 (Deprecated)
```
hamishwillee, line 1: FYI I decided that "with MAVROS" is not actually all that interesting any more. I guess it used to be because ROS1 had other transports for other systems, and may this was for those developers to understand why we don't use those. But anyhow, since we're deprecating this it really isn't relevant.

When ROS1 is really deprecated we should make decision about whether it is "supported" anymore.

en/ros2/index.md
```diff
@@ -1,38 +1,43 @@
...
-To use the [ROS 2](../ros2/user_guide.md) over XRCE-DDS effectively, you must (at time of writing) have a reasonable understanding of the PX4 internal architecture and conventions, which differ from those used by ROS.
-In the near term future we plan to provide ROS 2 APIs to abstract PX4 conventions, along with examples demonstrating their use.
-
-The main topics in this section are:
-- [ROS 2 User Guide](../ros2/user_guide.md): A PX4-centric overview of ROS 2, covering installation, setup, and how to build ROS 2 applications that communicate with PX4.
-- [ROS 2 Offboard Control Example](../ros2/offboard_control.md)
-
 ::: info
-ROS 2 is officially supported only on Linux platforms.
-Ubuntu 20.04 LTS is the official supported distribution.
```
hamishwillee, line 24: FYI decided not to mention this here. That way we can maintain support info that we want in the guide.

beniaminopozzan (commented):
Thanks @hamishwillee ,
This is a nice step toward better ROS 2 documentation!

I was worried about not finding any reference to the ROS 2 version of MAVROS, but there they are!

en/robotics/index.md
```diff
@@ -30,17 +30,45 @@ The main difference are:
...
   - Runs on Linux
-  - ROS 2 is the latest version, which connects via DDS.
+  - ROS 2 is the latest version, which connects via XRCE-DDS.
     - DDS interface layer allows deep integration into any aspect of PX4 that is exposed as a UORB topic (almost everything).
     - Can provide much lower latency.
-    - Still under development. At time of writing requires a deeper understanding of PX4 than ROS 1
-  - ROS 1 is the older version that can connect over MAVLink or via ROS 2.
-    Users are encouraged to update to ROS 2 for new projects.
+    - Still under development:
+      - At time of writing requires a deeper understanding of PX4 than ROS 1.
```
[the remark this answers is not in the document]
hamishwillee: @beniaminopozzan No. This says ROS2 requires a deeper understanding of PX4 than ROS 1. Am I missing something?

en/ros/offboard_control.md
```diff
@@ -14,7 +14,7 @@ There are two things you want to setup on the firmware side before starting offb
 
 ### Map an RC switch to offboard mode activation
 
-To do this, load up the parameters in *QGroundControl* and look for the RC_MAP_OFFB_SW parameter to which you can assign the RC channel you want to use to activate offboard mode.
+To do this, load up the parameters in _QGroundControl_ and look for the RC_MAP_OFFB_SW parameter to which you can assign the RC channel you want to use to activate offboard mode.
```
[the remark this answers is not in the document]
hamishwillee: Thanks. Slight mod.

```suggestion
In _QGroundControl_ you can modify the [RC_MAP_OFFB_SW](../advanced_config/parameter_reference.md#RC_MAP_OFFB_SW) parameter to set the RC channel that will be used to activate offboard mode.
```

en/ros/offboard_control.md
```diff
@@ -14,7 +14,7 @@ There are two things you want to setup on the firmware side before starting offb
 
 ### Map an RC switch to offboard mode activation
 
-To do this, load up the parameters in *QGroundControl* and look for the RC_MAP_OFFB_SW parameter to which you can assign the RC channel you want to use to activate offboard mode.
+In _QGroundControl_ you can set the [RC_MAP_OFFB_SW](../advanced_config/parameter_reference.md#RC_MAP_OFFB_SW) parameter to the RC channel that will be used to activate offboard mode.
 It can be useful to map things in such a way that when you fall out of offboard mode you go into position control.
```
hamishwillee, line 18: @beniaminopozzan This doesn't make much sense - why does this help you go into position control when you fall out of offboard mode? All it does is say that you can enable/disable offboard mode?
Or are you saying this is an easy way to disable offboard mode if needed, and by default you'll fall into position mode?

en/ros/offboard_control.md
```diff
@@ -14,7 +14,7 @@ There are two things you want to setup on the firmware side before starting offb
...
 
 Although this step isn't mandatory since you can activate offboard mode using a MAVLink message.
```
hamishwillee, line 20: @beniaminopozzan I think you're saying here that you don't need to map a switch because you can do so in other ways? I.e. using GCS or MAVLink? 

Depending on your answer I think I'll modify this as a post process.

hamishwillee:
Thanks for the review @beniaminopozzan . I'll merge but I've added a few questions for "post processing".

en/robotics/index.md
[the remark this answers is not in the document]
beniaminopozzan: oh yes, yes. My bad, I misunderstood.

en/ros/offboard_control.md
beniaminopozzan: @hamishwillee , it is indeed a bit misleading.
The RC switch allows you to enter and exit offboard mode, when lo leave it you'll return to the mode set by the mode switch in the RC which is not necessary position mode.
So line 18 should be "It can be useful to map things in such a way that when you fall out of offboard mode you go into **manual** control."
An even better approach is to use https://docs.px4.io/main/en/advanced_config/parameter_reference.html#COM_RC_OVERRIDE so the user just need to move the sticks to exit offboad though.

en/ros/offboard_control.md
beniaminopozzan: @hamishwillee I was not saying that but you are right, you can enter offboard mode via GCS, via MAVLink command, via ROS 2 and of course via RC.

en/ros/offboard_control.md
hamishwillee: Thank you! See https://github.com/PX4/PX4-user_guide/pull/3372

PR #3360, opened 2024-08-22
Realtime Debugging

Claudio-Chies:
added a section on realtime debugging with plotjuggler and ROS2.

- [x] Test the prerequisites and if everything needed is covered


Claudio-Chies:
@hamishwillee would be happy for a quick feedback regarding the location. as its more for debugging than for ROS. i thought this is the right place.

hamishwillee:
Thank you. ON list to look at tomorrow.

hamishwillee:
@Claudio-Chies This is extremely cool. I think it is a good place for the document, but I have cross linked from the uORB messaging topic so that people are more likely to find it. 

I've given this quite a big subedit. 

The big hole I see at the moment is that is isn't clear how PlotJugger gets access to the PX4 definitions. My assumption is that you build a ROS 2 workspace but that isn't mentioned. 

This will need a re-review from you for accuracy. Once you're finished we'll also get a look from Benja.

Claudio-Chies:
@hamishwillee 
Plotjuggler and ROS doesnt know uORB. this is why we need the message definitions in [px4_msgs](https://github.com/PX4/px4_msgs).
so what we are displaying arent the uORB messages, but their ROS counterpart. as the definitions are (from what i can see) 1:1 the same, its a bit the question if we want to talk about uORB messages in the scope of plotjuggler and ros, or about ROS messages.

en/middleware/uorb.md
```diff
@@ -120,8 +119,15 @@ sensor_accel                         1    1  249    43 1
 sensor_baro                          0    1   42     0 1
 sensor_combined                      0    6  242   636 1
 ```
+
 The columns are: topic name, multi-instance index, number of subscribers, publishing frequency in Hz, number of lost messages per second (for all subscribers combined), and queue size.
 
+## Plotting realtime changes in topics
```
hamishwillee, line 125:
```suggestion
## Plotting Changes in Topics
```

en/middleware/uorb.md
```diff
@@ -120,8 +119,15 @@ sensor_accel                         1    1  249    43 1
...
+
+Topic changes can be plotted in realtime using PlotJuggler and the PX4 ROS 2 integration.
+For more information see: [Realtime uORB Topic Debugging](../debug/realtime_uorb_debugging.md).
```
hamishwillee, line 128:
```suggestion
Topic changes can be plotted in realtime using PlotJuggler and the PX4 ROS 2 integration (note that this actually plots ROS topics that correspond to uORB topics, but the effect is the same).

For more information see: [Realtime uORB Topic Debugging](../debug/realtime_uorb_debugging.md).
```

hamishwillee:
> @hamishwillee
Plotjuggler and ROS doesnt know uORB. this is why we need the message definitions in [px4_msgs](https://github.com/PX4/px4_msgs) and then do a colcon build.

Yes, I fully understand the interactions between uorb, xrcedds, the client, the agent, and a ROS workspace :-). The only thing I didn't understand was that you're running plotjuggler as a ROS node (effectively) in a workspace that you've build the px4_messages in. Now that's cleared up how this works is clean.

>what we are displaying arent the uORB messages, but their ROS counterpart. as the definitions are (from what i can see) 1:1 the same, its a bit the question if we want to talk about uORB messages in the scope of plotjuggler and ros, or about ROS messages. i think ros messages might be more accurate, but cause more confusion?

IMO there isn't a great deal of confusing (though I just added a note in the uORB topic to clarify this point). We're logging the ROS representation of internal uORB topics. I can add a note to further clarify this in the main doc if you want, but not until next week.

The mapping is more or less 1:1 in this case "as a reader". If your publishing to at topic in PX4 things can get more complex.

Claudio-Chies:
@hamishwillee ok, good to know, then you might know more than me :)

hamishwillee:
@Claudio-Chies I think we might rename this. Debugging is also a bit generic. What do you think about "Realtime Charting of uORB topics? or "Dynamic Visualisation of uORB Topics", or some variant using the term "Plotting"? 

Debugging kind of implies you have a problem, but you might use this for all sorts of reasons.

I will go over this again next week. From your perspective, what else does this need?

hamishwillee:
W.r.t. "ROS 2" and uORB I perhaps overstated my understanding. Just FYI In particular you mentioned that perhaps  there is a 1:1 relationship between ROS topics and uORB topics. My understanding is that it is a little more nuanced. 

Specifically, there is a 1:1 relationship outbound from PX4. In other words, if you have a UORB message like [VehicleOpticalFlowVel](https://docs.px4.io/main/en/msg_docs/VehicleOpticalFlowVel.html) it will declare a bunch of topics like this:
```
# TOPICS estimator_optical_flow_vel vehicle_optical_flow_vel
```

Each one of those topics has independent existence in PX4, and if in the yaml, will have a corresponding ROS2 message. So ROS2 subscribers such as plotjuggler in this case do have a 1:1 mapping from uorb.

However it is different if going the other way, where ROS2 is publishing to PX4. In this case it is possible that PX4 is set up to create a **new** uORB topic for the inbound ROS2 messages.
That is needed because otherwise there is no way to differentiate updates from an external ROS source and internal source. 

This is what the stuff about subscriptions_multi means in https://docs.px4.io/main/en/middleware/uxrce_dds.html#dds-topics-yaml.

Claudio-Chies:
> @Claudio-Chies I think we might rename this. Debugging is also a bit generic. What do you think about "Realtime Charting of uORB topics? or "Dynamic Visualisation of uORB Topics", or some variant using the term "Plotting"? 
> 
> 
> 
> Debugging kind of implies you have a problem, but you might use this for all sorts of reasons.
> 
> 
> 
> I will go over this again next week. From your perspective, what else does this need?

Yea, i thought the same but it was what came to my mind.
"Realtime Monitoring of uORB topics" might also be an good fit, what do you think about that?

I'll have to test the dependencies, but for that i need a VM or a fresh maschine, i'll try to get it done by end of next week

hamishwillee:
> Realtime Monitoring of uORB topics

I really liked this idea, but you could argue that just looking at the `uorb top` values is "realtime monitoring". So I settled on **Plotting uORB Topic Data in Real Time using PlotJuggler**

I've given it another subedit. The main issue I had looking at this again is that as a complete newbie, ROS 2 is hard. I didn't want to  duplicate "everything" in the ROS 2 guide, but I wanted a bit more context. I'll add a couple of notes.

en/debug/plotting_realtime_uorb_data.md
```diff
@@ -0,0 +1,117 @@
...
+If you're using unmodified PX4, the definitions from [PX4/px4_msgs](https://github.com/PX4/px4_msgs) can be used.
+
+::: info
+This is the same process as covered in [Build ROS 2 Workspace](../ros2/user_guide.md#build-ros-2-workspace) in _ROS 2 Installation & Setup_.
+:::
+
+Assuming your ROS 2 workspace is `~/ros2_ws/`, fetch and build the `px4_msgs` package in a terminal as shown:
+
+```sh
+cd ~/ros2_ws/
```
hamishwillee, line 45:
```suggestion
cd ~/ros2_ws/src/
```

en/debug/plotting_realtime_uorb_data.md
```diff
@@ -0,0 +1,117 @@
...
+
+::: info
+This is the same process as covered in [Build ROS 2 Workspace](../ros2/user_guide.md#build-ros-2-workspace) in _ROS 2 Installation & Setup_.
+:::
+
+Assuming your ROS 2 workspace is `~/ros2_ws/`, fetch and build the `px4_msgs` package in a terminal as shown:
+
+```sh
+cd ~/ros2_ws/src/
+git clone https://github.com/PX4/px4_msgs.git
```
hamishwillee, line 46:
```suggestion
git clone https://github.com/PX4/px4_msgs.git
cd ..
```

en/debug/plotting_realtime_uorb_data.md
```diff
@@ -0,0 +1,118 @@
...
+cd ..
```
hamishwillee, line 47: I added this, because the instructions in the user guide do NOT use `ros2_ws` as a workspace, and it is a lot for a newbie to remember how they access and use a workspace.

en/debug/plotting_realtime_uorb_data.md
```diff
@@ -0,0 +1,118 @@
...
+colcon build
+source install/setup.bash
```
hamishwillee, line 49: Is this correct? The build guide seems to indicate I should do this in `src`
```
colcon build
```

Then the commands here: https://docs.px4.io/main/en/ros2/user_guide.html#running-the-example

Can you sanity check. The requirement is that this should work for a NEW workspace and fresh terminal after they have already installed ROS etc.

en/debug/plotting_realtime_uorb_data.md
```diff
@@ -0,0 +1,118 @@
...
+```
+
+To start sending ROS 2 topics from PX4 the uXRCE-DDS **client** has to be running on PX4, and the `MicroXRCEAgent` has to be running on the same computer as PlotJuggler.
+If you're using the Simulator then the client is started automatically.
+Assuming you are using the [Gazebo](../sim_gazebo_gz/index.md) simulator, you might start it in another terminal like this:
+
+```sh
+make px4_sitl gz_x500
+```
+
```
hamishwillee, line 65: I added a simulator example here. Never sure how much you should say, but if we want this to be useful for "dummies" then it doesn't hurt to spell out at least one example that should work out of the box by following instructions end to end.

en/debug/plotting_realtime_uorb_data.md
Claudio-Chies: yes, `source /opt/ros/humble/setup.bash` is for the ros environment, but you also need to source your workspace after building.
`source /opt/ros/humble/setup.bash` is not needed here as in the ros2 guide you're adding it to your `.bashrc`
but you could also add it here just in case.
I quickly tested it in a new install, and all dependencies are correct.
Only think I'm adding is to move to the PX4-Autopilot folder before `make px4_sitl gz_x500`

en/debug/plotting_realtime_uorb_data.md
Claudio-Chies: yea, there i'm not sure if its good that we use `ws_sensor_combined` in the ROS2 HowTo, as the officials ros HowTo's always use `ros2_ws `

en/debug/plotting_realtime_uorb_data.md
Claudio-Chies: good idea!

en/debug/plotting_realtime_uorb_data.md
hamishwillee: Yeah. I recall doing this for the reason that it wasn't clear in any of the ROS docs whether ros2_ws was a magic location or  just a convention. I quite like having something in our docs that is different.

en/debug/plotting_realtime_uorb_data.md
hamishwillee: Good idea. I did that.

Also split usage into two subsection sim/hardware. That makes it a little more digestable.

hamishwillee:
Thanks very much for this @Claudio-Chies - I think it's a really useful thing to add. Prior to this you could only really look at a small subset of the data via MAVLink, with all the associated delays from the processing pipeline. If you wanted to look at anything else it was a real pain.

PR #3366, opened 2024-08-28
Update user_guide.md

Kametor:
The gazebo-ros bridge installation step it says:
To install the bridge for use with ROS 2 "Humble" and Gazebo Garden (on Ubuntu 22.04): sudo apt install ros-humble-ros-gzgarden

However the installed developer environment for Ubuntu 22.04 is not gazebo garden it is gazebo harmonic, so version is wrong

### DIFF
--- a/en/ros2/user_guide.md
+++ b/en/ros2/user_guide.md
@@ -464,10 +464,10 @@ Use the following commands to install the correct ROS 2/gz interface packages (n
 :::: tabs
 
 ::: tab humble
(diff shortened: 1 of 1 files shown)


hamishwillee:
Thank you. That happened during the week, so you are very quick. I will try look at this tomorrow but the changes are more significant than this.

hamishwillee:
Thank you, but replacing this with #3367 - and continuing discussion there.

PR #3368, opened 2024-08-29
GPS information - DOP vs EPH/EPV

hamishwillee:
This adds a little information about some of the terminology around GPS information, as discussed in https://discord.com/channels/1022170275984457759/1032055573753118742/1255712384215941130

It is useful to have because there are many questions about this in the community that have popped up over time. It also helps to make clear that PX4 doesn't necessarily get all information that a GPS supplies, or use it. We also point to MAVLink and uORB for the interesting set.

@ryanjAA Look OK to you?

### DIFF
--- a/en/gps_compass/index.md
+++ b/en/gps_compass/index.md
@@ -185,7 +185,7 @@ Some of GNSS terms that are useful for interpreting the data include:
 ### DOP vs EPH/EPV
 
 DOP is a measure of the potential for high accuracy based on satellite positions.
-EPH/EPV are more comprehensive, as they consider both satellite geometry and other error sources like signal noise and atmospheric effects.
+EPH/EPV are more comprehensive: they are direct estimates of the GPS position error and consider both satellite geometry and other error sources like signal noise and atmospheric effects.
 It is possible to have low DOP (good satellite geometry) but still have high EPH/EPV if there is significant signal noise or atmospheric interference.
 
 EPH/EPV values therefore provide a more immediate and practical estimate of the actual GPS accuracy you can expect under current conditions.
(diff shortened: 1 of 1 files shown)


ryanjAA:
Ya, totally good.

I liked this below (change it if you want) in there as well simply because it gives less technical people a high-level basic comparison but all looks good:

DOP is a measure of the geometric quality of satellite positions and their effect on the precision of the GPS receiver's calculations wheras EPH and EPV are direct estimates of the GPS position error.

en/gps_compass/index.md
```diff
@@ -167,6 +167,29 @@ This is documented in [RTK GPS > Configuring GPS as Yaw/Heading Source](../gps_c
...
+  This is a measure of the geometric quality of satellite positions and their effect on the precision of the GPS receiver's calculations.
+- `EPH`: Standard deviation of horizontal position error (metres).
+  This represents the the uncertainty in the GPS fix latitude and longitude.
+- `EPV`: Standard deviation of vertical position error (metres).
+  This represents the the uncertainty in the GPS fix altitude.
+
+### DOP vs EPH/EPV
+
+DOP is a measure of the potential for high accuracy based on satellite positions.
+EPH/EPV are more comprehensive, as they consider both satellite geometry and other error sources like signal noise and atmospheric effects.
```
hamishwillee, line 188:
```suggestion
EPH/EPV are more comprehensive: they are direct estimates of the GPS position error and consider both satellite geometry and other error sources like signal noise and atmospheric effects.
```

hamishwillee:
Thanks @ryanjAA 

I've essentially split your suggested sentence in two, because I found it easier to understand that DOP tells you how accurate the signal could potentially be, but only uses a subset of information. 

> DOP is a measure of the potential for high accuracy based on satellite positions.
> EPH/EPV are more comprehensive: they are direct estimates of the GPS position error and consider both satellite geometry and other error sources like signal noise and atmospheric effects.

PR #3371, opened 2024-09-03
Rover fixes

hamishwillee:
Fix up rover parameter name changes broken by https://github.com/PX4/PX4-Autopilot/pull/23629

@chfriedrich98 Whenever you change/add a param in code, it is worth doing:
1. a search/replace in docs
2. quick check to see this change does not break the flow of text or instructions.
3. For new params, consider if they are worth highlighting in the docs (some are not).

I've done the first two for the param name change, and I have tried to catch the changes for the descriptions (they were OK in the body of the doc). Can you do a quick sanity check for the whole document?


chfriedrich98:
This looks good to me, thanks for the reminder!
I will try to update the docs more quickly after changes have been merged.

PR #3376, opened 2024-09-12
Update rtk_gps.md Heading Info

vincentpoont2:
Add unicore dual antenna heading

### DIFF
--- a/en/gps_compass/rtk_gps.md
+++ b/en/gps_compass/rtk_gps.md
@@ -38,7 +38,7 @@ Device | GPS | Compass | [DroneCAN](../dronecan/index.md) | [GPS Yaw](#configuri
 [Holybro H-RTK F9P Rover Lite](../gps_compass/rtk_gps_holybro_h-rtk-f9p.md) | F9P | IST8310 | | | | |
 [Holybro DroneCAN H-RTK F9P Rover](https://holybro.com/products/dronecan-h-rtk-f9p-rover) | F9P | BMM150 | | | &check; |
 [Holybro H-RTK M8P GNSS](../gps_compass/rtk_gps_holybro_h-rtk-m8p.md) | M8P | IST8310 | | |
-[Holybro H-RTK Unicore UM982 GPS](../gps_compass/rtk_gps_holybro_unicore_um982.md) | UM982 | IST8310 | | | |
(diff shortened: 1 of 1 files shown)


en/gps_compass/rtk_gps.md
```diff
@@ -38,7 +38,7 @@ Device | GPS | Compass | [DroneCAN](../dronecan/index.md) | [GPS Yaw](#configuri
 [Holybro H-RTK F9P Rover Lite](../gps_compass/rtk_gps_holybro_h-rtk-f9p.md) | F9P | IST8310 | | | | |
 [Holybro DroneCAN H-RTK F9P Rover](https://holybro.com/products/dronecan-h-rtk-f9p-rover) | F9P | BMM150 | | | &check; |
 [Holybro H-RTK M8P GNSS](../gps_compass/rtk_gps_holybro_h-rtk-m8p.md) | M8P | IST8310 | | |
-[Holybro H-RTK Unicore UM982 GPS](../gps_compass/rtk_gps_holybro_unicore_um982.md) | UM982 | IST8310 | | | |
+[Holybro H-RTK Unicore UM982 GPS](../gps_compass/rtk_gps_holybro_unicore_um982.md) | UM982 | IST8310 | | Unicore Dual Antenna Heading | |
```
hamishwillee, line 41: @vincentpoont2 This is good, but at the top of the table is a link "GPS Yaw" to https://docs.px4.io/main/en/gps_compass/rtk_gps.html#configuring-gps-as-yaw-heading-source

Do those instructions cover Unicore Dual Antenna Heading setup? (probably not, but we should point to some right?)

vincentpoont2: Maybe we can do something similar to this page?
https://ardupilot.org/copter/docs/common-gps-for-yaw.html#unicore-um982

hamishwillee: Yes, that's why I pointed to https://docs.px4.io/main/en/gps_compass/rtk_gps.html#configuring-gps-as-yaw-heading-source

In the first instance just add two headings, and add the params/setup requirements like this:
image

We can always have a yaw setup doc on its own if we get more information that can sensibly be in a sub section.

Can you sort that?

hamishwillee: @vincentpoont2 If you can, would be great. Note, if you do respond I am away next week, so there would be some wait time.

hamishwillee (approved):
OK, merging as per discussion. May pull the yaw stuff out at some point. TBD.

PR #3384, opened 2024-09-18
SITL:GZ: fixed typo

Claudio-Chies:
the command was wrong, fixed it. and with https://github.com/PX4/PX4-Autopilot/pull/23687 it should then also actually work


hamishwillee (approved):
Thank you!

PR #3385, opened 2024-09-18
Releases: update index page

mrpollo:
It looks like we missed one thing!

### DIFF
--- a/en/releases/index.md
+++ b/en/releases/index.md
@@ -3,7 +3,7 @@
 A list of PX4 release notes, they contain a list of the changes that went into each release, explaining the included features, bug fixes, deprecations and updates in detail.
 
 - [main](../releases/main.md) (changes since v1.15)
-- [v1.15](../releases/1.15.md) <Badge type="info" text="Stable"/>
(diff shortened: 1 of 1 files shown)


en/releases/index.md
```diff
@@ -3,8 +3,8 @@
 A list of PX4 release notes, they contain a list of the changes that went into each release, explaining the included features, bug fixes, deprecations and updates in detail.
 
 - [main](../releases/main.md) (changes since v1.15)
-- [v1.15](../releases/1.15.md) <Badge type="warning" text="Beta"/>
-- [v1.14](../releases/1.14.md) <Badge type="info" text="Stable"/>
+- [v1.15](../releases/1.15.md) <Badge type="info" text="Stable"/>
```
hamishwillee, line 6: Thanks. Have decided to remove this entirely. We note the stable version in other places, and this just gets confusing if we forget to update it when this version becomes v1.16 and later.
Upshot, no notes in these places.
In future I'll try work out a way to extract some of this kind of metadata dynamically from a file in the main branch.


```suggestion
- [v1.15](../releases/1.15.md)
```

PR #3390, opened 2024-09-20
Added MAVProxy in testing of a custom MAVLink message

wasaybaig:
I had a discussion with @hamishwillee that there is another easier way of testing your custom mavlink message. We can use MAVProxy which automatically uses messages in Pymavlink. However, if we clone the mavlink github repo, define our custom message, then generate the headers and then install pymavlink using setup.py in the mavlink repo, we can easily view our custom message on the MAVProxy console.

I find this method was easier and you could test it with MAVProxy without the need of building QGroundControl. Once this is tested on MAVProxy, you import the same headers in QGroundControl and build if required.

### DIFF
--- a/en/middleware/mavlink.md
+++ b/en/middleware/mavlink.md
@@ -500,8 +500,8 @@ There are several approaches you can use to view MAVLink traffic:
   cd pymavlink
   sudo python setup.py install
   ```
-  Now connect your autopilot to MAVProxy by following the [MAVProxy docs](https://ardupilot.org/mavproxy/index.html)
-  In the MAVPRoxy console use the watch command to dislay your custom message
+  Now connect your autopilot to MAVProxy by following the [MAVProxy docs](https://ardupilot.org/mavproxy/index.html).
+  In the shell, use the watch command to dislay your custom message
   ```bash
   watch custom_message_name
   ```
--- a/en/middleware/mavlink.md
+++ b/en/middleware/mavlink.md
@@ -491,7 +491,7 @@ There are several approaches you can use to view MAVLink traffic:
   git clone https://github.com/mavlink/mavlink.git --recursive
   cd mavlink
   ```
-  In the [message_definitions/v1.0](https://github.com/mavlink/mavlink/tree/master/message_definitions/v1.0) add your custom message and then generate the headers
+  In the [message_definitions/v1.0](https://github.com/mavlink/mavlink/tree/master/message_definitions/v1.0) folder, add your custom message and then generate the headers
   ```bash
   python3 -m pymavlink.tools.mavgen --lang=C --wire-protocol=2.0 --output=generated message_definitions/v1.0/your_dialect.xml
   ```
@@ -500,7 +500,7 @@ There are several approaches you can use to view MAVLink traffic:
   cd pymavlink
   sudo python setup.py install
   ```
-  Now connect your autopilot to MAVProxy using by following the [MAVProxy docs](https://ardupilot.org/mavproxy/index.html)
+  Now connect your autopilot to MAVProxy by following the [MAVProxy docs](https://ardupilot.org/mavproxy/index.html)
   In the MAVPRoxy console use the watch command to dislay your custom message
   ```bash
   watch custom_message_name
--- a/en/middleware/mavlink.md
+++ b/en/middleware/mavlink.md
@@ -476,7 +476,36 @@ There are several approaches you can use to view MAVLink traffic:
   :::tip
   It is much easier to generate a wireshark plugin and inspect traffic in Wireshark, than to rebuild QGroundControl with your dialect and use MAVLink Inspector.
   :::
-
+- Use [MAVProxy](https://ardupilot.org/mavproxy/) and [Pymavlink](https://mavlink.io/en/mavgen_python/).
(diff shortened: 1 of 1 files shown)


hamishwillee:
Thank you very much. I have restructured to put the update part in a separate section, and linking from where you had it. This makes the testing section a little more digestible. FWIW I prefer wireshark to this because it gives you traffic in both directions, but if you just want to test that the message is being sent, this is certainly pretty quick.

I think some of the instructions may not be quite right, but I'll add a separate comment.

en/middleware/mavlink.md
```diff
@@ -519,3 +525,55 @@ However if you use your own dialect then it should include ArduPilotMega.xml (or
...
+   ```sh
+   git clone https://github.com/mavlink/mavlink.git --recursive
+   cd mavlink
+   ```
+
+1. Copy your custom message XML into the [message_definitions/v1.0](https://github.com/mavlink/mavlink/tree/master/message_definitions/v1.0) folder
+1. Generate the headers:
+
+   ```sh
+   python3 -m pymavlink.tools.mavgen --lang=C --wire-protocol=2.0 --output=generated message_definitions/v1.0/your_dialect.xml
```
hamishwillee, line 562: @wasaybaig Are you sure you need this step?
This builds the C headers in mavlink/mavlink, not Python headers so I don't think this would be used.

en/middleware/mavlink.md
```diff
@@ -519,3 +525,55 @@ However if you use your own dialect then it should include ArduPilotMega.xml (or
...
+   ```
+
+1. Install Pymavlink using
+
+   ```sh
+   cd pymavlink
+   sudo python setup.py install
```
hamishwillee, line 569: I am not 100% certain this will work. Normally Pymavlink package is built with the ArduPilot fork of mavlink/mavlink headers.
What makes me think it probably does work is:

- The developer instructions https://github.com/ArduPilot/pymavlink?tab=readme-ov-file#for-developers  indicate that by default if you cloned MAVLink the location of the headers comes from the parent directoy.
- I assume you tested this :-)

wasaybaig: @hamishwillee yes I have tested it

en/middleware/mavlink.md
wasaybaig: @hamishwillee it works this way also but we can change the language to Python. But it is important to generate the headers with your new message or MAVProxy will not be able to understand the message.

hamishwillee: But MAVProxy uses Python so this is a null operation - you haven't built anything that MAVProxy can use.

hamishwillee: I guess I'll have to try it.

en/middleware/mavlink.md
julianoes: Running this with `sudo` is asking to mess up your Python dependencies. I don't think that's something we should have in the docs.

What are we trying to do here? And why can't we just use upstream pymavlink?

en/middleware/mavlink.md
hamishwillee: This is not needed. setup.py pulls any xml files in the parent definitions folder and builds them. Tested.

en/middleware/mavlink.md
hamishwillee: Yes, this does indeed screw up your installation - as I have just verified. I can get this to "work" (somewhat) but I can't undo it. Will have to delete my installation and start again :-(

What we are trying to do, ultimately, is have an easy way to test a custom message, since building QGC with a custom dialect is a bit of a pain. 

The solution described here is to install mavproxy and then delete the pymavlink that it depends on. Then rebuild pymavlink from the mavlink repo clone with your new xml definitions added. This step is easy because you just copy the xml files in and run setup.py and it creates a new installation/.egg with the .py files for the definitions.
Mavproxy can then display the message if it is received.

1. We definitely don't need the step above building the files from the definitions. the setup.py does that for you.
2. This still doesn't work though due to the error MAV_TYPE_VTOL_DUOROTOR not defined. That is correct - it has been removed from Mavlink, but it seems that mavproxy expects it in https://github.com/ArduPilot/MAVProxy/blob/master/MAVProxy/modules/mavproxy_link.py#L744
   - @wasaybaig How did you work your way around that? It's a bug in mavproxy.
3. I presume we can completely avoid all use of sudo if we install mavproxy and run setup from inside a virtual env?

hamishwillee: FYI, let's see if we can get MAVProxy to update https://github.com/ArduPilot/MAVProxy/issues/1470

peterbarker:
Just thought I'd mention my workflow for using modified message definitions.

```
pbarker@fx:~/rc/pymavlink(master)$ MDEF=~/rc/ardupilot/modules/mavlink/message_definitions pip install .
```

Running `setup.py` is deprecated and is likely to break in the next few years - I'm retraining my brain, I suggest other people start to do likewise :-)

hamishwillee:
Thanks @peterbarker ! I'll add that to the docs.

hamishwillee:
We're closing this - docs have moved into the PX4-Autopilot repo under /docs. We effectively decided this was more complicated than it looked

peterbarker:
@wasaybaig feel free to raise a PR against the MAVProxy wiki directly!

PR #3393, opened 2024-09-22
Fix typo in mavros_installation.md

Gaurav-021:
Says: "ROS Noetic (Ubuntu 22.04)", should be "ROS Noetic (Ubuntu 20.04)" as focal fossa is the officially supported distro


hamishwillee:
Thank you.

PR #3395, opened 2024-09-27
Update hb jetson image

vincentpoont2:
Fix mistake, it should say 

>  If you need to power the Jetson with a battery greater than **4S**,you must use an external UBEC.

Original image say **6S**.


vincentpoont2:
@hamishwillee Can we merge this?

hamishwillee:
Hi @vincentpoont2 

Thanks, am on holiday. For future reference, I prefer png files. At some point I'll run an image shrink tool over these, and PNG usually end up same size as jpg but with less distortion artifacts.

PR #3401, opened 2024-10-07
Added documentation for PX4 ULog encryption/decryption

jnomikos:
This PR contains documentation on how to setup/use the log encryption functionality on PX4. 

I noticed this was an area where there was not a lot of documentation / info on, and there were people in the forums asking about it.

Credits to @jlaitine for adding log encryption functionality in PX4


en/dev_log/log_encryption.md
```diff
@@ -0,0 +1,165 @@
+# Log Encryption
+
+The [system logger](../modules/modules_system.md#logger) is able to provide encrypted logs which can be decrypted manually before analysis. Currently, to have the logger create encrypted logs, you will need to flash firmware on your flight controller with this functionality supported. 
+
+::: info
+Currently, the only build that has log encryption enabled by default is `px4-fmu-v5_cryptotest`. If you wish to add it to other flight controllers, it will require you to modify the build to enable this feature. See below for instructions on how to do that.
+:::
+
+The default encryption algorithm used is XChaCha20, and the default wrapping algorithm used is RSA2048-OAEP. AES is also supported. The encryption algorithm used can be changed by modifying [SDLOG_ALGORITHM](../advanced_config/parameter_reference.md#SDLOG_ALGORITHM).
```
jlaitine, line 9: I wouldn't mention AES here; It doesn't work with current upstream PX4. It is a placeholder where one can add an own AES backend implementation.

jlaitine (approved):
Nice work, well explained! I added a note about AES support, which really isn't there

en/dev_log/log_encryption.md
hamishwillee: FYI both, I have just removed that sentence.

jnomikos: Thank you!

hamishwillee: Can we have a modification to the source code for `SDLOG_ALGORITHM` to indicate that it is currently not supported?

en/dev_log/log_encryption.md
```diff
@@ -0,0 +1,179 @@
+# Log Encryption
+
+The [System Logger](../modules/modules_system.md#logger) can be used to create encrypted logs, which may then be decrypted manually before analysis.
```
hamishwillee, line 3: What is the first version of PX4 in which this is supported?

en/dev_log/log_encryption.md
```diff
@@ -0,0 +1,179 @@
...
+However at time of writing, only XChaCha20 can actually be used (AES can be selected, but there is no implementation).
+:::
+
+## How ULog Encryption Works
+
+When the ULog file is first opened:
+
+- If [SDLOG_ALGORITHM](../advanced_config/parameter_reference.md#SDLOG_ALGORITHM) is set to default, a symmetric XChaCha20 key is generated and wrapped (encrypted) with RSA2048 public key encryption
+- This wrapped key is stored on the SD Card with the ".ulgk" suffix.
+
```
hamishwillee, line 23: I'd quite like to make this generic of XChaCha20 if possible. 

I understand that the system generates a symmetric key that is encrypted using the owners public key and stored on the SDK card. The symmetric key is used to encrypt the log. Only people who have the private key can decrypt the symmetric key file to get the key, and then use it to decrypt the log.

1. Is the symmetric key generated each time a new log is it created to go with the log, or is/can the same key used for multiple logs? I assume a new key each time!
2. If you used AES (were it supported) would the same process apply, but with the AES symmetric key? I.e. an AES key would be encrypted etc. Would the file names be the same in this case - .e. ".ulgk"?

en/dev_log/log_encryption.md
```diff
@@ -0,0 +1,179 @@
...
+However, these files will both have the normal `.ulg` suffix if downloading from QGroundControl.
+
+## Creating a Flight Controller Build that contains Log Encryption
+
+Crypto uses large amounts of flash memory, and is therefore not included in PX4 builds by default (the exception is the make target `px4-fmu-v5_cryptotest`).
+Below we show how to add log encryption functionality to a flight controller target, using the `px4-fmu-v5` board as an example.
+
+Reference the `cryptotest.px4board` file in `boards/px4/fmu-v5` for the arguments you will need to add to the `.px4board` file associated with the flight controller you are targeting.
+You could also copy and paste `cryptotest.px4board` to the other flight controller's `boards/` location if you want this to be a separate build entirely (in that case, your `make` command would include `_cryptotest` at the end to add log encryption).
```
hamishwillee, line 41: I think you are saying in the second line "You could" here, that _generically_ if there is a config file with name `_foo.px4board` in `boards/px4/fmu-v6` (say) you could build  `make px4-fmu-v6_foo` and that would generate firmware `px4-fmu-v5_foo` - right?
So in this case you're saying that you can build firmware with this name by copying the config file for cryptotest.

In the first line are you saying you should just copy the content of tge `cryptotest.px4board` file into the default board config?

Why I am confused is why you would do the copying. I THINK that you need to do this before you do the "Adding log encryption to NuttX config" so that the options appear in the kconfig UI. Is that correct.

So the steps are "add info about the config variables" to the config folder, then run kconfig and set their values. Right?

en/dev_log/log_encryption.md
```diff
@@ -0,0 +1,179 @@
...
+
+### Crypto .px4board arguments
+
+| Argument                     | Description                                                                                                                        |
+| ---------------------------- | ---------------------------------------------------------------------------------------------------------------------------------- |
+| CONFIG_BOARD_CRYPTO          | Requirement for log encryption functionality to be enabled.<br />= `y`: Enable log encryption.<br />= `n`: Disable log encryption. |
+| CONFIG_DRIVERS_SW_CRYPTO     | Enables the built-in PX4 crypto backend library to handle logfile encryption.<br />= `y`: Enable<br />= `n`: Disable               |
+| CONFIG_DRIVERS_STUB_KEYSTORE | Enables the built-in PX4 stub keystore driver.<br />= `y`: Enable<br />= `n`: Disable                                              |
```
hamishwillee, line 49: So these all need to be set to y for the functionality to work? Aren't they all Y if you copy the values from `cryptotest.px4board`? I.e. why do you need to run kconfig then other than to set the location of the keys.

en/dev_log/log_encryption.md
```diff
@@ -0,0 +1,179 @@
...
+| CONFIG_PUBLIC_KEY0           | Location of public key for keystore index 0.<br />= `{path to key0}`                                                               |
+| CONFIG_PUBLIC_KEY1           | Location of public key for keystore index 1.<br />= `{path to key1}`                                                               |
```
hamishwillee, line 51: Does this value define default keys that you can use, and if so, where can you find the corresponding private key for testing?

1. Where should these files be stored? Is there a standard location.
2. Is this everything you need to do to fetch the file and include it in firmware?
3. What is the difference between the two keystores? Do you need both? Which is used?

en/dev_log/log_encryption.md
```diff
@@ -0,0 +1,179 @@
...
+
+::: warning
+Crypto uses a lot of flash memory, and many builds are close to their maximum capacity.
+If you run into a build error telling you that you have gone above the maximum flash memory, you will need to disable other features in the `.px4board` file you are working on, or in the `default.px4board` file.
+Be careful not to disable something you need.
+
+As an example, you could disable SIH mode by setting `CONFIG_MODULES_SIMULATION_SIMULATOR_SIH=n` which could free up enough flash memory to allow crypto to be added.
```
hamishwillee, line 58: This is excellent advice :-)

en/dev_log/log_encryption.md
```diff
@@ -0,0 +1,179 @@
...
+The symmetric key is normally pretty small (around 300 bytes).
+:::
+
+The process for downloading log files is the same, except you will download two separate files per log.
+Each time you fly, a log file, and a symmetric key are generated.
+The only way to tell them apart is by file size.
+
+As normal, go to **Analyze Tools -> Log Download** and select the log you wish to download.
+
+Make sure to download both the encrypted log and symmetric key that were generated.
```
hamishwillee, line 107: Should we perhaps note that even though they have the same filename, which sadly isn't shown, you can tell which is which based on file size?

en/dev_log/log_encryption.md
```diff
@@ -0,0 +1,179 @@
...
+
+![QGroundControl ULog Download](../../assets/qgc/analyze/encrypted_log.png)
+
+This photo shows both a generated symmetric key and the associated encrypted ULog file.
```
hamishwillee, line 111:
```suggestion
This screenshot shows both a generated symmetric key and the associated encrypted ULog file.
```

en/dev_log/log_encryption.md
```diff
@@ -0,0 +1,179 @@
...
+/home/john/Downloads/log_23_2024-10-6-23-39-48.ulg \
+new_keys/private_key.pem
+```
+
+There will be no printed output on a successful decryption, and a new file is created with the ".ul" suffix instead of ".ulg".
+Rename this back to a .ulg file and it is now ready for flight review!
+
+### Generate Your Own Keys
+
+In a production environment, you should your own generated keys rather than using the example keys.
```
hamishwillee, line 153: Ah, so there are example keys. Where can they be found?

en/dev_log/log_encryption.md
```diff
@@ -0,0 +1,179 @@
...
+
+```sh
+# Convert private_key.pem to a DER file
+openssl rsa -pubout -in private_key.pem -outform DER -out public_key.der
+# From the DER file, generate a public key in hex format, seperated by commas
+xxd -p public_key.der | tr -d '\n' | sed 's/\(..\)/0x\1, /g' > public_key.pub
+```
+
+Now, you would modify your `.px4board` file to point CONFIG_PUBLIC_KEY0 to the path of `new.pub` that was generated via `cryptotools.py`, and set CONFIG_PUBLIC_KEY1 to `rsa2048.pub`.
+The private key generated can be used in the decryption environment.
```
hamishwillee, line 179: So again, not clear what the two keys are for - why you need both. Also a bit confused about this  new.json file here.

Can you provide a real example of the kconfig values for these keys? E.g. screenshot in kconfig menu? 
Are there any restrictions on the locations? I assume you must build the keys somewhere under the PX4 root, for example?

en/dev_log/log_encryption.md
hamishwillee: Is it "Enable" or "include in firmware"? I assume the later - and once in firmware it is SDLOG_ALGORITHM that determines whether the feature is enabled or not? Also that this is default on if the module is installed?

en/dev_log/log_encryption.md
```diff
@@ -0,0 +1,179 @@
+# Log Encryption
```
hamishwillee, line 1: Can we cross link to this _from_ https://docs.px4.io/main/en/dev_log/logging.html ? It's a cool feature so good to make it easier to find.

I was thinking perhaps in the introduction 

> The output log format is [ULog](https://docs.px4.io/main/en/dev_log/ulog_file_format.html)."

 TO 

> The output log format is [ULog](https://docs.px4.io/main/en/dev_log/ulog_file_format.html), and this may also be [encrypted](LINK).

hamishwillee (commented):
Thanks. Looks great - really appreciate this. A few questions inline, after which I will do a further minor subedit.

en/dev_log/log_encryption.md
jnomikos: > I think you are saying in the second line "You could" here, that generically if there is a config file with name _foo.px4board in boards/px4/fmu-v6 (say) you could build make px4-fmu-v6_foo and that would generate firmware px4-fmu-v5_foo - right?

I think my wording here was a bit confusing. I was essentially trying to illustrate the point that you don't HAVE to copy and paste the cryptotest.px4board file to a different board type, and instead, you could just copy the parameters in that file into a different one. 

> Why I am confused is why you would do the copying.

If you have multiple builds for a particular board and want your log encryption build to be separate from other builds, it would make sense to copy the cryptotest.px4board file to the other board so you can make it by doing make {board}_cryptotest. As an example, if you get rid of a feature due to flash memory constraints, you may want multiple variations of the firmware build so that you can keep that feature in some.

Is it necessary for us to mention this in the documentation? Probably not. 

> So the steps are "add info about the config variables" to the config folder, then run kconfig and set their values. Right?

Yes, but I don't believe the order at which you do so matters.

en/dev_log/log_encryption.md
jnomikos: `So these all need to be set to y for the functionality to work? Aren't they all Y if you copy the values from cryptotest.px4board? I.e. why do you need to run kconfig then other than to set the location of the keys.`

The values set in kconfig add CONFIG_CRYPTO=y and CONFIG_CRYPTO_RANDOM_POOL=y  to defconfig when saving. I know at least CONFIG_CRYPTO being enabled is necessary.

When writing this, I believe I misunderstood the purpose of kconfig. If we can just set these in the .px4board without doing kconfig to add them to defconfig, then kconfig isn't necessary. I was just replicating what was done in the px4_fmu_v5 build, as it also has CONFIG_CRYPTO=y and CONFIG_CRYPTO_RANDOM_POOL=y set in defconfig.

jnomikos: > Is it "Enable" or "include in firmware"? I assume the later - and once in firmware it is SDLOG_ALGORITHM that determines whether the feature is enabled or not? Also that this is default on if the module is installed?

I also think the latter.

hamishwillee: > When writing this, I believe I misunderstood the purpose of kconfig. If we can just set these in the .px4board without doing kconfig to add them to defconfig, then kconfig isn't necessary.

I _think_ you're right - Kconfig is a GUI tool for editing config files. I think the structure is fine, we just have to say that this is what we are doing/these are the alternatives

hamishwillee, line 49: I think it is best to say what they do more precisely. Not sure about the CONFIG_DRIVERS_STUB_KEYSTORE and CONFIG_CRYPTO_RANDOM_POOL isn't mentioned in the table but you mention it on the other thread here.

```suggestion
| CONFIG_BOARD_CRYPTO          | Include crypto module in firmware.<br />= `y`: Enable log encryption.<br />= `n`: Disable log encryption. |
| CONFIG_DRIVERS_SW_CRYPTO     | Include the built-in PX4 crypto backend library (used by above library).<br />= `y`: Enable<br />= `n`: Disable               |
| CONFIG_DRIVERS_STUB_KEYSTORE | Includes the built-in PX4 stub keystore driver.<br />= `y`: Enable<br />= `n`: Disable                                              |
```

en/dev_log/log_encryption.md
jnomikos: I know it works on 1.14. 

I am unsure about specifics about the first version where this functionality is supported and is stable. I think [this PR ](https://github.com/PX4/PX4-Autopilot/pull/17814) added it all the way back in 2021, but there were other PRs since then by @jlaitine related to it.  

I did some commit searches and I think 1.12.0 is the earliest version where the crypto commits exist, but I am unsure how it works on that version, since I only tested it on 1.14.

en/dev_log/log_encryption.md
jnomikos: > Is the symmetric key generated each time a new log is it created to go with the log, or is/can the same key used for multiple logs? I assume a new key each time!

A new key is generated for each new log. Basically, each log generates two files, the encrypted log (.ulgc), and the associated symmetric key (.ulgk).

> If you used AES (were it supported) would the same process apply, but with the AES symmetric key? I.e. an AES key would be encrypted etc. Would the file names be the same in this case - .e. ".ulgk"?

I am not well versed enough into encryption algorithms to know the answer to this question. Since they are both symmetric key algorithms, I would assume the implementation would be similar, but there is likely nuance to it that I am missing.

> How is the symmetric key protected on the device? I..e. is it amenable to being found by someone (other than the encrypted versions), can it be replaced somehow so that a "bad" key is always used?

The symmetric key itself is encrypted, so you would need to have access to the private rsa key to unwrap the symmetric key to decrypt the log.

en/dev_log/log_encryption.md
jnomikos: I have an info panel right above this that mentions it

hamishwillee: You are way ahead of me. Sorry!

I ran out of time on this. I will put it on my list to have another look next week.

jnomikos: No worries. I'll address these comments more on the weekend / make edits

hamishwillee: Thank you very much. I'm back Weds on this stuff but I'll try be responsive.

en/dev_log/log_encryption.md
hamishwillee: @jnomikos OK, so merging this, but :
1. YOu mentioned CONFIG_CRYPTO_RANDOM_POOL - is it something we should mention?
2. Can you check those descriptions are OK.

hamishwillee: I think the `CONFIG_DRIVERS_STUB_KEYSTORE` text "includes the built-in PX4 stub keystore driver" is correct - but it doesn't tell us much. I would guess that PX4 is suppoed to have a secure keystore, perhaps in a secure part. However that depends on hardware, so by default the keystore probably must maps to somewhere in the firmware or the SITL filesystem.

en/dev_log/log_encryption.md
hamishwillee: FYI, I have added https://github.com/PX4/PX4-Autopilot/pull/23816 to make it clear the AES option is not supported yet.

en/dev_log/log_encryption.md
hamishwillee: Thanks. The parameter appears in v1.14 first, so I have added a badge for that.

en/dev_log/log_encryption.md
hamishwillee: I have rewritten this as though it is generic, prefixed with a note that it is _likely_ to be the same for other algorithms

hamishwillee:
I'm still going with subedit based on what we "know".

en/dev_log/log_encryption.md
jnomikos: Just pushed up change that clarifies this

en/dev_log/log_encryption.md
jnomikos: I agree. Just pushed that change

en/dev_log/log_encryption.md
```diff
@@ -0,0 +1,223 @@
...
+Then you can create a public key from this private key:
+
+```sh
+# Convert private_key.pem to a DER file
+openssl rsa -pubout -in private_key.pem -outform DER -out public_key.der
+# From the DER file, generate a public key in hex format, seperated by commas
+xxd -p public_key.der | tr -d '\n' | sed 's/\(..\)/0x\1, /g' > public_key.pub
+```
+
+To use these keys you would modify your `.px4board` file to point `CONFIG_PUBLIC_KEY0` to the path of `new.pub` that was generated via `cryptotools.py`, and set `CONFIG_PUBLIC_KEY1` to point to the file location of `rsa2048.pub`.
```
hamishwillee, line 222: What is the purpose of new.pub? We have an RSS key generated with open ssl, so what is the file for/what does it do?

en/dev_log/log_encryption.md
jnomikos: @hamishwillee 

I did some digging into PX4 source code to find an answer to this. It makes sense to me now.

The current implementation of Crypto contains a keystore that can hold 4 separate keys. 
These are contained in CONFIG_PUBLIC_KEY0, CONFIG_PUBLIC_KEY1, CONFIG_PUBLIC_KEY2, CONFIG_PUBLIC_KEY3

Two params exist in PX4

[SDLOG_EXCH_KEY](https://docs.px4.io/main/en/advanced_config/parameter_reference.html#SDLOG_EXCH_KEY), and [SDLOG_KEY](https://docs.px4.io/main/en/advanced_config/parameter_reference.html#SDLOG_KEY). 

By default, SDLOG_EXCH_KEY is set to 1, and SDLOG_KEY is set to 2.

So by default, CONFIG_PUBLIC_KEY0 is our exchange key, and CONFIG_PUBLIC_KEY1 is our encryption key.

_______________
_**Note of a concern I have about these parameters:**_
 SDLOG_EXCH_KEY and SDLOG_KEY defaults are 1 and 2. Yet the parameter itself has a minimum value of 0. And when I'm looking in the source code, I cannot find anywhere that these param values are getting subtracted by 1 when indexing the keystore. I am wondering if there is an unintended bug here when running the default cryptotest example.
I am going to do more digging here
___________________
 

(take this with a grain of salt because cryptography is not my area of expertise) I believe that the exchange key exists so that the encryption key can be stored in an encrypted format on the SD card.


> Can you provide a real example of the kconfig values for these keys? E.g. screenshot in kconfig menu?

I don't think kconfig has an area to change these keys.

> Are there any restrictions on the locations? I assume you must build the keys somewhere under the PX4 root, for example?

Not that I can tell. As a test, I placed key1 in my home directory and changed the .px4board file and it built just fine.

hamishwillee:
@jnomikos I've gone through this and made all the changes I think are appropriate. Can you please review and then we'll get @jlaitine to have a quick scan and confirm.

I think there is just one open question https://github.com/PX4/PX4-user_guide/pull/3401/files#r1804098147

Note, I now "mostly" understand how kconfig works and hence why you have done it this way. I am exploring this [in discord](https://discord.com/channels/1022170275984457759/1022185721450213396/threads/1296323326679191582). Some notes "FYI":
1. The config system is called `kconfig` the tool for graphical editing is `menuconfig`. 
2. We use menuconfig to enable crypto because that is not set in the custom make target. 
   - In theory it should be possible to set the extra variables in the default.px4config but that doesn't seem to work. 
   - menuconfig appears to make a local cached copy of your change. So if you were to sync to git, I am not sure that you'd be able to restore the config changes.
  
Anyway, now this is worded as "if you're working on FMUv5" vs "other boards". If I get confirmation on how the board config items might be set in text files we can choose to update this.

en/dev_log/log_encryption.md
hamishwillee: @jnomikos Thanks! Looks like we have had some crossover in the last hour. Let me digest this.

hamishwillee: OK, so I think I get what you are saying.
- Keystore has up to 4 places for keys CONFIG_PUBLIC_KEY0, CONFIG_PUBLIC_KEY1, CONFIG_PUBLIC_KEY2, CONFIG_PUBLIC_KEY3
- We store, `new.pub` in `CONFIG_PUBLIC_KEY0` and our rsapub key in  `CONFIG_PUBLIC_KEY1`. We could also store a notional `new2.pub` in  `CONFIG_PUBLIC_KEY2` and rsapub2 key in  `CONFIG_PUBLIC_KEY3`.
- The  [SDLOG_EXCH_KEY](https://docs.px4.io/main/en/advanced_config/parameter_reference.html#SDLOG_EXCH_KEY), and [SDLOG_KEY](https://docs.px4.io/main/en/advanced_config/parameter_reference.html#SDLOG_KEY) are used to select the _index_ of the keys in the store that are supposed to be used in calculations.
- SDLOG_EXCH_KEY defaults to 1 and is supposed to store the location of the public key that is used to encypt the symmetric key.
  - It is not clear if this is indexed against the keystore or the set of log keys (?) - i.e. the two possible keys. Should be stated in the docs.
- SDLOG_KEY default to 2 and is supposed to store the location of the symmetric key that is generated.
  - It is not clear if this is indexed against the keystore or the set of log keys (?)
- Assuming zero index key store and that the params above are both relative to the key store, that means
  - SDLOG_KEY (symmetric key)=2=CONFIG_PUBLIC_KEY3= not set by us, which is fine.
  - SDLOG_EXCH_KEY (rsa key)=1=CONFIG_PUBLIC_KEY1= our RSA key
  - CONFIG_PUBLIC_KEY0 and CONFIG_PUBLIC_KEY2 are not used.
  - (new.pub) is not used
  
I think this will work because from what I can tell `new.pub` serves no purpose. My assumption is that as long as our RSA key is mapped to the correct key that we set in the config, and _nothing else is trying to write to `CONFIG_PUBLIC_KEY3`_ which we are using for accessing the symmetric key, we should be fine.

It would be great to confirm so that we can remove setting and using new.pub and that script.

jnomikos:
> @jnomikos I've gone through this and made all the changes I think are appropriate. Can you please review and then we'll get @jlaitine to have a quick scan and confirm.
> 
> I think there is just one open question https://github.com/PX4/PX4-user_guide/pull/3401/files#r1804098147
> 
> Note, I now "mostly" understand how kconfig works and hence why you have done it this way. I am exploring this [in discord](https://discord.com/channels/1022170275984457759/1022185721450213396/threads/1296323326679191582). Some notes "FYI":
> 
>     1. The config system is called `kconfig` the tool for graphical editing is `menuconfig`.
> 
>     2. We use menuconfig to enable crypto because that is not set in the custom make target.
>        
>        * In theory it should be possible to set the extra variables in the default.px4config but that doesn't seem to work.
>        * menuconfig appears to make a local cached copy of your change. So if you were to sync to git, I am not sure that you'd be able to restore the config changes.
> 
> 
> Anyway, now this is worded as "if you're working on FMUv5" vs "other boards". If I get confirmation on how the board config items might be set in text files we can choose to update this.

Nice. I have some minor changes I want to make. One that clarifies the steps for generating your own keys, and one that 

> @jnomikos I've gone through this and made all the changes I think are appropriate. Can you please review and then we'll get @jlaitine to have a quick scan and confirm.
> 
> I think there is just one open question https://github.com/PX4/PX4-user_guide/pull/3401/files#r1804098147
> 
> Note, I now "mostly" understand how kconfig works and hence why you have done it this way. I am exploring this [in discord](https://discord.com/channels/1022170275984457759/1022185721450213396/threads/1296323326679191582). Some notes "FYI":
> 
>     1. The config system is called `kconfig` the tool for graphical editing is `menuconfig`.
> 
>     2. We use menuconfig to enable crypto because that is not set in the custom make target.
>        
>        * In theory it should be possible to set the extra variables in the default.px4config but that doesn't seem to work.
>        * menuconfig appears to make a local cached copy of your change. So if you were to sync to git, I am not sure that you'd be able to restore the config changes.
> 
> 
> Anyway, now this is worded as "if you're working on FMUv5" vs "other boards". If I get confirmation on how the board config items might be set in text files we can choose to update this.

Haha I actually just wrote up something that clarifies what new.pub is. Pushing soon

en/dev_log/log_encryption.md
jnomikos: new.pub is the exchange key header file that is associated with the exchange key, while the RSS key generated with openssl is the encryption key.

hamishwillee:
> > Anyway, now this is worded as "if you're working on FMUv5" vs "other boards". If I get confirmation on how the board config items might be set in text files we can choose to update this.
> 
> Haha I actually just wrote up something that clarifies what new.pub is. Pushing soon

@jnomikos Great. See my comment https://github.com/PX4/PX4-user_guide/pull/3401#discussion_r1804127899 - I still infer it is useless :-). 

W.r.t. comments, you should be able to make github suggestions for changes if you like. Up to you. I am now hands off for a week.

jnomikos:
> W.r.t. comments, you should be able to make github suggestions for changes if you like. Up to you. I am now hands off for a week.

Yeah I'll throw in some suggestions rather than pushing

en/dev_log/log_encryption.md
```diff
@@ -0,0 +1,223 @@
...
+
+::: details Overview of the keys
+
+| Argument                     | Description                                                                                                     |
+| ---------------------------- | --------------------------------------------------------------------------------------------------------------- |
+| CONFIG_BOARD_CRYPTO          | Include crypto module in firmware.<br />= `y`: Enable log encryption.<br />= `n`: Disable log encryption.       |
+| CONFIG_DRIVERS_SW_CRYPTO     | Include the built-in PX4 crypto backend library (used by above library).<br />= `y`: Enable<br />= `n`: Disable |
+| CONFIG_DRIVERS_STUB_KEYSTORE | Includes the built-in PX4 stub keystore driver.<br />= `y`: Enable<br />= `n`: Disable                          |
+| CONFIG_PUBLIC_KEY0           | Location of public key for keystore index 0.<br />= `{path to key0}`                                            |
+| CONFIG_PUBLIC_KEY1           | Location of public key for keystore index 1.<br />= `{path to key1}`                                            |
```
jnomikos, line 92: We probably should make it clear that the keystore has room for more keys, and that the SDLOG_EXCH_KEY and SDLOG_KEY params affect which key is the exchange key and which key is the encryption key.

```suggestion
| Argument                     | Description                                                                                                                                                                                     |
| ---------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| CONFIG_BOARD_CRYPTO          | Include crypto module in firmware.<br />= `y`: Enable log encryption.<br />= `n`: Disable log encryption.                                                                                       |
| CONFIG_DRIVERS_SW_CRYPTO     | Include the built-in PX4 crypto backend library (used by above library).<br />= `y`: Enable<br />= `n`: Disable                                                                                 |
| CONFIG_DRIVERS_STUB_KEYSTORE | Includes the built-in PX4 stub keystore driver.<br />= `y`: Enable<br />= `n`: Disable                                                                                                          |
| CONFIG_PUBLIC_KEY0           | Location of public key for keystore index 0. By default, this is set as the exchange key by [SDLOG_EXCH_KEY](../advanced_config/parameter_reference.md#SDLOG_EXCH_KEY).<br />= `{path to key0}` |
| CONFIG_PUBLIC_KEY1           | Location of public key for keystore index 1. By default, this is set as the encryption key by [SDLOG_KEY](../advanced_config/parameter_reference.md#SDLOG_KEY).<br />= `{path to key1}`         |
| CONFIG_PUBLIC_KEY2           | Location of public key for keystore index 2.<br />= `{path to key2}`                                                                                                                            |
| CONFIG_PUBLIC_KEY3           | Location of public key for keystore index 3.<br />= `{path to key3}`                                                                                                                            |
```

en/dev_log/log_encryption.md
hamishwillee: The exchange key is the key used to encrypt the symmetric key right? So new.pub is what encrypts the symmetric key?

What's the encryption key? That's a new term.

en/dev_log/log_encryption.md
```diff
@@ -0,0 +1,223 @@
...
+/home/john/Downloads/log_23_2024-10-6-23-39-48.ulgk \
+new_keys/private_key.pem
+```
+
+On success the decrypted log file is created with the `.ul` suffix instead of `.ulg`.
+Rename the file back to `.ulg` and it is now ready for flight review.
+
+## Generate RSA Public & Private Keys
+
+In a production environment, you should your own generated keys rather than using the example keys.
```
jnomikos, line 198:
```suggestion
The `cryptotest` example implementation uses example keys which are found in `Tools/test_keys`. In a production environment, you should your own generated keys rather than using the example keys. 
```

en/dev_log/log_encryption.md
```diff
@@ -0,0 +1,223 @@
...
+
+To generate the `CONFIG_PUBLIC_KEY0` public key, you can run the `Tools/cryptotools.py` Python program.
```
jnomikos, line 200:
```suggestion
To generate the exchange key, you can run the `Tools/cryptotools.py` Python program.
```

hamishwillee:
Thanks very much for all your work on this. To be honest I think this would be usable now - just there would be still quite a bit of underlying "magic". Good to pull it apart properly, because at the end we should understand key stores as well as kconfig.

en/dev_log/log_encryption.md
```diff
@@ -0,0 +1,223 @@
...
+For example, this command generates files called `new.pub` and `new.json` file in `Tools/test_keys`:
+
+```sh
+python3 Tools/cryptotools.py  --genkey Tools/test_keys/new
+```
+
```
jnomikos, line 206:
```suggestion
You can now reference the path to this public exchange key in the `CONFIG_PUBLIC_KEY` keystore on the .px4board file. 
By default, the exchange key is associated with `CONFIG_PUBLIC_KEY0`, but this can be modified with the [SDLOG_EXCH_KEY](../advanced_config/parameter_reference.md#SDLOG_EXCH_KEY) parameter.
```

en/dev_log/log_encryption.md
jnomikos, line 222:
```suggestion
The path to the generated public key can now be referenced in the `CONFIG_PUBLIC_KEY` keystore on the .px4board file.
By default, the encryption key is associated with `CONFIG_PUBLIC_KEY1`, but this can be modified with the [SDLOG_KEY](../advanced_config/parameter_reference.md#SDLOG_KEY) parameter.
```

jnomikos:
> Thanks very much for all your work on this. To be honest I think this would be usable now - just there would be still quite a bit of underlying "magic". Good to pull it apart properly, because at the end we should understand key stores as well as kconfig.

Appreciate all the effort on your end as well

en/dev_log/log_encryption.md
jnomikos: The encryption key is used to encrypt the log itself. The CONFIG_PUBLIC_KEY param associated with it is determined by [SDLOG_KEY](https://docs.px4.io/main/en/advanced_config/parameter_reference.html#SDLOG_KEY).  However, I do see that [SDLOG_KEY](https://docs.px4.io/main/en/advanced_config/parameter_reference.html#SDLOG_KEY) says,

> Selects the key in keystore, used for encrypting the log. When using a symmetric encryption algorithm, the key is generated at logging start and kept stored in this index. For symmetric algorithms, the key is volatile and valid only for the duration of logging. The key is stored in encrypted format on the sdcard alongside the logfile, using an RSA2048 key defined by the SDLOG_EXCHANGE_KEY

Since XChaCha20 is a symmetric key algorithm, would this mean that it is useless to set an encryption key at the associated CONFIG_PUBLIC_KEY keystore, since encryption keys are volatile and change each time a new log is created?

hamishwillee: This is precisely my point. The only key we should to supply is the RSA key that encrypts the symmetric key (which is generated). 

I had assumed that this key was the key generated using open SSL (just because that sounds like the right way to do public key infrastructure than some random script we supply). But you're saying it is new.pub based on the default settings from encrypt. 

We should find out the intent - it might be that the openssl key is actually intended to be used as the exchange key, or that one of the keys has a completely different intended purpose and happens to be mapped by one of these two params. We really can't "guess".

hamishwillee: @jlaitine Can you shed light on this?

In summary, [`cryptotest.px4board`](https://github.com/PX4/PX4-Autopilot/blob/main/boards/px4/fmu-v5/cryptotest.px4board) sets the following keys which are stored in the key store.

```config
CONFIG_PUBLIC_KEY0="../../../Tools/test_keys/key0.pub"
CONFIG_PUBLIC_KEY1="../../../Tools/test_keys/rsa2048.pub"
```

Which of these is supposed to be used for the SDLOG exchange key (which encrypts the symmetric key), and what is the purpose of the other file?

The [SDLOG_EXCH_KEY ](https://docs.px4.io/main/en/advanced_config/parameter_reference.html#SDLOG_EXCH_KEY) param is value `1` which you would expect to map to `CONFIG_PUBLIC_KEY1` (0 indexed), so I think rsa2048.pub is the exchange key.

As per [this discussion ](https://github.com/PX4/PX4-user_guide/pull/3401#discussion_r1804127899) I think the other key is unrelated?

jlaitine (approved):
This looks very good to me, big thanks for this work!

About the menuconfig question, it *should* be possible to set the PX4 CONFIG_ flags in .px4board files. Either in default.px4board or some variant (just like it is done for fmuv5 cryptotest.px4board). There is no magic involved!

I can try to help if there is some issue where the flags are not working properly - then we can see if documentation needs any additions for that.

hamishwillee:
Thanks @jlaitine - I'll try the config variable tests again and get back to you.

Can you answer the question on this thread? https://github.com/PX4/PX4-user_guide/pull/3401#discussion_r1804208933 - I don't think we need new.pub in this.

en/dev_log/log_encryption.md
jlaitine: Hi!

This is a good point. The "key0.pub" is indeed not related to the logfile encryption at all! This "key0.pub" actually originates from being the key used by the bootloader to verify the firmware signature. And obviously, when the public keys are compiled in to the code like it is done in the "stub_keystore", it doesn't make much sense in this context. I will still shed some light on why it was left in here..

- I some of the projects in which I am working, there is a real secure keystorage in the device (the stub_keystore is replaced by another hw driver which accesses keys stored in hw); that is, both the signature verification key and the RSA KEK are stored in the same place, thus they are accessed the same way and indexed always with different indexes no matter whether you are in bootloader or in PX4.
- When I started this work, I wanted to model this even when having the keys compiled in to the image - I just configured the "stub_keystore" the same way both in nuttx-based bootloader and in px4; so using the keys looked as if there was a real keystore with two keys
- Since key0.pub is an ed25519 elliptic curve signature verification key, it can be used for signature verification also inside PX4, not just in bootloader. The support for signature verification is actually also there built in to the sw_crypto driver, so it can be directly used in other application code, if needed.

I hope this clarifies, and I apologize the confusion which this caused. I understand that currently the "key0.pub" is not used for anything in upstrem PX4 - and it has nothing to do with logfile encryption.

en/dev_log/log_encryption.md
```diff
@@ -0,0 +1,225 @@
...
+| CONFIG_BOARD_CRYPTO          | Include crypto module in firmware.<br />= `y`: Enable log encryption.<br />= `n`: Disable log encryption.                                                                                       |
+| CONFIG_DRIVERS_SW_CRYPTO     | Include the built-in PX4 crypto backend library (used by above library).<br />= `y`: Enable<br />= `n`: Disable                                                                                 |
+| CONFIG_DRIVERS_STUB_KEYSTORE | Includes the built-in PX4 stub keystore driver.<br />= `y`: Enable<br />= `n`: Disable                                                                                                          |
+| CONFIG_PUBLIC_KEY0           | Location of public key for keystore index 0. By default, this is set as the exchange key by [SDLOG_EXCH_KEY](../advanced_config/parameter_reference.md#SDLOG_EXCH_KEY).<br />= `{path to key0}` |
+| CONFIG_PUBLIC_KEY1           | Location of public key for keystore index 1. By default, this is set as the encryption key by [SDLOG_KEY](../advanced_config/parameter_reference.md#SDLOG_KEY).<br />= `{path to key1}`         |
```
jnomikos, line 92:
```suggestion
| CONFIG_PUBLIC_KEY0           | Location of public key for keystore index 0.  |
| CONFIG_PUBLIC_KEY1           | Location of public key for keystore index 1. By default, this is set as the exchange key by [SDLOG_EXCH_KEY](../advanced_config/parameter_reference.md#SDLOG_EXCH_KEY).<br />= `{path to key1}`        |
```

I believe my first suggestion had it wrong. Since both SDLOG_KEY, SDLOG_EXCH_KEY and the CONFIG_PUBLIC_KEY params are zero indexed, I believe CONFIG_PUBLIC_KEY1 maps to SDLOG_EXCH_KEY by default. CONFIG_PUBLIC_KEY0 would map to nothing.

(SDLOG_EXCH_KEY defaults to 1. SDLOG_KEY defaults to 2)

en/dev_log/log_encryption.md
```diff
@@ -0,0 +1,225 @@
...
+
+The `cryptotest` example implementation uses example keys which are found in `Tools/test_keys`. In a production environment, you should your own generated keys rather than using the example keys. 
+
+To generate the `CONFIG_PUBLIC_KEY0` public key, you can run the `Tools/cryptotools.py` Python program.
+For example, this command generates files called `new.pub` and `new.json` file in `Tools/test_keys`:
+
+```sh
+python3 Tools/cryptotools.py  --genkey Tools/test_keys/new
+```
+
```
jnomikos, line 208:
```suggestion
```

en/dev_log/log_encryption.md
```diff
@@ -0,0 +1,225 @@
...
+Then you can create a public key from this private key:
+
+```sh
+# Convert private_key.pem to a DER file
+openssl rsa -pubout -in private_key.pem -outform DER -out public_key.der
+# From the DER file, generate a public key in hex format, seperated by commas
+xxd -p public_key.der | tr -d '\n' | sed 's/\(..\)/0x\1, /g' > public_key.pub
+```
+
+To use these keys you would modify your `.px4board` file to point `CONFIG_PUBLIC_KEY0` to the path of `new.pub` that was generated via `cryptotools.py`, and set `CONFIG_PUBLIC_KEY1` to point to the file location of `rsa2048.pub`.
```
jnomikos, line 224:
```suggestion
To use this key you would modify your `.px4board` file to point `CONFIG_PUBLIC_KEY1` to the file location of `public_key.pub`.
```

en/dev_log/log_encryption.md
```diff
@@ -0,0 +1,225 @@
...
+The relevant keys in that file are reproduced below:
+
+```config
+CONFIG_BOARD_CRYPTO=y
+CONFIG_DRIVERS_STUB_KEYSTORE=y
+CONFIG_DRIVERS_SW_CRYPTO=y
+CONFIG_PUBLIC_KEY0="../../../Tools/test_keys/key0.pub"
+CONFIG_PUBLIC_KEY1="../../../Tools/test_keys/rsa2048.pub"
+```
+
```
jnomikos, line 83:
```suggestion

::: info
In the current PX4 implementation, `key0.pub` is unused and can be ignored.
:::

```

en/dev_log/log_encryption.md
```diff
@@ -0,0 +1,225 @@
...
+| CONFIG_PUBLIC_KEY0           | Location of public key for keystore index 0. By default, this is set as the exchange key by [SDLOG_EXCH_KEY](../advanced_config/parameter_reference.md#SDLOG_EXCH_KEY).<br />= `{path to key0}` |
+| CONFIG_PUBLIC_KEY1           | Location of public key for keystore index 1. By default, this is set as the encryption key by [SDLOG_KEY](../advanced_config/parameter_reference.md#SDLOG_KEY).<br />= `{path to key1}`         |
+| CONFIG_PUBLIC_KEY2           | Location of public key for keystore index 2.<br />= `{path to key2}`                                                                                                                            |
+| CONFIG_PUBLIC_KEY3           | Location of public key for keystore index 3.<br />= `{path to key3}`                                                                                                                            |
+
+:::
+
+You can use the above keys for testing, or replace `CONFIG_PUBLIC_KEY0` and `CONFIG_PUBLIC_KEY1` with the path to your own keys in the file (see [Generate RSA Public & Private Keys](#generate-rsa-public-private-keys)).
```
jnomikos, line 98:
```suggestion
You can use the above keys for testing, or replace `CONFIG_PUBLIC_KEY1` with the path to your own public key in the file (see [Generate RSA Public & Private Keys](#generate-rsa-public-private-keys)).
```

jnomikos (commented):
Added some suggestions, given the info that CONFIG_PUBLIC_KEY0 is unused

en/dev_log/log_encryption.md
```diff
@@ -0,0 +1,233 @@
...
+CONFIG_DRIVERS_SW_CRYPTO=y
+CONFIG_PUBLIC_KEY1="../../../Tools/test_keys/rsa2048.pub"
+```
+
+::: info
+The file also sets `CONFIG_PUBLIC_KEY0` to a key named `key0.pub`.
+This is not used in the current PX4 implementation and can be ignored.
+:::
+
+::: details Overview of crypto-relevant keys
```
hamishwillee, line 88: FYI @jnomikos I accepted all your changes.
However I have extended this bit to explain how the keys work "according to my understanding".

Essentially the board config copies the keys and required drivers to access them into firmware. But the actual keystore lives in software, so it is seeded by the values in the board config, but after that can modify them.

Not sure if this is also true for hardware-keystore - it might be, but then maybe they also store volatile key data.

hamishwillee: PS I think this is good to go now otherwise. Just retesting if I can add the values to the config rather than via menuconfig.

hamishwillee: @jlaitine I'm still stuck with setting the board config value. I think that perhaps they are being set, but I am missing some keys. 

I added the following config to the fmuv4 tree and rebuilt (`>` shows changes to config)

```
> CONFIG_MODULES_SIMULATION_SIMULATOR_SIH=n
CONFIG_MODULES_ATTITUDE_ESTIMATOR_Q=n
CONFIG_MODULES_FW_AUTOTUNE_ATTITUDE_CONTROL=n
CONFIG_MODULES_MC_AUTOTUNE_ATTITUDE_CONTROL=n
> CONFIG_CRYPTO=y
> CONFIG_CRYPTO_RANDOM_POOL=y
> CONFIG_CRYPTO_BLAKE2S
CONFIG_DRIVERS_STUB_KEYSTORE=y
CONFIG_BOARD_CRYPTO=y
CONFIG_DRIVERS_SW_CRYPTO=y
CONFIG_PUBLIC_KEY0="../../../Tools/test_keys/key0.pub"
CONFIG_PUBLIC_KEY1="../../../Tools/test_keys/rsa2048.pub"
```

But I get the error below. I would have thought that undefined reference to `px4_get_secure_random' due to `CONFIG_CRYPTO_RANDOM_POOL` being missing, but I added it above. Can you advise on how to debug?

```
rivers/stub_keystore/libkeystore_backend.a  src/lib/crypto/libmonocypher.a  src/lib/crypto/liblibtomcrypt.a  platforms/nuttx/src/px4/common/libpx4_random.a  NuttX/nuttx/crypto/libcrypto.a  src/lib/crypto/liblibtommath.a  platforms/common/uORB/libuORB.a  NuttX/nuttx/mm/libmm.a  src/lib/cdev/libcdev.a  NuttX/nuttx/fs/libfs.a  msg/libuorb_msgs.a  src/lib/heatshrink/libheatshrink.a  platforms/common/px4_work_queue/libpx4_work_queue.a  NuttX/nuttx/libs/libxx/libxx.a  NuttX/nuttx/drivers/libdrivers.a  NuttX/nuttx/libs/libc/libc.a  NuttX/nuttx/drivers/libdrivers.a  NuttX/nuttx/libs/libc/libc.a  NuttX/nuttx/sched/libsched.a  -lm  -lgcc && :
Memory region         Used Size  Region Size  %age Used
           flash:     1982641 B      2032 KB     95.28%
            sram:       32972 B       192 KB     16.77%
          ccsram:          0 GB        64 KB      0.00/opt/gcc-arm-none-eabi-9-2020-q2-update/bin/../lib/gcc/arm-none-eabi/9.3.1/../../../../arm-none-eabi/bin/ld: src/drivers/sw_crypto/libcrypto_backend.a(crypto.c.obj): in function `crypto_open':
/home/ubuntu/github/PX4-Autopilot/src/drivers/sw_crypto/crypto.c:188: undefined reference to `px4_get_secure_random'
/opt/gcc-arm-none-eabi-9-2020-q2-update/bin/../lib/gcc/arm-none-eabi/9.3.1/../../../../arm-none-eabi/bin/ld: src/drivers/sw_crypto/libcrypto_backend.a(crypto.c.obj): in function `crypto_generate_key':
/home/ubuntu/github/PX4-Autopilot/src/drivers/sw_crypto/crypto.c:345: undefined reference to `px4_get_secure_random'
/opt/gcc-arm-none-eabi-9-2020-q2-update/bin/../lib/gcc/arm-none-eabi/9.3.1/../../../../arm-none-eabi/bin/ld: src/lib/crypto/liblibtomcrypt.a(libtomcrypt_wrappers.c.obj): in function `rng_get_bytes':
/home/ubuntu/github/PX4-Autopilot/src/lib/crypto/libtomcrypt_wrappers.c:41: undefined reference to `px4_get_secure_random'
collect2: error: ld returned 1 exit status
```

Also, my understanding is that changes made in menuconfig UI are local - i.e. they don't affect any files saved to git right?

hamishwillee:
@jnomikos @jlaitine I've read this again and I think it is pretty good, and usable.

It would be _better_ if questions like https://github.com/PX4/PX4-user_guide/pull/3401#discussion_r1811934992 were answered, but not worth holding this up for.

So I am going to merge, and if you find your way to looking at those in future, it would be much appreciated.

Thank to you both for all your work on this.

PR #3402, opened 2024-10-08
Add to Remote id info

Aaron1356:
This should not be merged in until after the Remote ID base information is merged int located here:

- [ ] https://github.com/PX4/PX4-user_guide/pull/3339

This just adds in some information about how to see that Remote ID over can is working properly and how to get QGC setup so that others don't run into the same issue that I did.

PR #3339, opened 2024-08-08
remote_id: add note about DroneCAN compatibility

julianoes:
With main and versions after v1.15 we will have support for remote ID over DroneCAN:

- [x] https://github.com/PX4/PX4-Autopilot/pull/23113
- [ ] Release note added to doc?

FYI @BluemarkInnovations: I have not tested your remote ID module over DroneCAN but only CubePilot's and Holybro's, and mentioned that in the docs.

PR #3404, opened 2024-10-10
Remote id info - subedited

hamishwillee:
This should not be merged in until after the Remote ID base information is merged int located here:

- [x] https://github.com/PX4/PX4-user_guide/pull/3339

This just adds in some information about how to see that Remote ID over can is working properly and how to get QGC setup so that others don't run into the same issue that I did.

FYI @Aaron1356 This replaces #3402 which I couldn't subedit.


en/peripherals/remote_id.md
```diff
@@ -8,43 +10,46 @@ Remote ID is a government mandated technology for UAVs in Japan, the United Stat
...
 
 ## Supported Hardware
 
-PX4 integrates with Remote ID hardware that supports the [Open Drone ID](https://mavlink.io/en/services/opendroneid.html) MAVLink protocol (Open Drone ID is an open source implementation of Remote ID).
+PX4 integrates with Remote ID hardware that supports:
+
+- [Open Drone ID](https://mavlink.io/en/services/opendroneid.html) MAVLink protocol<Badge type="tip" text="PX4 v1.14" />
+- Remote ID over CAN<Badge type="tip" text="PX4 main (v1.16)" />
+
+  ::: warning
```
hamishwillee, line 23: FYI I moved your warning down here from the top to be a little less confronting - to easy to misread as all remote ids not supported except in main.

en/peripherals/remote_id.md
```diff
@@ -8,43 +10,46 @@ Remote ID is a government mandated technology for UAVs in Japan, the United Stat
...
+  DroneCAN Remote IDs hardware is only supported on `main` branch builds (builds _after_ PX4 v1.15).
```
hamishwillee, line 24:
```suggestion
  Remote ID hardware can only be connected via DroneCAN on `main` branch builds (builds _after_ PX4 v1.15).
```

en/peripherals/remote_id.md
```diff
@@ -168,10 +201,48 @@ Some known issues are:
...
+
+  ```sh
+  uavcan status
+  ```
+
+  The remote id should appears in the list as shown below (there maybe other nodes in the list too).
+
+  ```plain
+  Online nodes (Node ID, Health, Mode):
+     125 OK         OPERAT
```
hamishwillee, line 231: Is this really what you see? Does that mean we're looking for a node with id 125? I.e. how do you know this node is a remote ID node - in particular if there are multiples?

Aaron1356: So Enter UAVCAN generates a lot of text that I didn't want to put all of it making the document too long. When only the Cube ID is attached then yes this is the only node that is present but there can be other nodes attached.

en/peripherals/remote_id.md
Aaron1356: This is from the original Remote ID Documentation that @julianoes wrote. I just pulled in those changes so that I didn't write over or rewrite any of the information that was already present.

en/peripherals/remote_id.md
```diff
@@ -168,10 +201,48 @@ Some known issues are:
...
+  ```
+
+- Check that the `OPEN_DRONE_ID_BASIC_ID` and `OPEN_DRONE_ID_LOCATION` messages appear in the QGroundControl [MAVLink Inspector](https://docs.qgroundcontrol.com/master/en/qgc-user-guide/analyze_view/mavlink_inspector.html) (QGC **Analyze Tools > MAVLink Inspector**).
+
+If the above are not true, then the Remote ID itself may need to be configured:
+
+1. Open QGroundControl
+2. Navigate to the [Application settings](https://docs.qgroundcontrol.com/master/en/qgc-user-guide/settings_view/general.html): **Application Settings > General > Miscellaneous**.
+3. Select `Enable Remote ID`.
```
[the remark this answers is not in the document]
Aaron1356: Well to be more clear it is in the miscellaneous section of the general tab, but yes there is a checkbox to enable remote ID on QGC which will then enable the Remote ID tab so that data like operator ID, self ID, and Basic ID can be set for the drone and ground station.

en/peripherals/remote_id.md
hamishwillee: :-) All good!

en/peripherals/remote_id.md
hamishwillee: That's cool, but it doesn't quite confirm what I need. Let me put it another way - is 125 always the node of any remote id? 

Otherwise, i I think you are saying that there is no way to tell if a remote id is present unless you diff what the list looks like before and after.

Aaron1356: The safest approach would be to diff between the before and after on the CAN bus. Not sure if all Remote ID modules will have 125 as there Node ID. I was only able to test out the Cube ID.

en/peripherals/remote_id.md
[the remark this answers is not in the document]
hamishwillee: @Aaron1356 Hmmm. On my QGC daily I have a dedicated tab as shown below. What version are you using?

Either way, it sounds like we send them to this page depending on the version of QGC used.

image

en/peripherals/remote_id.md
hamishwillee: Thanks @Aaron1356 . I have edited this to put the message check first - if that works you don't need to do anything. Then for the `uavcan status` check I have suggested a simple node count as first option, and differencing as a better check. I think this is OK.

hamishwillee:
@Aaron1356 OK, so just one open issue https://github.com/PX4/PX4-user_guide/pull/3404#discussion_r1802300021

Any idea with the remote ID pr this is conditional on is expected to merge?

en/peripherals/remote_id.md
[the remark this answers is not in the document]
hamishwillee: I think I'll just note some of the options

hamishwillee:
@Aaron1356 I've merged this on top of @julianoes PR. I think it works well. Thank you very much!

PR #3406, opened 2024-10-11
Holybro DroneCAN ZED F9P setup instructions

dirksavage88:
These AP Periph dronecan nodes have different params that need to be enabled to have GPS heading when two gps's are used.

This PR addresses these params specific to the ZED-F9P dronecan gps from Holybro and also documents configuring the RTK gps module in general (non dual gps heading setup).

Rationale:
Moving baseline and GPS heading is not intuitive to set up for these AP Periph based dronecan nodes.

@vincentpoont2


hamishwillee (approved):
Thanks @dirksavage88 - I have subedited it for style, and to make it even more clear that this doc is just the DroneCAN variant, and that there are other variants.
Can you please sanity check my changes?

@vincentpoont2 Are you happy with this? Note that the RTK GPS table now has two links, one for this variant and another to your purchase page, which covers setup for the UART ports.
We could just make this cover all variants if you prefer (the other variants would just say "see parent doc for setup".

dirksavage88:
@hamishwillee the formatting looks a lot cleaner, thanks.

I think the technical content regarding the dual F9P for heading _may_ change. Need to talk with px4 maintainers to see how I can get the fix in. 

TLDR; PX4 dronecan gps heading message fields are currently handled differently than AP Periph gps heading message fields, since PX4 cannode sends the heading as a hack in an un-related ecef position message, and AP periph sends heading in an actual relative position heading message (per the dronecan dsdl).

hamishwillee:
Thanks very much for the information @dirksavage88 . I think this is going to be helpful.

FYI, [according to the docs](https://docs.px4.io/main/en/dronecan/#sensors) (and as you probably know) the `UAVCAN_SUB_GPS_R` (GPS relative) was added in v1.15 but is only used for logging currently.

If that changes, or any of this changes, can you please help me also update that linked doc. It is very useful IMO to try and have a central place to explain and list the dronecan subscriptions and publications, because the concept is not intuitive if you're used to just plugging and playing your UART component, perhaps just enabling one param.

@vincentpoont2 No comment from you, but I'm going to merge this as is for now. We can come back to it if you are not completely happy with it - PRs welcome :-)

en/gps_compass/rtk_gps_holybro_h-rtk-f9p.md
```diff
@@ -8,7 +13,7 @@ It also has an integrated safety switch for a simple and convenient operation.
 There are three models of Holybro H-RTK F9P to choose from, each with different antenna design to meet different needs.
 Refer to [Specification and Model Comparison section](#specification-and-model-comparison) for more details.
 
-Using RTK allows PX4 to get its position with centimeter-level accuracy, which is much more accurate than can be provided by a normal GPS.
+Using RTK allows PX4 to get its position with centimetre-level accuracy, which is much more accurate than can be provided by a normal GPS.
```
vincentpoont2, line 16: Not sure if I agree! 😂

PR #3408, opened 2024-10-15
Fixed vertiq.md documentation links that were 404'd

vertiq-luca:
We updated our docs organization so now these links need to be updated if possible


hamishwillee (approved):
Thank you very much!

PR #3410, opened 2024-10-21
Fix typos in mission.md

matthieu-db:
I assume these are the commands meant, feel free to close if I am mistaken


hamishwillee (approved):
Thanks! FYI these probably "should" be named with the do prefix by convention, but they aren't.

PR #3419, opened 2024-10-30
Clarify arming is powering and props might start

hamishwillee:
This add some clarifications falling out of the discussion in https://github.com/PX4/PX4-Autopilot/pull/23822#issuecomment-2431481264

Essentially on arming the motors may not start turning, but they can do so at any time without user input. This makes that difference clear. For MC it notes that the props will start turning except for during throw launch.

### DIFF
--- a/en/getting_started/px4_basic_concepts.md
+++ b/en/getting_started/px4_basic_concepts.md
@@ -280,7 +280,7 @@ For more information see: [Payloads & Cameras](../payloads/index.md)
 ## Arming and Disarming
 
 A vehicle is said to be _armed_ when all motors and actuators are powered, and _disarmed_ when nothing is powered.
-There is also a _prearmed_ state when only actuators are powered, which is primarily used for testing.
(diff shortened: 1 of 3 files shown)


en/getting_started/px4_basic_concepts.md
```diff
@@ -280,10 +280,12 @@ For more information see: [Payloads & Cameras](../payloads/index.md)
 ## Arming and Disarming
 
 A vehicle is said to be _armed_ when all motors and actuators are powered, and _disarmed_ when nothing is powered.
-There is also a _prearmed_ state when only actuators are powered.
+There is also a _prearmed_ state when only actuators are powered, which is primarily used for testing.
```
sfuhrer, line 283:
```suggestion
There is also a _prearmed_ state when only servo actuators are powered, which is primarily used for testing.
```

hamishwillee: Thanks for the review!

PR #3429, opened 2024-11-04
Update Lightware Lidar Docs

Claudio-Chies:
This step seems to not be necessary any more. Never had to do it on the LW20/C, and also checked lightware studio, and there isn't event the option for a "pixhawk compatability mode" any more. 

Cant say anything about the other lightware sensors like the SF11/C in the screenshot.
Does someone have one of these to test?

### DIFF
--- a/en/SUMMARY.md
+++ b/en/SUMMARY.md
@@ -237,7 +237,7 @@
       - [TFSlot Airspeed Sensor](sensor/airspeed_tfslot.md)
     - [Barometers](sensor/barometer.md)
     - [Distance Sensors \(Rangefinders\)](sensor/rangefinders.md)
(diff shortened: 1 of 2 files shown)


hamishwillee:
> Cant say anything about the other lightware sensors like the SF11/C in the screenshot.
> Does someone have one of these to test?

@Claudio-Chies I don't. I've emailed lightware support. Can you post on discuss or discord and cross link here if that doesn't result in an answer.

hamishwillee:
Thanks @Claudio-Chies - Lightware have confirmed this only applies to SF11/C.
I have undone your change and just made this section specific to that lidar.

Claudio-Chies:
@hamishwillee thanks alot for confirming with lightware!

PR #3465, opened 2024-11-20
Message versioning and ROS 2 message translation

GuillaumeLaine:
This documents the newly versioned subset of PX4 messages, and the ROS 2 translation node that enables seamless communication between PX4 and ROS 2 applications that may be using different message definition versions.

### Related Work

- [PX4-Autopilot] Implementing ROS 2 translation node and CI checks: https://github.com/PX4/PX4-Autopilot/pull/24113
- [PX4-Autopilot] Introducing versioned PX4 messages: https://github.com/PX4/PX4-Autopilot/pull/23850
- [px4_msgs] Updating build paths for versioned messages: https://github.com/PX4/px4_msgs/pull/46

### To Do

- [ ] Complete `TODO`s in the markdown of this PR


en/middleware/uorb.md
```diff
@@ -141,10 +141,27 @@ Make sure not to mix `orb_advertise_multi` and `orb_advertise` for the same topi
 
 The full API is documented in [platforms/common/uORB/uORBManager.hpp](https://github.com/PX4/PX4-Autopilot/blob/main/platforms/common/uORB/uORBManager.hpp).
 
+## Message Versioning
```
hamishwillee, line 145: Just FYI, this is how I mark up new content. On release I'll know the version number and replace this. You should also add a release note in https://docs.px4.io/main/en/releases/main.html , ideally with a link to any associated PRs.

I'll review properly when you tell me this is ready and/or when this moves out of draft. At that time would be good to link the various proposals that lead to this in the description.

Glad you are doing this and not me :-)

```suggestion

<Badge type="tip" text="main (PX4 v1.16+)" />

```

GuillaumeLaine: @hamishwillee thanks for the early review (: I've applied your suggestions, and with some of the related PRs getting merged I'm coming back to these docs.

There's essentially 2 things to document here:
1. The new versioned messages (already started here)
2. The ROS 2 translation node (to do)

The README in [this PR](https://github.com/PX4/PX4-Autopilot/pull/24113) documents point 2 well, I will now try to capture those points here. Once that's done I will open this PR for review

en/ros2/user_guide.md
```diff
@@ -34,6 +34,9 @@ The generator uses the uORB message definitions in the source tree: [PX4-Autopil
 ROS 2 applications need to be built in a workspace that has the _same_ message definitions that were used to create the uXRCE-DDS client module in the PX4 Firmware.
 You can include these by cloning the interface package [PX4/px4_msgs](https://github.com/PX4/px4_msgs) into your ROS 2 workspace (branches in the repo correspond to the messages for different PX4 releases).
 
+Starting from PX4 x.x.x in which [message versioning](../middleware/uorb.md#message-versioning) was introduced, ROS2 applications may use a different version of message definitions than those used to build PX4.
```
GuillaumeLaine, line 37: Do we have a way to add a badge just for this paragraph, not the whole section?

GuillaumeLaine:
@hamishwillee I think the PR is in a review ready state. I left a few `TODO`s in the markdown to resolve before merging. Mostly things that I need to go back to once the last related PR is merged

en/releases/main.md
```diff
@@ -41,6 +41,7 @@ Please continue reading for [upgrade instructions](#upgrade-guide).
   Mission commands that may take some time to complete, such as those for controlling gimbals, winches, and grippers, will progress to the next item when either feedback is received or the timeout expires.
   This is often used to provide a minimum delay for hardware that does not provide completion feedback, so that it can reach the commanded state before the mission progresses.
   ([PX4-Autopilot#23960](https://github.com/PX4/PX4-Autopilot/pull/23960)).
+- **[uORB]** Introduce a version field for a subset of uORB messages ([PX4-Autopilot#23850](https://github.com/PX4/PX4-Autopilot/pull/23850))
```
hamishwillee, line 44: Make this a link to perhaps uorb.html#message-versioning

en/releases/main.md
```diff
@@ -75,7 +76,7 @@ Please continue reading for [upgrade instructions](#upgrade-guide).
 
 ### uXRCE-DDS / ROS2
 
-- TBD
+- **[Feature]** ROS 2 Translation Node to translate PX4 messages from one defintion version to another dynamically ([PX4-Autopilot#24113](https://github.com/PX4/PX4-Autopilot/pull/24113))
```
hamishwillee, line 79: Add link to the translation node docs from "ROS 2 Translation Node"

en/ros2/user_guide.md
```diff
@@ -34,6 +34,9 @@ The generator uses the uORB message definitions in the source tree: [PX4-Autopil
 ROS 2 applications need to be built in a workspace that has the _same_ message definitions that were used to create the uXRCE-DDS client module in the PX4 Firmware.
 You can include these by cloning the interface package [PX4/px4_msgs](https://github.com/PX4/px4_msgs) into your ROS 2 workspace (branches in the repo correspond to the messages for different PX4 releases).
 
+Starting from PX4 x.x.x in which [message versioning](../middleware/uorb.md#message-versioning) was introduced (TODO: specify PX4 version), ROS2 applications may use a different version of message definitions than those used to build PX4.
```
hamishwillee, line 37: What version are you thinking. If it is "the next version" we don't know what that is yet. In which case use "PX4 v1.16 (main)". When this appears in a release version I plan on updating all instances of that string to whatever the actual version is.

en/ros2/user_guide.md
```diff
@@ -34,6 +34,9 @@ The generator uses the uORB message definitions in the source tree: [PX4-Autopil
...
+This requires the [ROS 2 Message Translation Node](../ros2/px4_ros2_msg_translation_node.md) to be running to ensure that messages can be converted and exchanged correctly.
```
hamishwillee, line 38: My thinking here is that we're going to want users to get used to this, and to almost do it by default - right? 

If so I think you're going to need to add instructions to install and use the versioned messages in this guide - probably in a section of https://docs.px4.io/main/en/ros2/user_guide.html#build-ros-2-workspace

Note that when I say "install and use", I really mean "install and run" - you would link to your instructions in the translation node doc for how to update versions and so on. There would be some duplication in that doc.

en/middleware/uorb.md
```diff
@@ -141,10 +141,29 @@ Make sure not to mix `orb_advertise_multi` and `orb_advertise` for the same topi
...
+
+Versioned messages are designed to remain more stable over time compared to their non-versioned counterparts, as they are intended to be used across multiple releases of PX4 and external systems, ensuring greater compatibility over longer periods.
+
+For the full list of versioned and non-versioned messages, refer to the full [uORB Message Reference](../msg_docs/index.md).
+
+For more on PX4 and ROS 2 communication, refer to the page about the [PX4-ROS 2 Bridge](../ros/ros2_comm.md).
 
 ## Message/Field Deprecation {#deprecation}
 
+(TODO: Is this section still relevant? Not seeing any `# DEPRECATED: ...` fields or msgs in latest PX4)
```
hamishwillee, line 165: Almost certainly still relevant. We don't want to break Flight Review or other tools. @bkueng will know.

en/middleware/uorb.md
```diff
@@ -141,10 +141,29 @@ Make sure not to mix `orb_advertise_multi` and `orb_advertise` for the same topi
...
 
+## Message Versioning
 
+<Badge type="tip" text="main (PX4 v1.16+)" />
+
+Message versioning was introduced to address the challenges of maintaining compatibility across systems where different versions of message definitions may be in use, such as between PX4 and external systems like ROS 2 applications.
+
+This versioning mechanism supports the [ROS 2 Message Translation Node](../ros2/px4_ros2_msg_translation_node.md), which enables seamless communication between PX4 and ROS 2 applications; when different versions of message definitions are in use, the ROS 2 translation node ensures that messages can be converted and exchanged correctly.
+
+Versioned messages are stored in the `msg/versioned/` directory, distinct from their non-versioned counterparts, which reside directly in the `msg/` directory.
```
hamishwillee, line 152: Nothing wrong with this, but I'm trying to make the distinction very clear, since I missed this first time.

```suggestion
Versioned messages are stored in the `msg/versioned/` directory (distinct from their non-versioned counterparts, which reside directly in the `msg/` directory).
```

en/middleware/uorb.md
```diff
@@ -141,10 +141,29 @@ Make sure not to mix `orb_advertise_multi` and `orb_advertise` for the same topi
...
+Each versioned message definition includes an additional field: `uint32 MESSAGE_VERSION = X`, where `X` corresponds to the current version of the message.
+When a versioned message definition is modified, the version number should be incremented to reflect changes in its structure or semantics.
+Additionally, [message translations](../ros2/px4_ros2_msg_translation_node.md#updating-a-message) must also be updated for versioned messages.
```
hamishwillee, line 155:
```suggestion
Additionally, [message translations](../ros2/px4_ros2_msg_translation_node.md#updating-a-message) must also be updated for versioned messages, in order to convert from the old version to the new structure.
```

en/middleware/uorb.md
```diff
@@ -141,10 +141,29 @@ Make sure not to mix `orb_advertise_multi` and `orb_advertise` for the same topi
...
+
+Versioned messages are designed to remain more stable over time compared to their non-versioned counterparts, as they are intended to be used across multiple releases of PX4 and external systems, ensuring greater compatibility over longer periods.
+
+For the full list of versioned and non-versioned messages, refer to the full [uORB Message Reference](../msg_docs/index.md).
```
hamishwillee, line 159:
```suggestion
For the full list of versioned and non-versioned messages, refer to the [uORB Message Reference](../msg_docs/index.md).
```

en/middleware/uorb.md
```diff
@@ -141,10 +141,29 @@ Make sure not to mix `orb_advertise_multi` and `orb_advertise` for the same topi
 
 The full API is documented in [platforms/common/uORB/uORBManager.hpp](https://github.com/PX4/PX4-Autopilot/blob/main/platforms/common/uORB/uORBManager.hpp).
 
+## Message Versioning
 
+<Badge type="tip" text="main (PX4 v1.16+)" />
+
+Message versioning was introduced to address the challenges of maintaining compatibility across systems where different versions of message definitions may be in use, such as between PX4 and external systems like ROS 2 applications.
+
+This versioning mechanism supports the [ROS 2 Message Translation Node](../ros2/px4_ros2_msg_translation_node.md), which enables seamless communication between PX4 and ROS 2 applications; when different versions of message definitions are in use, the ROS 2 translation node ensures that messages can be converted and exchanged correctly.
```
hamishwillee, line 150: This is a little messy and perhaps can be simplified. In particular
- ROS2 Message Translation node supports versioning, not the versioning supporting the translation node.
 
Perhaps something like:

```suggestion
Message versioning has been in introduced in PX4 v1.16 to make it easier to maintain compatibility between PX4 and ROS 2 versions.
This versioning mechanism uses the [ROS 2 Message Translation Node](../ros2/px4_ros2_msg_translation_node.md) to convert between different versions of messages, allowing seamless communication between PX4 and ROS 2 applications; when different versions of message definitions are in use.
```

Perhaps add a diagram showing where the magic happens?

It would also be good "somewhere here" to mention any forward thinking about the ROS 2 native translation system that they have been talking about for a couple of years. Perhaps as a note at the end.

en/releases/main.md
hamishwillee, line 79: Perhaps add the experimental badge here too.

en/ros2/px4_ros2_msg_translation_node.md
```diff
@@ -0,0 +1,110 @@
+# PX4 ROS 2 Message Translation Node
+
+<Badge type="tip" text="main (PX4 v1.16+)" /> <Badge type="warning" text="Experimental" />
+
+The message translation node allows to run applications that are compiled with one set of message versions against a PX4 with another set of message versions, without having to change either the application or the PX4 side.
+
+Specifically for this to work, topic publication/subscription/service names contain a message version in the form of `<topic_name>_v<version>`.
```
hamishwillee, line 7: I don't understand what this means. Specifically, at this point if I look at https://github.com/PX4/PX4-Autopilot/blob/main/msg/versioned/VehicleLandDetected.msg it does not have this format - I suspect this is all ROS2 stuff.

I don't think you need to include this line here. I would include information about how this works lower down, maybe in the implementation section. Not needed here.

en/ros2/px4_ros2_msg_translation_node.md
```diff
@@ -0,0 +1,110 @@
...
+
+The translation node knows about all existing message versions, and dynamically monitors the publications, subscriptions and services, and then creates translations as needed.
```
hamishwillee, line 9:
```suggestion
The translation node knows about all message versions previously defined by PX4, and dynamically monitors the publications, subscriptions and services, in order to create translations as needed.
```

en/ros2/px4_ros2_msg_translation_node.md
```diff
@@ -0,0 +1,110 @@
+# PX4 ROS 2 Message Translation Node
+
+<Badge type="tip" text="main (PX4 v1.16+)" /> <Badge type="warning" text="Experimental" />
+
+The message translation node allows to run applications that are compiled with one set of message versions against a PX4 with another set of message versions, without having to change either the application or the PX4 side.
```
hamishwillee, line 5:
```suggestion
The message translation node allows applications that were compiled against different versions of the PX4 messages to interwork with newer versions of PX4, and vice versa, without having to change either the application or the PX4 side.
```

en/ros2/px4_ros2_msg_translation_node.md
```diff
@@ -0,0 +1,110 @@
...
+
+Specifically for this to work, topic publication/subscription/service names contain a message version in the form of `<topic_name>_v<version>`.
+
+The translation node knows about all existing message versions, and dynamically monitors the publications, subscriptions and services, and then creates translations as needed.
+
+## Installation and First Test
```
hamishwillee, line 11: First test is implied IMO.

```suggestion
## Installation
```

en/ros2/px4_ros2_msg_translation_node.md
```diff
@@ -0,0 +1,110 @@
...
+
```
hamishwillee, line 12: I think it is worth saying when you should do this. I presume for every workspace? 

So something like: "For every workspace you create, it is recommended to install and run the node:", followed by the steps.

en/ros2/px4_ros2_msg_translation_node.md
```diff
@@ -0,0 +1,110 @@
...
+source /path/to/ros_ws/install/setup.bash
+```
+
+Finally, run the translation node. You should see an output similar to:
+
+```bash
+ros2 run translation_node translation_node_bin
+[INFO] [1734525720.729530513] [translation_node]: Registered pub/sub topics and versions:
+[INFO] [1734525720.729594413] [translation_node]: Registered services and versions:
+```
```
hamishwillee, line 39: I like lists, in particular if you add the prefix I suggest above. So

````suggestion
1. Create a ROS 2 workspace in which to build the message translation node and its dependencies.

   ```sh
   mkdir -p /path/to/ros_ws/src
   ```

2. Run the following helper script to copy the message definitions and translation node into your ROS workspace directory.

   ```sh
   cd /path/to/ros_ws
   /path/to/PX4-Autopilot/Tools/copy_to_ros_ws.sh .
   ```

3. Build and source the workspace.

   ```sh
   colcon build
source /path/to/ros_ws/install/setup.bash
   ```
4. Finally, run the translation node:

   ```sh
   ros2 run translation_node translation_node_bin
   ```

    You should see an output similar to:

   ```sh
   [INFO] [1734525720.729530513] [translation_node]: Registered pub/sub topics and versions:
   [INFO] [1734525720.729594413] [translation_node]: Registered services and versions:
   ```
````

File: en/ros2/px4_ros2_msg_translation_node.md [hunk 18]
```diff
@@ -0,0 +1,110 @@
...
+
+## Updating a Message
+
+When changing a message, a new version needs to be added.
+
+The steps include:
```
hamishwillee, line 45:
```suggestion
The steps are:
```

en/ros2/px4_ros2_msg_translation_node.md
```diff
@@ -0,0 +1,110 @@
...
+
+- Copy the versioned message file (`.msg`/`.srv`) to `px4_msgs_old/msg/` (or `px4_msgs_old/srv/`) and add the current version to the file name.
```
hamishwillee, line 47: > Copy the versioned message file (`.msg`/`.srv`)`

- what is .srv here?
- what is `px4_msgs_old`? How do they get that?

en/ros2/px4_ros2_msg_translation_node.md
```diff
@@ -0,0 +1,110 @@
...
+2. Generic case: this allows a translation between N input topics and M output topics.
+  This can be used for merging or splitting a message.
+  Or for example when moving a field from one message to another, a single translation should be added with the two older message versions as input and the two newer versions as output.
+  This way there is no information lost when translating forward or backward.
+
+::: info
+If a nested message definition changes, all messages including that message also require a version update.
+This is primarily important for services.
+:::
+
```
hamishwillee, line 71: What do I do if I have custom messages I want to version?

en/ros2/px4_ros2_msg_translation_node.md
```diff
@@ -0,0 +1,110 @@
...
+## Updating a Message
+
+When changing a message, a new version needs to be added.
+
+The steps include:
+
+- Copy the versioned message file (`.msg`/`.srv`) to `px4_msgs_old/msg/` (or `px4_msgs_old/srv/`) and add the current version to the file name.
+  For example `msg/versioned/VehicleAttitude.msg` becomes `px4_msgs_old/msg/VehicleAttitudeV3.msg`.
+  Update the existing translations that use the current topic version to the now old version.
```
hamishwillee, line 49: > Update the existing translations that use the current topic version to the now old version.

1. What's a translation? Needs a definition.
2. Am I replacing or adding a new one of these "whatever they are". I guess I'm asking if I make 3 versions of my message do I have three files? Or one that is updated three times?

en/ros2/px4_ros2_msg_translation_node.md
```diff
@@ -0,0 +1,110 @@
...
+  For example `px4_msgs::msg::VehicleAttitude` becomes `px4_msgs_old::msg::VehicleAttitudeV3`.
```
hamishwillee, line 50: This example doesn't help answer my questions above  - in what file would it be?

en/ros2/px4_ros2_msg_translation_node.md
```diff
@@ -0,0 +1,110 @@
...
+- Increment `MESSAGE_VERSION` and update the message fields as desired.
```
hamishwillee, line 51:
```suggestion
- Increment `MESSAGE_VERSION` in the current message definition (the one in `msg/versioned/`).
  Make the other changes to message fields that prompted the version change.
```

hamishwillee: Note ^^^ Just trying to be precise about which file/message is being updated.

en/ros2/px4_ros2_msg_translation_node.md
```diff
@@ -0,0 +1,110 @@
...
+1. Direct translations: these translate a single topic between two different versions.
+  This is the simpler case and should be preferred if possible.
+2. Generic case: this allows a translation between N input topics and M output topics.
+  This can be used for merging or splitting a message.
+  Or for example when moving a field from one message to another, a single translation should be added with the two older message versions as input and the two newer versions as output.
+  This way there is no information lost when translating forward or backward.
+
+::: info
+If a nested message definition changes, all messages including that message also require a version update.
+This is primarily important for services.
```
hamishwillee, line 69: I don't know anything about nested message definitions. What are they? 

Either way, perhaps a warning or tip?

en/ros2/px4_ros2_msg_translation_node.md
```diff
@@ -0,0 +1,110 @@
...
+- Include the added header in [`translations/all_translations.h`](https://github.com/PX4/PX4-Autopilot/blob/message_versioning_and_translation/msg/translation_node/translations/all_translations.h).
+
+For the second last step and for topics, there are two options:
+
+1. Direct translations: these translate a single topic between two different versions.
+  This is the simpler case and should be preferred if possible.
+2. Generic case: this allows a translation between N input topics and M output topics.
+  This can be used for merging or splitting a message.
+  Or for example when moving a field from one message to another, a single translation should be added with the two older message versions as input and the two newer versions as output.
+  This way there is no information lost when translating forward or backward.
```
hamishwillee, line 65: I haven't looked at the example, but it seems to me that this might require some worked examples of each type. If the headers are specific to each type, perhaps move them under the following sections?

en/ros2/px4_ros2_msg_translation_node.md
```diff
@@ -0,0 +1,110 @@
...
+```c++
+topic_name + "_v" + std::to_string(T::MESSAGE_VERSION)
+```
+
+Where `T` is the message type, e.g. `px4_msgs::msg::VehicleAttitude`.
+
+The DDS client in PX4 automatically adds the version suffix if a message contains the field `uint32 MESSAGE_VERSION = x`.
+
+::: info Version 0 of a topic means that no `_v<version>` suffix should be added.
+:::
```
hamishwillee, line 85:
```suggestion
::: info
Version 0 of a topic means that no `_v<version>` suffix should be added.
:::
```

en/ros2/px4_ros2_msg_translation_node.md
```diff
@@ -0,0 +1,111 @@
...
+If a nested message definition changes, all messages including that message also require a version update.
+This is primarily important for services.
+:::
+
+## Usage in ROS
+
+The message version can be added generically to a topic like this:
+
+```c++
+topic_name + "_v" + std::to_string(T::MESSAGE_VERSION)
```
hamishwillee, line 77: So this is something you have to do for all topics?

I think I might make this a bit more "real" by having a several line real-world example afterwards.

en/ros2/px4_ros2_msg_translation_node.md
```diff
@@ -0,0 +1,111 @@
...
+Therefore on a topic update, the graph is traversed using a shortest path algorithm.
+When moving from one node to the next, the message translation method is called with the current topic data.
+If a node contains an instantiated publisher (because it previously detected an external subscriber), the data is published.
+Thus, multiple subscribers of any version of the topic can be updated with the correct version of the data.
+
+For translations with multiple input topics, the translation continues once all input messages are available.
+
+## Limitations
+
+- The current implementation depends on a service API that is not yet available in ROS Humble, and therefore does not support services when built for ROS Humble.
```
hamishwillee, line 106: So what versions will it work for?

en/ros2/px4_ros2_msg_translation_node.md
```diff
@@ -0,0 +1,111 @@
...
+- Services only support a linear history, i.e. no message splitting or merging
```
hamishwillee, line 107: Do you mean services as in something different than message/topics? Perhaps this is what you mean by .svr above? Is there information on what those mean anywhere?

Either way, so are you saying that services require a linear history, but you can split or merge messages as you like?

en/ros2/px4_ros2_msg_translation_node.md
```diff
@@ -0,0 +1,111 @@
...
+- Having both publishers and subscribers for two different versions of a topic is currently not handled by the translation node and would trigger infinite circular publications.
```
hamishwillee, line 108: So a ROS application must compile against just one version of a topic? What if there are two ROS applications running on the network compiled against different versions - is this the problem case?

hamishwillee (commented):
@GuillaumeLaine 

OK, lots of comments here. They don't reflect that this is pretty good structure. Just I'm using myself as the straw man, and I don't know much ROS 2. 

When you've had a look at the comments and tried address them, can you loop in Beat and Benja to have a look? I ask because I probably will be on holiday when you do this. Back around week of Jan 20th.

en/ros2/user_guide.md
hamishwillee: Not really.

en/middleware/uorb.md
bkueng: Right, it's still relevant but it usually gets forgotten to be added.

en/ros2/px4_ros2_msg_translation_node.md
GuillaumeLaine: Awesome thanks, I agree the list is much neater. Regarding the prefix what do you mean, are you referring to "For every workspace you create, it is recommended to install and run the node:"?

hamishwillee: Yes!

mrpollo:
Maybe worth adding to the 1.16 release wip page? @hamishwillee

en/ros2/px4_ros2_msg_translation_node.md
GuillaumeLaine: Good points, I've elaborated on this.

> I guess I'm asking if I make 3 versions of my message do I have three files?

Exactly, 3 new versions of a message means adding three new translation files: each translation file bridging one version to the next

en/ros2/px4_ros2_msg_translation_node.md
GuillaumeLaine: You're right, I've gone and added a worked example also to make the steps clearer

en/ros2/px4_ros2_msg_translation_node.md
GuillaumeLaine: Nested messages are message definitions that include fields referencing other message types instead of using only the classic data types. For instance [PositionSetpointTriplet](https://github.com/PX4/PX4-Autopilot/blob/main/msg/PositionSetpointTriplet.msg).

I've added the example and made the paragraph into a warning

en/ros2/px4_ros2_msg_translation_node.md
GuillaumeLaine: Correct, ROS distinguishes between services and topics like so: https://docs.ros.org/en/foxy/How-To-Guides/Topics-Services-Actions.html. In PX4, we have the [srv/ services](https://github.com/PX4/PX4-Autopilot/tree/message_versioning_and_translation/srv) folder vs the [msg/ topics](https://github.com/PX4/PX4-Autopilot/tree/message_versioning_and_translation/msg) folder.

Messages define the data format used to communicate over _either_ topics or via services. The confusing part is that `.srv` files define service messages, and `.msg` files define topic messages (the file extension introduces confusion and make them seem more generic)

And finally yes, currently services require a linear history to translate while topics (`.msg`) can be split and merged

Thanks for the comment, I realized I was also confused by this. I'll do my best to make this clear in the docs

en/ros2/px4_ros2_msg_translation_node.md
GuillaumeLaine: That would not be a problem, those custom messages would just need the `MESSAGE_VERSION` field to be added and the file moved to the corresponding `versioned/` folder. Updating those message would follow the same steps as updating vanilla px4 messages

en/ros2/px4_ros2_msg_translation_node.md
GuillaumeLaine: The existing translations would be under `msg/translation_node/translations/*.h`. I've reworked the section to make it clearer

en/ros2/user_guide.md
GuillaumeLaine: Good point, I've added some optional steps in that section, also linking to the page on translation node

en/middleware/uorb.md
GuillaumeLaine: Good idea, added a diagram and a note concerning ROS native translation.

> ROS2 Message Translation node supports versioning, not the versioning supporting the translation node.

I may be thinking of this wrong, but in my mind message versioning works as a standalone feature (it's just about some messages having a version number incremented, others not). The translation node builds on top of versioned messages. In that sense, the translation supporting the versioning sounds backwards to me. How do you see this?

en/ros2/px4_ros2_msg_translation_node.md
bkueng: Jazzy or later

en/ros2/px4_ros2_msg_translation_node.md
bkueng: > So a ROS application must compile against just one version of a topic?

Yes

> What if there are two ROS applications running on the network compiled against different versions - is this the problem case?

No, only a specific subset, like:
- app 1: pub topic_v1, sub topic_v1
- app 2: pub topic_v2, sub topic_v2

In practice there's rarely (if any) setup that will run into this.

bkueng (commented):
Nice work, thanks for going through this

GuillaumeLaine:
@hamishwillee Thanks for the thorough review! I've expanded on what was there, hopefully this addresses most of your comments. Let me know if parts are still unclear when you're back.

In the meantime @beniaminopozzan would you like to have a look? Hamish told me might be interested in being kept in the loop

en/middleware/uorb.md
```diff
@@ -141,7 +141,33 @@ Make sure not to mix `orb_advertise_multi` and `orb_advertise` for the same topi
...
 The full API is documented in [platforms/common/uORB/uORBManager.hpp](https://github.com/PX4/PX4-Autopilot/blob/main/platforms/common/uORB/uORBManager.hpp).
 
+## Message Versioning
 
+<Badge type="tip" text="main (PX4 v1.16+)" />
+
+Message versioning has been in introduced in PX4 v1.16 (main) to make it easier to maintain compatibility between PX4 and ROS 2 versions.
+
+Versioned messages include an additional field `uint32 MESSAGE_VERSION = x`, where `x` corresponds to the current version of the message.
+Versioned message files are stored separetly from their non-versioned counterpart.
```
beniaminopozzan, line 151:
```suggestion
Versioned message files are stored separately from their non-versioned counterpart.
```

en/middleware/uorb.md
```diff
@@ -141,7 +141,33 @@ Make sure not to mix `orb_advertise_multi` and `orb_advertise` for the same topi
...
+
+Updating a versioned message involves more steps compared to updating a non-versioned one.
+Please see the section on [Updating a Versioned Message](../ros2/px4_ros2_msg_translation_node.md#updating-a-versioned-message).
+
+For the full list of versioned and non-versioned messages, refer to the [uORB Message Reference](../msg_docs/index.md).
+
+For more on PX4 and ROS 2 communication, refer to the page about the [PX4-ROS 2 Bridge](../ros/ros2_comm.md).
+
+::: note
+ROS 2 plans to natively support message versioning in the future, but this is not implmemented yet.
```
beniaminopozzan, line 167:
```suggestion
ROS 2 plans to natively support message versioning in the future, but this is not implememented yet.
```

en/ros2/px4_ros2_msg_translation_node.md
```diff
@@ -0,0 +1,434 @@
+# PX4 ROS 2 Message Translation Node
+
+<Badge type="tip" text="main (PX4 v1.16+)" /> <Badge type="warning" text="Experimental" />
+
+The message translation node allows ROS 2 applications that were compiled against different versions of the PX4 messages to interwork with newer versions of PX4, and vice versa, without having to change either the application or the PX4 side.
+
+## Overview
+
+The translation of messages from one definition version to another is possbile thanks to the introduction of [message versioning](../middleware/uorb.md#message-versioning).
```
beniaminopozzan, line 9:
```suggestion
The translation of messages from one definition version to another is possible thanks to the introduction of [message versioning](../middleware/uorb.md#message-versioning).
```

en/ros2/px4_ros2_msg_translation_node.md
```diff
@@ -0,0 +1,434 @@
...
+
+The translation node has access to all message versions previously defined by PX4. It dynamically observes the DDS data space, monitoring the publications, subscriptions and services originating from either PX4 via the [uXRCE-DDS Bridge](../middleware/uxrce_dds.md), or ROS2 applications. When necessary, it converts messages to the current versions expected by both applications and PX4, ensuring compatibility.
```
beniaminopozzan, line 11:
```suggestion
The translation node has access to all message versions previously defined by PX4. It dynamically observes the DDS data space, monitoring the publications, subscriptions and services originating from either PX4 via the [uXRCE-DDS Bridge](../middleware/uxrce_dds.md), or ROS 2 applications. When necessary, it converts messages to the current versions expected by both applications and PX4, ensuring compatibility.
```

en/ros2/px4_ros2_msg_translation_node.md
```diff
@@ -0,0 +1,434 @@
...
+
+::: tab Python
+
+```python
+import rclpy
+from rclpy.node import Node
+from px4_msgs.msg import VehicleCommand, VehicleAttitude
+
+# Helper function to get the message version suffix
+# The correct message version is directly inferred from the message defintion
```
beniaminopozzan, line 145:
```suggestion
# The correct message version is directly inferred from the message definition
```

en/ros2/px4_ros2_msg_translation_node.md
```diff
@@ -0,0 +1,434 @@
...
+
+### Definitions
+
+A __message__ defines the data format used for communication, whether over a topic or a service.
+Therefore a message can be either a _topic_ message defined by a `.msg` file, or a _service_ message defined by a `.srv` file.
+
+A __versioned message__ is a message for which changes are tracked and each change results in a version bump, with the previous state of the definition being stored in history.
+The latest version of every message is stored in `msg/versioned/` for topics (or `srv/versioned` for services), and all older versions are stored in `msg/px4_msgs_old/msg/` (or `msg/px4_msgs_old/srv/`). 
+
+A __version translation__ defines a bidrectional mapping of the contents of one or more message definition across different versions.
```
beniaminopozzan, line 196:
```suggestion
A __version translation__ defines a bidirectional mapping of the contents of one or more message definition across different versions.
```

en/ros2/px4_ros2_msg_translation_node.md
```diff
@@ -0,0 +1,434 @@
...
+    └── VehicleCommand.srv         # e.g. MESSAGE_VERSION = 2
+```
+
+#### Directory `px4_msgs_old/`
+
+- Archives the history of all versioned messages, including both topic and service messages (resp. under `msg/` and `srv/` subdirectories).
+- Each file includes a `MESSAGE_VERSION` field.
+- File names reflect the message's version with a suffix (e.g., `V1`, `V2`).
+
+Example directory structure (coherant with example above):
```
beniaminopozzan, line 249:
```suggestion
Example directory structure (coherent with example above):
```

en/ros2/px4_ros2_msg_translation_node.md
```diff
@@ -0,0 +1,434 @@
...
+      └── VehicleCommandV1.srv
+```
+
+#### Directory `translation_node/`
+
+- Contains headers for translating between all different versions of messages.
+- Each translation (direct or generic) is a single `.h` header file.
+- The header `all_translation.h` acts as the main header, and includes all subsequent translation headers.
+
+Example directory structure (coherant with example above):
```
beniaminopozzan, line 269:
```suggestion
Example directory structure (coherent with example above):
```

beniaminopozzan (commented):
Thanks @GuillaumeLaine !
I just pointed out a few typo. Nothing else for me to report - just exited to see it finished!

en/ros2/px4_ros2_msg_translation_node.md
hamishwillee: Ah, probably we need to update https://docs.px4.io/main/en/middleware/uorb.html#adding-a-new-topic with a new heading on nesting - this mentions nesting, but has no example and is easy to miss. That would be separate job, but if you're able to to it would be appreciated.

en/ros2/user_guide.md
```diff
@@ -264,6 +267,13 @@ To create and build the workspace:
    git clone https://github.com/PX4/px4_ros_com.git
    ```
 
+1. **(Optional)** From PX4 v1.16 (main), include the [Message Translation Node](../ros2/px4_ros2_msg_translation_node.md) into your workspace by running the following script.
+   This step is needed only if using `px4_msgs` that do not match your target version of PX4.
```
hamishwillee, line 271: I modified this line. Please check it is correct.

FWIW I think the updates to this doc work well.

en/middleware/uorb.md
```diff
@@ -141,7 +141,40 @@ Make sure not to mix `orb_advertise_multi` and `orb_advertise` for the same topi
...
+<Badge type="tip" text="main (PX4 v1.16+)" />
+
+Optional message versioning was introduced in PX4 v1.16 (main) to make it easier to maintain compatibility between PX4 and ROS 2 versions.
+Versioned messages are designed to remain more stable over time compared to their non-versioned counterparts, as they are intended to be used across multiple releases of PX4 and external systems, ensuring greater compatibility over longer periods.
+
+Versioned messages include an additional field `uint32 MESSAGE_VERSION = x`, where `x` corresponds to the current version of the message.
+
+Versioned and non-versioned messages are separated in the file system:
+
+- Non-versioned topic message files and service message files remain in the [`msg/`](https://github.com/PX4/PX4-Autopilot/tree/main/msg) and [`srv/`](https://github.com/PX4/PX4-Autopilot/tree/main/srv) directories, respectively.
```
hamishwillee, line 155: I've subedited this a little, mainly "to my taste".

It is still a problem for me that this is the first and only mention of "topic message files and service message files" and that there is no definition of what these are for, or where they are from. If you're from ROS-land perhaps you know, but I can assure you that many readers will not.

Can you create a section in this topic which outlines the differences/similarities, and when you should create one or the other. Can be a different PR.

en/ros2/px4_ros2_msg_translation_node.md
hamishwillee: Ah, so you only need the translation node running in one workspace - this observes the DDS data space. So you could run this node in the same workspace as an application or a different workspace. Right?

I see that in the User guide you have included the translation node in the same workspace as the example workspace you are building. I took that to mean "if you build a workspace and it needs translation, then the node must be running in that workspace" AND "if you have multiple workspaces that need translation, they would each need to run the note.

So what I would do is leave this page alone.

In the **user guide** I would remove the Optional steps you added into their own section "Starting the translation node". I'd put that after the example and have some very simple text and a link to the installation section below. The point being that you want to forward reference that you might need to use a translation node, but actually this is not something you need to do for this example, and not something you need to do every workspace.

The text might start something like ...

This example is built with PX4 and ROS2  versions that use the same message definitions. 
If you were to use incompatible [message versions] you would need to install and run a [translation node] as well, before running the example.



Does that make sense?

en/ros2/px4_ros2_msg_translation_node.md
```diff
@@ -0,0 +1,434 @@
...
+
+The message translation node allows ROS 2 applications that were compiled against different versions of the PX4 messages to interwork with newer versions of PX4, and vice versa, without having to change either the application or the PX4 side.
+
+## Overview
+
+The translation of messages from one definition version to another is possible thanks to the introduction of [message versioning](../middleware/uorb.md#message-versioning).
+
+The translation node has access to all message versions previously defined by PX4. It dynamically observes the DDS data space, monitoring the publications, subscriptions and services originating from either PX4 via the [uXRCE-DDS Bridge](../middleware/uxrce_dds.md), or ROS 2 applications. When necessary, it converts messages to the current versions expected by both applications and PX4, ensuring compatibility.
+
+![Overview ROS 2 Message Translation Node](../../assets/middleware/ros2/px4_ros2_interface_lib/translation_node.svg)
```
hamishwillee, line 13: I understand this diagram, but it took me a little while to understand what it meant - specifically what versions are supported. Do you think this makes it easier?

image

en/ros2/px4_ros2_msg_translation_node.md
```diff
@@ -0,0 +1,445 @@
...
+      ├── translation_vehicle_command_v1.h          # Direct translation v0 <-> v1
+      └── translation_vehicle_command_v2.h          # Direct translation v1 <-> latest (v2)
+```
+
+### Updating a Versioned Message
+
+This section provides a step-by-step walkthrough and a basic working example of what the process of changing a versioned message looks like.
+
+Before making changes to a versioned message (in `msg/versioned/` or `srv/versioned/`), we first create an archived copy of it, and update existing translation code to point to the archived version definition.
+We then update versioned message definition to increment the version number, add any new fields or other changes, and then add a new translation file to convert between it and the previous (now archived) version.
```
hamishwillee, line 295:
```suggestion
We then update the versioned message definition to increment the version number, add any new fields or other changes, and then add a new translation file to convert between it and the previous (now archived) version.
```

en/ros2/px4_ros2_msg_translation_node.md
```diff
@@ -0,0 +1,445 @@
...
+
+   For example, append the following line to `all_translation.h`:
+
+   ```c++
+   #include "translation_vehicle_attitude_v4.h"
+   ```
+
+Note that in the example above and in most cases, step 4 only requires the developer to create a direct translation for the definition change.
+This is because the changes only involved a single message.
+In more complex cases of splitting, merging and/or moving definitions then a generic translation must be created.
```
hamishwillee, line 411: It would be good to have an example of a generic translation and a service translation. This could be minimal, or a pointer to an example in the codebase once one exists.

en/ros2/px4_ros2_msg_translation_node.md
hamishwillee: PS If this only works with ROS Jazzy, we should say that too, and it makes more sense to do that in a separate section.

en/ros2/px4_ros2_msg_translation_node.md
hamishwillee: @GuillaumeLaine So ROS Jazzy is a dependency - if this is never going to be in ROS Humble we should maybe say

```suggestion
- The current implementation depends on a service API that is present from ROS Jazzy (ROS applications build for Humble and earlier cannot be used with the node).
```

I added a version for ROS Jazzy above on this doc. We should also add it in the guide, as noted in https://github.com/PX4/PX4-user_guide/pull/3465/files#r1924601823

en/ros2/px4_ros2_msg_translation_node.md
GuillaumeLaine: >  So you could run this node in the same workspace as an application or a different workspace. Right?

Correct

> Does that make sense?

It does, I've added such a section in the ROS 2 User Guide 👍🏻

en/ros2/px4_ros2_msg_translation_node.md
```diff
@@ -0,0 +1,445 @@
...
+      ├── translation_vehicle_command_v1.h          # Direct translation v0 <-> v1
+      └── translation_vehicle_command_v2.h          # Direct translation v1 <-> latest (v2)
+```
+
+### Updating a Versioned Message
+
+This section provides a step-by-step walkthrough and a basic working example of what the process of changing a versioned message looks like.
+
+Before making changes to a versioned message (in `msg/versioned/` or `srv/versioned/`), we first create an archived copy of it, and update existing translation code to point to the archived version definition.
+We then update the versioned message definition to increment the version number, add any new fields or other changes, and then add a new translation file to convert between it and the previous (now archived) version.
```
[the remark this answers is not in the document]
GuillaumeLaine: You're right I'm deleting it

en/ros2/px4_ros2_msg_translation_node.md
GuillaumeLaine: @hamishwillee I think there is some confusion with this bullet-point, let me try to clarify:

- On ROS Humble and older, the translation node will work but only for the translation of topic messages, not services messages. Service message translation requires a service API that is present starting from ROS Jazzy
- On ROS Jazzy and later, the translation will work to its full extent, both topic and service translation. 

Currently translating topic messages is the main use-case for the translation node, so in that sense ROS Jazzy isn't a mandatory dependency. But clearly this point wasn't clear, I'll update the docs now

en/middleware/uorb.md
hamishwillee: You are correct. Fortunately this has been fixed by my subedit to my taste.

en/ros2/px4_ros2_msg_translation_node.md
hamishwillee: Ah, I didn't understand - that's pretty much because to me service messages are still a magical nothing that is undefined.

The fix here would be:

 ```suggestion
- Translation of service messages does not work on ROS Humble, but does on ROS Jazzy.
  This is because the current implementation depends on a service API that is not yet available in ROS Humble.
  Translation of topic messages is fully supported.
```

Then undo the badging I added for ROS Jazzy.

This is better because it starts by saying exactly what does not work, rather than the missing dependency.

Make sense?

File: en/ros2/px4_ros2_msg_translation_node.md (hunk 30 above)
Comment (hamishwillee) [reply en/ros2/px4_ros2_msg_translation_node.md:439]: > Having both publishers and subscribers for two different versions of a topic is currently not handled by the translation node and would trigger infinite circular publications.

OK, I think I get it. This fails because you have "both publishers and subscribers for two different versions of a topic"
```
app 1: pub topic_v1, sub topic_v1
app 2: pub topic_v2, sub topic_v2
```
But this would not fail, because you have publishers of a different topic, but all subscribers subscribe to the same topic.
```
app 1: pub topic_v1,  sub topic_v2
app 2: pub topic_v2, sub topic_v2
```

File: en/ros2/px4_ros2_msg_translation_node.md [hunk 45]
```diff
@@ -0,0 +1,442 @@
...
+When moving from one node to the next, the message translation method is called with the current topic data.
+If a node contains an instantiated publisher (because it previously detected an external subscriber), the data is published.
+Thus, multiple subscribers of any version of the topic can be updated with the correct version of the data.
+
+For translations with multiple input topics, the translation continues once all input messages are available.
+
+## Limitations
+
+- The current implementation depends on a service API that is not yet available in ROS Humble, and therefore does not support services when built for ROS Humble.
+- Services only support a linear history, i.e. no message splitting or merging.
```
hamishwillee, line 438: I think this means 
```suggestion
- Service messages only support a linear history, i.e. no message splitting or merging.
```

It matters because _services_ is a very overloaded term. We should at least make it specific so anyone who doesn't understand what a service message is (i.e. me) knows they can ignore this limitation.

en/ros2/px4_ros2_msg_translation_node.md
hamishwillee: Yes, that's the only change. 

Initially I was thinking of marking up the nodes with the versions of messages that they sub/pub because I wanted to be able to show what the translation node does and why. With colour coding I think we can avoid this.

hamishwillee:
@GuillaumeLaine Thanks for your comments. I can see we're getting pretty close. Any ETA on when https://github.com/PX4/PX4-Autopilot/pull/24113 will get in?

en/ros2/px4_ros2_msg_translation_node.md
bkueng: Exactly

en/ros2/px4_ros2_msg_translation_node.md
GuillaumeLaine: A few lines above we provide links to "version translation templates" which are example translation header files for each scenario: direct topic translation, generic topic translation, and direct service translation. 


Did you have something more elaborate in mind?

GuillaumeLaine:
@hamishwillee The implementation PR is ready to merge in my opinion, it's all good from a user workflow perspective when I tested it. There hasn't been much activity from the community but FWIW I've just given my approval, I think we can expect it to get merged soon

en/ros2/px4_ros2_msg_translation_node.md
hamishwillee: Given I had so much trouble with this, perhaps insert the example? YOur call.

````suggestion
- Having both publishers and subscribers for two different versions of a topic is currently not handled by the translation node and would trigger infinite circular publications:

  ```
  app 1: pub topic_v1, sub topic_v1
  app 2: pub topic_v2, sub topic_v2
  ```

  Note that the far more common case, where you have multiple apps subscribing to the same topic and different publishers (or vice versa), are fine:
  ```

  app 1: pub topic_v1, sub topic_v2
  app 2: pub topic_v2, sub topic_v2
  ```

````

File: en/ros2/px4_ros2_msg_translation_node.md (hunk 43 above)
Comment (hamishwillee) [reply en/ros2/px4_ros2_msg_translation_node.md:408]: A little more ...

I was thinking perhaps show code here for the generic case. But as you imply, actually the template shows exactly this case you have outlined in the example below except the actual moving of the field.

> For example when moving a field from one message to another, a single generic translation should be added with the two older message versions as input, and the two newer versions as output.
> This ensures there is no information lost when translating forward or backward.

The problem is that this wasn't obvious to me, perhaps because I am not a developer. The only comment was  `// Translate ExampleTopic and OtherTopic v1 <--> v2` but that was a bit terse for me.

So perhaps after the "For example" add a comment like:

> This is exactly the approach shown by the [Generic Topic Message Translation Template](https://github.com/PX4/PX4-Autopilot/blob/message_versioning_and_translation/msg/translation_node/translations/example_translation_multi_v2.h), omitting only the code for actually modifying fields in the `fromOlder()` and `toOlder()` methods.

**Comment by hamishwillee** [comment]
Thanks for the update @GuillaumeLaine - really appreciate your attention to detail. I am happy for this to go in. There are two more suggestions for your consideration:
- https://github.com/PX4/PX4-user_guide/pull/3465#discussion_r1934951052
- https://github.com/PX4/PX4-user_guide/pull/3465#discussion_r1934939991

File: en/ros2/px4_ros2_msg_translation_node.md (hunk 43 above)
Comment (GuillaumeLaine) [reply en/ros2/px4_ros2_msg_translation_node.md:408]: Good point, I've applied this change to connect the paragraph to the code example

File: en/ros2/px4_ros2_msg_translation_node.md (hunk 30 above)
Comment (GuillaumeLaine) [reply en/ros2/px4_ros2_msg_translation_node.md]: I've added the example, it certainly helps clarify.

I've changed your suggestion a bit; the reason why the circular configuration is unlikely is because ROS topics from the FMU (carrying the messages that might be translated) are intended to be directional (i.e. /fmu/out vs /fmu/in). So the common case would be even simpler, more like:
```
  app 1: pub topic1_v1, sub topic2_v3
  app 2: pub topic1_v2, sub topic2_v5
```
i.e. apps don't sub to topic1, apps don't pub to topic2

And just to avoid any misconception, I think your 2nd example isn't possible:
  ```
  app 1: pub topic_v1, sub topic_v2
  app 2: pub topic_v2, sub topic_v2
  ```
Namely an app only has a single definition/version for a message, so `app1` could not subscribe to the same topic as `v1` and publish as `v2`

File: en/ros2/px4_ros2_msg_translation_node.md [hunk 46]
```diff
@@ -438,7 +440,13 @@ For translations with multiple input topics, the translation continues once all
...
-- Having both publishers and subscribers for two different versions of a topic is currently not handled by the translation node and would trigger infinite circular publications.
-  This could be extended if required.
+- Having both publishers and subscribers for two different versions of the same topic is currently not handled by the translation node and would trigger infinite circular publications.
+  This refers to the following problematic configuration:
+  ```
+  app 1: pub topic_v1, sub topic_v1
+  app 2: pub topic_v2, sub topic_v2
+  ```
+  In practice this configuration is unlikely to occur because ROS topics shared with the FMU are intended to be directional (e.g. `/fmu/out/vehicle_status` or `/fmu/in/trajectory_setpoint`), therefore apps typically do not publish and subscribe simultaneously to the same topic.
+  The translation node could be extended to handle this corner-case if required.
```
hamishwillee, line 450:
````suggestion
- Having both publishers and subscribers for two different versions of the same topic is currently not handled by the translation node and would trigger infinite circular publications.
  This refers to the following problematic configuration:
  
  ```
  app 1: pub topic_v1, sub topic_v1
  app 2: pub topic_v2, sub topic_v2
  ```

  In practice this configuration is unlikely to occur because ROS topics shared with the FMU are intended to be directional (e.g. `/fmu/out/vehicle_status` or `/fmu/in/trajectory_setpoint`), therefore apps typically do not publish and subscribe simultaneously to the same topic.
  The translation node could be extended to handle this corner-case if required.
````

en/ros2/px4_ros2_msg_translation_node.md
hamishwillee: Thanks - your explanation is much better than mine (obviously!).

hamishwillee:
@GuillaumeLaine I don't have any other comments. Looks good to go in when the code PR goes in.

en/ros2/px4_ros2_msg_translation_node.md
GuillaumeLaine: I missed this comment until now, I've opened a PR for it:
https://github.com/PX4/PX4-user_guide/pull/3562

GuillaumeLaine:
@hamishwillee Awesome thanks! Really appreciate you taking the time to refine things until they're right 🙌🏻

mrpollo:
Once this PR is in, we need a blog post on px4.io, who would like to co-author with me?

hamishwillee:
@GuillaumeLaine I missed that the associated PR merged! Yee har!

I made a few more changes. Updated this first section to make it clear that that the file might not go into the top level.

image

Did restructure to move the topics such as versioning above the stuff around plotting and publishing - they are all associated with message definition and fields, so they don't belong at the end of the doc. 

I am not sure whether versioning should go under message definition - it does contain field information. But since its a bigger topic, and new, its at the top level still.

Merging now - if you think this was a mistake we can revert :-)

hamishwillee:
> Once this PR is in, we need a blog post on px4.io, who would like to co-author with me?

@mrpollo I assume @GuillaumeLaine would like to volunteer. If not, I can help next week.

bkueng:
I can also help. Guillaume should be back next week.

GuillaumeLaine:
@mrpollo I'm can help if there's anything I can do

mrpollo:
@GuillaumeLaine lets coordinate on discord

PR #3460, opened 2024-11-14
Ark docs

patrikpordi:
Added/ Updated ARK Hardware in the Docs
Tested it, looks nice on my end.

Let me know if anything has to change.
Thanks in advance for the merge!

PR #3477, opened 2024-12-03
arkv6x revert page rename to fix broken links

dakejahl:
The page was rename in https://github.com/PX4/PX4-user_guide/pull/3460 and it breaks existing links. Easiest to revert this rename rather than updating links everywhere, especially since the link is a top hit on google.


hamishwillee:
@dakejahl Closing this - fixed with redirects in #3478 - will rebuild tomorrow probably.

Note, you should fix your links up "over time". At some point I will likely clean the redirects - might be years, but it will happen.

PR #3483, opened 2024-12-09
Add Holybro M10 GPS to the list

HB-SP-P:
Added Holybro M10 GPS to the table.
Committed by Peter C. (Holybro support).
support@holybro.com


hamishwillee (approved):
Thanks very much @HB-SP-P

PR #3495, opened 2024-12-12
Mission command timeout

hamishwillee:
Some hardware can take time to reach the state commanded by a corresponding mission item, such as gimbals, grippers, winches.

Ideally you'd trigger transition to the next step on feedback from a sensor on the hardware, but many grippers etc don't have that, and in any case it is possible you will miss it. So for grippers the `MIS_PD_TO` timeout was created - it basically ensures that if you miss feedback (or if there is none) the mission will wait this amount of time before progressing.

https://github.com/PX4/PX4-Autopilot/pull/23960 renames `MIS_PD_TO`  with `MIS_COMMAND_TOUT`, which applies to all of these long running mission items - it also uses this delay to give the gimbal time to move into position on a mission item. 

This provides the corresponding docs.

- https://github.com/PX4/PX4-Autopilot/pull/23960

### DIFF
--- a/en/advanced/gimbal_control.md
+++ b/en/advanced/gimbal_control.md
@@ -75,6 +75,18 @@ For example, you might have the following settings to assign the gimbal roll, pi
 The PWM values to use for the disarmed, maximum and minimum values can be determined in the same way as other servo, using the [Actuator Test sliders](../config/actuators.md#actuator-testing) to confirm that each slider moves the appropriate axis, and changing the values so that the gimbal is in the appropriate position at the disarmed, low and high position in the slider.
 The values may also be provided in gimbal documentation.
 
+## Gimbal Control in Missions
+
+[Gimbal Manager commands](https://mavlink.io/en/services/gimbal_v2.html#gimbal-manager-messages) may be used in missions if supported by the vehicle type.
+For example [MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW) is supported in [multicopter mission mode](../flight_modes_mc/mission.md).
+
+In theory you can address commands to a particular gimbal, specifying its component id using the "Gimbal device id" parameter.
+However at time of writing (December 2024) this is [not supported](https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/gimbal/input_mavlink.cpp#L889): all commands are sent to the gimbal with id [MAV_COMP_ID_GIMBAL (154)](https://mavlink.io/en/messages/common.html#MAV_COMP_ID_GIMBAL)
+
+Gimbal movement is not immediate.
+To ensure that the gimbal has time to move into position before the mission progresses to the next item (if gimbal feedback is not provided or lost), you should set [MIS_COMMAND_TOUT](../advanced_config/parameter_reference.md#MIS_COMMAND_TOUT) to be greater than the time taken for the gimbal to traverse its full range.
+After this timeout the mission will proceed to the next item.
+
 ## SITL
 
 The [Gazebo Classic](../sim_gazebo_classic/index.md) simulation [Typhoon H480 model](../sim_gazebo_classic/vehicles.md#typhoon-h480-hexrotor) comes with a preconfigured simulated gimbal.
--- a/en/advanced/gimbal_control.md
+++ b/en/advanced/gimbal_control.md
@@ -80,9 +80,6 @@ The values may also be provided in gimbal documentation.
 [Gimbal Manager commands](https://mavlink.io/en/services/gimbal_v2.html#gimbal-manager-messages) may be used in missions if supported by the vehicle type.
 For example [MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW) is supported in [multicopter mission mode](../flight_modes_mc/mission.md).
 
-In theory you can address commands to a particular gimbal, specifying its component id using the "Gimbal device id" parameter.
-However at time of writing (December 2024) this is [not supported](https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/gimbal/input_mavlink.cpp#L889): all mission commands are sent to the gimbal with id defined in the parameter [MNT_MAV_COMPID](../advanced_config/parameter_reference.md#MNT_MAV_COMPID), which is set by default to [MAV_COMP_ID_GIMBAL (154)](https://mavlink.io/en/messages/common.html#MAV_COMP_ID_GIMBAL).
-
 Gimbal movement is not immediate.
 To ensure that the gimbal has time to move into position before the mission progresses to the next item (if gimbal feedback is not provided or lost), you should set [MIS_COMMAND_TOUT](../advanced_config/parameter_reference.md#MIS_COMMAND_TOUT) to be greater than the time taken for the gimbal to traverse its full range.
 After this timeout the mission will proceed to the next item.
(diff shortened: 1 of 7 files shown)


en/advanced/gimbal_control.md
```diff
@@ -76,6 +76,18 @@ For example, you might have the following settings to assign the gimbal roll, pi
...
+For example [MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW) is supported in [multicopter mission mode](../flight_modes_mc/mission.md).
+
+In theory you can address commands to a particular gimbal, specifying its component id using the "Gimbal device id" parameter.
+However at time of writing (December 2024) this is [not supported](https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/gimbal/input_mavlink.cpp#L889): all commands are sent to the gimbal with id [MAV_COMP_ID_GIMBAL (154)](https://mavlink.io/en/messages/common.html#MAV_COMP_ID_GIMBAL)
+
+Gimbal movement is not immediate.
+To ensure that the gimbal has time to move into position before the mission progresses to the next item (if gimbal feedback is not provided or lost), you should set [MIS_COMMAND_TOUT](../advanced_config/parameter_reference.md#MIS_COMMAND_TOUT) to be greater than the time taken for the gimbal to traverse its full range.
+After this timeout the mission will proceed to the next item.
```
hamishwillee, line 89: @julianoes Is this accurate? Specifically, my assumption is that all gimbal commands are sent to MAV_COMP_ID_GIMBAL  but I couldn't find where.

en/flight_modes_mc/mission.md
```diff
@@ -191,6 +191,16 @@ Please add an issue report or PR if you find a missing/incorrect message.
...
 
+## Mission Command Timeouts
+
+Some mission commands/items can take time to complete, such as a gripper opening and closing, a winch extending or retracting, or a gimbal moving to point at a region of interest.
+
+Where provided PX4 may use sensor feedback from the hardware to determine when the action has completed and then move to the next mission item.
+If not provided, or if the feedback is lost, a mission command timeout can be used to ensure that these kinds of actions will progress to the next mission item rather than blocking progression.
+
+The timeout is set using the [MIS_COMMAND_TOUT](../advanced_config/parameter_reference.md#MIS_COMMAND_TOUT) parameter.
+This should be set to be a small amount greater than the time required for the longest long-running action in the mission to complete.
```
hamishwillee, line 202: Note, add this to all the vehicle mission mode sections.

en/releases/main.md
```diff
@@ -37,7 +37,11 @@ Please continue reading for [upgrade instructions](#upgrade-guide).
 - [Log Encryption](../dev_log/log_encryption.md) now generates an encrypted log that contains the public-key-encrypted symmetric key that can be used to decrypt it, instead of putting the key into a separate file.
   This makes log decryption much easier, as there is no need to download or identify a separate key file.
   ([PX4-Autopilot#24024](https://github.com/PX4/PX4-Autopilot/pull/24024)).
-  
+- The generic mission command timeout [MIS_COMMAND_TOUT](../advanced_config/parameter_reference.html#MIS_COMMAND_TOUT) parameter replaces the delivery-specific `MIS_PD_TO` parameter.
```
hamishwillee, line 40:
```suggestion
- The generic mission command timeout [MIS_COMMAND_TOUT](../advanced_config/parameter_reference.md#MIS_COMMAND_TOUT) parameter replaces the delivery-specific `MIS_PD_TO` parameter.
```

en/advanced/gimbal_control.md
julianoes: I had no idea this param existed.

I would be surprised if this worked at all. Something to check...

In terms of timeout: Usually the gimbals just answer with Ack, and then go to the setpoint, however slow they want. This behavior makes sense when used as a command - where you might want to send a new setpoint while it's still moving - but less as part of a mission, unfortunately. It might requiring adding in a wait command explicitly after. 

In my mind, if a command is not answered immediately, it should be followed up by a IN_PROGRESS to avoid a timeout, as per protocol.

hamishwillee: Its a new param added in https://github.com/PX4/PX4-Autopilot/pull/23960

Essentially it copies what was done for payload delivery - if you have feedback then the mission item completes on getting that, but if you don't then it completes after a timeout - ensuring the gimbal has time to reach your target setpoint, or your gripper has time to open/closed.

This is what is done, and will work. Is it the right thing? Depends on whether the gimbal protocol provides feedback on reaching the target as you suggest or just acknowledges the command. We don't define the camera commands as long running, so I suspect the former is what happens. Is there any other way to get feedback?

In any case, can you confirm that commands are sent to MAV_COMP_ID_GIMBAL ?

julianoes: I will have to test this. On my list now. Being able to set the timeout can be good of course.

en/advanced/gimbal_control.md
```diff
@@ -76,6 +76,18 @@ For example, you might have the following settings to assign the gimbal roll, pi
...
 The values may also be provided in gimbal documentation.
 
+## Gimbal Control in Missions
+
+[Gimbal Manager commands](https://mavlink.io/en/services/gimbal_v2.html#gimbal-manager-messages) may be used in missions if supported by the vehicle type.
+For example [MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW) is supported in [multicopter mission mode](../flight_modes_mc/mission.md).
+
+In theory you can address commands to a particular gimbal, specifying its component id using the "Gimbal device id" parameter.
+However at time of writing (December 2024) this is [not supported](https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/gimbal/input_mavlink.cpp#L889): all commands are sent to the gimbal with id [MAV_COMP_ID_GIMBAL (154)](https://mavlink.io/en/messages/common.html#MAV_COMP_ID_GIMBAL)
```
hamishwillee, line 85: Note may be no more true than the last version - the system might follow the gimbal device id. However it is more accurate in that IF the id is hard coded, it is hard coded to the parameter value, not a particular value.


```suggestion
However at time of writing (December 2024) this is [not supported](https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/gimbal/input_mavlink.cpp#L889): all mission commands are sent to the gimbal with id defined in the parameter [MNT_MAV_COMPID](../advanced_config/parameter_reference.md#MNT_MAV_COMPID), which is set by default to [MAV_COMP_ID_GIMBAL (154)](https://mavlink.io/en/messages/common.html#MAV_COMP_ID_GIMBAL).
```

en/advanced/gimbal_control.md
```diff
@@ -75,6 +75,18 @@ For example, you might have the following settings to assign the gimbal roll, pi
...
 
+## Gimbal Control in Missions
+
+[Gimbal Manager commands](https://mavlink.io/en/services/gimbal_v2.html#gimbal-manager-messages) may be used in missions if supported by the vehicle type.
+For example [MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW) is supported in [multicopter mission mode](../flight_modes_mc/mission.md).
+
+In theory you can address commands to a particular gimbal, specifying its component id using the "Gimbal device id" parameter.
+However at time of writing (December 2024) this is [not supported](https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/gimbal/input_mavlink.cpp#L889): all mission commands are sent to the gimbal with id defined in the parameter [MNT_MAV_COMPID](../advanced_config/parameter_reference.md#MNT_MAV_COMPID), which is set by default to [MAV_COMP_ID_GIMBAL (154)](https://mavlink.io/en/messages/common.html#MAV_COMP_ID_GIMBAL).
+
+Gimbal movement is not immediate.
```
hamishwillee, line 86: Removing this para, as it blocks me merging. The rest is accurate. Will add in another issue.
```suggestion
For example [MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW) is supported in [multicopter mission mode](../flight_modes_mc/mission.md).

Gimbal movement is not immediate.
```

en/advanced/gimbal_control.md
hamishwillee: @julianoes The timeout stuff is accurate and I don't want to wait on your testing to get this PR in. So what I have done is removed the text about addressing and recreated the problematic part in https://github.com/PX4/PX4-user_guide/pull/3514

PR #3496, opened 2024-12-12
Adding gimbal Gazebo description

Perrrewi:
Adding descriptors of the gimbal model

- [x] Depends on https://github.com/PX4/PX4-gazebo-models/pull/47
- [x] Depends on https://github.com/PX4/PX4-gazebo-models/pull/70
- [x] Depends on https://github.com/PX4/PX4-Autopilot/pull/23382


en/sim_gazebo_gz/vehicles.md
```diff
@@ -104,6 +104,22 @@ make px4_sitl gz_x500_lidar_2d
 The sensor information is written to the [ObstacleDistance](../msg_docs/ObstacleDistance.md) UORB message used by collision prevention.
 :::
 
+### X500 Quadrotor with Gimbal (Front-facing)
```
hamishwillee, line 107: @Perrrewi This all looks great - but we need a little detail on what/how you test with it.

1. I assume this simulates a MAVLink gimbal? What device id/component ID does the gimbal have? We should state this.
2. Can we link from https://docs.px4.io/main/en/advanced/gimbal_control.html#sitl so that people working on gimbals can find this
3. Also update the testing section below it: https://docs.px4.io/main/en/advanced/gimbal_control.html#testing
   Might be worth having separate sections for Gazebo and Gazebo Classic there so that we can easily delete the classic version when this goes EOL (depends on how it looks when tried).
4. Further, in the testing section, is there anything else we can do? I.e. If I define a mission with ROIs will the gimbal be used for them? In QGC is there a model where I can see the gimbal in operation - i.e. streamed video, and move the gimbal? Such as you might do in Auterion's AMC?
5. Perhaps note here that you need to configure PX4 as per [(find link in](https://docs.px4.io/main/en/advanced/gimbal_control.html#mavlink-gimbal-mnt-mode-out-mavlink) (presumably)?

In other words, stuff to make this findable and usable if people arrive at this topic, or at the gimbal topic.

hamishwillee: PS I presume that this can't merge until all the PRs in the description are merged.

PR #3500, opened 2024-12-12
Update ark_12s_pab_power_module.md

patrikpordi:
Adjusted Shunt

### DIFF
--- a/en/power_module/ark_12s_pab_power_module.md
+++ b/en/power_module/ark_12s_pab_power_module.md
@@ -12,9 +12,9 @@ Order this module from:
 
 ## Hardware Specifications
 
-- **TI INA226 Digital Power Monitor**
+- **TI INA238 Digital Power Monitor**
 
-  - 0.1mOhm Shunt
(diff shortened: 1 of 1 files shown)


en/power_module/ark_12s_pab_power_module.md
```diff
@@ -12,9 +12,9 @@ Order this module from:
 
 ## Hardware Specifications
 
-- **TI INA226 Digital Power Monitor**
+- **TI INA238 Digital Power Monitor**
 
-  - 0.1mOhm Shunt
+  - 0.0001 Ohm Shunt
```
hamishwillee, line 17: Just FMI, why is it better to present as 0,0001 ohm than 0.1 mOhm?

hamishwillee (approved):
Thanks.

en/power_module/ark_12s_pab_power_module.md
patrikpordi: You have to set the param in Ohm, so we just want to minimize the error from the user side, so they can just copy paste it!

PR #3513, opened 2024-12-18
Gimbal control - clarify range for component ids

hamishwillee:
This clarifies the multi-gimbal case. Specifically that you can have additional gimbals but they are effectively independent of PX4.

FYI @julianoes

### DIFF
--- a/en/advanced/gimbal_control.md
+++ b/en/advanced/gimbal_control.md
@@ -51,13 +51,12 @@ For example, if the `TELEM2` port on the flight controller is unused you can con
 PX4 can automatically create a gimbal manager for a connected PWM gimbal or the first MAVLink gimbal device with the same system id it detects on any interface.
 It does not automatically create gimbal manager for any other MAVLink gimbal devices that it detects.
 
-You can support additional gimbals provided that they:
+You can support additional MAVLink gimbals provided that they:
 
-- implement the gimbal _manager_ protocol
+- Implement the gimbal _manager_ protocol.
 - Are visible to the ground station and PX4 on the MAVLink network.
   This may require that traffic forwarding be configured between PX4, the GCS, and the gimbal.
-- Each gimbal must have a unique component id.
-  For a PWM connected gimbal this will be the component ID of the autopilot
+- Have a unique component id, and this component id must be in the range 7 - 255.
 
 ## Gimbal on FC PWM Output (MNT_MODE_OUT=AUX)
 


en/advanced/gimbal_control.md
```diff
@@ -51,13 +51,12 @@ For example, if the `TELEM2` port on the flight controller is unused you can con
...
-You can support additional gimbals provided that they:
+You can support additional MAVLink gimbals provided that they:
 
-- implement the gimbal _manager_ protocol
+- Implement the gimbal _manager_ protocol.
 - Are visible to the ground station and PX4 on the MAVLink network.
   This may require that traffic forwarding be configured between PX4, the GCS, and the gimbal.
-- Each gimbal must have a unique component id.
-  For a PWM connected gimbal this will be the component ID of the autopilot
+- Have a unique component id, and this component id must be in the range 7 - 255.
```
hamishwillee, line 59: To match the restrictions of connected devices in MAVLink

en/advanced/gimbal_control.md
```diff
@@ -51,13 +51,12 @@ For example, if the `TELEM2` port on the flight controller is unused you can con
...
 
-You can support additional gimbals provided that they:
+You can support additional MAVLink gimbals provided that they:
 
-- implement the gimbal _manager_ protocol
+- Implement the gimbal _manager_ protocol.
 - Are visible to the ground station and PX4 on the MAVLink network.
   This may require that traffic forwarding be configured between PX4, the GCS, and the gimbal.
-- Each gimbal must have a unique component id.
-  For a PWM connected gimbal this will be the component ID of the autopilot
```
hamishwillee, line 60: FYI @julianoes - this is not relevant in a discussion about the additional gimbals. Note that we should require that the device ID for the autopilot connected camera is set to 1 by default, and reject other values.

hamishwillee:
PS We should also say something about mission addressing support, but we need to test that.

en/advanced/gimbal_control.md
julianoes: In theory that's there:
https://github.com/PX4/PX4-Autopilot/blob/fd87cd682db5360de4408aefa1b93bad1eaa1d3d/src/modules/gimbal/output_rc.cpp#L65-L66

julianoes: In practice it seems to be correct for GIMBAL_MANAGER_STATUS:

image

But wrong for GIMBAL_DEVICE_ATTITUDE_STATUS:

image

And here is a fix: https://github.com/PX4/PX4-Autopilot/pull/24123

hamishwillee: > In theory that's there: https://github.com/PX4/PX4-Autopilot/blob/fd87cd682db5360de4408aefa1b93bad1eaa1d3d/src/modules/gimbal/output_rc.cpp#L65-L66

That defines it in `OutputRC` - I assume that is the oddly named equivalent of the `AUX` (PWM) gimbal? Cool!

We probably don't yet report errors on non-1 values in the mission though, do we? [don't bother looking if you don't know :-)]

hamishwillee: Thanks for this. Good to get it clear.

PR #3531, opened 2024-12-25
fix typo in vehicles.md

jonasdlindner:
It looks like the text "frac" is rendered instead of the interpreting the next chars as frac.


hamishwillee (approved):
Good catch! Thank you.

PR #3532, opened 2025-01-06
Refactor magnetometer fusion and add mag_type "init"

bresch:
Mag type "init" was added in https://github.com/PX4/PX4-Autopilot/pull/23185 but documentation was missing


en/advanced_config/tuning_the_ecl_ekf.md
```diff
@@ -127,26 +127,46 @@ This minimum data set is required for all EKF modes of operation. Other sensor d
...
-Setting '3-axis' mode all the time (`EKF2_MAG_TYPE=2`) is more error-prone, and requires that all the IMUs are well calibrated.
-
-The option is available to operate without a magnetometer, either by replacing it using [yaw from a dual antenna GPS](#yaw-measurements) or using the IMU measurements and GPS velocity data to [estimate yaw from vehicle movement](#yaw-from-gps-velocity).
+Three axis body fixed magnetometer data at a minimum rate of 5Hz is required to be considered by the estimator.
+
+::: info
+
+- The magnetometer **biases** are only observable while the drone is rotating
+- The true heading is observable when the vehicle is accelerating (linear acceleration) while absolute position or velocity measurements are fused (e.g. GPS).
+  This means that magnetometer heading measurements are optional after initialization if those conditions are met often enough to constrain the heading drift (caused by gyro bias).
```
hamishwillee, line 136: I understand that we can get heading from GPS-measured position/velocity when moving so we don't need magnetometer readings when moving, as long as we're getting enough info to constrain the heading drift - right. Presumably there might be some drift when we stop in position?

When you say "optional after initialization" do you mean "before arming"? I mean, why do we need them on initialization/or does the measurement matter until we're ready to fly?

en/advanced_config/tuning_the_ecl_ekf.md
```diff
@@ -127,26 +127,46 @@ This minimum data set is required for all EKF modes of operation. Other sensor d
...
+     See [Estimate yaw from vehicle movement](#yaw-from-gps-velocity) below.
+6. Init only:
+   - Magnetometer data is only used to initialize the heading estimate.
+     This is useful when the data can be used before arming but not afterwards (e.g.: high current after the vehicle is armed).
+   - After initialization, the heading is constrained using other observations.
+   - Unlike mag type `None`, when combined with GPS measurements, this method allows position controlled modes to run directly during takeoff.
+
+The following selection tree can be used to select the right option:
+
+![EKF mag type selection tree](../../assets/config/ekf/ekf_mag_type_selection_tree.png)
```
hamishwillee, line 169: FYI, Images are best in-tree (done). I also ran prettier and fixed a few typos.

hamishwillee:
@bresch Did this go into PX4 v1.15 or main? I want to add a release note for this.

Looks great. A few questions mostly FMI here https://github.com/PX4/PX4-user_guide/pull/3532/files#r1926316422 but I'm going to merge.

en/advanced_config/tuning_the_ecl_ekf.md
bresch: > Presumably there might be some drift when we stop in position?

Yes, the heading can drift if you hover in place but usually there are enough disturbances to always provide the tiny bit of motion required to constrain the heading drift. Also, when the position controller is active, an incorrect heading makes the drone "toilet bowl", which makes the heading observable, so unless really unlucky, the heading cannot really drift away by a lot without causing disturbances that makes the heading observable.

bresch: > When you say "optional after initialization" do you mean "before arming"? I mean, why do we need them on initialization/or does the measurement matter until we're ready to fly?

We often just need the magnetometer to align the heading and start fusing GNSS. So even before arming, if the drone is not moving, the heading won't change and the drift is really low, and if it's moving the heading is observable so we don't need a mag at all to constrain the drift

bresch:
> Did this go into PX4 v1.15 or main?

That feature was merged after 1.15

en/advanced_config/tuning_the_ecl_ekf.md
bresch: Ok, good to know, thanks!

hamishwillee:
Cheers. Added a relnote in https://github.com/PX4/PX4-user_guide/pull/3552

PR #3534, opened 2025-01-07
add page for statistical flight log analysis

bkueng:
The question for license for public flight logs came up on https://discuss.px4.io/t/license-for-the-dataset/42819, and in the process of [clarifying that](https://github.com/PX4/flight_review/pull/302), I thought it makes sense to add a short page about the option for statistical log analysis to the docs.

### DIFF
--- a/en/dev_log/flight_log_analysis_statistical.md
+++ b/en/dev_log/flight_log_analysis_statistical.md
@@ -1,0 +1,18 @@
+# Statistical Flight Log Analysis
+
+Flight Review hosts a large set of publicly available log files.
+These can be used for example for statistical analysis or machine learning.
+
+The dataset contains a set of different:
+- vehicle types
+- PX4 versions (including development versions)
+- boards
+- flight modes
+
+The logs are accessible on [logs.px4.io/browse](https://logs.px4.io/browse) and are licensed under [CC-BY PX4](https://creativecommons.org/licenses/by/4.0/).
+
+Log files can also be downloaded in bulk with the [download_logs.py](https://github.com/PX4/flight_review/blob/main/app/download_logs.py) script.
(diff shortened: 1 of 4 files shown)


en/dev_log/flight_log_analysis_statistical.md
```diff
@@ -0,0 +1,18 @@
+# Statistical Flight Log Analysis
```
hamishwillee, line 1: @bkueng It seems to me that this topic is really "how to download flight review logs", which can then be used for any purpose.
A topic named Statistical Flight Log analysis would seem to me to be about "how you go about doing such an analysis", and it might include information such as where to get public data sets, such as the ones in Flight review.

So proposing that we move this content into [Log Analysis using Flight Review](https://docs.px4.io/main/en/log/flight_review.html). This would be a section at the bottom "Downloading Pubic Flight Logs". In the introduction perhaps a link or note `Public Flight Review logs can be downloaded for statistical analysis or machine learning: See XXx.

What do you think? I can do this if you think my point is reasonable?


```suggestion
# Statistical Flight Log Analysis
```

bkueng: You're right, it's a better description for what it is about.
My reasoning was that I wanted to maximize the chance for someone searching in general to find it (maybe not even knowing about flight review), and that the statistical analysis is the primary use-case for downloading those logs in bulk.
I also thought about adding it to the Log Analysis page, but concluded it's not a good fit as that page is specifically about analyzing a single (or specific set of) logs with the flight review plotting page.
But I'm also not specifically against adding it there if you still think it makes the most sense.

hamishwillee: All good points. I've left as is, but added an introduction to the page that makes it clear that this page could grow with other resources when we have them. Still discoverable, but now the title and content match better.

PR #3554, opened 2025-01-29
video_streaming - Update QGC streaming doc link

### LINKED ISSUE #3430: Documentation 'Video Streaming (Companion Computer/QGroundControl)' page: external link to VideoReceiver is broken
Please update page 'Video Streaming (Companion Computer/QGroundControl)' https://github.com/PX4/PX4-user_guide/blob/main/en/companion_computer/video_streaming.md because external link to https://github.com/mavlink/qgroundcontrol/blob/master/src/VideoReceiver/index.md is broken

hamishwillee:
Fixes #3430

### DIFF
--- a/en/companion_computer/video_streaming.md
+++ b/en/companion_computer/video_streaming.md
@@ -12,7 +12,41 @@ _QGroundControl_ uses GStreamer 1.18.1 and a stripped down version of _QtGstream
 
 ## Companion Computer Setup
 
-General instructions for installing _GStreamer_ and starting the stream on a companion computer are provided in the [QGroundControl VideoReceiver GStreamer README](https://github.com/mavlink/qgroundcontrol/blob/master/src/VideoManager/VideoReceiver/GStreamer/README.md).
+You will need to install the _GStreamer_ packages that match those used by QGC.
+
+For a Ubuntu companion, a minimal set might be:
+
+```sh
+sudo apt install gstreamer1.0-plugins-bad gstreamer1.0-libav gstreamer1.0-gl -y
+```
+
+For the full set you can mirror the QGC dependencies installed by [/tools/setup/install-dependencies-debian.sh](https://github.com/mavlink/qgroundcontrol/blob/master/tools/setup/install-dependencies-debian.sh).
+At time of writing this is:
+
+```sh
+DEBIAN_FRONTEND=noninteractive apt-get -y --quiet install \
+    libgstreamer1.0-dev \
+    libgstreamer-plugins-bad1.0-dev \
+    libgstreamer-plugins-base1.0-dev \
+    libgstreamer-plugins-good1.0-dev \
+    libgstreamer-gl1.0-0 \
+    gstreamer1.0-plugins-bad \
+    gstreamer1.0-plugins-base \
+    gstreamer1.0-plugins-good \
+    gstreamer1.0-plugins-ugly \
+    gstreamer1.0-plugins-rtp \
+    gstreamer1.0-gl \
+    gstreamer1.0-libav \
+    gstreamer1.0-rtsp \
+    gstreamer1.0-vaapi \
+    gstreamer1.0-x
+```
+
+::: tip
+For other companion platforms you can mirror the instructions in the [corresponding installation scripts](QGC).
+:::
+
+General instructions for starting the stream on a companion computer are provided in the [QGroundControl VideoReceiver GStreamer README](https://github.com/mavlink/qgroundcontrol/blob/master/src/VideoManager/VideoReceiver/GStreamer/README.md).
 
 The setup of cameras and data links depends on many factors.
 Examples in this library are listed below (note, these are options, not "recommended"):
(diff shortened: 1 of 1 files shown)


en/companion_computer/video_streaming.md
```diff
@@ -30,7 +30,7 @@ To setup and use video steaming with QGC:
    sudo apt install gstreamer1.0-plugins-bad gstreamer1.0-libav gstreamer1.0-gl -y
```
hamishwillee, line 30: @DonLakeFlyer @HTRamsey This is the source for the PX4 instructions for installing the GStreamer on Ubuntu for working with QGC. 

The linked https://github.com/mavlink/qgroundcontrol/blob/master/src/VideoManager/VideoReceiver/GStreamer/README.md suggests something like this

```suggestion  
   list=$(apt-cache --names-only search ^gstreamer1.0-* | awk '{ print $1 }' | sed -e /-doc/d | grep -v gstreamer1.0-hybris)
   sudo apt-get install $list
   sudo apt-get install libgstreamer-plugins-base1.0-dev libgstreamer-plugins-bad1.0-dev -y
```

Is the PX4 variant "wrong" - i.e. for Ubuntu can I leave it as it is, or should we be using the same instructions as in QGC tree?

HTRamsey: The PX4 variant just includes runtime dependencies (which are included in the AppImage in v5, unlike v4.X) while the QGC one includes build-time libs but also unnecessarily installs literally everything gstreamer related. We've just been relying on a build script (https://github.com/mavlink/qgroundcontrol/blob/master/tools/setup/install-dependencies-debian.sh) that we can update without having to update instructions like these in the future.

There is also a good likelihood that the main/default video receiver switches to QGC provided ffmpeg libs which would mean no additional libs are required and would resolve all the problems that have existed due to GStreamer's difficulty to work with.

Not 100% sure how to handle this at the moment.

hamishwillee: Thanks very much @HTRamsey ! 

OK, so I have added a note that this step is not required for daily builds and releases after QGC4 - is that correct? (or is it only correct for Appimages)

1. So what of step 1 in https://docs.qgroundcontrol.com/master/en/qgc-user-guide/getting_started/download_and_install.html#ubuntu is no longer required for daily builds? As this is "main" we should document the daily build behaviour - but with a note that for release builds you still have to do the old steps.
2. Does it make sense to update https://github.com/mavlink/qgroundcontrol/blob/master/src/VideoManager/VideoReceiver/GStreamer/README.md to refer to https://github.com/mavlink/qgroundcontrol/blob/master/tools/setup/install-dependencies-debian.sh as well?

en/companion_computer/video_streaming.md
```diff
@@ -24,13 +24,18 @@ Examples in this library are listed below (note, these are options, not "recomme
 To setup and use video steaming with QGC:
 
 1. Install GStreamer before running QGC.
-   On Ubuntu this is done with the command:
+
+   ::: tip
+   This is not needed for the latest QGroundControl _daily builds_ on Ubuntu (and won't be required for releases after version 4).
```
hamishwillee, line 29:
```suggestion
   This step is not needed for the latest QGroundControl _daily builds_ on Ubuntu (and won't be required for releases after version 4).
```

en/companion_computer/video_streaming.md
HTRamsey: Right now we only provide AppImages so it should be fine. If we ever just provide a zip or something then we'd have to update this. I'd probably reference the script in the readme. Actually there's a lot that's out of date in that readme. @DonLakeFlyer comments on this?

DonLakeFlyer: I'll update to remove gstreamer install instructions from the video stream section and just rely on the top level dependencies install stuff.

hamishwillee: Thanks very much. I have adjusted this accordingly:
- Companion computer section:
  - Shows QGC binary minimum set for companion computer (as per QGC 4). This will probably work for most users.
  - Link to the QGC setup scripts for the full set of dependencies and show the current set for Ubuntu. (In case minimum set doesn't work)
  - Still link to the VideoReceiver guide for more info.
- QGC
  - Install states you only need to do Gstreamer for Ubuntu release builds prior to 4. Otherwise you can skip.

en/companion_computer/video_streaming.md
```diff
@@ -8,11 +8,45 @@ You can't video stream directly from a camera connected to PX4.
...
+    gstreamer1.0-plugins-rtp \
+    gstreamer1.0-gl \
+    gstreamer1.0-libav \
+    gstreamer1.0-rtsp \
+    gstreamer1.0-vaapi \
+    gstreamer1.0-x
+```
+
+::: tip
+For other companion platforms you can mirror the instructions in the [corresponding installation scripts](QGC).
```
hamishwillee, line 46:
```suggestion
For other companion platforms you can mirror the instructions in the [corresponding installation scripts](https://github.com/mavlink/qgroundcontrol/tree/master/tools/setup).
```

PR #3560, opened 2025-01-30
Offboard control Python links

### LINKED ISSUE #1675: px4-ros2 offboard control with pyrhon example
There is a cpp script for offboard control example, but it would be also helpful to provide some launchable python codes for the same example as ros2 usually supports both languages.

hamishwillee:
Add links to examples of offboard Python.

Ideally we'd provide some worked examples in the docs for working with Python. We don't have offers to contribute those, but we do have some reasonable links to resources which we can add. If @Jaeyoung-Lim or someone else wants to add an in doc page for that, would be great, but this is better than nothing.

Fixes #1675

### DIFF
--- a/en/ros2/offboard_control.md
+++ b/en/ros2/offboard_control.md
@@ -7,7 +7,7 @@ While simple, it shows the main principles of how to use offboard control and ho
 
 It has been tested on Ubuntu 20.04 with ROS 2 Foxy and PX4 v1.14.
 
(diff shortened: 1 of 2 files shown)


hamishwillee (commented):
@Jaeyoung-Lim I don't see you having any likely issue with this, so merging. Can correct if you decide you don't like this approach.

PR #3562, opened 2025-01-31
uorb: add section on nested messages

GuillaumeLaine:
Follow up to: https://github.com/PX4/PX4-user_guide/pull/3465#discussion_r1924522578

### DIFF
--- a/en/middleware/uorb.md
+++ b/en/middleware/uorb.md
@@ -15,27 +15,73 @@ New uORB topics can be added either within the main PX4/PX4-Autopilot repository
 For information on adding out-of-tree uORB message definitions, please see [this section](../advanced/out_of_tree_modules.md#out-of-tree-uorb-message-definitions).
 
 To add a new topic, you need to create a new **.msg** file in the `msg/` directory and add the file name to the `msg/CMakeLists.txt` list.
-From this, the needed C/C++ code is automatically generated.
+The section [Message Definitions](#message-definitions) below describes the message format.
 
-Have a look at the existing `msg` files for supported types.
-A message can also be used [nested in other messages](#nested-messages).
+From the message definitions, the needed C/C++ code is automatically generated.
 
 To each generated C/C++ struct, a field `uint64_t timestamp` will be added.
 This is used for the logger, so make sure to fill it in when publishing the message.
 
-To use the topic in the code, include the header:
+To use the topic in the code, first include the generated header, which will be named using the snake-cased version of the (CamelCase) message definition file name.
+For example, for a message named `VelocityLimits` you would include `velocity_limits.h` as shown:
 
 ```cpp
-#include <uORB/topics/topic_name.h>
+#include <uORB/topics/velocity_limits.h>
 ```
 
-By adding a line like the following in the `.msg` file, a single message definition can be used for multiple independent topics:
+In code your refer to the topic using its id, which by default is also the snake_cased version of the message file name: `ORB_ID(velocity_limits)`.
+If needed you can have multiple topics defined from the same message, as described below in [Multi-Topic Messages](#multi-topic-messages).
 
-```cpp
-# TOPICS mission offboard_mission onboard_mission
+## Message Definitions
+
+The message definition should start with a descriptive _comment_ (a comment starts with the `#` symbol and goes to the end of the line).
+The message will then define one or more fields, which are defined with a _type_, such as `bool`, `uint8`, and `float32`, followed by a _name_.
+By convention, each field is followed by a descriptive _comment_, which is any text from the `#` symbol to the end of the line.
+
+For example the [VelocityLimits](../msg_docs/VelocityLimits.md) message definition shown below has a descriptive comment, followed by a number of fields, which each have a comment.
+
+```
+# Velocity and yaw rate limits for a multicopter position slow mode only
+
+uint64 timestamp # time since system start (microseconds)
+
+# absolute speeds, NAN means use default limit
+float32 horizontal_velocity # [m/s]
+float32 vertical_velocity # [m/s]
+float32 yaw_rate # [rad/s]
 ```
 
-Then in the code, use them as topic id: `ORB_ID(offboard_mission)`.
+By default this message definition will be compiled to a single topic with an id `velocity_limits`, a direct conversion from the CamelCase name to a snake-case version.
+
+This is the simplest form of a message.
+See the existing [`msg`](../msg_docs/index.md) files for other examples of how messages are defined.
+
+### Multi-Topic Messages
+
+Sometimes it is useful to use the same message definition for multiple topics.
+This can be specified at the end of the message using a line prefixed with `# TOPICS `, followed by space-separated topic ids.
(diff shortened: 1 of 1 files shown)


hamishwillee:
@GuillaumeLaine You're going to hate me, but I just did a fairly significant restructure of the starting section on defining messages. This is to fix the fact that nested messages was "floating" and it wasn't clear to me the difference between messages and topics.

The reason you'll hate me is that this will require a rebase of the other PR and a little restructure on that.
The good news is that you'll be able to add versioning into that structure fairly easily I think.

Let me know what you think? Note, I wrote this late in the day. I'm pretty sure the structure is good, but there might be typos and grammar errors.

en/middleware/uorb.md
```diff
@@ -15,27 +15,73 @@ New uORB topics can be added either within the main PX4/PX4-Autopilot repository
...
+uint64 timestamp # time since system start (microseconds)
+
+# absolute speeds, NAN means use default limit
+float32 horizontal_velocity # [m/s]
+float32 vertical_velocity # [m/s]
+float32 yaw_rate # [rad/s]
 ```
 
-Then in the code, use them as topic id: `ORB_ID(offboard_mission)`.
+By default this message definition will be compiled to a single topic with an id `velocity_limits`, a direct conversion from the CamelCase name to a snake-case version.
```
GuillaumeLaine, line 54:
```suggestion
By default this message definition will be compiled to a single topic with an id `velocity_limits`, a direct conversion from the CamelCase name to a snake_case version.
```

GuillaumeLaine (commented):
@hamishwillee I think the restructure makes a lot of sense, I'm happy to rebase the other PR once this one goes in 👍🏻

en/middleware/uorb.md
```diff
@@ -15,27 +15,73 @@ New uORB topics can be added either within the main PX4/PX4-Autopilot repository
...
 
-Have a look at the existing `msg` files for supported types.
-A message can also be used nested in other messages.
+From the message definitions, the needed C/C++ code is automatically generated.
 
 To each generated C/C++ struct, a field `uint64_t timestamp` will be added.
 This is used for the logger, so make sure to fill it in when publishing the message.
 
-To use the topic in the code, include the header:
+To use the topic in the code, first include the generated header, which will be named using the snake-cased version of the (CamelCase) message definition file name.
```
GuillaumeLaine, line 25:
```suggestion
To use the topic in the code, first include the generated header, which will be named using the snake_cased version of the (CamelCase) message definition file name.
```

en/middleware/uorb.md
```diff
@@ -15,27 +15,73 @@ New uORB topics can be added either within the main PX4/PX4-Autopilot repository
...
+By default this message definition will be compiled to a single topic with an id `velocity_limits`, a direct conversion from the CamelCase name to a snake-case version.
+
+This is the simplest form of a message.
+See the existing [`msg`](../msg_docs/index.md) files for other examples of how messages are defined.
+
+### Multi-Topic Messages
+
+Sometimes it is useful to use the same message definition for multiple topics.
+This can be specified at the end of the message using a line prefixed with `# TOPICS `, followed by space-separated topic ids.
+For example, the [ActuatorOutputs](../msg_docs/ActuatorOutputs.md) message definition is used to define the topic IDs as shown:
```
GuillaumeLaine, line 63:
```suggestion
For example, the [ActuatorOutputs](../msg_docs/ActuatorOutputs.md) message definition is used to define the topic ids as shown:
```
more nitpicking, just to be consistent either using ID or id

en/middleware/uorb.md
```diff
@@ -4,38 +4,90 @@
...
+A message definition file can define one or more _topics_, which all have the same fields and structure.
+By default a definition maps to a single topic that is named using a snake_case version of the message definition file name (for example, `TopicName.msg` would define a topic `topic_name`).
+You can also specify multiple topics to be created by the message definition, which is useful when you need several topics that have the same fields and structure (see [Multi-Topic Messages](#multi-topic-messages) below).
+
+The section [Message Definitions](#message-definitions) below describes the message format.
+
+From the message definitions, the needed C/C++ code is automatically generated.
 
 To each generated C/C++ struct, a field `uint64_t timestamp` will be added.
 This is used for the logger, so make sure to fill it in when publishing the message.
```
hamishwillee, line 30: @GuillaumeLaine I've accepted your feedback (thanks), done further updates, and I'm nearly ready to merge. 

The only thing blocking me is this paragraph. The message definitions all seem to have a field `uint64_t timestamp` defined in them. My thinking is that this is not quite correct - because it implies this timestamp is added magically in the generated header. I think it means this:

```suggestion
All message definitions **must** include the `uint64_t timestamp` field as shown, and this should be filled in when publishing the associated topic(s).
This field is needed in order for the logger to be able to record UORB topics.
```

Is that correct? If you don't, will it break anything? (i.e. is it a "should")?

Note I would actually move this in the Message definitions section, I've just suggested it here for convenience

en/middleware/uorb.md
```diff
@@ -4,38 +4,90 @@
...
+```text
+# Global position setpoint triplet in WGS84 coordinates.
+# This are the three next waypoints (or just the next two or one).
+
+uint64 timestamp		# time since system start (microseconds)
+
+PositionSetpoint previous
+PositionSetpoint current
+PositionSetpoint next
+```
```
hamishwillee, line 91: @GuillaumeLaine A few lines below it says

> The `listener` command is only available on Pixracer (FMUv4) and Linux / OS X.

Is that still true? How can I check the supported boards?

en/middleware/uorb.md
GuillaumeLaine: Oh good catch, I wasn't aware either but it seems you're right:
https://github.com/PX4/PX4-Autopilot/blob/0115a0b3d87262674f86bdbd4771c958b993c399/Tools/msg/px_generate_uorb_topic_files.py#L124-L130

If you don't include that exact field in your message, then PX4 will fail to build

en/middleware/uorb.md
GuillaumeLaine: I don't know, I'm not even sure how to check. I **think** the [CONFIG_SYSTEMCMDS_TOPIC_LISTENER=y](https://github.com/PX4/PX4-Autopilot/blob/8d296a50f9b3cf2842dfb5e49037fbbd3ecbe532/boards/airmind/mindpx-v2/default.px4board#L96) board config enables the `listener` command, but I'm not sure.

If this is case, this config does seem enabled for most boards.

What I can tell is that the line you quoted was written in the docs earlier (4 years ago) than the listener board config was enabled in most boards (3 years ago). Seems possible that docs were not updated after the fact

hamishwillee: Nice forensics! Also interesting that the boards after fmuv4 are enabled, but the older ones aren't. I'm going to update this as though we were certain.

hamishwillee (commented):
OK, I think this is good. Merging. Much appreciated @GuillaumeLaine

Issue #3565, opened 2025-02-04
Refresh system architecture diagram


mrpollo:
The current system architecture diagram, while still accurate, might not fully represent the actual capabilities of PX4, and it might be worth updating to present the newer additions and capabilities, including the custom flight modes, thanks to the uXRCE-DDS interface and the ROS 2 interface library.

Mostly, what I think should taken into account is that most modern drone topologies include a mission computer and a flight controller and that a pure "flight controller only" architecture is now the exception.

https://docs.px4.io/main/en/concept/px4_systems_architecture.html

![](https://docs.px4.io/main/assets/px4_arch_fc.Cn1UOZ0K.svg)

hamishwillee:
<details>
<summary>A bit of a whinge</summary>
My concern with this is that if we present the topology of FC (RTOS) + App layer compute (Mission computer) "first" then people are going to be very surprised about how "not turnkey" that option is when working with PX4.

That makes me uncomfortable, particularly as nothing else about PX4 supports this priority - we don't have much (if any) ROS-side code, we don't have a reference platform for this architecture, we don't test any code we do have full stack in the same way as we do PX4 on SITL and Pixhawk. The docs are very PX4 setup centric - probably because companion setup is not at all standardized, except PX4 side.

I guess we could move in this direction though. We do need to support this case better, and why not here.
</details>

Further down the page we have a section that broadly covers the architecture you are talking about: https://docs.px4.io/main/en/concept/px4_systems_architecture.html#fc-and-companion-computer

As a small start we might put that section first, updating the image to state how ROS connects them. We might refer to this as "Advanced Drone Architecture" or similar and explain that drones are moving towards this architecture and what PX4 does to support it - including info about ROS2 and the inbuilt modes for referencing the interface layer. Maybe also about things like MAVLink cameras, which are pass through into companions.

Then the stand alone PX4 would appear second.

Does this make sense?

- Other than custom flight modes, what do you think we have that supports this topology?
- Did you have any other thought on what you wanted?
- Were you expecting anyone in particular (other than me) to act on this?

bkueng:
I agree with your points. I still think it makes sense to move into that direction, especially now that we have boards like the ARK Jetson.

I'm thinking something into this direction (not too different from what we have, but I think it better displays how things integrate):

![Image](https://github.com/user-attachments/assets/13f1f1a8-c0c1-4414-a14f-aa17dcdec163)

hamishwillee:
[...]

PR #3577, opened 2025-02-12
Flight modes dev guide updates

### LINKED ISSUE #3564: Update Flight Modes Diagram
The current flight modes diagram is based on v1.12 and needs an update.

https://docs.px4.io/main/en/concept/flight_modes.html

hamishwillee:
This updates the flight mode "dev" guide - https://docs.px4.io/main/en/concept/flight_modes.html

- Removes the diagram
- Removes info about specific modes - links to other docs.
- Adds info about Standard modes and mavlink integration

Lots still to do. This is draft.

Fixes #3564

### DIFF
--- a/en/concept/flight_modes.md
+++ b/en/concept/flight_modes.md
@@ -7,7 +7,7 @@ The pilot transitions between flight modes using switches on the remote control 
 Modes can be implemented as [PX4 internal modes](#px4-internal-modes) running on the flight controller, or as [PX4 external (ROS2) modes](#px4-external-modes) running on a companion computer.
 From the perspective of a ground station (MAVLink), the origin of a mode is indistinguishable.
 
-This topic links to documentation for the supported modes, compares PX4 internal and external modes, provides implementation hints, and provides an overview of how PX4 modes can be used with MAVLink.
+This topic links to documentation for the supported modes, compares PX4 internal and external modes, provides implementation hints, and provides links to how PX4 modes can be used with MAVLink.
 
 ## Supported Modes
 
@@ -56,10 +56,18 @@ PX4 external modes, are written in ROS 2 using the [PX4 ROS 2 Control Interface]
 
 ## PX4 Internal Modes
 
+<!--
 The specific control behaviour of a mode at any time is determined by a [Flight Task](../concept/flight_tasks.md).
 A mode might define one or more tasks that define variations of the mode behavior, for example whether inputs are treated as acceleration or velocity setpoints.
 
 The task that is used is normally defined in a parameter, and selected in [src/modules/flight_mode_manager/FlightModeManager.cpp](https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/flight_mode_manager/FlightModeManager.cpp#L266-L285)
+
+
+Name the relevant modules in which code directly related to flight modes is defined.
+Name any base classes that modes must/should derive from
+Explain the core things you need to do to make a mode work
+Very high level architecture
+-->
 
 ### Mode Restrictions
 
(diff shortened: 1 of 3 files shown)


en/concept/flight_modes.md
```diff
@@ -1,111 +1,130 @@
...
+Something here about what they are, how they related to [Flight Tasks](../concept/flight_tasks.md), where they are located.
+How they are added.
 
-"Manual" modes are those where the user has direct control over the vehicle via the RC control (or joystick).
-Vehicle movement always follows stick movement, but the level/type of response changes depending on the mode.
-For example, experienced fliers can use modes that provide direct passthrough of stick positions to actuators, while beginners will often choose modes that are less responsive to sudden stick-position changes.
+## MAVLink Integration
 
-- **Rovers / Boats:**
+### Standard Modes Protocol
```
hamishwillee, line 57: @bkueng This section is underway.

Some questions you may choose to answer, though it is early for a review.

- I can see that when you add a new mode, you don't need anything special for the current mode to update. 
- Do you need to do anything if you add a new mode to trigger the mode monitor?
- I recall for ROS2 registration will add a mode to the list of available modes. Do you need to do anything in P4 to add a new mode to the list of available modes?

en/concept/flight_modes.md
```diff
@@ -1,111 +1,215 @@
...
+## PX4 ROS 2 Flight Modes (Companion)
+
+[PX4 ROS 2 Control Interface](../ros2/px4_ros2_control_interface.md) documents how to create flight modes running on a companion computer.
 
+## PX Flight Modes (FC)
+
+The specific control behaviour of a flight mode at any time is determined by a [Flight Task](../concept/flight_tasks.md).
+A flight mode might define one or more tasks that define variations of the mode behavior.
+The actual behaviour that is used is normally defined in a parameter, and selected in [src/modules/flight_mode_manager/FlightModeManager.cpp](https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/flight_mode_manager/FlightModeManager.cpp#L266-L285)
+
```
hamishwillee, line 68: @MaEtUgR Can I get some hints for flight modes. It looks like the bit of the code that does something is the Flight Task, and that a mode might have a few of these all that do a particular variant of the mode behaviour - e.g. different types of position control.

I also understand the mode restrictions, which is in the next section (I think).

But not clear 
- how I declare an entirely new mode "HamishMode".
- Where I have to add the code for that.
- Whether I have to add that to the list of modes exported to MAVLink or that happens magically
- If there are any base classes - i.e. What is a mode.

Any chance of some bullet points of the key concepts?

en/concept/flight_modes.md
```diff
@@ -1,111 +1,215 @@
...
+::: tip
+Readers may note that this image is from [PX4 ROS2 Control Interface > Failsafes and mode requirements](../ros2/px4_ros2_control_interface.md#failsafes-and-mode-requirements).
+The requirements and concepts are essentially the same (though defined in different places).
+The main difference is that ROS 2 modes _infer_ the correct requirements to use, while modes in PX4 source code must explicitly specify them.
 :::
 
-## Flight Mode Summary
+## MAVLink Integration
+
+### Standard Modes Protocol
```
hamishwillee, line 116: FYI I might move this whole section under https://docs.px4.io/main/en/middleware/mavlink.html

I do plan to:
- [ ] Add a mavlink section to the mode docs explaining how you can engage the standard modes. Custom modes.

en/concept/flight_modes.md
```diff
@@ -1,111 +1,215 @@
...
 - [Flight Modes (VTOL)](../flight_modes_vtol/index.md)
 - [Drive Modes (Differential Rover)](../flight_modes_rover/differential.md)
 - [Drive Modes (Ackermann Rover)](../flight_modes_rover/ackermann.md)
+- [Basic Configuration > Flight Modes](../config/flight_mode.md)
+
+::: info
+At time of writing there are no PX4 ROS 2 flight modes that are considered "supported as an integral part of the PX4 open source offering".
+:::
+
+## PX4 FC Modes vs PX4 ROS2 Flight Modes
```
hamishwillee, line 29: FYI, this section put me in contortions - how do I refer to a flight mode that runs in the FC?

It was the reason I wrote this https://github.com/PX4/PX4-user_guide/issues/3565#issuecomment-2655407569

sfuhrer: Should we label them with "internal" (vs. "external"=running on ROS2)

sfuhrer (commented):
Looks concise. I unfortunately don't have enough exposure to external flight modes yet to highlight common pitfalls that we should cover with the docs.
One thing I find a bit unfortunate with the new structure (and with the Standard Modes) is that it's hard to distinguish "auto" from "manual" modes.

en/concept/flight_modes.md
mrpollo: I like Internal vs External @sfuhrer

en/concept/flight_modes.md
```diff
@@ -1,111 +1,215 @@
...
 
 - [Flight Modes (Multicopter)](../flight_modes_mc/index.md)
 - [Flight Modes (Fixed-Wing)](../flight_modes_fw/index.md)
 - [Flight Modes (VTOL)](../flight_modes_vtol/index.md)
 - [Drive Modes (Differential Rover)](../flight_modes_rover/differential.md)
 - [Drive Modes (Ackermann Rover)](../flight_modes_rover/ackermann.md)
+- [Basic Configuration > Flight Modes](../config/flight_mode.md)
+
+::: info
+At time of writing there are no PX4 ROS 2 flight modes that are considered "supported as an integral part of the PX4 open source offering".
```
mrpollo, line 26: I wonder if this is needed, I don't think we will ever get to the point where there's an external flight mode that is required for PX4 to run. What's the intention here?

en/concept/flight_modes.md
```diff
@@ -1,111 +1,215 @@
...
+:::
+
+## PX4 FC Modes vs PX4 ROS2 Flight Modes
+
+With some exceptions it is possible to write a mode in either the FC and companion computer.
```
mrpollo, line 31: Maybe a good time to introduce the Internal vs External concept, and use that troughout the doc

en/concept/flight_modes.md
```diff
@@ -1,111 +1,215 @@
...
+- Modes that require low-level access, strict timing, and/or high update rate requirements.
+  For example, a multicopter mode that implements direct motor control.
+- Safety critical modes, such as [Return mode](../flight_modes_mc/return.md).
+- When you can't use ROS (for any reason).
+
+PX4 ROS 2 modes should be considered for all other cases.
+They have the following benefits:
+
+- Easier to implement as there is no need to deal with low-level embedded constraints and requirements (such as restricted stack sizes).
+- Easier to maintain as the integration API is small, well defined, and stable (porting custom PX4 modes on the flight controller between PX4 versions is much harder).
```
mrpollo, line 46: Maybe add a note separate from this bullet point with the warning under parenthesis, and extend on the reasons

en/concept/flight_modes.md
```diff
@@ -1,111 +1,215 @@
...
+- Crashing the app does not crash the vehicle.
```
mrpollo, line 47: While this is technically true, a poorly written mode can still crash the vehicle, maybe rewrite to say something like "Crashing the process results in a fallback to internal flight modes" or something similar

en/concept/flight_modes.md
```diff
@@ -1,111 +1,215 @@
...
+- They can override existing modes to provide more advanced features.
+  You can even override a safety-critical mode with a better versions: if the ROS 2 mode crashes the original mode will be engaged.
+- High-level functionality is available, including a better-feature programming environment, and many useful Linux and ROS libraries.
+- More available compute to do more advanced processing (e.g. computer vision).
+
+Note that there are some negatives to keep in mind:
```
mrpollo, line 53: I think we should open an issue where we track these three bullet points and link to this instead of listing them here where they will likely get stale over time

en/concept/flight_modes.md
```diff
@@ -1,111 +1,215 @@
...
+
+- There is no CI environment for testing ROS 2 applications with PX4 in CI (it is easier to automate testing of PX4 flight modes running on the FC).
+- The [PX4 ROS 2 Control Interface](../ros2/px4_ros2_control_interface.md) that is used to define the modes is experimental
+- There is no formal "PX4 companion computer flight stack", so there is nowhere to contribute PX4 ROS2 modes such that they are become a part of the standard offering available to all users.
+
+## PX4 ROS 2 Flight Modes (Companion)
+
+[PX4 ROS 2 Control Interface](../ros2/px4_ros2_control_interface.md) documents how to create flight modes running on a companion computer.
 
+## PX Flight Modes (FC)
```
mrpollo, line 63:
```suggestion
## PX4 Flight Modes (FC)
```

mrpollo (commented):
I hope this helps

en/concept/flight_modes.md
hamishwillee: @mrpollo @sfuhrer I "like" Internal vs External too, but I'm not certain it is future proof terminology. 

Assuming you read https://github.com/PX4/PX4-user_guide/issues/3565#issuecomment-2655407569, are you suggesting that if in future we have a flight stack that spans FC and Companion, then we will refer to the parts as _PX4 Internal_ (runs on FC) and _PX4 External_ (runs on companion)?

It's a little clunky in that it will make any code on the companion a second class citizen. That might be OK. 
I proposed PX4 RTOS modes / PX4 GTOS modes, but if you are both happy with internal external for this, we can give it a go.

en/concept/flight_modes.md
hamishwillee: There have been statements made at various points that PX4 will eventually offer a software stack that spans the companion computer. That doesn't mean "required to run" so much as "officially supported software".

We're not there yet, but since we are moving towards an architecture in which companion is an expected part of the drone, isn't that the direction you wish to take?

Upshot, perhaps no one would assume there might be. But I'd like to understand where we are going.

en/concept/flight_modes.md
hamishwillee: Yes, as long as we agree the terminology - https://github.com/PX4/PX4-user_guide/pull/3577#discussion_r1960875336

en/concept/flight_modes.md
hamishwillee:
```suggestion
- Process termination of a ROS 2 mode results in a fallback to an internal flight mode (while termination of an internal mode may well crash the vehicle).
```

en/concept/flight_modes.md
hamishwillee: @mrpollo My _assumption_ here is that PX4 dev team makes little effort in any respect support custom forks, and may well break any dependency of a custom flight mode - with the possible exception of core things such as the uorb interface itself, and from now on things like uorb messages.

If that is what is alluded to here, then how about?


```suggestion
- Easier to maintain as the integration API is small, well defined, and stable.
- Porting custom PX4 modes on the flight controller between PX4 versions can be much harder, as often flight modes use interfaces that are considered internal, and allowed to change. 
```

en/concept/flight_modes.md
hamishwillee: @mrpollo I propose you do that - the point about the control interface being experimental is relevant here but probably wouldn't go in the same issue. We do need to talk to @bkueng about what is needed to get it to the point it is considered "stable".

I did create this one just for the CI:
- https://github.com/PX4/PX4-Autopilot/issues/24363

en/concept/flight_modes.md
hamishwillee: For now I will try Internal vs External, but I would like you to think about ^^^^

hamishwillee:
@mrpollo @sfuhrer OK, updated to use external/internal. I explained above the potential issue with this, but I think we can live with it for now. Also removed the bits that were worrying you @mrpollo.

What worries me about this is that you still couldn't create a flight mode with this information - i.e. we need an answer to https://github.com/PX4/PX4-user_guide/pull/3577#discussion_r1953934066

en/concept/flight_modes.md
sfuhrer: > "Internal" implies "you've got to be on the flight stack to really be part of PX4".

But that's how it is - we (PX4 community) do not maintain the ROS flight modes, so the user is on it's own when they use them.

I btw fully support you in spending the time now to think this through - it will be very hard to change in the future once people get used to the naming.

hamishwillee: Yes, that is exactly the problem.

If I knew we'd have "full stack" PX4 in a year or so I'd push for `PX4 RTOS modes` and `PX4 GCOS modes` (or replacing RTOS/GCOS with RT, NuttX, GC, Linux).

hamishwillee:
@mrpollo @sfuhrer @MaEtUgR 
- I removed the information on how to create PX4 Internal modes other than the mode requirement section (which is accurate)
- I moved the standard modes integration docs into a standalone MAVLink doc, because you might want to know about that without knowing about the rest of this stuff - and linked

This still can't be used for creating internal modes, but it is not "incorrect" and it is more useful than what it replaces.

Merging for now, but I still need the extra info.

PR #3596, opened 2025-02-22
Add PX4 ROS 2 service servers docs

### LINKED ISSUE #2876: Warning: uXRCE-DDS now supports services
After https://github.com/PX4/PX4-Autopilot/pull/22313 you have to use ROS2 services to send VehicleCommands.
Doc update is required.

Example of usage is here: https://github.com/PX4/px4_ros_com/pull/207

beniaminopozzan:
With this PR, PX4 service servers documentation is added. The documentation consists on a very general overview on the service servers capabilities and a in-dept description of the `VehicleCommand` service (which is also the only service currently implemented on PX4).

The service description is completed with a references to the available `offboard_control_srv` node, which uses the service to perform offboard control.

Fixes #2876


en/middleware/uorb.md
```diff
@@ -125,7 +125,7 @@ Versioned messages include an additional field `uint32 MESSAGE_VERSION = x`, whe
 
 Versioned and non-versioned messages are separated in the file system:
 
-- Non-versioned topic message files and service message files remain in the [`msg/`](https://github.com/PX4/PX4-Autopilot/tree/main/msg) and [`srv/`](https://github.com/PX4/PX4-Autopilot/tree/main/srv) directories, respectively.
+- Non-versioned topic message files and [server service](../ros2/user_guide.md#px4-ros-2-service-servers) message files remain in the [`msg/`](https://github.com/PX4/PX4-Autopilot/tree/main/msg) and [`srv/`](https://github.com/PX4/PX4-Autopilot/tree/main/srv) directories, respectively.
```
hamishwillee, line 128: FYI, cross linking to docs now we have a definition.

en/middleware/uxrce_dds.md
```diff
@@ -490,6 +490,17 @@ Add a topic to the `subscriptions_multi` section to:
...
+## DDS (ROS 2) Services
+
+PX4 uXRCE-DDS middleware supports [ROS 2 services](https://docs.ros.org/en/jazzy/Concepts/Basic/About-Services.html).
+These are remote procedure calls, from one node to another, which return a result.
+They simplify communication between ROS 2 nodes and PX4 by encapsulating and associating the request and response topics, ensuring that replies are only returned to the specific requesting user.
+
+A service server is the entity that will accept a remote procedure request, perform some computation on it, and return the result.
+For example, the `/fmu/vehicle_command` service server defined in [`px4_msgs::srv::VehicleCommand`](https://github.com/PX4/px4_msgs/blob/main/srv/VehicleCommand.srv) can be called by ROS 2 applications to send PX4 [VehicleCommand](../msg_docs/VehicleCommand.md) uORB messages and receive PX4 [VehicleCommandAck](../msg_docs/VehicleCommandAck.md) uORB messages in response.
+
+For a list of services, details and examples see the [service documentation](../ros2/user_guide.md#px4-ros-2-service-servers) in the ROS 2 User Guide.
```
hamishwillee, line 502: @beniaminopozzan I modified this so that we just have one place where all services are documented "in detail". The ROS2 user guide is the right place for that.

This is a little bit of duplication, but shouldn't need further change once you're happy with it.

en/ros2/px4_ros2_msg_translation_node.md
```diff
@@ -318,15 +318,15 @@ The example describes the process of updating the `VehicleAttitude` message defi
...
+   ```txt
    uint32 MESSAGE_VERSION = 3
    uint64 timestamp
    ...
    ```
 
    to
 
-   ```.msg
+   ```txt
```
hamishwillee, line 329: FYI, just removing unsupported code block types (unrelated to this PR)

en/ros2/user_guide.md
```diff
@@ -751,6 +751,117 @@ will generate topics under the namespaces:
 
 :::
 
+## PX4 ROS 2 Service Servers
```
hamishwillee, line 754: OK, so I have modified this to have the structure
- Intro 
  - what a ros 2 service is and why is it cool
  - List of services in PX4
 - VehicleCommand service
   - Example
 - More services as they come along ....
 
 This is where we centralise service docs.

en/ros2/user_guide.md
```diff
@@ -751,6 +751,115 @@ will generate topics under the namespaces:
...
+
+<Badge type="tip" text="PX4 v1.15" />
+
+PX4 uXRCE-DDS middleware supports [ROS 2 services](https://docs.ros.org/en/jazzy/Concepts/Basic/About-Services.html).
+Services are remote procedure calls, from one node to another, that return a result.
+
+A service server is the entity that will accept a remote procedure request, perform some computation on it, and return the result.
+They simplify communication between ROS 2 nodes and PX4 by encapsulating and associating the request and response topics, and ensuring that replies are only returned to the specific requesting user.
+This is much easier that publishing the request, subscribing to the reply, and filtering out any unwanted responses.
```
hamishwillee, line 763: I pulled this out of your Vehicle command docs below, because I presume it is true in all cases. Removed text this comes from was:

> This is an helpful alternative to having to create a topic publisher for `/fmu/in/vehicle_command` and a topic subscriber for `/fmu/out/vehicle_command_ack`, and then having to filter out unwanted `VehicleCommandAck` messages.

en/ros2/user_guide.md
```diff
@@ -751,6 +751,115 @@ will generate topics under the namespaces:
...
+
+The service servers that are built into the PX4 [uxrce_dds_client](../modules/modules_system.md#uxrce-dds-client) module include:
+
+- `/fmu/vehicle_command` (definition: [`px4_msgs::srv::VehicleCommand`](https://github.com/PX4/px4_msgs/blob/main/srv/VehicleCommand.srv).)
```
hamishwillee, line 767: 1. What is the terminology/name for `/fmu/vehicle_command`? I.e. `/fmu/out/` is a publication, and `fmu/in` is a subscription that you an publish to. Is it just a "server"? Is it a "topic" or a "message"?

2. What is the name for `px4_msgs::srv::VehicleCommand` ? YOu called it a service Type - where does "Type" come from? I have called it a definition here, because that's what is linked.

[Ultimately I'm asking if this line is OK!)

3. `VehicleCommand.srv` is a service server message definition. Should we be auto-generating docs for this, as we do for normal uorb messages?

en/ros2/user_guide.md
```diff
@@ -751,6 +751,115 @@ will generate topics under the namespaces:
...
+
+  This service can be called by ROS 2 applications to send PX4 [VehicleCommand](../msg_docs/VehicleCommand.md) uORB messages and receive PX4 [VehicleCommandAck](../msg_docs/VehicleCommandAck.md) uORB messages in response.
+
+Details and specific examples are provided in the following sections.
+
+
+### VehicleCommand service
+
+This can be used to send commands to the vehicle, such as "take off", "land", change mode, and "orbit", and receive a response.
```
hamishwillee, line 776: I thought we should say what this is for, as readers might not understand what vehicle commands are.

en/ros2/user_guide.md
```diff
@@ -751,6 +751,115 @@ will generate topics under the namespaces:
...
+
+The service is available under `/fmu/vehicle_command`, which can be changed through custom namespaces just like the other PX4 topics.
```
hamishwillee, line 778: "which can be changed through custom namespaces just like the other PX4 topics." - I don't know what that means. Can you link it.

en/ros2/user_guide.md
```diff
@@ -751,6 +751,115 @@ will generate topics under the namespaces:
...
+
+```txt
+VehicleCommand request
+---
+VehicleCommandAck reply
+```
+
+Users can make service requests by sending [VehicleCommand](../msg_docs/VehicleCommand.md) messages, and receive a [VehicleCommandAck](../msg_docs/VehicleCommandAck.md) message in response.
+The service ensures that only the `VehicleCommandAck` reply generated for the specific request made by the user is sent back.
+
```
hamishwillee, line 789: Not every command can be sent - in fact most can't. Should/can we list the subset that will work?

en/ros2/user_guide.md
```diff
@@ -751,6 +751,115 @@ will generate topics under the namespaces:
...
+#### VehicleCommand Service Offboard Control Example
+
+A complete _offboard control_ example using the VehicleCommand service is provided by the [offboard_control_srv](https://github.com/PX4/px4_ros_com/blob/main/src/examples/offboard/offboard_control_srv.cpp) node available in the `px4_ros_com` package.
+
+The example closely follows the _offboard control_ example described in [ROS 2 Offboard Control Example](../ros2/offboard_control.md) but uses the `VehicleCommand` service to request mode changes, vehicle arming and vehicle disarming.
+
+Like for any other ROS 2 service client, the ROS 2 application needs to declare a service client of type `px4_msgs::srv::VehicleCommand`:
```
hamishwillee, line 796:
```suggestion
First the ROS 2 application needs to declare a service client, just like for any other ROS2 service client.
In this case the the client has type: `px4_msgs::srv::VehicleCommand`:
```

hamishwillee: ^^^^ assumption is `rclcpp::Client()` is the "normal way", so we should spell it out.

en/ros2/user_guide.md
```diff
@@ -751,6 +751,115 @@ will generate topics under the namespaces:
...
+
+```cpp
+rclcpp::Client<px4_msgs::srv::VehicleCommand>::SharedPtr vehicle_command_client_;
+```
+
+and then initialize it to the right ROS 2 service.
+As the application assumes the standard PX4 namespace is used, this results in
```
hamishwillee, line 803:
```suggestion
Then the client is initialized to the right ROS 2 service (`/fmu/vehicle_command`).
As the application assumes the standard PX4 namespace is used, the code to do this looks like this:
```

en/ros2/user_guide.md
```diff
@@ -751,6 +751,115 @@ will generate topics under the namespaces:
...
+
+```cpp
+vehicle_command_client_{this->create_client<px4_msgs::srv::VehicleCommand>("/fmu/vehicle_command")}
+```
+
+After that, the client can be used to send any vehicle command request.
```
hamishwillee, line 809: The code seems to also require that you first wait for the service:

```
while (!vehicle_command_client_->wait_for_service(1s)) {
			if (!rclcpp::ok()) {
```

Where does `wait_for_service()` come from?

hamishwillee (commented):
Looks good. I've added some slight restructure and some questions.

en/ros2/user_guide.md
beniaminopozzan: More info added in 890a999c250424e52a535a953195e834717c583a

en/ros2/user_guide.md
beniaminopozzan: The ROS 2 node that is making the service request is treated as an internal PX4 module (unless https://github.com/PX4/px4_msgs/blob/912244a6e3a64324f01086569ac9805df854f5d9/msg/VehicleCommand.msg#L195 is set to true). Therefore the user can use any command.

As far as I know, the only commands that has no effect are the ones that deal with PX4 parameters (you cannot set, get, list) parameters.

beniaminopozzan (commented):
Thanks @hamishwillee for the review!

I hope I answered all your questions :)

en/ros2/user_guide.md
hamishwillee: This was resolved, but "thumbsed up". So I am unresolving and merging.

en/ros2/user_guide.md
```diff
@@ -751,6 +751,115 @@ will generate topics under the namespaces:
...
 
+## PX4 ROS 2 Service Servers
+
+<Badge type="tip" text="PX4 v1.15" />
+
+PX4 uXRCE-DDS middleware supports [ROS 2 services](https://docs.ros.org/en/jazzy/Concepts/Basic/About-Services.html).
+Services are remote procedure calls, from one node to another, that return a result.
+
+A service server is the entity that will accept a remote procedure request, perform some computation on it, and return the result.
+They simplify communication between ROS 2 nodes and PX4 by encapsulating and associating the request and response topics, and ensuring that replies are only returned to the specific requesting user.
```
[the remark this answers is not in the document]
hamishwillee: @beniaminopozzan I guess by encapsulate I mean "serve the same purpose as". The service is actually defined in terms of the command and command ack messages, which is what I was getting at - need to get a consistent vocabulary in my head. Anyway, I'm trying to capture the value of this service concept, which as I see it is handling the command/response and filtering for you. Is this better (if not, can you counter-propose an alternative)?

[for example, where I write "by associating the request and response", I'd like to append "topics", but I think that is wrong. Who knows].

```suggestion
They simplify communication between ROS 2 nodes and PX4 by associating the request and response, and ensuring that replies are only returned to the specific requesting user.
```

en/ros2/user_guide.md
hamishwillee: Thank you very much for this. I'm going to email myself with a link and think about it a bit more.

en/ros2/user_guide.md
hamishwillee: Shouldn't we say that? Note that I have no idea where this method is defined (?) I assume `wait_for_service()` is some kind of standard interface/API? Or is it our own interface that hides some other magic?

Assuming it is standard, you might do something like this:

```suggestion
::: info
It is good practice to make sure the that the server server is ready to accept requests before sending them. You can use `wait_for_service()` to check if the server is ready.
:::

After that, the client can be used to send any vehicle command request.
```

en/middleware/uxrce_dds.md
```diff
@@ -490,6 +490,17 @@ Add a topic to the `subscriptions_multi` section to:
 You can arbitrarily change the configuration.
 For example, you could use different default namespaces or use a custom package to store the message definitions.
 
+## DDS (ROS 2) Services
+
+PX4 uXRCE-DDS middleware supports [ROS 2 services](https://docs.ros.org/en/jazzy/Concepts/Basic/About-Services.html).
+These are remote procedure calls, from one node to another, which return a result.
+They simplify communication between ROS 2 nodes and PX4 by encapsulating and associating the request and response topics, ensuring that replies are only returned to the specific requesting user.
```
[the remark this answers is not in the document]
hamishwillee:
```suggestion
They simplify communication between ROS 2 nodes and PX4 by grouping the request and response behaviour, and ensuring that replies are only returned to the specific requesting user.
```

en/ros2/user_guide.md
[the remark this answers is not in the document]
hamishwillee: @beniaminopozzan I need to merge this so we don't have to start again when sources move into PX4-Autopilot. So I am accepting my change here and in https://github.com/PX4/PX4-user_guide/pull/3596/files#r1990496995

Hope you can do a follow up review!

hamishwillee:
@beniaminopozzan Thanks for all of this. I want to merge, because most of this is definitely clear.

As a post process, would appreciate it if you could look at 
- https://github.com/PX4/PX4-user_guide/pull/3596/files#r1990483208 and https://github.com/PX4/PX4-user_guide/pull/3596#discussion_r1990504251

PR #3597, opened 2025-02-22
Update Micro XRCE-DDS Agent required version

beniaminopozzan:
Since Eprosima's release of Micro XRCE-DDS Agent version 3.0.0, cloning `main` breaks compatibility with PX4.

This PR warns about the isse and suggest the right version to use (`v2.4.2`) to have full compatibility with ROS 2 Humble and Foxy.

Note that this is not the **latest** `v2.4.x` vesion as `v2.4.3` exists. However, this version is incompatible with building the Agent with colcon in a ros workspace as internally it uses `Fast CDR` version `2.2.x` while both ROS 2 Humble and Foxy use version `1.x`

### DIFF
--- a/en/middleware/uxrce_dds.md
+++ b/en/middleware/uxrce_dds.md
@@ -49,11 +49,15 @@ Note that all code generation associated with the messages is handled by ROS 2.
 ## Micro XRCE-DDS Agent Installation
 
 The Micro XRCE-DDS Agent can be installed on the companion computer using a binary package, built and installed from source, or built and run from within a ROS 2 workspace.
-All of these methods fetch _all_ the dependencies needed to communicate with the client (such as FastCDR)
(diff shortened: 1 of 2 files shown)


hamishwillee (approved):
Thanks very much @beniaminopozzan 

1. Is this likely to change again any time soon?
2. Will this affect v1.15? 

Essentially I'm wondering if we need to backport this to v1.15 because we'll otherwise get lots of questions. That would make sense if the instructions would "always" work with v1.15 (and the old ones wont).

beniaminopozzan:
@hamishwillee I'm not aware of plans of updating the PX4 client side to `v3.x`, so Agent `v2.x` will remain. Moreover, even the latest ROS 2 LTS (Jazzy) is still using FastDDS 2.x. Things may change with the next LTS but for that we'll need to wait a year.

Yeah, it requires backport on v1.15

hamishwillee:
Thank you very much!

PR #3618, opened 2025-03-05
Gazebo supports liftdrag now

hamishwillee:
According to @dakejahl the plugin is now supported: https://discord.com/channels/1022170275984457759/1031894342731829288/1345492981620277269. 

A few small questions inline:
- Doc says

  > For more detail on the lift calculations for the Advanced Plane, see [PX4-SITL_gazebo-classic/src/liftdrag_plugin/README.md](https://github.com/PX4/PX4-SITL_gazebo-classic/blob/main/src/liftdrag_plugin/README.md)

  That's **PX4-SITL_gazebo-classic** (not Gz). Where is this information for GZ?

- Will this work out of the box for 1.15 as well?


dakejahl:
> That's PX4-SITL_gazebo-classic (not Gz). Where is this information for GZ?

I guess it doesn't exist.

> Will this work out of the box for 1.15 as well?

Not sure about 1.15... it won't work on Garden. I think 1.15 uses Garden?

hamishwillee:
Thanks. Then in it goes, as is.

PR #3620, opened 2025-03-05
Battery load compensation fixes

hamishwillee:
This improves the battery load compensation in response to questions from @DonLakeFlyer in https://discord.com/channels/1022170275984457759/1345826197585662043

Specifically it makes clear that `BAT1_R_INTERNAL` determines the kind of load compensation but not whether compensation is used or not. 

@chfriedrich98 A few questions:
1. Is it the presence of the "Amps per volt divider" setting that enables current load compensation?
2. Assuming yes above, this enables for both manual and automatic estimate of internal resistance right?
3. Do we actually ever measure the current in our power module, or is the estimate always from an initial load test? 
4. What I'm really confirming here is that you don't need a power module that can measure current - right? We use the voltage measurements to estimate the current and also the internal resistance.
5. When/why would you set this resistance manually? My guess is if you can - if the value is measurable then you would trust that more than our estimate?

### DIFF
--- a/en/config/battery.md
+++ b/en/config/battery.md
@@ -185,11 +185,6 @@ Using a real time estimate allows the compensation to adapt to changes in the in
 The internal resistance can also be measured and [set manually](../advanced_config/parameters.md) in [BAT1_R_INTERNAL](../advanced_config/parameter_reference.md#BAT1_R_INTERNAL).
 A positive value in this parameter will be used for the internal resistance instead of the estimated value (`0` disables load compensation altogether).
 
-::: tip
-The default approach of using the real-time estimate is preferred.
-While in theory this might not converge to a value for the internal resistance, testing has not show this to be a problem.
-:::
-
 :::info
 There are LiPo chargers that can measure the internal resistance of your battery.
 A typical value for LiPo batteries is 5mΩ per cell but this can vary with discharge current rating, age and health of the cells.
--- a/en/config/battery.md
+++ b/en/config/battery.md
@@ -7,7 +7,7 @@ These instructions require that the vehicle has a [Power Module (PM)](../power_m
 :::
 
 ::: tip
-This tuning is not needed for [Smart/MAVLink Batteries](../smart_batteries/index.md).
+This tuning is not needed for [Smart/MAVLink Batteries](../smart_batteries/index.md): batteries that can supply a reliable indication of remaining charge.
 :::
 
 ## Overview
--- a/en/config/battery.md
+++ b/en/config/battery.md
@@ -69,7 +69,7 @@ This sets the number of cells connected in series in the battery.
 Typically this will be written on the battery as a number followed by "S" (e.g "3S", "5S").
 
 ::: info
-The voltage across a single galvanic battery cell is dependent on the chemical properties of the battery type.
+The voltage across a single galvanic battery cell is dependent on the [chemical properties of the battery type](../power_systems/battery_chemistry.md).
 Lithium-Polymer (LiPo) batteries and Lithium-Ion batteries both have the same _nominal_ cell voltage of 3.7V.
 In order to achieve higher voltages (which will more efficiently power a vehicle), multiple cells are connected in _series_.
 The battery voltage at the terminals is then a multiple of the cell voltage.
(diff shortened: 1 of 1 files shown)


chfriedrich98:
I agree with the changes that you made here, they make it easier to understand.
In regards to your questions:
1./2. Yes, in addition to the `Basic Battery Settings`, load compensation only further requires a current measurement, regardless whether the internal resistance is estimated (`BAT1_R_INTERAL = -1`) or manually set (`BAT1_R_INTERAL > 0`).
3. The current is measured throughout the flight and in case of the online internal resistance estimation it is continously used to update the estimate (IR values can change slightly based on temperature, SoC etc.).
4. Well, we require a value for the current, whether that is measured using a shunt resistor (Amps per volt divider) or any other method does not matter in principle.
5. That is up to the user. The estimation is enabled by default because most users never properly set up the value for IR and just used the default value, which can be quite far off depeding on the battery used. The estimation has the further advantage that it can adapt to changes to the IR due to temperature, SoC etc. during the flight and slow changes to the IR due to degredation of the battery over multiple uses.
However, since the estimation has no guarantee of convergence, setting the value manually if it is available _could_ be considered the safter option (even though during testing I have never had issues so far).

chfriedrich98 (approved):

hamishwillee (commented):
Thanks @chfriedrich98 - much appreciated. Can I please ask for a new scan of this doc, because some of the broader context was back to front. I have tried to provide more context on the voltage and current divider.

Specifically, 
- you use the voltage divider to calibrate the voltage measurements from the power module against an external measurement.
- you use the amps to volts divider to calibrate the current measurement.
- What it was saying before was that you do the calibration in order to enable load compensation/current integration. That is incorrect - load compensation will happen by default anyway if you are providing a current measurement - you must do this or the estimate will be wrong. 

This also says that you don't need to do the calibrations if you have a smart battery, which makes sense because the provided values are expected to be accurate/precalibrated. Can you make that assumption with any other type of Power module? For example, would a CAN power module still have to be calibrated like this?

en/config/battery.md
```diff
@@ -164,25 +168,31 @@ This setting corresponds to parameter(s): [BAT1_A_PER_V](../advanced_config/para
...
+PX4 uses current-based load compensation based on a _real-time estimate_ of the internal resistance of the battery by default (real time estimates are enabled if [BAT1_R_INTERNAL=-1](../advanced_config/parameter_reference.md#BAT1_R_INTERNAL)).
+Using a real time estimate allows the compensation to adapt to changes in the internal resistance of the battery due to temperature changes during flight, as well as over time as the battery degrades.
+
+The internal resistance can also be measured and [set manually](../advanced_config/parameters.md) in [BAT1_R_INTERNAL](../advanced_config/parameter_reference.md#BAT1_R_INTERNAL).
+A positive value in this parameter will be used for the internal resistance instead of the estimated value (`0` disables load compensation altogether).
+
+::: tip
+The default approach of using the real-time estimate is preferred.
+While in theory this might not converge to a value for the internal resistance, testing has not show this to be a problem.
+:::
```
chfriedrich98, line 191:
```suggestion
::: tip
The default approach of using the real-time estimate is preferred.
While in theory this might not converge to a value for the internal resistance, testing has not show this to be a problem.
:::
```
I would ommit this tip, since the real-time estimate is the default anyway and "not converging" sounds quite scary. 
We also list the advantages 2 paragraphs above, so someone that has enough knowledge about battieries to understand internal resistance and measuring it, they are able to make that call themselves and for your average user the default setting works and i think this tip would just unsettle them.
But if you prefer leaving it in, that also works for me.

chfriedrich98 (commented):
I really like the changes I think this streamlines the setup process nicely.

chfriedrich98:
> This also says that you don't need to do the calibrations if you have a smart battery, which makes sense because the provided values are expected to be accurate/precalibrated. Can you make that assumption with any other type of Power module? For example, would a CAN power module still have to be calibrated like this?

I think smart batteries in PX4 refer to batteries that supply information through `MAVLink` and that should not require any calibration (even though I'm not sure on this because I did not work with any smart batteries yet).
In regards to other types of power modules, I don't know since I have not worked with many different ones yet.
Maybe @MaEtUgR has more insight on this topic ?

en/config/battery.md
hamishwillee: No, I think you are right. Not needed.

en/config/battery.md
```diff
@@ -164,25 +168,31 @@ This setting corresponds to parameter(s): [BAT1_A_PER_V](../advanced_config/para
...
+
```
hamishwillee, line 192:
```suggestion
A positive value in this parameter will be used for the internal resistance instead of the estimated value (`0` disables load compensation altogether).

```

hamishwillee:
> I think smart batteries in PX4 refer to batteries that supply information through MAVLink and that should not require any calibration (even though I'm not sure on this because I did not work with any smart batteries yet).
In regards to other types of power modules, I don't know since I have not worked with many different ones yet.

OK, so you're mostly right about smart batteries. Not all MAVLink batteries are smart batteries - only those  that supply a reliable capacity level. When you have that capacity level you don't need to calculate it using voltage or current. That said, right now there is no way to be sure whether the data you are getting from a mavlink battery for capacity is accurate or not - that will wait on [BATTERY_STATUS_V2](https://mavlink.io/en/messages/development.html#BATTERY_STATUS_V2) with the flag `MAV_BATTERY_STATUS_FLAGS_CAPACITY_RELATIVE_TO_FULL`. 

[in retrospect, I should have designed this like [FUEL_STATUS](https://mavlink.io/en/messages/common.html#FUEL_STATUS) where only measured values are sent - that way a GCS can decide what it wants to infer, and how.]

I guess I was really wondering if a CAN battery can be assumed to be smart. In retrospect that doesn't make sense - CAN is no different from MAVLink - unless the message says "Hey I'm a smart battery" then there is no reason to assume this.

en/config/battery.md
```diff
@@ -4,7 +4,9 @@ This topic explains how to configure power settings so that PX4 can estimate the
 
 ::: info
 These instructions require that the vehicle has a [Power Module (PM)](../power_module/index.md), or other hardware that can measure the battery voltage and (optionally) the current.
+:::
 
+::: tip
 This tuning is not needed for [Smart/MAVLink Batteries](../smart_batteries/index.md).
```
hamishwillee, line 10:
```suggestion
This tuning is not needed for [Smart/MAVLink Batteries](../smart_batteries/index.md): batteries that can supply a reliable indication of remaining charge.
```

en/config/battery.md
```diff
@@ -4,8 +4,10 @@ This topic explains how to configure power settings so that PX4 can estimate the
 
 ::: info
 These instructions require that the vehicle has a [Power Module (PM)](../power_module/index.md), or other hardware that can measure the battery voltage and (optionally) the current.
+:::
 
-This tuning is not needed for [Smart/MAVLink Batteries](../smart_batteries/index.md).
+::: tip
+This tuning is not needed for [Smart/MAVLink Batteries](../smart_batteries/index.md): batteries that can supply a reliable indication of remaining charge.
```
hamishwillee, line 10: @MaEtUgR Should a battery than can reliably supply discharge rates - max and/or average - do so in [BATTERY_STATUS_V2](https://mavlink.io/en/messages/development.html#BATTERY_STATUS_V2)?

Right now we supply a battery level and we trigger on percentage levels. If we supply a capacity and an average discharge rate then you would have flight time min and average remaining, which would allow a fair bit less guesswork.

Just thinking aloud - perhaps this is overkill. Not going to block merging this.

PR #3622, opened 2025-03-05
Test Card Cleanup

afwilkin:
We are doing more of the flight testing. we are using these test cards as a way to track what we are doing. Wanted to clean them up a bit so we can expand on them as well


hamishwillee:
@mrpollo These seem better than what they replace. However test cards are really a dev team question - are you happy with this? Can you add to the call to sign off?

dagar (approved):

hamishwillee:
Thanks @afwilkin !

PR #3630, opened 2025-03-06
Update ssd size in Jetson baseboard documentation

MightyPiggie:
The official documentation only specifies the 2242 size, which is wrong. It also supports the 2280 size, which is also shown on images on the same page.

I have notified Holybro of the issue as well.


MightyPiggie:
Holybro has now updated there own documentation to reflect the correct size as well

hamishwillee (approved):
Thanks very much!

PR #3640, opened 2025-03-12
holybro_pm02d.md: add parameter for config

sfuhrer:
During an integration it came up that the setup of this power module is not straight forward as the built in sensor (according to PX4 guide) is an INA226 but the parameter that has to be enabled for the high voltage version is INA228.

### DIFF
--- a/en/power_module/holybro_pm02d.md
+++ b/en/power_module/holybro_pm02d.md
@@ -1,6 +1,7 @@
 # Holybro PM02D Power Module
 
 The Holybro PM02D digital power module provides regulated power to flight controller and power distribution board, and sends information to the autopilot about battery voltage and current supplied to the flight controller and the motors.
+There are two versions: high voltage (12S/14S) and low voltage (6S).
 
 The power module is connected using the I2C protocol.
 It is designed for flight controllers based on the Pixhawk FMUv5X and FMUv6X open standard, including the [Pixhawk 5X](../flight_controller/pixhawk5x.md).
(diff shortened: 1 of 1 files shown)


niklaut (approved):
Yup, that's correct now.

en/power_module/holybro_pm02d.md
```diff
@@ -37,3 +38,18 @@ The PM is **NOT** compatible with flight controllers that require an analog powe
 ![pm02d_pinout](../../assets/hardware/power_module/holybro_pm02d/pm02d_pinout.png)
 
 Additional wiring and connection information can be found in: [Holybro Pixhawk 5x Wiring Quick Start](../assembly/quick_start_pixhawk5x.md).
+
+## Configuration
```
hamishwillee, line 42:
```suggestion
## PX4 Configuration
```

hamishwillee:
Also thanks for the review @niklaut

hamishwillee:
@vincentpoont2 FYI, note this fix. Are there really two versions of the PM02D?

vincentpoont2:
Yes. The original pm02d uses the Ina226, it was changed to ina228 maybe around 2 years ago. We used to have two selection on the website, but I think it was removed by my colleagues not too long ago.

hamishwillee:
> Yes. The original pm02d uses the Ina226, it was changed to ina228 maybe around 2 years ago. We used to have two selection on the website, but I think it was removed by my colleagues not too long ago.

Thanks. Tidying this further then in https://github.com/PX4/PX4-Autopilot/pull/24539

PR #3650, opened 2025-03-19
Aerium Lumen/Pascal

yaadch:
Added links for Aerium baseboards for NVIDIA Jetson and Raspberry Pi CM4/CM5


hamishwillee:
Thanks. This repo is being archived - can you please recreate this PR in https://github.com/PX4/PX4-Autopilot/tree/main/docs/en ?

NOte, 
1. you have the right page, but these should be under the heading further up in the page "Pixhawk Autopilot Bus Carrier Boards with Companion Computers"
2. This is similar to the ARK and Holybro boards. As you can see from https://docs.px4.io/main/en/companion_computer/ark_jetson_pab_carrier.html#ark-jetson-pab-carrier we normally have setup information for those on the site - in particular if there are particular needs for configuring things like ethernet. Do you intend to create those.

